PR #7, opened 2023-03-02
Add first iteration of Caliptra Runtime Firmware spec

jhand2:
The Caliptra Runtime Firmware spec defines the responsibilities and behavior of the Caliptra Runtime Firmware.

This is the same as https://github.com/chipsalliance/caliptra-sw/pull/1, which was closed due to recent updates to main and cannot be re-opened.


runtime_firmware.md
```diff
@@ -0,0 +1,418 @@
...
+
+* `Extend(pcr_index, INPUT_DATA)`
+
+The DPE ExtendTci command will exercise the following PCR operations:
+
+* `Extend(pcr_index, INPUT_DATA)`
+
+The DPE DestroyContext command will exercise the following PCR operations:
+
+* `Clear(pcr_index)`
```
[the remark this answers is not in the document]
Bryankel: I think perhaps we conflate possession of key with measurement and assuming equal outcome if the key is lost or measurements are cleared.  The context is also a weak security mechanism for preventing forgery of measurement.

Also agree with the point on partial reset and challenges with clean state.

runtime_firmware.md
```diff
@@ -0,0 +1,418 @@
...
+
+| Field                          | Sub Field   | Value
+| -------------                  | ---------   | ---------
+| Version                        | v3          | 2
+| Serial Number                  |             | First 20 bytes of sha256 hash of DPE Alias public key
+| Issuer Name                    | CN          | Caliptra Runtime Alias
+|                                | serialNumber | First 20 bytes of sha384 hash of Runtime Alias public key
+| Validity                       | notBefore   | February 10th, 2023
+|                                | notAfter    | 99991231235959Z
+| Subject Name                   | CN          | Caliptra DPE Leaf
```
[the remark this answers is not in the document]
varuns-nvidia: I'm starting to understand now. @Bryankel , your concern is that there is no way to ask DPE to report all measurements? DPE only reports measurements associated with a node's TCB, where TCB is defined by the boot chain of the ancestor notes.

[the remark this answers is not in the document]
bluegate010: There is a mechanism for allowing DPE to report all measurements it currently has in its tree - TagTci / GetTaggedTci. What may be missing is a mechanism that allows DPE to report all measurements that it was _ever_ told about, in cases where it has to prune the tree to support a partial reset. The tree would only be pruned if doing so is an accurate reflection of the sub-tree's TCB being reset from some known starting value. _Not_ pruning the tree could potentially lead to all manner of unintended consequences - similarly as if a TPM's PCRs don't reset when the host reboots. We need to take those cases into account.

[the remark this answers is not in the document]
jhand2: 1. Updated certificate to include PAUSER of each TCB component
2. Added a list of opens to the bottom, including discussion of a Quote API. I would like to make this as a separate PR so we can discuss it individually.

runtime_firmware.md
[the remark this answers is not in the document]
jhand2: Put an open at the bottom regarding PCR clearing. I think a follow-up PR describing a quote API (using non-clearable PCRs) may help address this.

runtime/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+Caliptra Runtime firmware accepts impactless updates which will update
+Caliptra’s firmware without resetting other cores in the SoC.
+
+### Applying Updates
+
+A Runtime Firmware update is triggered by the `CALIPTRA_FW_LOAD` command. Upon
+receiving this command, Runtime Firmware will:
+
+1. Write-lock mailbox
```
varuns-nvidia, line 247: There'll be more to do this because of multi-pass -- LMS will make the payload larger than the mailbox. But ok to tackle in a subsequent PR

runtime/README.md
```diff
@@ -0,0 +1,488 @@
...
+* When invoked via JTAG, ROM will write the certificate to the mailbox so it can
+  be read out over JTAG
+
+Table: `GET_LDEV_CERT` output arguments
+
+| **Name**  | **Type**      | **Description**
+| --------  | --------      | ---------------
+| data      | u8[...]       | DER-encoded LDevID Certificate
+
+### ECDSA384\_SIGNATURE\_VERIFY
```
varuns-nvidia, line 147: need an open on whether this should take the hash from the SHA block in accordance with https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#soc-fw-authentication-flows

runtime/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+Table: `INVOKE_DPE_COMMAND` output arguments
+
+| **Name**    | **Type**      | **Description**
+| --------    | --------      | ---------------
+| crc         | u32           | CRC over other output arguments, computed by Caliptra. Little endian.
+| result      | u32           | Result code. Little endian.
+| data        | u8[...]       | DPE response structure as defined in the DPE iRoT profile.
+
+## CRC
```
varuns-nvidia, line 216: Since this is not the definition of a CRC, should we call it checksum instead or use CRC32?

runtime/README.md
```diff
@@ -0,0 +1,488 @@
...
+### GET\_IDEV\_CSR
+
+ROM exposes a command to get a self-signed IDEVID CSR.
+GET\_IDEV\_CSR is not exposed by runtime firmware.
+
+Command Code: `0x4944_4556` ("IDEV")
+
+`GET_IDEV_CSR` takes no input arguments.
+
+* When invoked via JTAG, ROM will write the CSR to the mailbox so it can
```
varuns-nvidia, line 120: Is this statement meaningful? The outcome is the same whether using JTAG or not.

runtime/README.md
```diff
@@ -0,0 +1,488 @@
...
+|                                | Type        | 4-byte TYPE field of TCI node
+|                                | VendorInfo  | Locality of the caller (analog for PAUSER)
+
+\*MultiTcbInfo ontains one TcbInfo for each TCI Node in the path from the
+current TCI Node to the root. Max of 24.
+
+# Opens
+
+The following items are still under discussion
+
```
varuns-nvidia, line 483: Summarizing my requested opens since I pressed the wrong github button:
* Firmware authentication vs just ECDSA verify as the API
* Handling of impactless updates when firmware + manifest size exceeds mailbox size
* Resource usage https://github.com/chipsalliance/caliptra-sw/issues/17

varuns-nvidia (changes requested):
Main thing I'd like fixed before approval is the CRC terminology, because once a word is used it's hard to un-use it

runtime/README.md
jhand2: Makes sense. Changed to Checksum.

runtime/README.md
jhand2: Added an open at the bottom

runtime/README.md
jhand2: Added to opens

runtime/README.md
jhand2: Covered these in opens

runtime/README.md
jhand2: Agreed, removed.

varuns-nvidia (approved):

andreslagarcavilla (approved):

andreslagarcavilla:
Supportive of launch and iterate on this spec, at this point.

varuns-nvidia (approved):

PR #28, opened 2023-03-09
Check for clippy warnings in CI workflow, and fix the warnings

korran:
Fixes hundreds of warnings across 70+ files.

New versions of clippy may introduce new warnings. To prevent these new warnings from blocking CI when GitHub updates the rust installation in the runner, update to a specific version of the    Rust toolchain near the beginning of the workflow.


.github/workflows/build-test.yml
```diff
@@ -95,6 +99,12 @@ jobs:
           (cd fmc && cargo make build)
           sccache --show-stats
 
+      # Clippy needs to build crates as part of the check, so do it after the
+      # build.
+      - name: Clippy lint check
+        run: |
+          cargo clippy -- -D warnings
```
rkr35, line 106: Change to `cargo clippy --all-targets -- -D warnings` to pick up tests

sw-emulator/lib/periph/src/hmac_sha384.rs
```diff
@@ -448,10 +448,8 @@ impl HmacSha384 {
             ),
         };
 
-        if key != None {
-            self.key
-                .data_mut()
-                .copy_from_slice(&key.unwrap()[..HMAC_KEY_SIZE]);
+        if let Some(ref key) = key {
```
rkr35, line 451: Don't need the `ref`

sw-emulator/lib/periph/src/hmac_sha384.rs
```diff
@@ -484,8 +482,8 @@ impl HmacSha384 {
             ),
         };
 
-        if data != None {
-            self.format_block(&data.unwrap());
+        if let Some(ref data) = data {
+            self.format_block(data);
```
rkr35, line 486: nit: instead of `ref`, we could borrow at the argument
`self.format_block(&data)`

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -870,14 +870,14 @@ impl SocRegistersImpl {
 
         for _ in (0..n).step_by(core::mem::size_of::<u32>()) {
             let buf = self.mailbox.read_dataout().unwrap();
-            file.write(&buf.to_le_bytes()).unwrap();
+            file.write_all(&buf.to_le_bytes()).unwrap();
```
rkr35, line 873: Not for this PR, but this function is `unwrap()` heavy.

.github/workflows/build-test.yml
korran: Done

sw-emulator/lib/periph/src/soc_reg.rs
korran: I'm not convinced this function should even exist.

sw-emulator/lib/periph/src/hmac_sha384.rs
korran: Done

sw-emulator/lib/periph/src/hmac_sha384.rs
korran: Done

rkr35 (approved):

PR #33, opened 2023-03-09
Add firmware upload  and trace enablement functionality to the HwModel trait

rusty1968:
The intent of this commit is to extend the functionality exposed by the HwModel trait.

1. Make the basic flow of uploading  Caliptra firmware into Mailbox SRAM buffer available to test suites in the caliptra_hw_model crate. 
2.  Expose the trace enablement interface.


hw-model/src/model_emulated.rs
```diff
@@ -91,4 +91,8 @@ impl crate::HwModel for ModelEmulated {
 
         &mut self.output
     }
+
+    fn tracing_hint(&mut self, _enable: bool) {
```
korran, line 95: These tracing changes aren't mentioned in the PR description. Perhaps you intend this to be a separate PR?

hw-model/src/lib.rs
```diff
@@ -216,4 +281,46 @@ mod tests {
             "expected output \"ha\", was \"hi\""
         );
     }
+
+    #[test]
+    pub fn test_upload_firmware() {
+        let firmware: Vec<u8> = [
```
korran, line 287: Can you add a comment describing what this firmware does?

korran: Oh wait, this doesn't appear to be used

rusty1968: This is a test routine for the upload_firmware method. I can add the comment.

hw-model/src/model_emulated.rs
rusty1968: Sure.

hw-model/src/lib.rs
korran: Oh, sorry, this is separate from the ROM used above, I'm crazy.

korran (approved):

korran (approved):

PR #37, opened 2023-03-10
[feat] Adding nonce input argument to Ecc384 firmware api

mhatrevi:
This change adds the capability to the Ecc384 firmware api to ingest a nonce.
This corresponds to the similar functionality added to the  ECC384 RTL crypto block.


sw-emulator/lib/crypto/src/ecc384.rs
```diff
@@ -146,14 +146,21 @@ impl Ecc384 {
...
         // Seed is received as a list of big-endian DWORDs. Changing them to little-endian.
         let mut seed_reversed = *seed;
         seed_reversed.to_little_endian();
 
-        let mut drbg = HmacDrbg::<Sha384>::new(&[], &[], &seed_reversed);
+        // Nonce is received as a list of big-endian DWORDs. Changing them to little-endian.
+        let mut nonce_reversed = *nonce;
+        nonce_reversed.to_little_endian();
+
+        let mut drbg = HmacDrbg::<Sha384>::new(&[], &nonce_reversed, &seed_reversed);
```
bluegate010, line 163: I'm not sure the argument order here is correct. The [implementation](https://docs.rs/rfc6979/0.3.0/src/rfc6979/lib.rs.html#77) indicates that the arguments are `entropy_input`, `nonce`, and `additional_data`. That third name is actually incorrect and is a bug in the rfc6979 library. Section 10.1.2.3 of [SP 800-90A](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf) describes the instantiation function and shows that this argument should actually be called `personalization_string`. `additional_data` is something provided during bit generation, so should be provided to the `fill_bytes` function of this library.

Since the instantiation function just concatenates all these together (`seed_material = entropy_input || nonce || personalization_string`), effectively this call here is using the provided nonce as the `entropy_input`, and the provided seed as the `nonce`, with no `personalization_string`. Prior to this PR the correct thing was actually being done under the hood; since no `entropy_input` or `nonce` was provided, the concatenation just ended up using the provided seed as `entropy_input`.

All in all, this should probably be `HmacDrbg::<Sha384>::new(&seed_reversed, &nonce_reversed, &[]);`

bluegate010: Fixed the name in the dependency: https://github.com/RustCrypto/signatures/pull/679

PR #41, opened 2023-03-13
[fix] X509 Rename, Perf & Panic related fixes

vsonims:
This commit contains various fixes for improve the performance, code size and panic removal (ROM must be panic free). Following are the fixes part of this commit:
 - Rename Device Serial Number to UEID to align with TCG
 - Rename Cert & CSR structs to indicate they are `To Be Signed` (TBS)
 - Rename TCB_INFO_FMC_HASH to TCB_INFO_FMC_TCI
 - Rename TCB_INFO_FMC_CONFIG to TCB_OWNER_PK_HASH
 - Convert Parameters to references to reduce number of compiler generated copies and code size
 - Refactoring to make Certificate Builder panic free
 - Refactoring to reduce Certificate Builder code size
 - Clippy cleanup


x509/build/build.rs
```diff
@@ -64,19 +64,19 @@ fn gen_fmc_alias_cert(out_dir: &str) {
...
-                    name: "TCB_INFO_FMC_HASH",
+                    name: "TCB_INFO_FMC_TCI",
                     fwid: Fwid {
                         hash_alg: asn1::oid!(/*sha384*/ 2, 16, 840, 1, 101, 3, 4, 2, 2),
                         digest: &[0xCD; 48],
                     },
                 },
                 FwidParam {
-                    name: "TCB_INFO_FMC_CONFIG",
+                    name: "TCB_INFO_OWNER_PK_HASH",
```
jhand2, line 79: The current spec says something about measuring security fuse state into FMC alias, although it's not super specific. Not sure if there are things other than FW hash and owner pk.

jhand2 (approved):

PR #49, opened 2023-03-14
Replace drivers/test-fw/test.sh with HwModel tests that can be run with "cargo test"

korran:
These tests can also be executed inside verilator with (cd drivers && cargo test --features=verilator).


builder/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+                "Process {:?} {:?} exited with status code {:?}",
+                cmd.get_program(),
+                cmd.get_args(),
+                status.code()
+            ),
+        ))
+    }
+}
+
+pub fn build_firmware_elf(fw_crate_name: &str, bin_name: &str) -> io::Result<Vec<u8>> {
```
rusty1968, line 31: Is this function tailored for test rom only or production rom as well?

builder/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+    fs::read(
+        Path::new(WORKSPACE_DIR)
+            .join("target")
+            .join(TARGET)
+            .join(PROFILE)
+            .join(bin_name),
+    )
+}
+
+pub fn build_firmware_rom(fw_crate_name: &str, bin_name: &str) -> io::Result<Vec<u8>> {
```
rusty1968, line 59: Is this function tailored for  the test harness ROM  or any ROM?

korran: This function takes any crate that has sections targeting the ROM address space and converts into a raw binary of the ROM. You could use it with the "real ROM" or with one of the test harness ROMs.

We'll need a similar function for the FMC, which ultimately will involve some code-signing and end up generating the fw blob that goes into the mailbox.

rusty1968 (dismissed):

PR #55, opened 2023-03-15
Add Caliptra mutable code crates (runtime, mutable)

rusty1968:
This crate contains types, macros and definitions that are common across FMC and Runtime firmware.

1.  Startup code.
2. Exception handling infra-structure.
3. NMI handling.


Cargo.toml
```diff
@@ -12,10 +12,12 @@ members = [
   "drivers",
   "drivers/test-fw",
   "fmc",
+  "fmc/test-rom",
   "hw-latest/verilated",
   "hw-model",
   "registers",
   "registers/bin/generator",
+  "rt",
```
jhand2, line 20: Is the idea that this would be shared between fmc and runtime firmware? Might be good to call this something else since there will also be a separate crate which implements runtime firmware.

rusty1968: You are correct. What should we call it? I am very bad with names.

FerralCoder: we also briefly discussed earlier in the ROM/FMC standup call that maybe it makes more sense to rename @jhand2's module to something else like "App FW" instead of "Runtime FW" since the "rt" nomenclature is already used in rust developmnent. But if not, I suggest calling this one "Common" or something along those lines. I am fine with either direction, what do you think @jhand2?

rt/src/hand_off.rs
```diff
@@ -0,0 +1,51 @@
+// Licensed under the Apache-2.0 license.
+
+const FHT_MAGIC: u32 = 0x54484643;
+#[repr(C)]
+#[derive(Clone, Debug, Default)]
+pub struct HandoffParams {
```
FerralCoder: call it FirmwareHandoffTable, stay consistent with spec

Same for all the fields, keep the naming in line with what is in the spec so there is no confusion. Its fine that you changed from camel case to snake case but don't deviate on the actual text

If there is a reason to change a name, it should be done here and in the spec simultaneously, as part of the same PR

rt/src/hand_off.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+impl HandoffParams {
+    pub fn is_valid(&self) -> bool {
+        self.magic == FHT_MAGIC
+    }
+}
+
+impl Default for HandoffParams {
+    // TODO : replace placholder with actual values.
```
FerralCoder: values will be populated by ROM/FMC as they set things up. There should not be a set of defaults already built in. The Indexes should start as 0xFF to indicate they are not valid (use something like INVALID_IDX = 0xFF) and all others should be set to 0. Nothing else should be assumed by FMC or Runtime FW

Cargo.toml
rusty1968: @jhand2
Jordan , I renamed it firmware crate.

FerralCoder (changes requested):

Cargo.toml
jhand2: Either way is good with me (either renaming Runtime FW -> Application FW, or renaming this crate to just be named "firmware")

rusty1968: I extracted the hand_off into a common crate and can rename the riscv-rt infrastructure support stuff to cpu , since the riscv stuff is not going to be used by rom. 

@jhand2  , would you like me to create an rt crate shell  as well ?

jhand2: > would you like me to create an rt crate shell as well ?

Ya, if you don't mind that would be great. Thanks!

rusty1968: I have added the runtime crate.

rt/src/hand_off.rs
rusty1968: This data structure has been moved to the common crate.

rt/src/hand_off.rs
rusty1968: This data structure now resides in the common crate in main branch.

jhand2 (approved):

drivers/test-fw/src/bin/error_reporter_tests.rs
```diff
@@ -15,17 +15,14 @@ Abstract:
...
-};
+use caliptra_lib::ErrorReporter;
 use caliptra_registers::soc_ifc;
 
 mod harness;
 
 fn test_report_fw_error() {
     let v: u32 = 0xdead0;
-    report_fw_error_non_fatal(0xdead0);
+    ErrorReporter::default().report_fw_error_non_fatal(0xdead0);
```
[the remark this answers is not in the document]
rusty1968: I will just revert my changes in the error reporter and discuss this in the wider audience. Currently all APIs follow the model above except for three modules. Associated functions versus modules. Which one should we pick?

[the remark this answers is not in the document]
jhand2: Ack. I don't feel strongly either way, but agreed we can do this in a follow-up PR.

FerralCoder (approved):

drivers/test-fw/src/bin/error_reporter_tests.rs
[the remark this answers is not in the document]
rusty1968: @jhand2  Jordan, can you merge? I don't have permissions.

PR #69, opened 2023-03-17
Move firmware upload to emulator application logic

rusty1968:
The intent of this commit is to create a callback mechanism so applications using the emulator library can customize the handling of the ready_for_fw event.

Validated with rom-dev branch using make run.

### DIFF
--- a/sw-emulator/app/src/main.rs
+++ b/sw-emulator/app/src/main.rs
@@ -14,7 +14,9 @@ Abstract:
 
 use caliptra_emu_bus::Clock;
 use caliptra_emu_cpu::{Cpu, RvInstr, StepAction};
-use caliptra_emu_periph::{CaliptraRootBus, CaliptraRootBusArgs, TbServicesCb};
+use caliptra_emu_periph::{
+    CaliptraRootBus, CaliptraRootBusArgs, Mailbox, ReadyForFwCb, TbServicesCb,
+};
 use clap::{arg, value_parser, ArgAction};
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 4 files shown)


sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -1142,7 +1106,11 @@ impl Bus for SocRegistersImpl {
     /// Called by Bus::poll() to indicate that time has passed
     fn poll(&mut self) {
         if self.timer.fired(&mut self.op_fw_write_complete_action) {
-            self.upload_firmware().unwrap();
+            (self.ready_for_fw_cb)(&mut self.mailbox, &self.firmware);
```
korran, line 1109: self.firmware? Should that come from the app?

sw-emulator/app/src/main.rs
```diff
@@ -145,6 +150,41 @@ fn main() -> io::Result<()> {
             0xFF => exit(0x00),
             _ => print!("{}", val as char),
         }),
+        ready_for_fw_cb: ReadyForFwCb::new(move |mailbox: &mut Mailbox, firmware: &Vec<u8>| {
```
korran, line 153: Prefer &[u8] to &Vec[u8]

sw-emulator/lib/periph/src/soc_reg.rs
rusty1968: it is passed as an argument by the app.

sw-emulator/app/src/main.rs
rusty1968: ok - will change.

sw-emulator/lib/periph/src/soc_reg.rs
rusty1968: Now it is coming from the app environment - removed the argument passing to the model

sw-emulator/app/src/main.rs
rusty1968: Done

korran (approved):

PR #74, opened 2023-03-20
[feat] Basic framework for triggering and scheduling reset actions in the emulator

mhatrevi:
This change adds the following features to the emulator:
	1. Uses soc register::generic_output_wires_1 to trigger a warm reset flow in the emulator.
	2. Adds a 'timer action' attribute to the timer callbacks for triggering different actions. Currently, poll, warm reset and update reset are the supported timer actions.

	TODO Items:
	1. Merge the warm and update reset callbacks into a single reset callback with reset reason as the argument.
	2. Reset the appropriate registers on various peripherals on their respective reset callbacks.

### DIFF
--- a/sw-emulator/lib/periph/src/soc_reg.rs
+++ b/sw-emulator/lib/periph/src/soc_reg.rs
@@ -17,7 +17,7 @@ use crate::{CaliptraRootBusArgs, Iccm, Mailbox};
 use caliptra_emu_bus::BusError::{LoadAccessFault, StoreAccessFault};
 use caliptra_emu_bus::{
     Bus, BusError, Clock, ReadOnlyMemory, ReadOnlyRegister, ReadWriteMemory, ReadWriteRegister,
-    Register, Timer, TimerAction,
+    Register, Timer, TimerAction, TimerActionType,
 };
 use caliptra_emu_types::{RvAddr, RvData, RvSize};
 use std::cell::RefCell;
@@ -31,6 +31,7 @@ use tock_registers::registers::InMemoryRegister;
 // Second parameter is schedule(ticks_from_now: u64, cb: Box<dyn FnOnce(&mut
 // Mailbox)>), which is called to schedule firmware writing in the future
 type ReadyForFwCallback = Box<dyn FnMut(ReadyForFwCbArgs)>;
+type UploadUpdateFwCallback = Box<dyn FnMut(&mut Mailbox)>;
 
 /// CPTRA_HW_ERROR_FATAL Register Start Address
 const CPTRA_HW_ERROR_FATAL_START: u32 = 0x0;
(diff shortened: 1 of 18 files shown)


sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -1102,7 +1136,15 @@ impl Bus for SocRegistersImpl {
             }
 
             INTERNAL_FW_UPDATE_RESET_START..=INTERNAL_FW_UPDATE_RESET_END => {
-                self.internal_fw_update_reset.write(size, val)
+                self.internal_fw_update_reset.write(size, val)?;
+
+                // Schedule a firmware update reset timer action.
+                self.op_reset_trigger_action = Some(
```
korran, line 1142: Does the Verilog code also do this? If not, I'd prefer to move this hacky "special case" logic out of the emulator library to somewhere else, as we might accidentally rely on it in production code.

mhatrevi: A write to internal_fw_update_reset will trigger an update reset in Caliptra; this is per documented behavior.

mhatrevi: This is specific to the emulator. Other option is to define a emulator specific register to trigger the warm reset.

rom/dev/Makefile
```diff
@@ -105,7 +105,7 @@ run: build-emu build-fw-image objcopy
 		--req-idevid-csr \
 		--idevid-key-id-algo sha1 \
 		--rom $(TARGET_DIR)/caliptra-rom.bin \
-		--firmware $(TARGET_DIR)/caliptra-rom-test-fw \
+		--current-firmware $(TARGET_DIR)/caliptra-rom-test-fw \
```
vsonims, line 108: can we leave it as firmware?

rom/dev/src/lock.rs
```diff
@@ -25,13 +25,17 @@ use crate::rom_env::RomEnv;
 /// * `env` - ROM Environment
 /// * `reset_reason` - Reset reason
 pub fn lock_registers(env: &RomEnv, reset_reason: ResetReason) {
+    cprintln!("[state] Locking DataVault");
     if reset_reason == ResetReason::ColdReset {
```
vsonims, line 29: This should be part of seperate PR>

vsonims (changes requested):

sw-emulator/lib/periph/src/soc_reg.rs
mhatrevi: Disabled the warm reset flow. We'll enable it on reaching design agreement.

ajisaxena (approved):

vsonims (approved):

PR #85, opened 2023-03-21
[feat] Add device lifecycle state to emulator

mhatrevi:


sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -679,6 +681,20 @@ impl SocRegistersImpl {
         }
     }
 
+    fn set_cptra_security_state_device_lifecycle(&mut self, args: &CaliptraRootBusArgs) {
+        let mut value = SecurityState::LIFE_CYCLE::UNKNOWN;
```
vsonims, line 685: Can we make unprovisioned default instead of unknown

vsonims (changes requested):

sw-emulator/lib/periph/src/soc_reg.rs
mhatrevi: Updated

vsonims (approved):

PR #93, opened 2023-03-22
Refactoring code and adding FHT load

rusty1968:


fmc/memory.x
```diff
@@ -2,11 +2,15 @@ ROM_SIZE  = 32752;
 META_SIZE = 16;
 ICCM_SIZE = 128K;
 DCCM_SIZE = 128K;
+DATA_SIZE = 100K;
 
-ROM_ORG  = 0x00000000;
-META_ORG = ROM_SIZE;
-ICCM_ORG = 0x40000000;
-DCCM_ORG = 0x50000000;
+ROM_ORG   = 0x00000000;
```
FerralCoder, line 7: is there a way we can make these definitions public/shared so that we don't have to duplicate them for each module? I see this as being something that may be missed during future maintenance, and not having the memory regions defined properly will result in nasty bugs.

FerralCoder (approved):

JohnTraverAmd (approved):

PR #95, opened 2023-03-22
Unify pre-rust startup code, NMI/Exception entry point  for FMC and RT

rusty1968:
This commit includes the following:

- Consolidates pre-rust code (trap, nmi, startup) and linker descriptor file for FMC and RT in the cpu crate.
- Loads and checks the FHT in both FMC and RT binaries.
- Adds a Makefile for FMC to automate the development flow :  building the FMC and RT binary crates, create the firmware image and check FMC can transition to RT.


jhand2 (approved):

rusty1968:
Why github drops approvals if we make any changes?

FerralCoder (dismissed):

jhand2:
> Why github drops approvals if we make any changes?

I think that's a Github setting to ensure that what people approved is what gets merged.

Feel free to tag me when you're ready for another round of review and I'll take another look.

rusty1968:
@jhand2 , ready for another round of review. Vishal recommended we share the minimum amount of code possible with ROM, so I removed the rental cell and the environment type from the common crate.

cpu/build.sh
korran, line 1: Is this file used?

korran (approved):

PR #102, opened 2023-03-25
Mark KeyUsage and DICE extensions as critical

jhand2:


x509/src/fmc_alias_cert.rs
```diff
@@ -17,19 +17,26 @@ include!(concat!(env!("OUT_DIR"), "/fmc_alias_cert_tbs.rs"));
 
 #[cfg(all(test, target_os = "linux"))]
 mod tests {
+
     use openssl::ecdsa::EcdsaSig;
```
vsonims, line 21: Since the keyusage extension is used in idevid and ldevid can we update those tests to make sure critical bit is set

vsonims (changes requested):

x509/src/fmc_alias_cert.rs
jhand2: Done

jhand2:
@vsonims anything else open here?

jhand2:
Closing this. Going to open a new PR which only changes KeyUsage (and adds tests). Need add more nuance in how we handle DICE extensions.

PR #105, opened 2023-03-27
Add ability to boot runtime firmware crate on Caliptra emulator

jhand2:


runtime/Makefile
```diff
@@ -0,0 +1,152 @@
...
+default: build
+
+build:
+	$(RUSTFLAGS) cargo build \
+		--profile firmware \
+		--target=riscv32imc-unknown-none-elf \
+		--no-default-features \
+		--features riscv
+
+build-emu:
```
korran, line 30: I'm skeptical this target does anything useful...

korran (approved):

runtime/Makefile
```diff
@@ -0,0 +1,152 @@
+##
```
korran, line 1: Do you actually need this? Doesn't "make run" from the fmc directory already build the app firmware?

jhand2: It does, but the FMC doesn't boot into RT right now. So I just wanted to replicate that Makefile here and boot directly into RT.

Maybe it would make more sense to

* Move fmc/Makefile to the root
* Add a target to boot into runtime directly

runtime/Makefile
jhand2: Ya you're right, cargo run should just build it if necessary

runtime/Makefile
korran: What am I missing here? Isn't this the runtime firmware?

```
fmc$ make run
<snip>
 / ___|__ _| (_)_ __ | |_ _ __ __ _  |  _ \_   _|
| |   / _` | | | '_ \| __| '__/ _` | | |_) || |
| |__| (_| | | | |_) | |_| | | (_| | |  _ < | |
 \____\__,_|_|_| .__/ \__|_|  \__,_| |_| \_\|_|
               |_|

[rt] FHT Marker: 0x54484643
<snip>
fmc$
```

korran: Also, "make run" isn't very useful, since you can't manipulate the mailbox there. Most testing should be done against the HwModel from Rust.

jhand2: Oh you're right. For whatever reason it wasn't working for me yesterday. Ok, will ditch this and just use the fmc makefile

PR #114, opened 2023-03-29
emu: ready_for_fw_cb should be called immediately.

korran:
Previously, special-case logic in the emulator library locked the mailbox, then called ready_for_fw_cb 1000 cycles later, with no way for users to opt-out of this logic.

In this commit, ready_for_fw_cb is called immediately, and provided an API for scheduling mailbox actions in the future. The emulator app uses this API to maintain the previous behavior, but other users don't have to.


rusty1968 (approved):

rusty1968 (approved):

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -804,15 +807,27 @@ impl SocRegistersImpl {
         if size != RvSize::Word {
             Err(BusError::StoreAccessFault)?
         }
+        println!("on_write_flow_status");
```
rusty1968, line 810: Was this for debug only or we want to keep this?

korran: Whoops, thanks for noticing!

rusty1968 (approved):

PR #117, opened 2023-03-29
[fix] Invalid FHT address fields should have the 0xFFFF_FFFF value.

rusty1968:
Invalid FHT address fields should have the 0xFFFF_FFFF value. This aligns to the how invalid address values  are expressed by ROM code (see fht::make_fht).

- Updated FMC specification to reflect the change above.
- Updated Default trait implementation of FHT table to the change above.

### DIFF
--- a/common/src/hand_off.rs
+++ b/common/src/hand_off.rs
@@ -3,11 +3,12 @@
 use zerocopy::{AsBytes, FromBytes};
 
 pub const FHT_MARKER: u32 = 0x54484643;
-pub const FHT_INVALID_IDX: u8 = 0xff;
+pub const FHT_INVALID_IDX: u8 = u8::MAX;
+pub const FHT_INVALID_ADDRESS: u32 = u32::MAX;
 
-/// The Firmware Handoff Table is a data structure that is resident
-/// at a well-known location in DCCM. It is initially populated by ROM and modified by FMC as a way to pass parameters and configuration information
-/// from one firmware layer to the next.
+/// The Firmware Handoff Table is a data structure that is resident at a well-known
+/// location in DCCM. It is initially populated by ROM and modified by FMC as a way
+/// to pass parameters and configuration information from one firmware layer to the next.
 #[repr(C)]
 #[derive(Clone, Debug, AsBytes, FromBytes)]
 pub struct FirmwareHandoffTable {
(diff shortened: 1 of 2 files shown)


JohnTraverAmd (dismissed):

common/src/hand_off.rs
```diff
@@ -120,9 +121,21 @@ impl Default for FirmwareHandoffTable {
...
-        self.fht_marker == FHT_MARKER && self.fmc_cdi_kv_idx != FHT_INVALID_IDX
+        self.fht_marker == FHT_MARKER
+            && self.fmc_cdi_kv_idx != FHT_INVALID_IDX
+            && self.manifest_load_addr != FHT_INVALID_ADDRESS
+            && self.fmc_pub_key_x_dv_idx != FHT_INVALID_IDX
+            && self.fmc_pub_key_y_dv_idx != FHT_INVALID_IDX
+            && self.fmc_cert_sig_r_dv_idx != FHT_INVALID_IDX
+            && self.fmc_cert_sig_s_dv_idx != FHT_INVALID_IDX
+            && self.rt_tci_dv_idx != FHT_INVALID_IDX
+            && self.rt_fw_entry_point != FHT_INVALID_ADDRESS
```
FerralCoder, line 135: please add additional checks....

1. self.rt_fw_load_addr != FHT_INVALID_ADDRESS
2. self.fips_fw_load_addr == FHT_INVALID_ADDRESS (since we don't yet support anything else)
3. anything else that ROM is required to fill per the FMC spec

common/src/hand_off.rs
```diff
@@ -120,9 +121,21 @@ impl Default for FirmwareHandoffTable {
 
 impl FirmwareHandoffTable {
     /// Perform valdity check of the table's data.
+    /// The fields below should have been populated by ROM with
+    /// valid data before it transfers control to mutable code.
     pub fn is_valid(&self) -> bool {
```
FerralCoder, line 126: should rename to is_valid_for_fmc(), since the FMC requirements are ot the same as RT requirements.

Then create new routine, is_valid_for_rt() than covers what is needed there.

FerralCoder (changes requested):

common/src/hand_off.rs
rusty1968: As of now , the is_valid refers to values populated by ROM before it exits as per  comments , but we can extend its functionality.

common/src/hand_off.rs
FerralCoder: Should also make sure the RT entry point is inside the locked ICCM memory, and that attacker is not somehow trying to get us to jump to an address in DCCM

common/src/hand_off.rs
rusty1968: As per conversation, with @FerralCoder we will keep "is_valid" within the scope of  checking if entries populated by ROM are not invalid indexes or addresses (as per agreed markers) and add more extensive check in the mutable code.

common/src/hand_off.rs
rusty1968: As per discussion with @FerralCoder  , we will add more extensive checks in the FMC and RT  and keep the scope of is_valid as light as possible , checking if fields have been populated by ROM to avoid feature creep inside the common crate.

FerralCoder (approved):

PR #118, opened 2023-03-30
Add panic detection tests to FMC and RT crates.

rusty1968:
This commit adds adds tests to prevent code that panics to be introduced in the mutable code modules. This ensures we are producing robust panic-free code.

### DIFF
--- a/fmc/src/main.rs
+++ b/fmc/src/main.rs
@@ -13,11 +13,13 @@ Abstract:
 --*/
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), no_main)]
+use core::hint::black_box;
 
 #[cfg(feature = "riscv")]
 core::arch::global_asm!(include_str!("transfer_control.S"));
 
 use caliptra_common::{cprintln, FirmwareHandoffTable};
+use caliptra_drivers::report_fw_error_non_fatal;
 pub mod fmc_env;
 pub mod fmc_env_cell;
 
@@ -71,11 +73,7 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
         trap_record.mscause,
         trap_record.mepc
     );
-
-    // Signal non-fatal error to SOC
-    caliptra_drivers::report_fw_error_non_fatal(0xdead0);
-
-    loop {}
+    report_error(0xdead);
 }
 
 #[no_mangle]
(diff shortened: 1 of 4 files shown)


fmc/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    error.rs
+
+Abstract:
+
+    File contains API and macros used by the ROM for error handling
```
JohnTraverAmd, line 11: Does this need to change to be the FMC?

rusty1968: Comments updated.

fmc/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+Abstract:
+
+    File contains API and macros used by the FMC for error handling
+
+--*/
+
+/// Caliptra Component
+pub enum FmcComponent {
+    /// Global Error
+    Global = 0x103,
```
korran, line 18: This might collide with the error space on the ROM?

We need to talk to Vishal during the next ROM standup about error namespace coordination.

rusty1968: Let's use 0x203 for FMC and 0x303 for RT.

rusty1968: Good hook to have a user guide with low level details for caliptra developers.

rusty1968: Error codes changed.

fmc/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+        #[derive(Debug, Copy, Clone, Eq, PartialEq)]
+        #[allow(clippy::enum_variant_names)]
+        pub enum $enum_name {
+            $($field_name = $field_val,)*
+        }
+
+        impl From<$enum_name> for u32 {
+            fn from(val: $enum_name) -> Self {
+                ((($crate::error::FmcComponent::$comp_name) as Self) << 24) | (val as Self)
```
korran, line 33: Doesn't `0x201 << 24` overflow a 32-bit number, making this 0x02000000, which colides with `0x101 << 24` in the ROM? @vsonims?

korran: I think we need to have a discussion about create a canonical registry of error codes. Why not just have a central crate with ALL the caliptra error codes, that can be referenced from test cases and caliptra clients? (including a Debug impl when compiled with the std feature)

rusty1968: I can move this code to common. Let me tweak that.

rusty1968: I trimmed the PR to just detect panic - we need to propose a collision-free partition for error namespace with Vishal and make it common.

korran (approved):

PR #122, opened 2023-03-30
[FMC][ROM] FHT.rt_fw_entry_point field should be the entry index into the DV

rusty1968:
FMC specification is defining FHT.rt_fw_entry_point as the physical address of the entry point of Runtime FW Module in ICCM SRAM, however ROM is already making this information available in the data vault.

The proposal is to pass RT entry point field as the entry index instead of a physical address.


common/src/hand_off.rs
```diff
@@ -31,14 +31,14 @@ pub struct FirmwareHandoffTable {
...
-    pub rt_fw_load_addr: u32,
-
-    /// Entry point of Runtime FW Module in ICCM SRAM.
-    pub rt_fw_entry_point: u32,
+    pub rt_fw_load_addr: u8,
 
     /// Index of FMC TCI value in the Data Vault.
     pub fmc_tci_dv_idx: u8,
 
+    /// Entry point of Runtime FW Module in ICCM SRAM.
```
FerralCoder, line 39: The order here is odd. Either move this back to where the other address are, or group with all the other RT fields. Either way, make this and the doc match.

If you did this for alignment, we can also add another reserved field after the group of address _idx fields. If you make the FIPS address field the last of those, we would have some room if we find that we need more data once that feature is supported.

fmc/README.md
```diff
@@ -113,8 +113,8 @@ fields may not be changed or removed). Table revisions with different Major Vers
 | fht_minor_ver         | 2            | ROM, FMC   | Minor version of FHT. Initially written by ROM but may be changed to a higher version by FMC.            |
 | manifest_load_addr    | 4            | ROM        | Physical base address of Manifest in DCCM SRAM.                                                          |
 | fips_fw_load_addr     | 4            | ROM        | Physical base address of FIPS Module in ROM or ICCM SRAM. May be 0xFFFF_FFFF if there is no discrete module.    |
-| rt_fw_load_addr       | 4            | ROM        | Physical base address of Runtime FW Module in ICCM SRAM.                                                 |
-| rt_fw_entry_point     | 4            | ROM        | Entry point of Runtime FW Module in ICCM SRAM.                                                           |
+| rt_fw_load_addr       | 4            | ROM        | Index of load address of Runtime FW Module value in data vault.SRAM.                                                 |
```
FerralCoder, line 116: be consistent in naming, since these are now index values, add the _idx suffix.

fmc/README.md
```diff
@@ -113,8 +113,8 @@ fields may not be changed or removed). Table revisions with different Major Vers
 | fht_minor_ver         | 2            | ROM, FMC   | Minor version of FHT. Initially written by ROM but may be changed to a higher version by FMC.            |
 | manifest_load_addr    | 4            | ROM        | Physical base address of Manifest in DCCM SRAM.                                                          |
 | fips_fw_load_addr     | 4            | ROM        | Physical base address of FIPS Module in ROM or ICCM SRAM. May be 0xFFFF_FFFF if there is no discrete module.    |
```
FerralCoder, line 115: I think it would make sense to change this one to an index as well, for consistency

fmc/README.md
```diff
@@ -113,8 +113,8 @@ fields may not be changed or removed). Table revisions with different Major Vers
 | fht_minor_ver         | 2            | ROM, FMC   | Minor version of FHT. Initially written by ROM but may be changed to a higher version by FMC.            |
 | manifest_load_addr    | 4            | ROM        | Physical base address of Manifest in DCCM SRAM.                                                          |
```
FerralCoder, line 114: @vsonims, would it make sense for this to also be in the DV, or should this stay as physical address?

common/src/hand_off.rs
rusty1968: This is probably rust-analyzer changing field order when I saved.

fmc/README.md
rusty1968: done

fmc/README.md
rusty1968: It is unused for gen1, so it would be a nop.

fmc/README.md
rusty1968: Let's merge with what ROM already stores in the DV and sort this out later as an issue assigned to Vishal.

JohnTraverAmd (approved):

PR #125, opened 2023-03-30
Implement caliptra_builder::build_and_sign_image.

korran:
This is used for building fmc/app-fw bundles for use in integration
tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,6 +275,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "caliptra-image-fake-keys"
+version = "0.1.0"
+dependencies = [
+ "caliptra-image-gen",
+ "caliptra-image-openssl",
+ "caliptra-image-types",
+]
+
+[[package]]
 name = "caliptra-image-gen"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,7 +110,14 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 name = "caliptra-builder"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
+ "caliptra-image-elf",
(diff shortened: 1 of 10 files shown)


rom/dev/tests/smoke_test.rs
```diff
@@ -0,0 +1,24 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_builder::{ImageOptions, APP_WITH_UART, FMC_WITH_UART, ROM_WITH_UART};
+use caliptra_hw_model::{HwModel, InitParams};
+
+#[test]
+fn smoke_test() {
```
rusty1968, line 7: Where should the integration test bench with the runtime and fmc crates reside?

rom/dev/tests/smoke_test.rs
```diff
@@ -0,0 +1,24 @@
...
+        &FMC_WITH_UART,
+        &APP_WITH_UART,
+        ImageOptions::default(),
+    )
+    .unwrap();
+    let mut hw = caliptra_hw_model::create(InitParams {
+        rom: &rom,
+        ..Default::default()
+    })
+    .unwrap();
```
rusty1968, line 19: Should we step until flow_status.ready_for_fw  is set?

korran: I want upload_firmware() to do that.

rom/dev/tests/smoke_test.rs
rusty1968: Agree. This is a very useful outcome.

korran: As there's still a lot of unimplemented stuff here, I've decided to remove this commit from the PR. Will follow up in a future PR.

rom/dev/tests/smoke_test.rs
korran: Removed from this PR.

rusty1968 (dismissed):

rusty1968 (approved):

PR #126, opened 2023-03-30
Add integration tests for runtime firmware

jhand2:
First commit moves the driver test harness to its own crate.

Second commit adds the ability to test two scenarios:

* Boot ROM + FMC, then run a self-contained test runtime image
* Boot ROM, FMC, and Runtime, then run end-to-end integration tests via
  soc registers and the mailbox.


runtime/tests/integration_tests.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// Build a runtime firmware bundle
+///
+/// If `test_bin_name` is None, build the regular runtime image
+fn build_runtime_bundle(test_bin_name: Option<&str>) -> Vec<u8> {
+    const WORKSPACE_DIR: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/..");
+    const TARGET: &str = "riscv32imc-unknown-none-elf";
+    const PROFILE: &str = "firmware";
+
+    caliptra_builder::build_firmware_elf("caliptra-fmc", "caliptra-fmc").unwrap();
```
rusty1968, line 22: Should we use FwIds here?

jhand2: Ah, that's much nicer. Done.

test-harness/src/lib.rs
```diff
@@ -42,13 +45,47 @@ pub fn _print(args: fmt::Arguments) {
...
 macro_rules! test_suite {
     ($($test_case: ident,)*) => {
         use core::arch::global_asm;
         use core::panic::PanicInfo;
-        use $crate::harness::Testable;
+        use caliptra_test_harness::{println, Testable};
 
+
+        #[cfg(feature = "rom")]
         global_asm!(include_str!("start.S"));
```
korran, line 89: Where is this file coming from?

jhand2: https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/test-fw/src/bin/start.S

But this is only for ROM. RW tests use https://github.com/chipsalliance/caliptra-sw/tree/main/cpu/src for early boot code.

Open to reorganizing this though.

korran (dismissed):

test-harness/src/lib.rs
jhand2: Moved start.S to be part of the test-harness crate

test-harness/src/start.S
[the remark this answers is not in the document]
jhand2: I was trying to avoid changing the behavior of the `drivers` tests with this PR, so I think I'd prefer to keep the existing start.S at least for now (this is just moved from `drivers/test-fw/src/bin`)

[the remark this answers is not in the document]
rusty1968: Makes sense.

rusty1968 (dismissed):

korran (changes requested):
There's something wrong with this PR with regards to UART output:

Before the PR:

```
$ cargo test -p caliptra-drivers test_sha384acc -- --nocapture
running 1 test
UART: sha384acc::test_kat...	[ok]
UART: sha384acc::test_digest_max_mailbox_size...	[ok]
UART: sha384acc::test_digest0...	[ok]
UART: sha384acc::test_digest1...	[ok]
UART: sha384acc::test_digest2...	[ok]
UART: sha384acc::test_digest_zero_size_buffer...	[ok]
* TESTCASE PASSED
```

After this PR

```
$ cargo test -p caliptra-drivers test_sha384acc -- --nocapture
running 1 test
* TESTCASE PASSED
test test_sha384acc ... ok
```

I'm not even sure if it's actually running the firmware test-cases. Please fix!

korran:
Looks to be related to the emu feature. This seems to help:

```
diff --git a/drivers/test-fw/Cargo.toml b/drivers/test-fw/Cargo.toml
index 4fd5499..1d86b51 100644
--- a/drivers/test-fw/Cargo.toml
+++ b/drivers/test-fw/Cargo.toml
@@ -14,7 +14,7 @@ caliptra-test-harness = { path = "../../test-harness" }
 cfg-if = "1.0.0"
 
 [features]
-emu = []
+emu = ["caliptra-test-harness/emu"]
 
 # This feature is used to filter all these binary targets during normal builds
 # (targets must be built with cargo arguments:
diff --git a/test-harness/Cargo.toml b/test-harness/Cargo.toml
index d7963e3..177d70b 100644
--- a/test-harness/Cargo.toml
+++ b/test-harness/Cargo.toml
@@ -13,3 +13,4 @@ runtime = []
 
 [dependencies]
 cfg-if = "1.0.0"
+caliptra-drivers = { path = "../drivers", features=["emu"] }
```

drivers/test-fw/Cargo.toml
```diff
@@ -10,17 +10,21 @@ edition = "2021"
...
 caliptra-registers = { path = "../../registers" }
 caliptra-kat = { path = "../../kat" }
+caliptra-test-harness = { path = "../../test-harness" }
 cfg-if = "1.0.0"
 
 [features]
 emu = []
 
+# Tells the test_suite macro that this is a ROM image, not a runtime image
+rom = []
```
korran: Can you explain why this is necessary?

test-harness/Cargo.toml
```diff
@@ -8,5 +8,8 @@ edition = "2021"
 [features]
 emu = []
 
+# Whether or not this is a harness for a runtime firmware image
+runtime = []
```
korran, line 12: Can you explain why this is necessary?

drivers/test-fw/Cargo.toml
jhand2: It's not, I accidentally removed it in the wrong commit. Fixed.

test-harness/Cargo.toml
jhand2: For the runtime tests I'm using the assembly files from `cpu/src` for early boot. This is the real entrypoint for runtime firmware, so I thought it was ideal to just use that.

For the driver tests, I kept it using the test `start.S` (which now lives in the test-harness crate). This is much simpler, it just sets up the memory regions and runs main.

So this feature is used to tell test-harness whether it is using the runtime early boot code. In that case, the test-harness crate needs to:

1. Set up some extra handlers
2. Doesn't need to include a start.S (it will be provided by the cpu crate)

If the caller doesn't want to use the runtime early boot code, the test-harness crate will just provide the simple `start.S`.

jhand2: It would potentially be acceptable to just make everyone use the boot code from the start.S crate. I was trying to avoid changing the behavior of the driver tests by keeping the same boot code, but I'm not sure it matters much.

korran: Are you booting to your tests through the ROM/FMC?

jhand2: Ya

korran: For any tests that you would want to use to debug the RTL (like the drivers), it's best that we keep the startup code as simple as possible, so the tests run in a reasonable amount of time, and the waveforms (needed to debug the RTL) don't get too big.

I might try doing a pass at some point to unify all this code, but it won't be using the normal boot ROM, instead just jumping straight to ICCM with no validation or anything.

korran (approved):

test-harness/Cargo.toml
jhand2: Sounds good. Thanks.

korran (approved):

Cargo.lock
```diff
@@ -78,9 +78,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.0.2"
+version = "2.1.0"
```
korran, line 81: Please keep Cargo.lock changes to the minimum required to implement the feature. I'm guessing you called `cargo update` by mistake?

korran (changes requested):

Cargo.lock
jhand2: Fixed

korran (approved):

PR #133, opened 2023-04-01
Handoff module of FMC

rusty1968:
Introduces a custom type to encapsulate knowledge of FHT fields as well as implement handoff behavior of FMC :
       - Retrieves FHT table from fixed address in DCCM.
       - Perform validity checks of FHT data.
       -  Retrieves information stored  in the data vault by ROM.
       - Transfers control to the runtime firmware.


fmc/src/hand_off.rs
```diff
@@ -0,0 +1,97 @@
...
+        fht.rt_fw_entry_point_idx
+    );
+    cprintln!(
+        "[fmc] FHT RT Entry Point: 0x{:08x}",
+        fht.rt_fw_load_addr_idx
+    );
+}
+
+impl HandOff {
+    pub fn from_rom() -> Option<HandOff> {
```
jhand2, line 65: It looks like RT can also use this function. Maybe `from_previous`? Or if there are any differences, could have a from_rom and from_fmc

jhand2 (approved):

jhand2 (approved):

PR #136, opened 2023-04-04
[feat] ROM integration tests - Header Verification

mhatrevi:


builder/src/lib.rs
```diff
@@ -209,6 +214,10 @@ pub fn build_and_sign_image(fmc: &FwId, app: &FwId, opts: ImageOptions) -> anyho
         vendor_config: opts.vendor_config,
         owner_config: opts.owner_config,
     })?;
+    Ok(image)
+}
+
+pub fn sign_image(image: &ImageBundle) -> anyhow::Result<Vec<u8>> {
```
korran, line 220: I don't think this function is named correctly; I don't think it's actually doing any signing...

hw-model/src/lib.rs
```diff
@@ -137,6 +137,36 @@ pub trait HwModel {
         }
     }
 
+    fn copy_output_until_fatal_error(
+        &mut self,
+        code: u32,
```
korran, line 142: call this expected_fw_error (or expected_fw_error_nonfatal)?

hw-model/src/lib.rs
```diff
@@ -137,6 +137,36 @@ pub trait HwModel {
...
+        mut w: impl std::io::Write,
+    ) -> std::io::Result<()> {
+        loop {
+            if self.soc_ifc().cptra_fw_error_non_fatal().read() == code {
```
korran, line 146: I'm a bit confused why a function called "*_until_fatal_error" is actually looking at a field named error_non_fatal? Can you add a comment explaining why this is?

hw-model/src/lib.rs
```diff
@@ -137,6 +137,36 @@ pub trait HwModel {
         }
     }
 
+    fn copy_output_until_fatal_error(
```
korran, line 140: The semantics of this function confuse me. My expectation was that it would only succeed if the firmware failed with an error code matching the argument, and fail if the firmware did an "exit success" or "exit with some other failure".

If you want the semantics as-written, perhaps it makes more sense to add the error-checking logic to copy_output_until_exit_success?

rom/dev/tests/test_image_validation.rs
```diff
@@ -0,0 +1,465 @@
...
+    // Set ecc_pub_key.y to zero.
+    image_bundle.manifest.preamble.vendor_pub_keys.ecc_pub_keys[vendor_ecc_pub_key_idx].x =
+        ecc_pub_key_x_backup;
+    image_bundle.manifest.preamble.vendor_pub_keys.ecc_pub_keys[vendor_ecc_pub_key_idx]
+        .y
+        .fill(0);
+    let image = caliptra_builder::sign_image(&image_bundle).unwrap();
+    hw.upload_firmware(&image).unwrap();
+    let result = hw.copy_output_until_fatal_error(VENDOR_PUB_KEY_DIGEST_INVALID_ARG, &mut output);
+    assert!(result.is_err());
```
korran, line 127: result.is_err() is also true when somebody calls ExitCtrl::exit(1). I think we need this assertion to be stricter.

hw-model/src/lib.rs
mhatrevi: Typo; fixed.

builder/src/lib.rs
mhatrevi: Updated the function name.

hw-model/src/lib.rs
mhatrevi: Fixed

hw-model/src/lib.rs
mhatrevi: Updated to only succeed if the error code matches.

builder/src/lib.rs
```diff
@@ -190,6 +190,11 @@ impl Default for ImageOptions {
 }
 
 pub fn build_and_sign_image(fmc: &FwId, app: &FwId, opts: ImageOptions) -> anyhow::Result<Vec<u8>> {
+    let image = build_image(fmc, app, opts)?;
+    get_image_bytes(&image)
+}
+
+pub fn build_image(fmc: &FwId, app: &FwId, opts: ImageOptions) -> anyhow::Result<ImageBundle> {
```
jhand2, line 197: This is also signing the image, it's just returning the whole image struct instead of just the bytes.

Seems like build_and_sign_image can just return the ImageBundle and then callers can call `to_bytes` as needed

hw-model/src/lib.rs
jhand2: > If you want the semantics as-written, perhaps it makes more sense to add the error-checking logic to copy_output_until_exit_success?

Not a huge deal, but this suggestion seems more clear rather than having different functions for checking different kinds of errors

builder/src/lib.rs
mhatrevi: Updated

builder/src/lib.rs
```diff
@@ -173,8 +173,8 @@ pub struct ImageOptions {
     fmc_svn: u32,
     app_min_svn: u32,
     app_svn: u32,
-    vendor_config: ImageGeneratorVendorConfig,
-    owner_config: Option<ImageGeneratorOwnerConfig>,
+    pub vendor_config: ImageGeneratorVendorConfig,
```
korran, line 176: Whoops, all these things should be public Sorry.

hw-model/types/src/lib.rs
```diff
@@ -107,6 +107,30 @@ impl From<U4> for u32 {
...
+            0b0111 => Self::X7,
+            0b1000 => Self::X8,
+            0b1001 => Self::X9,
+            0b1010 => Self::Xa,
+            0b1011 => Self::Xb,
+            0b1100 => Self::Xc,
+            0b1101 => Self::Xd,
+            0b1110 => Self::Xe,
+            0b1111 => Self::Xf,
+            16_u32..=u32::MAX => Self::X0,
```
korran, line 129: This should be TryFrom<Error=()>, and this variant should return Err(())

korran (approved):

hw-model/src/lib.rs
mhatrevi: Let's go with the current change; we'll revisit if this ends up into too many copy_output_* functions.

PR #140, opened 2023-04-04
Update to the latest RTL.

korran:


korran:
Verilator failure occurs because the mailbox semantics were changed in https://github.com/chipsalliance/caliptra-rtl/commit/a786ec637808f3bd5aa76a05f43652edce27d94b such that the sender can no longer read out of the mailbox. We need to update this test to work from both sides of the mailbox.

korran:
Obsolete.

PR #142, opened 2023-04-05
hw-model: Set fuses via soc_ifc, change HwModel::new() semantics

korran:
This PR attempts to unify fuses between HwModelEmulated and HwModelVerilated, by removing all the fuse-related configuration from CaliptraRootBusArgs, and instead using the external soc_ifc bus to set them during early boot.

Also, caliptra_hw_model::create() is replaced by caliptra_hw_model::new(), which has slightly different semantics. This function is intended to be used by most test cases to set the fuses, start the boot FSM, and (optionally) upload the firmware to the mailbox. A bare-bones caliptra_hw_model::new_unbooted() function is provided for test cases that need to manipulate the external SoC->Caliptra APB bus directly during early boot in a non-standard way.


hw-latest/verilated/caliptra_verilated.sv
```diff
@@ -78,10 +80,6 @@ module caliptra_verilated (
     logic [MBOX_DATA_AND_ECC_W-1:0] mbox_sram_wdata_bitflip;
     logic [MBOX_DATA_AND_ECC_W-1:0] mbox_sram_rdata;
 
-
-    //TIE-OFF device lifecycle
-    security_state_t security_state = '{device_lifecycle: DEVICE_PRODUCTION, debug_locked: 1'b1};
```
rusty1968, line 83: Why was this removed?

sw-emulator/app/Cargo.toml
```diff
@@ -9,10 +9,14 @@ edition = "2021"
 
 [dependencies]
 clap = { version = "3.2.14", default-features = false, features = ["std"] }
+caliptra-registers = { path = "../../registers" }
 caliptra-emu-bus = { path = "../lib/bus" }
 caliptra-emu-cpu = { path = "../lib/cpu" }
 caliptra-emu-periph = { path = "../lib/periph" }
 caliptra-emu-types = { path = "../lib/types" }
+caliptra-hw-model = { path = "../../hw-model" }
```
rusty1968, line 17: Do we have any risk for circular dependencies? emu's peripheral lib is a dependency of hw-model crate's emulated model. Should we aim for an emulator app to be built on top of hw-emulated-model

korran: This is a binary, so it can't be a library dependency (and thus no risk of circular deps).

Regardless, eventually I'd like to either delete this app or re-implement it on top of HwModel and move it somewhere inside hw-model/.

hw-latest/verilated/caliptra_verilated.sv
korran: As part of https://github.com/chipsalliance/caliptra-sw/pull/142/commits/03973c6fa59de441fbac9c0d8cb292e867331402 :

This is not longer hard-coded, but a module input (from C++/rust); see the diff near the top of the file.

korran: FWIW, I strongly recommend reviewing this PR commit-by-commit, as the reasoning should be clearer that way.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -431,7 +442,11 @@ impl Bus for SocRegisters {
     /// Write data of specified size to given address
     fn write(&mut self, size: RvSize, addr: RvAddr, val: RvData) -> Result<(), BusError> {
         match addr {
-            Self::CALIPTRA_REG_START_ADDR..=Self::CALIPTRA_REG_END_ADDR => {
+            FUSE_START_ADDR..=FUSE_END_ADDR => {
```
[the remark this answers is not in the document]
korran: I thought about writing a unit test for this, but I'm leaning towards writing new tests against HwModel directly, so we can compare the test output between the RTL and sw-emulator directly.

sw-emulator/lib/periph/src/helpers.rs
```diff
@@ -13,18 +13,122 @@ Abstract:
 --*/
 #![allow(unused)]
 
-pub fn words_from_bytes_le(arr: &[u8; 48]) -> [u32; 12] {
-    let mut result = [0u32; 12];
-    for i in 0..result.len() {
-        result[i] = u32::from_le_bytes(arr[i * 4..][..4].try_into().unwrap())
+use std::ops::{Index, IndexMut, RangeFrom};
```
rusty1968, line 16: This may have utility value beyond the sw-emulator.

sw-emulator/lib/periph/src/soc_reg.rs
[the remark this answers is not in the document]
rusty1968: Makes sense.

sw-emulator/lib/periph/src/helpers.rs
rusty1968: Agreed.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -361,8 +362,8 @@ struct SocRegistersImpl {
     #[register(offset = 0x00b0)]
     cptra_timer_config: ReadWriteRegister<u32>,
 
-    #[register(offset = 0x00b4)]
-    cptra_bootfsm_go: ReadOnlyRegister<u32>,
+    #[register(offset = 0x00b4, write_fn = on_write_bootfsm_go)]
+    cptra_bootfsm_go: u32,
```
rusty1968, line 366: Why are we using a primitive type u32 instead of making a ReadWriteRegister<u32>?

rusty1968 (approved):

sw-emulator/lib/periph/src/soc_reg.rs
korran: ReadWriteRegister doesn't really do anything useful over a primitive. ReadOnlyRegister is handy since it lets you prevent writes without having to write any code. In this case, we have a write_fn that handles the write behavior, and the default read behavior of "return the primitive" is what we want.

PR #148, opened 2023-04-07
Refactor DICE computation in ROM

bluegate010:
This commit makes the following changes:

* Captures fuse state and FMC state as one hash, rather than a chain of PCR extensions.
* Corrects comments referring to the measurement as the UDS.
* Centralizes FMC measurement generation in fmc_alias.rs, rather than pcr.rs.
* Captures the generated FMC measurement in the data vault.
* Represents the generated FMC measurement in the FMC alias cert.
* Deprecates usage of PCR1.
* Uses data vault helper functions which lock slots as they are written.
* Updates README.md accordingly.

### DIFF
--- a/rom/dev/src/lock.rs
+++ b/rom/dev/src/lock.rs
@@ -25,6 +25,8 @@ use crate::rom_env::RomEnv;
 /// * `env` - ROM Environment
 /// * `reset_reason` - Reset reason
 pub fn lock_registers(env: &RomEnv, reset_reason: ResetReason) {
+    // Note that these registers should already have been locked when they
+    // were first written; locking is performed here as a precaution.
     if reset_reason == ResetReason::ColdReset {
         lock_cold_reset_reg(env);
         lock_warm_reset_reg(env);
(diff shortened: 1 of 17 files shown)


bluegate010:
I'm not 100% sure about some of the changes in this commit - particularly the deprecation of PCR1. I do think we need to capture all CDI inputs in the cert, but whether those are placed in Data Vault or PCR1 doesn't make that much difference - except that it matters for what the reference measurements must have - either the actual measurements, or the result of a hash-extend.

rom/dev/src/lock.rs
vsonims, line 1: This is a problem. Locks need to happen at one place. We had things floating around before ever where and centralized it in one place for maintainace. 

We should chat about this change.

vsonims (changes requested):

rom/dev/src/lock.rs
bluegate010: I'm worried about someone writing something sensitive to the data vault in one file, and then forgetting to lock it down in a separate file. I almost made that mistake in this commit. With this change, the only way anything gets written to the data vault is by also locking it immediately after writing.

That being said, this part of the commit is not something I feel incredibly strongly about.

bluegate010: I've updated the PR to restore locking. We now lock opportunistically when data is written, and again as a safety measure at the end.

bluegate010:
Note that tests are failing because this commit allocates a new Data Vault slot, which RTL does not provide. Will need to either make do with the current number of DV slots, or go allocate at least one more in RTL.

bluegate010:
Sorry for the churn, gonna close this PR and open a new one so I can do some refactoring.

PR #163, opened 2023-04-10
Refactor DICE layering.

bluegate010:
This commit clarifies inputs to each DICE layer, and captures each layer's outputs. A future commit will stash an output of the FMC layer's DICE computation in the FHT, without traversing Data Vault.


rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -54,97 +50,48 @@ rom_err_def! {
...
+        cprintln!("[idev] SUBJECT.KEYID = {}", KEY_ID_IDEVID_PRIV_KEY as u8);
+        cprintln!("[idev] UDS.KEYID = {}", KEY_ID_UDS as u8);
 
         // Derive the DICE CDI from decrypted UDS
-        let cdi = Self::derive_cdi(env, uds, input.cdi)?;
+        Self::derive_cdi(env, KEY_ID_UDS, KEY_ID_CDI)?;
 
         // Derive DICE Key Pair from CDI
-        let key_pair = Self::derive_key_pair(env, cdi, input.subj_priv_key)?;
+        let idevid = Self::derive_key_pair(env, KEY_ID_CDI, KEY_ID_IDEVID_PRIV_KEY)?;
```
mhatrevi, line 73: Consider renaming to idevid_key_pair.

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -35,79 +37,72 @@ rom_err_def! {
...
+        cprintln!("[ldev] FE.KEYID = {}", KEY_ID_FE as u8);
 
-        // Derive the DICE CDI from decrypted UDS
-        let cdi = Self::derive_cdi(env, uds, input.cdi)?;
+        // Derive the DICE CDI from decrypted field entropy
+        Self::derive_cdi(env, KEY_ID_FE, KEY_ID_CDI)?;
 
         // Derive DICE Key Pair from CDI
-        let key_pair = Self::derive_key_pair(env, cdi, input.subj_priv_key)?;
+        let ldevid_key_pair = Self::derive_key_pair(env, KEY_ID_CDI, KEY_ID_LDEVID_PRIV_KEY)?;
```
mhatrevi, line 65: Consider naming this to ldevid_key_pair.

rom/dev/src/flow/cold_reset/idev_id.rs
bluegate010: Done

rom/dev/src/flow/cold_reset/ldev_id.rs
bluegate010: Done

vsonims (changes requested):
Want to understand the rational for these changes.

bluegate010:
@vsonims, rationales are the following:

- Due to the way DICE layering is currently stitched together in [ColdResetFlow::run](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/mod.rs#L45), it is currently not possible to pass any data from [FmcAliasLayer::derive](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/fmc_alias.rs#L49) to [fht::make_fht](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/fht.rs#L28) without having that data first be written to the Data Vault. This is problematic because there is evidence we would like to add to the FMC alias certificate that will not fit in Data Vault. The intent is to instead write it to FHT directly, so firmware can reconstitute the certificate.
- The DICE layering enforces that all layers receive the same input and produce the same output. This does not correspond with the actual needs of the different layers: IDevID needs UDS, LDevID needs field entropy, and FMC Alias needs FMC. But as it currently stands, the [DiceInput](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/dice.rs#L2) struct contains a superset of all data needed by all layers, so we are passing the UDS and field entropy key IDs to all layers, even those that don't need it.
- The DICE layering leverages a [compose_layers](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/dice.rs#L112) helper, which produces a new DiceInput from a DiceOutput and an older DiceInput. This helper toggles each layer's subject and auth keypairs and destination private key IDs. It is therefore not obvious when viewing the [initial input](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/mod.rs#L54-L66) in ColdResetFlow::run that IDevID goes in slot 7, LDevID goes in slot 5, and FMC alias goes in slot 7. If the number of certificate layers were even instead of odd, the FMC alias key would not end up in the slot identified by `KEY_ID_PRIV_KEY`.
- The `DiceInput` structure is hard to reason about. When created in ColdResetFlow::run, it is not used to provide input subject key, auth key, auth sn/keyid, UDS, or field entropy to any DICE layers. Rather, it is declaring where these values should be written by `InitDevIdLayer::derive`. It is initialized in the manner that it is so that when it is later used by `compose_layers` to produce a new DiceInput, the right things happen. Again, hard to follow.
- It is not necessary to pass the CDI key ID from input to output to input again. There is only one key ID that ever holds the CDI, so it can be hard-coded in each layer's derivation routine.
- Overall, the data flow in ColdResetFlow::run can be made easier to reason about by removing the layer composition mechanics and simply invoking each layer's derivation function in turn, rather than introducing layers of abstraction for arbitrary DICE layer composition. This decouples the output provided by each layer; as noted above, in a follow-up PR we can have `FmcAliasLayer::derive` return a different type than `LocalDevIdLayer::derive`, because the former will produce evidence that needs to be passed to `fht::make_fht`.
- UDS and field entropy decryption are moved out to `ColdResetFlow::run` because it's a bit odd having the IDevID layer be dealing with field entropy, as that's only needed by the LDevID layer. I see why UDS and field entropy decryption should happen together (so we can erase obfuscation secrets immediately afterwards), so I figured it's cleaner to move both out. Also, in a follow-up PR, we can centralize erasure of UDS and field entropy. Right now erasure is scattered between [InitDevIdLayer::derive_cdi](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/idev_id.rs#L161) and [LocalDevIdLayer::derive_cdi](https://github.com/chipsalliance/caliptra-sw/blob/56f333336ff042299368812658f8c4b6633ee8dd/rom/dev/src/flow/cold_reset/ldev_id.rs#L102). As with DV slot locking, erasure can be centralized.
- Several helper functions, such as Crypto::hmac384_mac and derive_cdi, return the same argument they are provided. This is unnecessary, as the caller already has the argument.

bluegate010:
Closing as per feedback.

PR #170, opened 2023-04-12
RT Alias Certificate

rusty1968:


x509/build/build.rs
```diff
@@ -87,3 +88,33 @@ fn gen_fmc_alias_cert(out_dir: &str) {
...
+            &[
+                FwidParam {
+                    name: "TCB_INFO_RT_TCI",
+                    fwid: Fwid {
+                        hash_alg: asn1::oid!(/*sha384*/ 2, 16, 840, 1, 101, 3, 4, 2, 2),
+                        digest: &[0xCD; 48],
+                    },
+                },
+                FwidParam {
+                    name: "TCB_INFO_OWNER_PK_HASH",
```
bluegate010, line 110: This evidence was already present in the FMC alias cert. Not sure we also need it in the RT alias cert.

x509/build/build.rs
```diff
@@ -87,3 +88,33 @@ fn gen_fmc_alias_cert(out_dir: &str) {
     let template = bldr.tbs_template("Caliptra FMC Alias", "Caliptra LDevID");
     CodeGen::gen_code("FmcAliasCertTbs", template, out_dir);
 }
+
+fn gen_rt_alias_cert(out_dir: &str) {
+    let mut usage = KeyUsage::default();
+    usage.set_key_cert_sign(true);
+    let bldr = cert::CertTemplateBuilder::<EcdsaSha384Algo>::new()
+        .add_basic_constraints_ext(true, 0)
```
jhand2, line 96: I think we can make the pathLenConstraint 1 (Runtime is only expected to sign DPE certs which are leaf certs).

x509/build/build.rs
```diff
@@ -87,3 +88,33 @@ fn gen_fmc_alias_cert(out_dir: &str) {
     let template = bldr.tbs_template("Caliptra FMC Alias", "Caliptra LDevID");
     CodeGen::gen_code("FmcAliasCertTbs", template, out_dir);
 }
+
+fn gen_rt_alias_cert(out_dir: &str) {
+    let mut usage = KeyUsage::default();
```
jhand2, line 93: I think we should include `digitalSignature` and `keyCertSign`.

FerralCoder: @jhand2 if I am understanding correctly, this build.rs file is just test code, correct? So you are suggesting that the test code should also produce as output the digitalSignature of the CERT. And then something you are calling keyCertSign? What do you mean by keyCertSign? My apologies if I am not correctly following what you meant.

jhand2: build.rs creates a template, and then ROM/FMC fill in anything at runtime that is variable (like TcbInfo fwids).

KeyUsage is an X.509 extension that says what the certificate can be used for. KeyUsage is essentially a bitfield, and two of those bits are:

* keyCertSign: This key can be used to sign other X.509 certificates
* digitalSignature: This key can be used to sign non-certificate data

Depending on the outcome of the quoting discussion, the RT cert may only need the keyCertSign bit set.

These attributes don't change at runtime, so they should be set during template creation.

rusty1968: Done

jhand2 (approved):

PR #173, opened 2023-04-13
Integrate mailbox into runtime firmware

jhand2:
Handle mailbox commands in runtime and add test which can run against
sw and verilator.

### DIFF
--- a/drivers/src/error.rs
+++ b/drivers/src/error.rs
@@ -53,6 +53,11 @@ pub enum CaliptraComponent {
 
     /// CSRNG
     Csrng = 13,
+
+    /// Runtime firmware
+    /// TODO: Once https://github.com/chipsalliance/caliptra-sw/pull/220 is
+    /// merged remove this and use RT error mechanism instead.
+    Runtime = 14,
 }
 
 #[macro_export]
--- a/rom/dev/tests/test_idevid_derivation.rs
+++ b/rom/dev/tests/test_idevid_derivation.rs
@@ -3,6 +3,7 @@
 use caliptra_builder::ImageOptions;
 use caliptra_drivers::{state::MfgFlags, IdevidCertAttr, X509KeyIdAlgo};
 use caliptra_hw_model::{Fuses, HwModel};
+use std::io::Write;
 
 pub mod helpers;
 
@@ -26,9 +27,12 @@ fn test_generate_csr() {
     hw.upload_firmware(&image_bundle.to_bytes().unwrap())
         .unwrap();
 
-    let result = hw.copy_output_until_exit_success(&mut output);
-    assert!(result.is_ok());
+    hw.step_until_output_contains("Caliptra RT listening for mailbox commands...")
+        .unwrap();
 
+    output
+        .write_all(hw.output().take(usize::MAX).as_bytes())
+        .unwrap();
     let output = String::from_utf8_lossy(&output);
     let csr_str = helpers::get_data("[idev] CSR = ", &output);
     let csr_uploaded = hex::decode(csr_str).unwrap();
@@ -37,7 +41,6 @@ fn test_generate_csr() {
 
 #[test]
 fn test_idev_subj_key_id_algo() {
-    let mut output = vec![];
     for algo in 0..(X509KeyIdAlgo::Fuse as u32 + 1) {
         let mut fuses = Fuses::default();
         fuses.idevid_cert_attr[IdevidCertAttr::Flags as usize] = algo;
@@ -47,7 +50,7 @@ fn test_idev_subj_key_id_algo() {
         hw.upload_firmware(&image_bundle.to_bytes().unwrap())
             .unwrap();
 
-        let result = hw.copy_output_until_exit_success(&mut output);
-        assert!(result.is_ok());
+        hw.step_until_output_contains("Caliptra RT listening for mailbox commands...")
+            .unwrap();
     }
 }
--- a/runtime/src/main.rs
+++ b/runtime/src/main.rs
@@ -36,6 +36,9 @@ pub extern "C" fn entry_point() -> ! {
     cprintln!("{}", BANNER);
 
     if let Some(_fht) = caliptra_common::FirmwareHandoffTable::try_load() {
+        cprintln!("Caliptra RT listening for mailbox commands...");
+        caliptra_runtime::handle_mailbox_commands();
+
         caliptra_drivers::ExitCtrl::exit(0)
     } else {
         caliptra_drivers::ExitCtrl::exit(0xff)
@@ -47,7 +50,7 @@ pub extern "C" fn entry_point() -> ! {
 #[allow(clippy::empty_loop)]
 extern "C" fn exception_handler(trap_record: &TrapRecord) {
     cprintln!(
-        "FMC EXCEPTION mcause=0x{:08X} mscause=0x{:08X} mepc=0x{:08X}",
+        "RT EXCEPTION mcause=0x{:08X} mscause=0x{:08X} mepc=0x{:08X}",
         trap_record.mcause,
         trap_record.mscause,
         trap_record.mepc
@@ -62,7 +65,7 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
 #[allow(clippy::empty_loop)]
 extern "C" fn nmi_handler(trap_record: &TrapRecord) {
     cprintln!(
-        "FMC NMI mcause=0x{:08X} mscause=0x{:08X} mepc=0x{:08X}",
+        "RT NMI mcause=0x{:08X} mscause=0x{:08X} mepc=0x{:08X}",
         trap_record.mcause,
         trap_record.mscause,
         trap_record.mepc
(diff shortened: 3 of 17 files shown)


runtime/tests/integration_tests.rs
```diff
@@ -31,6 +32,35 @@ fn run_rt_test(test_bin_name: Option<&str>) {
...
+    model.soc_mbox().cmd().write(|_| cmd_id);
+    model.soc_mbox().dlen().write(|_| arg_buf.len().try_into().unwrap());
+
+    for word in arg_buf {
+        model.soc_mbox().datain().write(|_| *word);
+    }
+    model.soc_mbox().execute().write(|exec_reg| exec_reg.execute(true));
+
+    model.step_until(|m| {
+        m.soc_mbox().status().read().status().cmd_complete()
```
korran, line 50: Should be:
model.step_until(|m| {
        !m.soc_mbox().status().read().status().cmd_busy()).

runtime/src/commands.rs
```diff
@@ -0,0 +1,35 @@
...
+pub fn handle_mailbox_commands() -> CaliptraResult<()> {
+    loop {
+        wait_for_cmd();
+
+        if let Some(mut recv_txn) = Mailbox::default().try_start_recv_txn() {
+            let cmd_id = recv_txn.cmd();
+            let len = recv_txn.dlen() as usize;
+            let mut buf = [0u32; 1024];
+            recv_txn.copy_request(0, &mut buf[..len]).unwrap();
+            recv_txn.complete(MboxStatusE::CmdBusy);
```
korran, line 23: This is wrong; mailbox response needs to be populated before calling complete.

The driver doesn't appear to have an API for that

runtime/src/commands.rs
```diff
@@ -0,0 +1,35 @@
...
+
+            // TODO: Actually handle command
+            cprintln!("[rt] Received command={}, len={}", cmd_id, len);
+
+            // Write response
+            let out_buf = [0xFFu8; 16];
+            let mut send_txn = Mailbox::default().try_start_send_txn().unwrap();
```
korran, line 30: You don't want to use try_start_send_txn to send a response; that's used for making an "RPC" request to the SoC.

drivers/src/mailbox.rs
```diff
@@ -70,10 +69,44 @@ impl Mailbox {
     /// * 'MailboxRecvTxn' - Object representing a receive operation
     pub fn try_start_recv_txn(&self) -> Option<MailboxRecvTxn> {
         let mbox = mbox::RegisterBlock::mbox_csr();
-        match mbox.status().read().mbox_fsm_ps() {
-            MboxFsmE::MboxExecuteUc => Some(MailboxRecvTxn::default()),
-            _ => None,
+        if mbox.status().read().mbox_fsm_ps().mbox_execute_uc() {
+            match mbox.status().read().status() {
```
korran, line 73: What is the purpose of this change?

drivers/src/mailbox.rs
```diff
@@ -70,10 +69,44 @@ impl Mailbox {
...
+        } else {
+            None
+        }
+    }
+}
+
+pub struct MailboxTxn;
+
+impl MailboxTxn {
+    pub(crate) fn enqueue(buf: &[u8]) -> CaliptraResult<()> {
```
korran, line 86: This should probably just be a top-level function.

drivers/src/mailbox.rs
jhand2: In the loop in `runtime/src/main.rs`, I'm continually calling `try_start_recv_txn` under the assumption that if there's a command in the mailbox and we're in execute state, we've received a new command.

The problem is that when we write data back to the mailbox, it's still in execute state. So until the SoC clears the execute register, runtime fw will keep looping and trying to execute the same command repeatedly.

I'm open to a different way of doing this though.

drivers/src/mailbox.rs
jhand2: Ack. Will do.

drivers/src/mailbox.rs
korran: In the RTL, once you write to mbox.status(), mbox.mbox_fsm_ps() switches from ExecuteUc to ExecuteSoc. Apparently this is not happening in the sw-emulator.

jhand2: I removed this. I still think the emulator needs to be updated, but that's not blocking this PR (mostly because the test steps until the status is set and then stops).

I'm happy to make this change in this PR or just do it in a follow-up.

drivers/src/mailbox.rs
jhand2: Moved runtime mailbox usage outside of this driver

drivers/src/mailbox.rs
jhand2: Resolving for now

drivers/src/error.rs
```diff
@@ -53,6 +53,9 @@ pub enum CaliptraComponent {
 
     /// CSRNG
     Csrng = 13,
+
+    /// Runtime firmware
+    Runtime = 14,
```
korran, line 58: I don't think this is where @vsonims intended the runtime errors to go. Maybe put a TODO here saying something like "move this somewhere more idiomatic".

@rusty1968 has plans to revamp how the errors are organized.

jhand2: Done (once https://github.com/chipsalliance/caliptra-sw/pull/220 is merged I'll switch to use that mechanism)

rom/dev/tests/test_idevid_derivation.rs
```diff
@@ -55,7 +58,7 @@ fn test_idev_subj_key_id_algo() {
         hw.upload_firmware(&image_bundle.to_bytes().unwrap())
             .unwrap();
 
-        let result = hw.copy_output_until_exit_success(&mut output);
-        assert!(result.is_ok());
+        // Step until runtime firmware denotes successful firmware upload
+        hw.step_until(|m| m.soc_mbox().status().read().status().cmd_complete());
```
korran, line 62: The ROM still has plenty of stuff to do after the mailbox transaction is completed. We should probably be waiting until the runtime firmware has printed something out to the UART, or perhaps just send the runtime firmware a mailbox command to make sure it is alive.

runtime/tests/integration_tests.rs
```diff
@@ -30,20 +47,84 @@ fn run_rt_test(test_bin_name: Option<&str>) {
...
+    let image = caliptra_builder::build_and_sign_image(
+        &FMC_WITH_UART,
+        &runtime_fwid,
+        ImageOptions::default(),
+    )
+    .unwrap();
+
+    model.step_until(|m| m.soc_ifc().cptra_flow_status().read().ready_for_fw());
+
+    model.upload_firmware(&image.to_bytes().unwrap()).unwrap();
```
korran, line 59: nitpick: you can get this part for free if you pass set BootParams::fw_image to Some(&image.to_bytes().unwrap()) above. See https://github.com/chipsalliance/caliptra-sw/blob/main/hw-model/src/lib.rs#L185

test/tests/smoke_test.rs
```diff
@@ -30,7 +31,13 @@ fn smoke_test() {
     })
     .unwrap();
     let mut output = vec![];
-    hw.copy_output_until_exit_success(&mut output).unwrap();
+
+    // Step until runtime firmware denotes successful firmware upload
+    hw.step_until(|m| m.soc_mbox().status().read().status().cmd_complete());
```
korran, line 36: Does this actually do what you think it does? hw.upload_firmware() already waits for the upload command to complete:

https://github.com/chipsalliance/caliptra-sw/blob/effcd075176fdb598ba103a43e7aa99ac9f06b09/hw-model/src/lib.rs#L385-L399

I don't think the strings we are looking for below will be found.

runtime/src/main.rs
```diff
@@ -31,11 +32,21 @@ const BANNER: &str = r#"
                |_|
 "#;
 
+fn report_fw_load_success() {
+    // Complete the firmware download mailbox command
+    let mbox = mbox::RegisterBlock::mbox_csr();
+    mbox.status()
+        .write(|w| w.status(|_| MboxStatusE::CmdComplete));
```
korran, line 39: This has already been done. The ROM completed the mailbox transaction write after it validated the signature. Remove this.

rom/dev/tests/test_idevid_derivation.rs
jhand2: Done.

test/tests/smoke_test.rs
jhand2: Discussed offline. Switched to checking for UART output from RT.

runtime/src/main.rs
jhand2: Done.

runtime/tests/integration_tests.rs
jhand2: Done.

korran (approved):

PR #176, opened 2023-04-13
ROM: Don't indicate failure after successful firmware download.

korran:
Also, have HwModel::download_firmware() check for mailbox success, and extract general purpose code into HwModel::mailbox_execute() function.

More tests are in the pipeline, but they are currently blocked on incorrect behavior of the emulator's mailbox peripheral.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -26,8 +26,6 @@ pub use model_emulated::ModelEmulated;
 #[cfg(feature = "verilator")]
 pub use model_verilated::ModelVerilated;
 
-use caliptra_emu_types::RvSize;
-
 /// Ideally, general-purpose functions would return `impl HwModel` instead of
 /// `DefaultHwModel` to prevent users from calling functions that aren't
 /// available on all HwModel implementations.  Unfortunately, rust-analyzer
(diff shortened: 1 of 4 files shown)


hw-model/src/lib.rs
```diff
@@ -330,44 +341,107 @@ pub trait HwModel {
...
+                &self.soc_mbox().dataout().read().to_le_bytes()[..usize::try_from(dlen).unwrap()],
+            );
         }
 
-        // Set Execute Bit
-        self.soc_mbox().execute().write(|w| w.execute(true));
+        self.soc_mbox().execute().write(|w| w.execute(false));
+        // mbox_fsm_ps isn't updated immediately after execute is cleared (!?),
+        // so step an extra clock cycle to wait for fm_ps to update
+        self.step();
```
jhand2, line 434: I assume you're seeing this in verilator, not the sw emulator?

jhand2 (dismissed):

hw-model/src/lib.rs
korran: Yes

mhatrevi (approved):

jhand2 (approved):

PR #178, opened 2023-04-14
Rebase caliptra-rtl submodule onto latest main branch.

korran:


hw-model/src/lib.rs
```diff
@@ -485,46 +477,4 @@ mod tests {
             "expected output \"ha\", was \"hi\""
         );
     }
-
-    #[test]
-    pub fn test_upload_firmware() {
```
jhand2, line 490: is this test removed because it no longer works with the new RTL?

jhand2: Never mind, I see your commit message says this

jhand2 (approved):

jhand2 (approved):

PR #179, opened 2023-04-14
[feat] Additonal checks in image verifications

mhatrevi:
This change adds the following checks:
	1. Check for validity of the load and entry point address of FMC and Runtime.
	2. SVN validity

Additional integration tests are added for testing these conditions.

### DIFF
--- a/drivers/src/fuse_bank.rs
+++ b/drivers/src/fuse_bank.rs
@@ -210,9 +210,12 @@ impl FuseBank {
     /// # Returns
     ///     fmc security version number
     ///
-    pub fn fmc_svn() -> u32 {
+    pub fn fmc_svn(&self) -> u32 {
         let soc_ifc_regs = soc_ifc::RegisterBlock::soc_ifc_reg();
-        soc_ifc_regs.fuse_fmc_key_manifest_svn().read()
+        32 - soc_ifc_regs
+            .fuse_fmc_key_manifest_svn()
+            .read()
+            .leading_zeros()
     }
 
     /// Get the runtime security version number.
@@ -223,9 +226,10 @@ impl FuseBank {
     /// # Returns
     ///     runtime security version number
     ///
-    pub fn runtime_svn() -> u64 {
+    pub fn runtime_svn(&self) -> u32 {
         let soc_ifc_regs = soc_ifc::RegisterBlock::soc_ifc_reg();
-        (soc_ifc_regs.fuse_runtime_svn().at(1).read() as u64) << 32
-            | soc_ifc_regs.fuse_runtime_svn().at(0).read() as u64
+        64 - ((soc_ifc_regs.fuse_runtime_svn().at(1).read() as u64) << 32
+            | soc_ifc_regs.fuse_runtime_svn().at(0).read() as u64)
+            .leading_zeros()
     }
 }
--- a/image/verify/src/verifier.rs
+++ b/image/verify/src/verifier.rs
@@ -53,6 +53,20 @@ caliptra_err_def! {
         UpdateResetVenPubKeyIdxMismatch = 30,
         UpdateResetFmcDigestMismatch = 31,
         UpdateResetVenPubKeyIdxOutOfBounds = 32,
+        FmcLoadAddrInvalid = 33,
+        FmcLoadAddrUnaligned = 34,
+        FmcEntryPointInvalid = 35,
+        FmcEntryPointUnaligned = 36,
+        FmcSvnGreaterThanMaxSupported = 37,
+        FmcSvnLessThanMinSupported = 38,
+        FmcSvnLessThanFuse = 39,
+        RuntimeLoadAddrInvalid = 40,
+        RuntimeLoadAddrUnaligned = 41,
+        RuntimeEntryPointInvalid = 42,
+        RuntimeEntryPointUnaligned = 43,
+        RuntimeSvnGreaterThanMaxSupported = 44,
+        RuntimeSvnLessThanMinSupported = 45,
+        RuntimeSvnLessThanFuse = 46,
     }
 }
 
@@ -437,6 +451,13 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
         Ok(info)
     }
 
+    // Check if SVN check is required
+    fn svn_check_required(&self, image: Env::Image) -> bool {
+        // If device is unprovisioned or if anti-rollback is disabled, don't check the SVN.
+        !(self.env.dev_lifecycle(image) == Lifecycle::Unprovisioned
+            || self.env.anti_rollback_disable(image))
+    }
+
     /// Verify FMC
     fn verify_fmc(
         &self,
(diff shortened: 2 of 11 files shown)


ajisaxena (dismissed):

ajisaxena (dismissed):

rom/dev/src/rom_env.rs
```diff
@@ -175,4 +195,9 @@ impl RomEnv {
     pub fn pcr_bank(&self) -> &EnvCell<PcrBank> {
         &self.pcr_bank
     }
+
+    /// Get ICCM Range
+    pub fn iccm_range(&self) -> &Range<u32> {
```
korran, line 200: FWIW, the compiler would probably have an easier time optimizing this if iccm_range(&self) returned a constant rather than a field.

rom/dev/src/rom_env.rs
```diff
@@ -70,10 +76,20 @@ pub struct RomEnv {
...
+    iccm_range: Range<u32>,
 }
 
 impl Default for RomEnv {
     fn default() -> Self {
+        let iccm_start: u32;
+        let iccm_size: u32;
+        unsafe {
+            iccm_start = (&ICCM_ORG as *const u32) as u32;
+            iccm_size = (&ICCM_SIZE as *const u32) as u32;
```
korran, line 90: To be sound, you need to use [addr_of!](https://doc.rust-lang.org/stable/std/ptr/macro.addr_of.html) instead of using an intermediate reference. Probably best to do this everywhere you use this pattern.

korran: To be honest, creating extern "C" variables for stuff that is not actually memory is probably asking for trouble, since folks can cause unsound things to happen without writing unsafe. I'd prefer to replace ICCM_SIZE with ICCM_END, or make ICCM_SIZE a normal const defined 100% in Rust code.

rom/dev/tests/test_image_validation.rs
```diff
@@ -36,6 +36,17 @@ const VENDOR_PUB_KEY_DIGEST_MISMATCH: u32 = 0x0B000005;
...
+const FMC_LOAD_ADDR_INVALID: u32 = 0x0B000021;
+const FMC_ENTRY_POINT_INVALID: u32 = 0x0B000022;
+const FMC_SVN_GREATER_THAN_MAX_SUPPORTED: u32 = 0x0B000023;
+const FMC_SVN_LESS_THAN_MIN_SUPPORTED: u32 = 0x0B000024;
+const RUNTIME_LOAD_ADDR_INVALID: u32 = 0x0B000025;
+const RUNTIME_ENTRY_POINT_INVALID: u32 = 0x0B000026;
+const RUNTIME_SVN_GREATER_THAN_MAX_SUPPORTED: u32 = 0x0B000027;
+const RUNTIME_SVN_LESS_THAN_MIN_SUPPORTED: u32 = 0x0B000028;
+
+const ICCM_START_ADDR: u32 = 0x40000000;
```
korran, line 48: These probably need to be defined somewhere centrally too (another TODO?)

mhatrevi: Agreed. Currently I have them under a TODO.

rom/dev/src/rom_env.rs
mhatrevi: Added two constants ICCM_START and ICCM_SIZE.

mhatrevi: > To be sound, you need to use [addr_of!](https://doc.rust-lang.org/stable/std/ptr/macro.addr_of.html) instead of using an intermediate reference. Probably best to do this everywhere you use this pattern.
> 
> LLVM's optimizer is allowed to issue phantom loads from anything a reference points to (since normally that is safe). But if the reference is to something other than "normal memory", that is very dangerous.

Makes sense; will do this as a separate PR.

rom/dev/src/rom_env.rs
mhatrevi: Updated.

image/verify/src/verifier.rs
```diff
@@ -455,10 +463,31 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
             raise_err!(FmcDigestMismatch)
         }
 
-        // TODO: Perform following Address checks
-        // 1. Load address is a valid ICCM Address
-        // 2. Entry Point is a valid ICCM Address
-        // 3. Entry Point is within the image
+        // TODO: Perform following Address check
+        // Entry Point is within the image
```
korran, line 467: What about alignment check? For both address and entry-point?

mhatrevi: Added.

rom/dev/tests/test_image_validation.rs
```diff
@@ -731,6 +887,150 @@ fn test_runtime_digest_mismatch() {
...
+        hw.upload_firmware(&image).unwrap_err()
+    );
+    let result =
+        hw.copy_output_until_non_fatal_error(RUNTIME_SVN_LESS_THAN_MIN_SUPPORTED, &mut output);
+    assert!(result.is_ok());
+}
+
+#[test]
+fn test_runtime_svn_less_than_fuse_svn() {
+    let fuse_svn: [u32; 4] = [10, 0, 0, 0];
```
korran, line 1015: Is this a valid SVN fuse value? The binary value (0b1010) does not have consecutive ones.

mhatrevi: Updated.

drivers/src/fuse_bank.rs
```diff
@@ -223,9 +226,10 @@ impl FuseBank {
     /// # Returns
     ///     runtime security version number
     ///
-    pub fn runtime_svn() -> u64 {
+    pub fn runtime_svn(&self) -> u32 {
         let soc_ifc_regs = soc_ifc::RegisterBlock::soc_ifc_reg();
-        (soc_ifc_regs.fuse_runtime_svn().at(1).read() as u64) << 32
-            | soc_ifc_regs.fuse_runtime_svn().at(0).read() as u64
+        64 - ((soc_ifc_regs.fuse_runtime_svn().at(1).read() as u64) << 32
```
korran, line 231: For the record, I think this leading-zero  stuff should be in its own function with unit tests to make it easier for folks to review (and trust).

image/verify/src/verifier.rs
```diff
@@ -455,10 +467,37 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
+
+        if !self.env.iccm_range().contains(&verify_info.entry_point) {
+            raise_err!(FmcEntryPointInvalid)
+        }
+        if verify_info.entry_point % 4 != 0 {
+            raise_err!(FmcEntryPointUnaligned)
+        }
+
+        if self.env.dev_lifecycle(image) != Lifecycle::Unprovisioned
+            || !self.env.anti_rollback_disable(image)
```
korran, line 487: Is this logic right? "Check the SVN if we're provisioned OR anti-rollback is enabled" 

Shouldn't that be AND?

drivers/src/fuse_bank.rs
mhatrevi: Will fix as a separate PR.

image/verify/src/verifier.rs
mhatrevi: Udpated

korran (approved):

PR #181, opened 2023-04-14
Add CSRNG driver

rkr35:
Contains enough code to get a handle to the CSRNG peripheral and generate `u32`s from a conditioned, physical entropy source.

Outstanding items:

- [x] Enable [`CALIPTRA_INTERNAL_TRNG`](https://github.com/chipsalliance/caliptra-rtl/blob/5d49ad1483467ee972b28b5563c074ee35a0bd43/src/integration/rtl/config_defines.svh#L19) (otherwise Verilated tests will hang).
- [ ] ~~Figure out how to simulate health checks failures so we can test querying them.~~ **I'll address in a future PR.**
- [ ] ~~If needed, add tests for [other commands](https://opentitan.org/book/hw/ip/csrng/doc/theory_of_operation.html#command-description) in the peripheral, e.g., Reseed and Update. Otherwise remove their driver code.~~ **I'll address in a future PR.**


drivers/src/csrng.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// let mut random_words = csrng.generate(num_words)?;
+    ///
+    /// // Rounds up to nearest multiple of four.
+    /// assert_eq!(random_words.len(), 4);
+    ///
+    /// for word in random_words {
+    ///     // Do something with `word`.
+    /// }
+    /// ```
+    pub fn generate(&mut self, num_words: NonZeroUsize) -> CaliptraResult<Iter> {
```
korran, line 131: The normal use-case will be to initialize blinding vectors; I wonder if codegen might be tighter if this were just

```
pub fn generate(&mut self, buf: &[mut u32]);
```

Have you done any experiments?

rkr35: Played around with the CSRNG test binary and pretty much concluded that you can get any result you want depending on how you set up the call site, change the frequency of functions called, or pick which functions are `#[inline(always/never)]`.

@korran and I discussed offline that it's more useful to examine the code size of the driver once it's embedded in the ROM.

Below are a few experiments just to rule out anything egregious.

## Current PR
* Iterator fills a buffer.
* Buffer asserted against expected values.
```
$ cargo size --locked --target riscv32imc-unknown-none-elf --features emu,riscv --no-default-features --profile firmware -p caliptra-drivers-test-bin --bin csrng -- --format=sysv

section                size         addr
.text                  9149          0x0
.data                  1336   0x50000000
```

9149 + 1336 = 10485 bytes

## Point-of-use iterator
* Each value of iterator used immediately.
* No storage (beyond a single `u32`) of yielded values.
```
section                size         addr
.text                  9125          0x0
.data                  1256   0x50000000
```

9125 + 1256 = 10381 bytes

```rust
let mut csrng = unsafe { Csrng::with_seed(SEED) }.expect("construct CSRNG");
let num_words = NonZeroUsize::new(EXPECTED_OUTPUT.len()).expect("non-zero num_words");
let words = csrng.generate(num_words).expect("generate 2");

assert!(words.len() >= EXPECTED_OUTPUT.len());

for (actual, expected) in zip(words, EXPECTED_OUTPUT) {
    // Point-of-use of each random word as opposed to store-in-buffer.
    assert_eq!(actual, expected);
}
```

## Fill buffer
* Populate a caller-provided buffer.
* Buffer asserted against expected values.
```
section                size         addr
.text                  9053          0x0
.data                  1360   0x50000000
```

9053 + 1360 = 10413 bytes

```rust
pub fn generate(&mut self, out: &mut [u32]) -> CaliptraResult<()> {
    // Round up to nearest multiple of 128-bit block.
    let num_128_bit_blocks = (out.len() + 3) / 4;
    send_command(Command::Generate { num_128_bit_blocks })?;

    for (i, word) in out.iter_mut().enumerate() {
        if i % WORDS_PER_GENERATE_BLOCK == 0 {
            // Wait for CSRNG to generate next block of 4 words.
            wait::until(|| registers().genbits_vld().read().genbits_vld());
        }

        *word = registers().genbits().read();
    }

    Ok(())
}
```

korran: Ok, let's plan to go with the output buffer. That can be done in a separate PR if you like.

korran: Ok, let's defer this decision until we see what the ROM codegen looks like.

drivers/src/csrng.rs
```diff
@@ -0,0 +1,346 @@
...
+            glen = 0;
+            extra_words = &[];
+            err = CsrngErr::Uninstantiate;
+        }
+    }
+
+    // Write mandatory 32-bit command header.
+    registers()
+        .cmd_req()
+        .write(|_| ((glen as u32) << 12) | ((flag0 as u32) << 8) | ((clen as u32) << 4) | acmd);
```
korran, line 324: Oh gross, I wonder why ureg didn't create a type for this...

korran, line 324: Can you mask the fields (I know this is probably fine as is with all the constants above, but I'd prefer to err on the safe side)?

rkr35: Done.

korran (approved):

PR #184, opened 2023-04-17
Fix emulator mailbox peripheral to support response data and add mailbox command test for all HwModels

korran:
test_mailbox_execute() runs as a presubmit against both verilator and sw-emulator.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -446,7 +446,8 @@ pub trait HwModel {
 
 #[cfg(test)]
 mod tests {
-    use crate::{mmio::Rv32GenMmio, BootParams, HwModel, InitParams};
+    use crate::{mmio::Rv32GenMmio, BootParams, HwModel, InitParams, ModelError};
+    use caliptra_builder::FwId;
     use caliptra_emu_bus::Bus;
     use caliptra_emu_types::RvSize;
     use caliptra_registers::soc_ifc;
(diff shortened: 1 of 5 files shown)


hw-model/test-fw/mailbox_responder.rs
```diff
@@ -0,0 +1,63 @@
...
+    let soc_ifc = caliptra_registers::soc_ifc::RegisterBlock::soc_ifc_reg();
+    soc_ifc.cptra_flow_status().write(|w| w.ready_for_fw(true));
+    let mbox = caliptra_registers::mbox::RegisterBlock::mbox_csr();
+
+    loop {
+        while !mbox.status().read().mbox_fsm_ps().mbox_execute_uc() {}
+        let cmd = mbox.cmd().read();
+
+        match cmd {
+            0x1000_0000 => {
```
jhand2, line 30: nit: Would be helpful for understanding the tests to add comments explaining what each command is supposed to do

hw-model/src/lib.rs
```diff
@@ -549,4 +550,65 @@ mod tests {
...
+                ..Default::default()
+            },
+            ..Default::default()
+        })
+        .unwrap();
+
+        // Send command that echoes the command and input message
+        assert_eq!(
+            model.mailbox_execute(0x1000_0000, &message),
+            Ok(Some(vec![
```
jhand2, line 577: nit: Might be mildly more readable to do something like:

```rust
[vec![0x00, 0x00, 0x00, 0x10], message].concat()
```

Up to you

jhand2 (approved):

hw-model/test-fw/mailbox_responder.rs
korran: Done

PR #188, opened 2023-04-20
[fix] [design] Handoff table fields need to fully describe storage location of handoff information in Caliptra.

rusty1968:
- The FHT struct fields providing a level of indirection for the actual handoff information,  formerly modeled as 8-bit indexes are now 32-bit handles.
- A 32-bit handle describes the storage location as the tuple (vault, regType, regNum).
- The element vault identifies the storage spaces in Caliptra (e.g. keyVault, dataVault)
- The element regType identifies the attribute of the register bank instantiated by the vault (e.g. sticky, non-sticky)
- The element regNum identifies the register entry in the bank.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -526,6 +526,7 @@ dependencies = [
 name = "caliptra_common"
 version = "0.1.0"
 dependencies = [
+ "bitfield",
  "caliptra-drivers",
  "ufmt",
  "zerocopy",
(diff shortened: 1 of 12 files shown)


FerralCoder (commented):
I'd like to see the fields in the FMC spec updated to match what the code now has.

FerralCoder (approved):

PR #190, opened 2023-04-21
Fix #175: driver should not reset execute register.

### LINKED ISSUE #175: Mailbox driver should not reset the execute register when receive transaction is over
https://github.com/chipsalliance/caliptra-sw/blob/4467bf1652ad18373dca33351240a0f89bc1c44f/drivers/src/mailbox.rs#L334-L336

According to https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/Caliptra_Integration_Specification.pdf, this is the responsibility of the sender, not the receiver:

> 
> Receiving data from the mailbox:
> 1. On mailbox_data_avail assertion, the receiver reads the COMMAND register.
> 2. Receiver reads the DLEN register.
> 3. Receiver reads the MBOX DATAOUT register.
>    * Continue reading MBOX DATAOUT register until DLEN bytes are read.
> 4. If a response is required, receiver can populate the mailbox with the response by writing DATAIN
> 5. Set the mailbox status register to hand control back to the sender to read the response
> 6. **The sender will reset the EXECUTE register** after reading the response.
>    * This releases the LOCK on the mailbox.

korran:
From #175:

> According to https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/Caliptra_Integration_Specification.pdf, this is the responsibility of the sender, not the receiver:
> 
> > 
> > Receiving data from the mailbox:
> > 1. On mailbox_data_avail assertion, the receiver reads the COMMAND register.
> > 2. Receiver reads the DLEN register.
> > 3. Receiver reads the MBOX DATAOUT register.
> >    * Continue reading MBOX DATAOUT register until DLEN bytes are read.
> > 4. If a response is required, receiver can populate the mailbox with the response by writing DATAIN
> > 5. Set the mailbox status register to hand control back to the sender to read the response
> > 6. **The sender will reset the EXECUTE register** after reading the response.
> >    * This releases the LOCK on the mailbox.


drivers/src/mailbox.rs
```diff
@@ -331,10 +353,6 @@ impl MailboxRecvTxn {
         let mbox = mbox::RegisterBlock::mbox_csr();
         mbox.status().write(|w| w.status(|_| status));
 
-        // Release the lock
-        let mbox = mbox::RegisterBlock::mbox_csr();
-        mbox.execute().write(|w| w.execute(false));
-
         self.state = MailboxOpState::Idle;
```
rusty1968, line 356: We are no longer transitioning to idle. We need to remove that line.

korran: How so? The write to the status() register transfer control back to the SoC. We're definitely not in an Execute state any longer.

korran: FWIW, the FSM state in the RTL transitions from EXECUTE_UC to EXECUTE_SOC.

FWIW, I'm considering getting rid of self.state and instead using the type system to maintain state. This will result in less code (no runtime state checks) and will be less error prone (invalid state transitions won't compile).

rusty1968: The write to the execute register unlocks the mailbox. Did that change?

rusty1968: statemachine! {
    transitions: {
        // CurrentState Event [guard] / action = NextState
        *Idle + RdLock(Owner) [is_not_locked] / lock = RdyForCmd,
        RdyForCmd  + CmdWrite(Cmd) / set_cmd = RdyForDlen,
        RdyForDlen + DlenWrite(DataLength) / init_dlen = RdyForData,
        RdyForData + DataWrite(DataIn) / enqueue = RdyForData,
        RdyForData + ExecSet = Exec,
        Exec + DataRead / dequeue = Exec,
        Exec + DlenWrite(DataLength) / init_dlen = Exec,
        Exec + DataWrite(DataIn) / enqueue = Exec,
        Exec + ExecClear [is_locked] / unlock = Idle
    }
}

korran: Yes, but that's done by the *sender*, not the *receiver*. From the perspective of the microcontroller, the transaction is over.

korran: The RTL makes a distinction between what the sender can do vs the receiver. The emulator peripheral does not make this distinction, and thus it allows things to happen that should not be allowed (I plan to fix this after I have robust tests against the RTL).

rusty1968 (dismissed):

drivers/src/mailbox.rs
```diff
@@ -75,6 +75,28 @@ impl Mailbox {
...
+
+    /// Aborts with failure any pending SoC->Uc transactions.
+    ///
+    /// This is useful to call from a fatal-error-handling routine.
+    ///
+    /// # Safety
+    ///
+    /// Callers must guarantee that no other code is interacting with the
+    /// mailbox at the time this function is called. (For example, all
+    /// MailboxRecvTxn and MailboxSendTxn instances have been destroyed or
```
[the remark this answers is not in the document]
korran: This commit is actually part of #189. I've just rebased this PR onto main to make it clearer what's in scope for review.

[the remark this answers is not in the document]
korran: In #189, I changed the wording to "For example, **any** MailboxRecvTxn and MailboxSendTxn"

mhatrevi (approved):

PR #200, opened 2023-04-25
In ROM start.S, mark ROM region as cacheable in MRAC

korran:
Even though our CPU isn't instantiated with the cache enabled, setting this bit significantly improves CPU performance, cutting the amount of time spent in a wait state by roughly half.

Before this change: 4.9ms to clear ICCM/DCCM ([ROM boot timing log](https://gist.github.com/korran/98667451225df7a7cbc0617ce60ab412))

cache-off

After this change: 2.3ms to clear memory ([ROM boot timing log](https://gist.github.com/korran/af8b1555634fdf37888d8c1f924efdb9))

cache-on

Also, based on what I'm seeing in the waveforms, if we unroll the memory-clearing loop we should be able to get even faster.


bharatpillilli:
@calebofearth is turning in this change through the RTL - already created a PR for it to run the tests and then turn it in - you can just open an enhancement on this @korran

korran:
> @calebofearth is turning in this change through the RTL - already created a PR for it to run the tests and then turn it in - you can just open an enhancement on this @korran

I'm confused. We still need to set MRAC in the ROM bootup code, no?

bharatpillilli:
Correct - yes, through your ROM code you have to...I was just talking about RTL turn-in that have our own 'ROM' set up for HW testing..I thought you made that change because our internal PR looked exactly the same. Its my misunderstanding on the repo! Sorry about that.

vsonims (approved):

PR #208, opened 2023-04-26
Take advantage of veer-specific features to reduce code size. 

korran:


korran:
+zbb and +relax have been added in a separate commit, which are the most useful flags.

PR #212, opened 2023-04-27
Refactor/Optimize the Mailbox driver to use the type-state pattern.

### LINKED ISSUE #206: Refactor mailbox API to use typestate pattern
A type-state implementation is more efficient  as we can get rid of the state variables and associated errors, besides having the transitions verified by the compiler.

rusty1968:
The type-state pattern enforces correctness of the Mailbox state machine implementation at compile time.
It also reduces the code size as well as it increases performance by eliminating the overhead of checking the
validity of state transitions during runtime.

### DIFF
--- a/drivers/src/mailbox.rs
+++ b/drivers/src/mailbox.rs
@@ -1,16 +1,9 @@
-/*++
-
-Licensed under the Apache-2.0 license.
-
-File Name:
-
-    mailbox.rs
-
-Abstract:
-
-    File contains API for Mailbox operations
-
---*/
+//! Licensed under the Apache-2.0 license.
+//! file : mailbox.rs
+//! Abstract:
+//!    File contains mailbox abstraction.
+//!   The mailbox is the interface for the SoC to communicate with the Caliptra iRot.
+//!   The mailbox is used to send commands to the Caliptra iRot and to receive responses.
 
 use crate::{caliptra_err_def, CaliptraResult};
 use caliptra_registers::mbox::enums::MboxFsmE;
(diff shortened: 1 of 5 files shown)


drivers/src/mailbox.rs
```diff
@@ -99,72 +128,42 @@ impl Mailbox {
...
+        // Write Data Length :
+        mbox.dlen().write(|_| dlen);
+        Ok(MailboxSendTxn { state: RdyForData })
     }
-
-    fn enqueue(&self, buf: &[u8]) -> CaliptraResult<()> {
+}
+impl MailboxSendTxn<RdyForData> {
+    pub fn try_write_data(&self, buf: &[u8]) -> CaliptraResult<MailboxSendTxn<RdyForData>> {
         let remainder = buf.len() % size_of::<u32>();
```
rkr35, line 167: We could simplify the 4-byte-chunking logic with [`chunks_exact`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.chunks_exact).

Something like:
```rust
let chunks = buf.chunks_exact(size_of::<u32>());
let remainder = chunks.remainder();

for word in chunks {
    // write `word`
}

if !remainder.is_empty() {
    let last_word = remainder
        .iter()
        .rev()
        .fold(0, |w, &byte| (w << 8) | u32::from(byte));
    // write `last_word`
}
```

drivers/src/mailbox.rs
```diff
@@ -22,57 +15,93 @@ caliptra_err_def! {
...
+pub struct RdyForData;
+pub struct Execute;
+#[derive(Default)]
 /// Caliptra mailbox abstraction
 pub struct Mailbox {}
 
 const MAX_MAILBOX_LEN: u32 = 128 * 1024;
 
 impl Mailbox {
+    pub fn send_request(&self, cmd: u32, data: &[u8]) -> CaliptraResult<MailboxSendTxn<Execute>> {
```
rkr35, line 60: I'm likely missing context, but why can't the send functionality be a single method?

```rust
pub fn send_request(&mut self, cmd: u32, data: &[u8]) -> CaliptraResult<()> {
    // Do sanity checks, e.g., is locked, is `data.len()` <= 128 KiB.

    // Block until mailbox is ready for a command.
    
    // Write out (cmd, dlen, data).
    
    // "Send" by setting `execute` flag.

    // Block until mailbox is done sending.

    // "Done send" by clearing execute flag.

    Ok(())
}
```
...adjusted accordingly depending on what the blocking behavior should be to the caller.

rusty1968: Hey Royce, 

Overall you want to overlap your I/O operation with something else.

The user can still block waiting for mailbox access by checking the return.
Checkout the usage model for  'send_request'  in  idev_id.rs and sha384acc_tests.rs.

drivers/src/mailbox.rs
rusty1968: Done. Thank you , Royce!

JohnTraverAmd:
Rejecting based on ROM freeze and this PR isn't a must have.  We could re-open later if needed for another stepping or changes just in Run-Time.

Issue #171, opened 2023-04-12
Runtime FW SVN number of fuse bits discrepancy


FerralCoder:
Main Caliptra spec says 128 bits, ROM spec says 64 bits

FerralCoder:
Per discussion on 4/14, we don't need so many fuses for RT firmware at this time. ROM will only support 64. However, RTL/fuses has support for 128 and this will not be changed at this time. Need to adjust main spec to account for this.

- Current status:
    - Fuses: 128
    - Main spec: 128
    - Rom spec: 64
- Updates per discussion:
    - Fuses: 128 (no change, allows for future spec update if needed)
    - Main spec: 64 (change to note support of 64 but that 128 fuses are reserved anyway)
    - Rom spec: 64 (no change)

PR #213, opened 2023-04-27
Add Runtime SVN bit clarification in ROM Spec

FerralCoder:
Clarify that HW reserves 128 fuse bits for Runtime SVN even though ROM only supports 64 bits in this generation.

Also cleaned up trailing whitespace


FerralCoder:
addresses https://github.com/chipsalliance/caliptra-sw/issues/171

PR #216, opened 2023-04-28
Workaround LLD linker relaxation bogus overflow error.

korran:
LLD incorrectly emits region overflow errors when linker relaxation would prevent an overflow. See https://github.com/llvm/llvm-project/issues/62423 for more information.

This workaround increases the size of the ROM region in the linker script, allowing lld to exceed the ROM size before relaxation optimizations are applied. To prevent legitimate overflows, caliptra_builder::elf2rom() will ensure that the sections fit within the true 32k ROM size.

### DIFF
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -160,14 +160,15 @@ pub fn elf2rom(elf_bytes: &[u8]) -> io::Result<Vec<u8>> {
         let Some(src_bytes) = elf_bytes.get(file_offset..file_offset + len) else {
             return Err(other_err(format!("segment at 0x{:x} out of file bounds", segment.p_offset)));
         };
-        if len > 0 {
-            let Some(dest_bytes) = result.get_mut(mem_offset..mem_offset + len) else {
-              return Err(other_err(format!(
-                    "segment at 0x{mem_offset:04x}..0x{:04x} exceeds the ROM region \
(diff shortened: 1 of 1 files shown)


builder/src/lib.rs
```diff
@@ -160,10 +160,14 @@ pub fn elf2rom(elf_bytes: &[u8]) -> io::Result<Vec<u8>> {
         let Some(src_bytes) = elf_bytes.get(file_offset..file_offset + len) else {
             return Err(other_err(format!("segment at 0x{:x} out of file bounds", segment.p_offset)));
         };
-        let Some(dest_bytes) = result.get_mut(mem_offset..mem_offset + len) else {
-            continue;
-        };
-        dest_bytes.copy_from_slice(src_bytes);
+        if len > 0 {
```
bluegate010, line 163: Optional: could do,

```
let len = segment.p_filesz as usize;
if len == 0 {
  continue;
}
let Some(src_bytes) = ...;
let Some(dest_bytes) = ...;
```
I assume the check against zero is needed because there are some out-of-bounds but zero-length segments that we don't want to trip over unnecessarily.

bluegate010 (approved):

bluegate010 (approved):

mhatrevi (approved):

Issue #100, opened 2023-03-23
ROM Task - support measurement command


varuns-nvidia:
Problem statement: SoCs may have ROM patches and firmware that they need to boot before Caliptra firmware. Want to submit those measurements to Caliptra before execution.

 

Solution:
- Caliptra ROM will add a mailbox command for storing measurement.
- Caliptra ROM will support this command prior to CALIPTRA_FW_LOAD command.
- Caliptra ROM will use these measurements to derive CDI.
- Caliptra ROM will store those measurements to either PCRs or to Data Vault (set of write lockable registers on pwrgood reset). Vishal prefers Data Vault because PCRs don't support write locking, but write locking is not a strict requirement.

varuns-nvidia:
Amending the ask: ROM patch measurements should NOT be used as part of CDI derivation for Alias FMC.

jhand2:
> Caliptra ROM will use these measurements to derive CDI

So the new proposal instead is that runtime will push these measurements into DPE during initialization, right?

FerralCoder:
Great, I like this proposal much better!

varuns-nvidia:
Discussion with @vsonims  and @mhatrevi :
* ROM to implement https://github.com/chipsalliance/caliptra-sw/tree/main/runtime#stash_measurement in a command loop
* Handle 8 measurements prior to CALIPTRA_FW_LOAD
* Caliptra ROM is producing a fixed size event log in DCCM. ROM will populate this event log with the metadata and measurement arguments provided by the STASH_MEASUREMENT calls. Later, Runtime FW can use this event log to build DPE TCIs
* Caliptra ROM to extend PCR[1] with the measurement arguments provided (NOT the metadata)
* @vsonims to discuss with @bharatpillilli whether ROM should perform LDevID key generation before or after the STASH_MEASUREMENT command loop. Current stance is to do it before (keep existing flow, unify command loops with CALIPTRA_FW_LOAD)

@jhand2 @FerralCoder does this look good?

bharatpillilli:
My suggestion/view is thatbefore we accept any payload from outside of Caliptra, we should complete the LDevID key gen operation.

mhatrevi:
1. ROM will implement the following command:
![image](https://user-images.githubusercontent.com/38512878/235182322-ca25f826-36ed-4712-84ba-56cf8231d56f.png)

2. ROM will extend the measurement field of the command into PCR1.
3. ROM will allow extending up to 8 measurements.
[...]

PR #217, opened 2023-04-28
Add definition for `DISABLE_ATTESTATION`

bluegate010:
As postulated in https://github.com/chipsalliance/caliptra-sw/issues/100#issuecomment-1527879744.


varuns-nvidia (approved):
A couple places where the word intended is misspelled but lgtm

bluegate010:
> A couple places where the word intended is misspelled but lgtm

Ah, thanks, fixed that.

runtime/README.md
```diff
@@ -187,6 +187,26 @@ Table: `STASH_MEASUREMENT` output arguments
...
 | dpe\_result | u32      | Result code of DPE DeriveChild command. Little endian.
 
+### DISABLE\_ATTESTATION
+
+Disable attestation by erasing the CDI and DICE key. This command is intended
+for callers who update infrequently and cannot tolerate a changing DPE API surface, and is intended for situations where Caliptra firmware cannot be loaded
+and the SoC must proceed with boot.
+
+Upon receipt of this command, the current CDI is replaced with zeroes, and the
+associated DICE key is re-derived from the zeroed CDI.
```
jhand2, line 197: > the associated DICE key is re-derived from the zeroed CDI

I don't quite understand this. This seems to imply Caliptra firmware would still wield a signing key for the caller. Presumably we want to (1) at least refuse to give out a cert/CSR and (2) consider not signing stuff. Right?

bluegate010: We could do that too. My thinking was that the low-level machinery all works the same so it may be easier to run testing flows in certain circumstances. Con is that it may be less obvious that something has gone wrong; certs will just be invalid. Could define some error state for DPE where it's told by the Caliptra boot stack that something has gone wrong and it should just fail to sign anything.

jhand2: Why will the certificate be wrong? It will still be signed by RT Alias, will still contain whatever measurements are made in the future. The key won't be stable but it'll still be valid to a verifier.

bluegate010: If we re-derive the current DICE key (either LDevID, AliasFMC, or AliasRT depending on when `DISABLE_ATTESTATION` is invoked) based on a zeroed CDI, then when DPE in RT fw goes to sign a DPE leaf cert, some link in that cert chain will break because the valid keypair endorsed by one cert won't be the zeroed key used to sign the next cert.

jhand2: Oh I see. I thought you meant the current DPE CDI.

Is there a specific KeyVault slot we can refer to which holds the "current" CDI? I guess it's different for each layer. Maybe would be good to specify "The CDI for the current Caliptra boot layer"

bluegate010: Each layer actually uses the [same slot](https://github.com/chipsalliance/caliptra-sw/blob/6581104f46041eea9796de047431faa6f0b7bb27/rom/dev/src/flow/cold_reset/mod.rs#L32) for its CDI.

jhand2: Oh cool. Maybe just change "current CDI" to "current Caliptra firmware CDI" for clarity?

bluegate010: Changed to "Caliptra's current CDI", since it might not be a firmware CDI yet (if this is called before firmware download).

jhand2 (approved):

JohnTraverAmd (approved):

PR #224, opened 2023-05-01
[WIP] Add Image LMS defs

satyamsetti:


image/types/src/lib.rs
```diff
@@ -45,6 +52,18 @@ pub struct ImageEccPubKey {
     pub y: ImageScalar,
 }
 
+#[repr(C)]
+#[derive(AsBytes, FromBytes, Default, Debug, Copy, Clone, Eq, PartialEq)]
+pub struct ImageLmsPublicKey {
+    pub tree_type: u32,
+
+    pub otstype: u32,
```
jhand2, line 60: should these be enums? can give the enums `#[repr(u32)]` just might make the code more readable

satyamsetti: Not sure how to get rid of it, but when doing so, Rust warns that we need to encode all 4 Billion variants.
Similar with u16 and u8 repr too. Leaving it as is .

PR #227, opened 2023-05-02
[fix] Remove automatic generation of IDEVID CSR der file in emulator 

### LINKED ISSUE #194: Annoying file rom/dev/caliptra_idevid_csr.der is created when running the tests
```
$ ls -l rom/dev/caliptra_idevid_csr.der
ls: cannot access 'rom/dev/caliptra_idevid_csr.der': No such file or directory
$ cargo test
<snip>
$ ls -l rom/dev/caliptra_idevid_csr.der
-rw-r----- 1 kor primarygroup 428 Apr 21 22:45 rom/dev/caliptra_idevid_csr.der
$ git status
On branch <snip>

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	rom/dev/caliptra_idevid_csr.der
```

I suspect this code is at fault:

https://github.com/chipsalliance/caliptra-sw/blob/735eb3ad75a328e7866dd322cef9e0f2ba9e0dcb/sw-emulator/lib/periph/src/soc_reg.rs#L724-L730

mhatrevi:


rom/dev/tests/test_idevid_derivation.rs
```diff
@@ -41,6 +41,9 @@ fn test_generate_csr() {
 
     let output = String::from_utf8_lossy(&output);
     assert_output_contains(&output, "[idev] CSR uploaded");
+
+    // Remove the CSR file.
+    let _ = fs::remove_file("caliptra_idevid_csr.der");
```
bluegate010, line 46: I suggest putting a deletion helper function over next to https://github.com/chipsalliance/caliptra-sw/blob/59c457b2e328ef89dfec05851e883fd035ab9e35/sw-emulator/lib/periph/src/soc_reg.rs#L724,L746 where these files are created, so the file paths, creation and deletion, is concentrated in one place.

Plus from that file it looks like there are two DER files that may be generated, though I'm not clear on why caliptra_ldevid_cert.der is not being created.

korran: Can we move this file into some junk directory that's in .gitignore, and out of the main source directories?

bluegate010: +1, if soc_reg.rs were tweaked to initially generate these files in an ignored directory that would be great.

rom/dev/tests/helpers.rs
```diff
@@ -27,3 +29,35 @@ pub fn build_hw_model_and_image_bundle(
 
     (hw, image_bundle)
 }
+
+pub fn get_data(to_match: &str, haystack: &str) -> String {
```
bluegate010, line 33: Document this?

rom/dev/tests/test_idevid_derivation.rs
mhatrevi: Removed auto-generation of the CSR file in the emulator. Tests now have to explicitly download the CSR from the mailbox.

rom/dev/tests/helpers.rs
```diff
@@ -27,3 +29,35 @@ pub fn build_hw_model_and_image_bundle(
...
+    let mut index = haystack.find(to_match).unwrap();
+    index += to_match.len();
+    let mut str = String::new();
```
bluegate010, line 36: Could this be something like,

```
let end = haystack[index..].find("\n");
haystack[index..end]
```

mhatrevi: Looks like with the suggested change, the search terminates prematurely, returning the partial string. Will need to look more into this. We can take this as a separate improvement fix.

bluegate010: Ack, could be an off-by-one error.

bluegate010 (approved):

korran (approved):

PR #232, opened 2023-05-03
Remove unnecessary image arg from trait functions.

bluegate010:


mhatrevi (changes requested):
We need additional metadata from the Image object (e.g.: image size), so we need that Trait.

bluegate010:
@mhatrevi is that a future enhancement? Currently it looks like the image verification routines just use the TOC to find the FMC and FW size.

mhatrevi:
> @mhatrevi is that a future enhancement? Currently it looks like the image verification routines just use the TOC to find the FMC and FW size.

Yes, there are couple of TODO checks in verifier.rs regarding this.

bluegate010:
Ah, understood.

PR #242, opened 2023-05-05
Introducing the error crate

rusty1968:
The error crate is a central error definition and management repository.  
1. Move Caliptra error definition to error crate.
2. Use newtype pattern for CaliptraError.
3. All components defined in error crate.


rusty1968:
@mhatrevi , @korran  : can you leave a review?

mhatrevi (commented):
Can the Integration tests (at rom/dev/tests) be also updated to use the error codes from the error crate?

rusty1968:
> Can the Integration tests (at rom/dev/tests) be also updated to use the error codes from the error crate?

HwModel does not use the error crate.

Issue #162, opened 2023-04-08
Report all FMC-relevant evidence in the FMC alias certificate


bluegate010:
Currently, the following information is used to derive the FMC CDI:

* Lifecycle state
* Debug lock state
* Anti-rollback disable state
* Vendor public key hash
* Owner public key hash
* Vendor ECC public key index
* FMC hash
* FMC SVN

However, only the following information is placed as evidence in the FMC alias certificate:

* FMC hash
* Owner public key hash

These two lists should be consistent. Any data used to derive the CDI, but not placed in the certificate over the resulting public key, cannot functionally be attested.

There are some options:

1. Add a third digest to the alias certificate, which represents a hash of all the data in the first bullet list. The two existing digests are present only as a matter of convenience.
2. Remove the two existing digests in favor of a single digest over all data in the first bullet list.
3. Leave the FMC hash digest in place in the cert, and repurpose the other digest to be over all data in the first bullet list.

Option 1 presents a complication: we will need some way to ferry the new hash from ROM to FMC. For the other data in the certificate, this is currently done by way of Data Vault slots. However, we are currently at capacity for Data Vault slots. We would need to modify RTL to allocate at least one more digest-sized slot to hold this new information that FMC will need to reconstitute the certificate.

Alternatively, we could stash this data in a PCR. The issue there is that the data that FMC pulls out and places in the certificate will not be SHA384(...); rather, it will be SHA384(0x0000 || SHA384(...)) - the result of a hash extend.

One other option would be to expand the handoff table structure to include this new data. There is no special reason for the data to be locked away in a Data Vault slot. If it mutates after ROM uses it, the certificate that FMC reconstitutes will not have a valid signature.

bluegate010:
Related question: In DICE, the SVN field is an ASN.1 Integer, meaning it's variable-length. If we want to represent the FMC's SVN in the cert, I see a few options:

* Do not use DICE's SVN field. Add the SVN into a hash of other values, and put the result in the cert as a digest.
* Add a stipulation that Caliptra SVNs never grow past 127.
* Rework the X.509 serialization library to allow for variable-length fields (not actually advocating for this but just noting it for completeness' sake).

varuns-nvidia:
> Related question: In DICE, the SVN field is an ASN.1 Integer, meaning it's variable-length. If we want to represent the FMC's SVN in the cert, I see a few options:

[...]

PR #243, opened 2023-05-06
[RFC] Expose fuse-svn from verifier logic and represent in DICE evidence.

bluegate010:
This commit also reorganizes the evidence across multiple DiceTcbInfos, to avoid overloading the `svn` field.

Partially addresses https://github.com/chipsalliance/caliptra-sw/issues/162


image/verify/src/verifier.rs
```diff
@@ -595,6 +609,21 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
             .sha384_digest(image, range.start, range.len() as u32)
             .map_err(|_| err_u32!(VendorPubKeyDigestFailure))
     }
+
+    /// Calculates the effective minimum SVN.
+    ///
+    /// If anti-rollback is disabled, the effective min-SVN is zero.
+    /// Otherwise, it is MAX(min-SVN, SVN-fuses), because SVN-fuses will
+    /// be updated to match min-SVN.
```
varuns-nvidia, line 617: Is this the same logic that BootROM will do for image authentication?

bluegate010: I would expect that boot ROM would not have to do the `MAX(min-SVN, SVN-fuses)` bit when authenticating the image.

This is where some feedback would be helpful. I went with MAX here because without it, if we just sample SVN-fuses, then on the first boot after min-SVN revs those fuses the DICE evidence may reflect what the fuses _were_, not what they are now (this assuming that SVN-fuses get blown after Caliptra boots, not before). A subsequent power cycle would cause the evidence to catch up to the fuse state.

varuns-nvidia: There's multiple layers here because what Caliptra sees are registers and not the fuses themselves. The registers are latched on Caliptra power-good and locked as part of the Boot FSM transition states. So there's no way for Caliptra ROM to know any fuse state change until the next power-good anyway.

Either ROM or FW acting as update authorizer needs to do the `MAX(min-SVN, SVN-fuses)` logic to handle hitless update cases, because min-SVN can rev on update reset via manifest, but SVN-fuses cannot. I prefer ROM does it so we're not implementing authorization logic in different places.

bluegate010: I see; it sounds like in either case, we should be writing this effective-min-SVN out to the DataVault.

Should we try and create a mechanism for Caliptra ROM to signal to SoC what the new value of SVN-fuses should be?

Also, I gather there's an assumption here that upon hitless update, while we can increment some "soft" min-SVN and stash it in the DataVault, we can't increment the "hard" min-SVN stored in fuses?

varuns-nvidia: > Should we try and create a mechanism for Caliptra ROM to signal to SoC what the new value of SVN-fuses should be?

That can be useful, but it's also possible for the SoC to just manage the fuses orthogonally from what Caliptra says.

> Also, I gather there's an assumption here that upon hitless update, while we can increment some "soft" min-SVN and stash it in the DataVault, we can't increment the "hard" min-SVN stored in fuses?

Yep. @vsonims @bharatpillilli for vis

bluegate010: Ack. So we should think about what kind of min-SVN we actually want to report. If we report min-SVN as present in the firmware header, then that's a directive to firmware saying "fuses should be at least this", but the fuses might not have been blown by the time Caliptra runs, and Caliptra can't see what its new value is, as it doesn't manage fuses. Also if fuses are X and min-SVN is X-1, that's no less trustworthy a state than if min-SVN were X. So ignoring fuses seems a bit weird.

We could report what the fuse values were at boot-up. If min-SVN caused fuses to increment in the current boot cycle, a power cycle would be needed to actually get the attested evidence to update.

We could report the soft-SVN latched in the data vault. We _can't_ report that on FMC's alias key cert, though, because we don't re-generate that across hitless updates. Would need FMC to sample it and put it in RT FW's alias cert in order for it to reflect reality.

I'm not sure what all would be interesting to a verifier. @ericeilertson, any thoughts?

bluegate010: Resolving this; we now use fuse-SVN as per https://github.com/chipsalliance/caliptra-sw/issues/269

FerralCoder (approved):
I think this looks good but I would like @fdamato to confirm some things that we had discussed earlier today before this gets merged. He said he will be able to do that sometime today.

image/verify/src/verifier.rs
```diff
@@ -589,6 +597,18 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
     }
+
+    /// Calculates the effective fuse SVN.
+    ///
+    /// If anti-rollback is disabled, the effective fuse-SVN is zero.
+    /// Otherwise, it is SVN-fuses.
+    fn effective_fuse_svn(fuse_svn: u32, anti_rollback_disable: bool) -> u32 {
+        if anti_rollback_disable {
+            0_u32
+        } else {
```
fdamato, line 608: I`m not sure I understand how this works. Per spec Runtime SVN is 128 bits, and FMC is 32 bits.
Also it gets later downcast to a u8

bluegate010: The bits start out in a unary representation, but the driver [converts](https://github.com/chipsalliance/caliptra-sw/blob/2e26d715dcbfa92c1e5e7eb3bcd2cda760db9792/drivers/src/fuse_bank.rs#L206-L236) them to numerical.

I picked u32 as the type just to mirror what's already present at this layer, but we could just change this to u8 here, to cut down on the number of different representations.

fdamato: I think sticking with either u32 or u8 might prevent any room for confusion...but I`m ok with that. Thanks

fdamato (approved):

image/verify/src/verifier.rs
bluegate010: Ack, will refactor to use u8 consistently in a follow-up PR (don't have bandwidth this week and I understand there's work that's blocked on this).

PR #244, opened 2023-05-06
Add KAT for LMS

### LINKED ISSUE #154: ROM KAT support for LMS
Known Answer Tests as a pre-condition for FIPS certifiabilityu

ericeilertson:
Fixes #154.


kat/src/lms_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+#[derive(Default, Debug)]
+pub struct LmsKat {}
+
+impl LmsKat {
+    /// This function executes the Known Answer Tests (aka KAT) for LMS.
+    pub fn execute(&self, lms: &Lms) -> CaliptraResult<()> {
+        self.kat_lms_24(lms)
+    }
+
+    fn _kat_lms_24(&self, _lms: &Lms) -> CaliptraResult<()> {
```
mhatrevi, line 39: Is this intended to be used?

mhatrevi (changes requested):

kat/src/lms_kat.rs
ericeilertson: No, it was left in there as a stub to essentially disable the LMS KAT since including the KAT code puts us right at the ROM size limit.  It can be removed if we think we are safely below the ROM limit, or people could remove the LMS kat in their on branches if they are hitting this limit.

ericeilertson: Will remove

mhatrevi (dismissed):

mhatrevi (approved):

mhatrevi (approved):

PR #247, opened 2023-05-08
[fix] Fix For Error Code Collision Due To 8-bit Component Code

ajisaxena:
The component names were 8-bit fields and the associated code for the components were more than 8-bits. Due to this, the components names greater than 256 were overflowing, rendering the error codes incorrect.


mhatrevi (dismissed):

rusty1968 (dismissed):

rom/dev/src/error.rs
```diff
@@ -43,7 +43,7 @@ macro_rules! rom_err_def {
         impl From<$enum_name> for core::num::NonZeroU32 {
             fn from(val: $enum_name) -> Self {
                 // Panic is impossible as long as the enums don't define zero.
-                core::num::NonZeroU32::new(((($crate::error::RomComponent::$comp_name) as u32) << 24) | (val as u32)).unwrap()
+                core::num::NonZeroU32::new(((($crate::error::RomComponent::$comp_name) as u32) << 16) | (val as u32)).unwrap()
```
korran, line 46: Shouldn't this be consistent with this macro in other files such as 

https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/error.rs#L76

?

ajisaxena: I looked at that code. The number of components in the drivers are very small and I thought that 8 bits should be enough for that. Actually, we first started with this way of defining errors from the driver module and carried over all other places and then introduced this error. I believe that drivers should be ok. But please let me know if we want to make it consistent across the board so that it is easier to decipher the error code.

korran: I think it needs to be consistent to avoid collisions? 

With your PR as is, InitDevId and DeobfuscationEngine components both have the same 16-bit prefix 0x0100

ajisaxena: Let me check and change it to 16 bits of component id.

rusty1968: The error encoding is going to change in the upcoming error crate.

mhatrevi (approved):

rom/dev/src/error.rs
ajisaxena: @rusty1968 I think that is the right way to go about this. Else it will be very hard to avoid collisions.

ajisaxena:
Closing the PR because the latest merge with main caused a panic to be introduced.

PR #249, opened 2023-05-09
Misc size optimizations

korran:


rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -58,9 +58,7 @@ impl Crypto {
     ///
     /// * `Array4x5` - Digest
     pub fn sha1_digest(env: &RomEnv, data: &[u8]) -> CaliptraResult<Array4x5> {
```
bluegate010, line 60: Not something to do as part of this commit, but it's arguable whether these helpers even need to exist.

korran: Ack. This isn't the only vestigial abstraction laying around. Thankfully they get inlined to nothing in the machine code.

bluegate010 (approved):

vsonims (approved):

PR #251, opened 2023-05-10
Add ECDSA verify mailbox command

jhand2:
Commit 1: Add support to sw-emulator for using the sha512acc from SoC in streaming mode

Commit 2: Add scaffolding more mailbox commands and implement the ECDSA verify command. The signature is assumed to be over the data hashed into the sha512acc peripheral.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -3,14 +3,14 @@
 #![no_std]
 
 mod mailbox;
+mod verify;
 
 use mailbox::Mailbox;
 
 use caliptra_common::cprintln;
 use caliptra_drivers::{caliptra_err_def, CaliptraResult};
-use caliptra_registers::mbox::enums::MboxStatusE;
-
-use core::mem::size_of;
+use caliptra_registers::{mbox::enums::MboxStatusE, soc_ifc};
+use zerocopy::{AsBytes, FromBytes};
 
 caliptra_err_def! {
     Runtime,
@@ -18,7 +18,60 @@ caliptra_err_def! {
     {
         // Internal
         InternalErr = 0x1,
+        UnimplementedCommand = 0x2,
+        InsufficientMemory = 0x3,
     }
+}
+
+#[derive(PartialEq, Eq)]
+pub struct CommandId(pub u32);
+
+impl CommandId {
+    pub const FIRMWARE_LOAD: Self = Self(0x46574C44); // "FWLD"
+    pub const GET_IDEV_CSR: Self = Self(0x49444556); // "IDEV"
+    pub const GET_LDEV_CERT: Self = Self(0x4C444556); // "LDEV"
+    pub const ECDSA384_VERIFY: Self = Self(0x53494756); // "SIGV"
+    pub const STASH_MEASUREMENT: Self = Self(0x4D454153); // "MEAS"
+    pub const INVOKE_DPE: Self = Self(0x44504543); // "DPEC"
+}
+impl From<u32> for CommandId {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+impl From<CommandId> for u32 {
+    fn from(value: CommandId) -> Self {
+        value.0
+    }
+}
+
+#[repr(C)]
+#[derive(AsBytes, FromBytes)]
+pub struct EcdsaVerifyCmd {
+    pub chksum: u32,
+    pub pub_key_x: [u8; 48],
+    pub pub_key_y: [u8; 48],
+    pub signature_r: [u8; 48],
+    pub signature_s: [u8; 48],
+}
+
+impl Default for EcdsaVerifyCmd {
+    fn default() -> Self {
+        Self {
+            chksum: 0,
+            pub_key_x: [0u8; 48],
+            pub_key_y: [0u8; 48],
+            signature_r: [0u8; 48],
+            signature_s: [0u8; 48],
+        }
+    }
+}
+
+#[repr(C)]
+#[derive(AsBytes, FromBytes)]
+pub struct EcdsaVerifyResponse {
+    pub chksum: u32,
+    pub result: u32,
 }
 
 fn wait_for_cmd() {
@@ -29,21 +82,40 @@ fn wait_for_cmd() {
     //}
 }
 
-fn handle_command() -> CaliptraResult<()> {
+fn handle_command() -> CaliptraResult<MboxStatusE> {
     let cmd_id = Mailbox::cmd();
+    let dlen = Mailbox::dlen() as usize;
     let dlen_words = Mailbox::dlen_words() as usize;
     let mut buf = [0u32; 1024];
     Mailbox::copy_from_mbox(buf.get_mut(..dlen_words).ok_or(err_u32!(InternalErr))?);
 
-    // TODO: Actually handle command
-    cprintln!("[rt] Received command={}, len={}", cmd_id, Mailbox::dlen());
+    if dlen > buf.len() * 4 {
+        // dlen larger than max message
+        return Err(err_u32!(InsufficientMemory));
+    }
 
-    // Write response
-    let out_buf = [0xFFFFFFFFu32; 4];
-    Mailbox::set_dlen((out_buf.len() * size_of::<u32>()) as u32);
-    Mailbox::copy_to_mbox(&out_buf);
+    let cmd_bytes = buf
+        .as_bytes()
+        .get(..dlen)
+        .ok_or(err_u32!(InsufficientMemory))?;
 
-    Ok(())
+    cprintln!(
+        "[rt] Received command=0x{:x}, len={}",
+        cmd_id,
+        Mailbox::dlen()
+    );
+    match CommandId::from(cmd_id) {
+        CommandId::FIRMWARE_LOAD => Err(err_u32!(UnimplementedCommand)),
+        CommandId::GET_IDEV_CSR => Err(err_u32!(UnimplementedCommand)),
+        CommandId::GET_LDEV_CERT => Err(err_u32!(UnimplementedCommand)),
+        CommandId::ECDSA384_VERIFY => {
+            verify::handle_ecdsa_verify(cmd_bytes)?;
+            Ok(MboxStatusE::DataReady)
+        }
+        CommandId::STASH_MEASUREMENT => Err(err_u32!(UnimplementedCommand)),
+        CommandId::INVOKE_DPE => Err(err_u32!(UnimplementedCommand)),
+        _ => Err(err_u32!(UnimplementedCommand)),
+    }
 }
 
 pub fn handle_mailbox_commands() {
@@ -51,10 +123,16 @@ pub fn handle_mailbox_commands() {
         wait_for_cmd();
 
         if Mailbox::is_cmd_ready() {
-            if handle_command().is_ok() {
-                Mailbox::set_status(MboxStatusE::DataReady);
-            } else {
-                Mailbox::set_status(MboxStatusE::CmdFailure);
+            match handle_command() {
+                Ok(status) => {
+                    Mailbox::set_status(status);
+                }
+                Err(e) => {
+                    let soc_ifc_regs = soc_ifc::RegisterBlock::soc_ifc_reg();
+                    soc_ifc_regs.cptra_fw_error_non_fatal().write(|_| e.into());
+
+                    Mailbox::set_status(MboxStatusE::CmdFailure);
+                }
             }
         }
     }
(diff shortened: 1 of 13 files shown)


runtime/tests/integration_tests.rs
```diff
@@ -60,15 +66,58 @@ fn run_rt_test(test_bin_name: Option<&str>) -> DefaultHwModel {
...
+
+    model.step_until(|m| m.soc_sha512_acc().status().read().valid());
+
+    let digest = model.soc_sha512_acc().digest().read();
+    for word in digest {
+        println!("0x{:x}", word);
+    }
+
+    model.soc_sha512_acc().lock().write(|w| w.lock(true)); // clear lock
+}
```
jhand2, line 99: @korran @mhatrevi when I run this on verilator, it hangs on line 91 (valid bit never gets set). I tried to poke around at the waveform to see if anything is going on, but it's hard to tell.

Anything I'm doing here look obviously wrong? (Note, I don't do anything in firmware to initialize the sha512acc peripheral, but it didn't seem like anything was necessary)

jhand2: Did some more digging: https://github.com/chipsalliance/caliptra-rtl/issues/84

I tweaked the RTL to just always set `valid_user`, and everything works. So either there's an issue in the RTL or there's something else I should be doing to ensure `valid_user` gets set.

jhand2: Resolved in https://github.com/chipsalliance/caliptra-rtl/issues/84

sw-emulator/lib/periph/Cargo.toml
```diff
@@ -19,3 +19,4 @@ smlang="0.6.0"
 lazy_static = "1.4.0"
 arrayref = "0.3.6"
 bitfield = "0.14.0"
+sha2 = { version = "0.10.2", default-features = false, features = ["compress"] }
```
korran, line 22: Is this being used?

sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -34,7 +36,7 @@ const SHA_ACC_OP_TICKS: u64 = 1000;
 
 const SHA512_BLOCK_SIZE: usize = 128;
 const SHA512_HASH_SIZE: usize = 64;
-const SHA384_HASH_SIZE: usize = 48;
+const _SHA384_HASH_SIZE: usize = 48;
```
korran, line 39: I don't think this needs to be _SHA384_HASH_SIZE.

runtime/src/lib.rs
```diff
@@ -29,32 +85,57 @@ fn wait_for_cmd() {
...
+    }
 
-    // Write response
-    let out_buf = [0xFFFFFFFFu32; 4];
-    Mailbox::set_dlen((out_buf.len() * size_of::<u32>()) as u32);
-    Mailbox::copy_to_mbox(&out_buf);
+    // Convert command buffer to byte slice this is safe since we have already
+    // confirmed the buffer is longer than dlen
+    let cmd_bytes: &[u8] =
+        unsafe { ::core::slice::from_raw_parts(buf.as_ptr() as *const u8, dlen) };
```
korran, line 103: This code is unsound because the lifetime of cmd_bytes is not tied to buf, allowing cmd_bytes to alias modifications to buf. (I can modify buf while the cmd_bytes reference is alive).

Please don't use unsafe code in app code. In this case, you should probably use zerocopy to do this:

```
let cmd_bytes = &buf.as_bytes()[..dlen];
```

korran: unsafe code should be placed into its own module, and have extensive test cases that execute under MIR to check for undefined behavior.

korran: FWIW, I would probably write this as:

```
    let Some(cmd_bytes) = buf.as_bytes().get(..dlen) else {
        return Err(err_u32!(InsufficientMemory));
    };
```

runtime/src/lib.rs
```diff
@@ -3,24 +3,80 @@
...
     RuntimeErr
     {
         // Internal
         InternalErr = 0x1,
+        UnimplementedCommand = 0x2,
+        InsufficientMemory = 0x3,
+    }
+}
+
+pub enum CommandId {
```
korran, line 26: I generally prefer the open enum pattern for this sort of thing:

```
pub struct CommandId(pub u32);
impl CommandId {
    pub const FIRMWARE_LOAD: Self = Self(0x46574C44);     // "FWLD"
    pub const GET_IDEV_CERT: Self = Self(0x49444556);       // "IDEV"
    pub const GET_LDEV_CERT: Self = Self(0x4C444556);      // "LDEV"
    pub const ECDSA384_VERIFY: Self = Self(0x53494756);   // "SIGV"
    pub const INVOKE_DPE: Self = Self(0x44504543);        // "DPEC"
}
impl From<u32> for CommandId {
    fn from(value: u32) -> Self {
        Self(value)
    }
}
impl From<CommandId> for u32 {
    fn from(value: CommandId) -> Self {
        value.0
    }
}
```

runtime/tests/integration_tests.rs
```diff
@@ -62,15 +65,55 @@ fn run_rt_test(test_bin_name: Option<&'static str>) -> DefaultHwModel {
...
+    model.step_until(|m| m.soc_sha512_acc().status().read().valid());
+
+    model.soc_sha512_acc().lock().write(|w| w.lock(true)); // clear lock
+}
+
+/// Sends a SoC -> uC command to the mailbox
+///
+/// If the command succeeds, returns the response buffer. On error, returns the
+/// SoC ifc firmware error code.
+fn send_mailbox_command(
```
korran, line 101: What's wrong with HwModel::mailbox_execute?

runtime/tests/integration_tests.rs
```diff
@@ -62,15 +65,55 @@ fn run_rt_test(test_bin_name: Option<&'static str>) -> DefaultHwModel {
...
-    model.soc_mbox().cmd().write(|_| cmd_id);
+    model.soc_mbox().cmd().write(|_| cmd as u32);
     model
         .soc_mbox()
         .dlen()
-        .write(|_| u32::try_from(arg_buf.len()).unwrap() * 4);
+        .write(|_| u32::try_from(cmd_buf.len()).unwrap());
 
-    for word in arg_buf {
+    let (prefix, words, suffix) = unsafe { cmd_buf.align_to::<u32>() };
```
korran, line 113: I don't think you need to hyper-optimize this with unsafe code. Use zerocopy, or use HwModel::mailbox_execute() directly.

runtime/tests/integration_tests.rs
korran: AFAICT, this works fine:

```
    let resp_bytes: Vec<u8> = model.mailbox_execute(CommandId::Ecdsa384Verify as u32, cmd.as_bytes()).unwrap().unwrap();
    let response = EcdsaVerifyResponse::read_from(&resp_bytes[..]).unwrap();
```

runtime/src/lib.rs
```diff
@@ -3,24 +3,80 @@
...
+    FirmwareLoad = 0x46574C44,     // "FWLD"
```
korran, line 27: Note: in "hexdump -C", this little-endian value will appear as "DLWF"; do we care?

IIRC, convention with FourCC is to ensure they are human-readable when serialized.

sw-emulator/lib/periph/src/sha512_acc.rs
jhand2: It's only used in tests (because Sha512AcceleratorRegs::_hash_len/_copy_hash are only used in tests). So rustc complains if it doesn't have the underscore.

runtime/src/lib.rs
jhand2: https://github.com/chipsalliance/caliptra-sw/issues/275

sw-emulator/lib/crypto/src/sha512.rs
```diff
@@ -118,14 +126,60 @@ impl Sha512 {
...
+    }
+
+    /// Update the hash with an arbitrary number to bytes
+    ///
+    /// `bytes` is expected to be big-endian data
+    pub fn update_bytes(&mut self, bytes: &[u8]) {
+        self.partial_block.extend_from_slice(bytes);
+
+        while self.partial_block.len() >= Self::BLOCK_SIZE {
+            // Safe to unwrap becasue slice is guaranteed to be correct size
```
korran, line 146: Nit: it's possible to use https://doc.rust-lang.org/std/primitive.slice.html#method.get_mut here to get rid of the unwrap

sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -456,31 +493,61 @@ impl Sha512Accelerator {
     }
 
     /// Get the length of the hash
-    pub fn hash_len(&self) -> usize {
+    pub fn _hash_len(&self) -> usize {
```
korran, line 496: Did you intend to delete these? If these are only used in tests perhaps #[cfg(test)] is appropriate?

sw-emulator/lib/periph/src/sha512_acc.rs
korran: Consider using #[cfg(test)]

runtime/src/mailbox.rs
```diff
@@ -42,6 +44,23 @@ impl Mailbox {
...
+    /// Write a word-aligned `buf` to the mailbox
+    pub fn write_response(buf: &[u8]) -> CaliptraResult<()> {
+        if buf.len() % 4 != 0 {
+            return Err(RuntimeErr::InternalErr.into());
+        }
+
+        // Convert command buffer to word slice. This is safe since we have already
+        // confirmed the buffer size is a multiple of word size.
+        let buf_words: &[u32] =
+            unsafe { ::core::slice::from_raw_parts(buf.as_ptr() as *const u32, buf.len()) };
```
korran, line 56: This is undefined behavior, as you haven't guaranteed buf_words has u32 alignment.

I think maybe there's something we can do here without unsafe using &[zerocopy::Unalign<u32>]. Now that we have target-feature=+unaligned-scalar-mem, the code-gen for that type should look pretty good.

korran (changes requested):

runtime/src/mailbox.rs
korran: Something like this should work:

```diff
diff --git a/runtime/src/mailbox.rs b/runtime/src/mailbox.rs
index c221d0fe..5494cbca 100644
--- a/runtime/src/mailbox.rs
+++ b/runtime/src/mailbox.rs
@@ -3,6 +3,7 @@
 use crate::RuntimeErr;
 use caliptra_drivers::CaliptraResult;
 use caliptra_registers::mbox::{self, enums::MboxStatusE};
+use zerocopy::{LayoutVerified, Unalign};
 
 pub struct Mailbox {}
 
@@ -38,25 +39,20 @@ impl Mailbox {
         }
     }
 
-    pub fn copy_to_mbox(buf: &[u32]) {
+    pub fn copy_to_mbox(buf: &[Unalign<u32>]) {
         for word in buf {
-            mbox::RegisterBlock::mbox_csr().datain().write(|_| *word);
+            mbox::RegisterBlock::mbox_csr().datain().write(|_| word.get());
         }
     }
 
     /// Write a word-aligned `buf` to the mailbox
     pub fn write_response(buf: &[u8]) -> CaliptraResult<()> {
-        if buf.len() % 4 != 0 {
+        let Some(buf_words) = LayoutVerified::new_slice_unaligned(buf) else {
+            // The size isn't a multiple of the word-size
             return Err(RuntimeErr::InternalErr.into());
-        }
-
-        // Convert command buffer to word slice. This is safe since we have already
-        // confirmed the buffer size is a multiple of word size.
-        let buf_words: &[u32] =
-            unsafe { ::core::slice::from_raw_parts(buf.as_ptr() as *const u32, buf.len()) };
-
+        };
         Self::set_dlen(buf.len() as u32);
-        Self::copy_to_mbox(buf_words);
+        Self::copy_to_mbox(&buf_words);
 
         Ok(())
     }
```

sw-emulator/lib/crypto/src/sha512.rs
jhand2: The borrow checker doesn't like

```rust
 while let Some(block) = self.partial_block.get_mut(..Self::BLOCK_SIZE) {
    self.update(block.try_into().unwrap());
}
```

Because we have a mutable reference to `self` (from get_mut) and `self.update` also takes a mutable reference to `self`. I'm not sure if there's a better way to do this.

korran: Yeah, the borrow checker doesn't seem to offer any mechanism to demote the &mut reference to a &.

korran (approved):

PR #258, opened 2023-05-10
[fix] Fix For Error Code Collision Due To 8-bit Component Code

ajisaxena:
The component names were 8-bit fields and the associated code for the components were more than 8-bits. Due to this, the components names greater than 256 were overflowing, rendering the error codes incorrect.


mhatrevi (dismissed):

ajisaxena:
Problems merging with main. Will raise another PR for this.

PR #259, opened 2023-05-11
[feat] ROM PCR logging

mhatrevi:
This change logs a PCR entry to DCCM when a PCR is extended.

### DIFF
--- a/common/src/pcr.rs
+++ b/common/src/pcr.rs
@@ -1,0 +1,66 @@
+/*++
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    pcr.rs
+
+Abstract:
+
+    PCR-related types.
+
+--*/
+
+use zerocopy::{AsBytes, FromBytes};
+
+// PcrLogEntryId is used to identify the PCR entry and
+// the size of the data in PcrLogEntry::pcr_data.
+//
+// For valid entries, it is also used as the index into the PCR log as per the formula:
+//      log_entry_index = pcr_entry_id - 1
+#[repr(u16)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PcrLogEntryId {
+    Invalid = 0,
+    DeviceLifecycle = 1,      // data size = 1 byte
+    DebugLocked = 2,          // data size = 1 byte
+    AntiRollbackDisabled = 3, // data size = 1 byte
+    VendorPubKeyHash = 4,     // data size = 48 bytes
+    OwnerPubKeyHash = 5,      // data size = 48 bytes
+    VendorPubKeyIndex = 6,    // data size = 1 byte
+    FmcTci = 7,               // data size = 48 bytes
+    FmcSvn = 8,               // data size = 1 byte
+}
+
+impl From<u16> for PcrLogEntryId {
+    /// Converts to this type from the input type.
+    fn from(id: u16) -> PcrLogEntryId {
+        match id {
+            1 => PcrLogEntryId::DeviceLifecycle,
+            2 => PcrLogEntryId::DebugLocked,
+            3 => PcrLogEntryId::AntiRollbackDisabled,
+            4 => PcrLogEntryId::VendorPubKeyHash,
+            5 => PcrLogEntryId::OwnerPubKeyHash,
+            6 => PcrLogEntryId::VendorPubKeyIndex,
+            7 => PcrLogEntryId::FmcTci,
+            8 => PcrLogEntryId::FmcSvn,
+            _ => PcrLogEntryId::Invalid,
+        }
+    }
+}
+
+/// PCR log entry
+#[repr(C)]
+#[derive(AsBytes, FromBytes, Default, Debug)]
+pub struct PcrLogEntry {
+    /// Entry identifier
+    pub id: u16,
+
+    /// PCR to which the data is being extended to.
+    pub pcr_id: u16,
+
+    // PCR data
+    pub pcr_data: [u32; 12],
+
+    pub reserved: [u8; 4],
+}
--- a/rom/dev/src/pcr.rs
+++ b/rom/dev/src/pcr.rs
@@ -22,7 +22,23 @@ Note:
 --*/
 
 use crate::rom_env::RomEnv;
+use caliptra_common::{PcrLogEntry, PcrLogEntryId};
 use caliptra_drivers::{Array4x12, CaliptraResult, PcrId};
+use caliptra_error::caliptra_err_def;
+use zerocopy::AsBytes;
+
+caliptra_err_def! {
+    RomGlobal,
+    GlobalErr
+    {
+        PcrLogInvalidEntryId = 0x4,
+        PcrLogUnsupportedDataLength = 0x5,
+    }
+}
+
+extern "C" {
+    static mut PCR_LOG_ORG: u8;
+}
 
 /// Extend PCR0
 ///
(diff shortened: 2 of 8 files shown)


common/src/dice.rs
```diff
@@ -12,7 +12,38 @@ Abstract:
...
+use zerocopy::{AsBytes, FromBytes};
+
 pub const FLAG_BIT_NOT_CONFIGURED: u32 = 1 << 0;
 pub const FLAG_BIT_NOT_SECURE: u32 = 1 << 1;
 pub const FLAG_BIT_DEBUG: u32 = 1 << 3;
 pub const FLAG_BIT_FIXED_WIDTH: u32 = 1 << 31;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PcrEntryId {
+    DeviceLifecycle = 0,
```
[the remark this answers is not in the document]
mhatrevi: Moved the pcr logging common types to a new file, pcr.rs.

rom/dev/tests/test_fmcalias_derivation.rs
```diff
@@ -1,8 +1,14 @@
...
+use caliptra_common::PcrLogEntry;
+use caliptra_common::PcrLogEntryId::*;
+use caliptra_hw_model::{BootParams, Fuses, HwModel, InitParams, ModelError, SecurityState};
+use caliptra_image_fake_keys::VENDOR_CONFIG_KEY_1;
+use caliptra_image_gen::ImageGenerator;
+use caliptra_image_openssl::OsslCrypto;
 use caliptra_image_types::IMAGE_BYTE_SIZE;
+use zerocopy::{AsBytes, FromBytes};
 
 pub mod helpers;
```
ajisaxena, line 14: [TODO] Use the error codes from the common library.
Can we resolve this as a part of this check in or you want to do it separately?

ajisaxena (dismissed):

common/src/pcr.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Entry identifier
+    pub id: u16,
+
+    /// PCR to which the data is being extended to.
+    pub pcr_id: u16,
+
+    // PCR data
+    pub pcr_data: [u32; 12],
+
+    /// PCR metadata
```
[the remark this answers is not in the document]
bluegate010: I suspect this is intended to map to the arg of the same name in `STASH_MEASUREMENT`. That arg was introduced to map to DPE's `type` arg.

Given that we intend that arg to carry evidentiary value and be attested along with the actual measurement, it's a little bit odd to have it here but _not_ have it be extended into the PCR. One option would be to remove it from here, and if we need to mix in that data we can do so by appending it into `pcr_data`.

rom/dev/src/pcr.rs
```diff
@@ -39,26 +55,97 @@ pub fn extend_pcr0(env: &RomEnv) -> CaliptraResult<()> {
...
+///
+/// # Return Value
+/// * `Ok(())` - Success
+/// * `Err(GlobalErr::PcrLogInvalidEntryId)` - Invalid PCR log entry ID
+/// * `Err(GlobalErr::PcrLogUpsupportedDataLength)` - Unsupported data length
+///
+pub fn log_pcr(
+    pcr_entry_id: PcrLogEntryId,
+    pcr_id: PcrId,
+    data: &[u8],
```
bluegate010, line 121: I don't see where in this function we're preserving the length of `data`, which will be important for verification. I see the test currently assumes the length of each `pcr_data` field.

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -178,3 +183,74 @@ fn copy_tbs(tbs: &mut [u8], ldevid_tbs: bool) {
...
+            // No-op
+        }
+    }
+}
+
+fn read_pcr_log(mbox: &caliptra_registers::mbox::RegisterBlock) {
+    let mut pcr_entry_count = 0;
+    loop {
+        let pcr_entry = get_pcr_entry(pcr_entry_count);
+        if PcrLogEntryId::from(pcr_entry.id) != PcrLogEntryId::Invalid {
```
bluegate010, line 224: Optional nit: consider inverting this logic and breaking early, can un-indent this whole loop.

```
loop {
  let pcr_entry = get_pcr_entry(pcr_entry_count);
  if PcrLogEntryId::from(pcr_entry.id) == PcrLogEntryId::Invalid {
    break;
  }
  pcr_entry_count += 1;
  ...
}

File: rom/dev/tests/test_fmcalias_derivation.rs [hunk 6]
```diff
@@ -53,3 +59,142 @@ fn test_firmware_gt_max_size() {
...
+
+    // Check PCR entry for FmcSvn.
+    pcr_log_entry_offset += core::mem::size_of::<PcrLogEntry>();
+    let pcr_log_entry =
+        PcrLogEntry::read_from_prefix(pcr_entry_arr[pcr_log_entry_offset..].as_bytes()).unwrap();
+    assert_eq!(pcr_log_entry.id, FmcSvn as u16);
+    assert_eq!(pcr_log_entry.metadata.as_bytes(), &[0u8; 4]);
+    assert_eq!(pcr_log_entry.pcr_id, 0);
+    assert_eq!(pcr_log_entry.pcr_data[0], FMC_SVN);
+}
```
bluegate010, line 200: Would it be possible to have a test that hashes the data from the event log and ensures that it corresponds with the current value of the hardware PCRs?

rom/dev/src/pcr.rs
mhatrevi: Data length is implied by the type of the log entry (identified by pcr_entry_id). If this is not sufficient, I can add a size field.

rom/dev/tests/test_fmcalias_derivation.rs
mhatrevi: Discussed offline; this TODO will be addressed with the common error crate changes.

common/src/pcr.rs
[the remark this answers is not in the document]
mhatrevi: This is correct; it's to map the metadata arg from STASH_MEASUREMENT command. IMO, we can keep this for now and come back to this when there is resolution on the measurement stashing ask.

rom/dev/tests/test_fmcalias_derivation.rs
mhatrevi: Yeah, should be doable.

common/src/pcr.rs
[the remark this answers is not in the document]
bluegate010: If we keep this field, could we call it 'reserved'?

rom/dev/src/pcr.rs
bluegate010: I don't _believe_ Caliptra will be accepting variable-length measurements, so having the size be implicit may work. Up to you. If the size is implicit then that should be documented where the IDs are defined.

rom/dev/src/pcr.rs
```diff
@@ -39,26 +55,97 @@ pub fn extend_pcr0(env: &RomEnv) -> CaliptraResult<()> {
     // Lock the PCR from clear
     pcr_bank.map(|p| p.set_pcr_lock(caliptra_drivers::PcrId::PcrId0));
 
-    let extend = |data: Array4x12| {
+    let extend = |data: Array4x12, pcr_entry_id: PcrLogEntryId| {
```
bluegate010, line 58: Optional but I feel like it might be useful to have the entry ID be measured as well, since it provides context for the bits that follow.

mhatrevi: We'll take this as a separate change. Can you file an issue for this?

common/src/pcr.rs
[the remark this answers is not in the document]
mhatrevi: Done.

rom/dev/src/pcr.rs
mhatrevi: Added size info.

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -178,3 +183,74 @@ fn copy_tbs(tbs: &mut [u8], ldevid_tbs: bool) {
...
 }
+
+fn get_pcr_entry(entry_index: usize) -> PcrLogEntry {
+    // Copy the pcr log entry from DCCM
+    let mut pcr_entry: [u8; core::mem::size_of::<PcrLogEntry>()] =
+        [0u8; core::mem::size_of::<PcrLogEntry>()];
+
+    let src = unsafe {
+        let offset = core::mem::size_of::<PcrLogEntry>() * entry_index;
+        let ptr = (&mut PCR_LOG_ORG as *mut u8).add(offset);
```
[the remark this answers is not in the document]
mhatrevi: Discussed offline, will document the DCCm regions. For additional PCR extensions, there will be some from processing the stash measurement command(s).

rom/dev/src/pcr.rs
bluegate010: Filed at https://github.com/chipsalliance/caliptra-sw/issues/290

rom/dev/src/pcr.rs
```diff
@@ -39,26 +55,96 @@ pub fn extend_pcr0(env: &RomEnv) -> CaliptraResult<()> {
...
+/// # Return Value
+/// * `Ok(())` - Success
+/// * `Err(GlobalErr::PcrLogInvalidEntryId)` - Invalid PCR log entry ID
+/// * `Err(GlobalErr::PcrLogUpsupportedDataLength)` - Unsupported data length
+///
+pub fn log_pcr(
+    pcr_entry_id: PcrLogEntryId,
+    pcr_id: PcrId,
+    data: &[u8],
+    _metadata: u32,
```
bluegate010, line 122: Optional: consider removing this for now. I suspect that when it gets wired up to `STASH_MEASUREMENT`, it will not be needed. Can always add it back without messing with the PCR log format.

common/src/pcr.rs
```diff
@@ -0,0 +1,64 @@
...
+
+Abstract:
+
+    PCR-related types.
+
+--*/
+
+use zerocopy::{AsBytes, FromBytes};
+
+// PcrLogEntryId is used to identify the PCR entry.
```
bluegate010, line 16: Consider, "used to identify the PCR entry's semantics and size."

mhatrevi: Updated.

rom/dev/src/pcr.rs
mhatrevi: Remmoved.

rom/dev/tests/test_fmcalias_derivation.rs
mhatrevi: I'll add this as a separate change; Filed https://github.com/chipsalliance/caliptra-sw/issues/296 to track it.

bluegate010 (dismissed):

bluegate010 (approved):

ajisaxena (approved):

PR #260, opened 2023-05-11
Emulator fixes

mhatrevi:
This change contains the fix to bring back CSR download to emulator app.

### DIFF
--- a/sw-emulator/app/src/main.rs
+++ b/sw-emulator/app/src/main.rs
@@ -14,9 +14,10 @@ Abstract:
 
 use caliptra_emu_bus::{Bus, Clock};
 use caliptra_emu_cpu::{Cpu, RvInstr, StepAction};
+use caliptra_emu_periph::soc_reg::DebugManufService;
 use caliptra_emu_periph::{
-    CaliptraRootBus, CaliptraRootBusArgs, MailboxInternal, ReadyForFwCb, TbServicesCb,
-    UploadUpdateFwCb,
+    CaliptraRootBus, CaliptraRootBusArgs, DownloadIdevidCsrCb, MailboxInternal, ReadyForFwCb,
+    TbServicesCb, UploadUpdateFwCb,
 };
 use caliptra_emu_types::{RvAddr, RvSize};
 use caliptra_hw_model::BusMmio;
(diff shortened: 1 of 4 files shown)


sw-emulator/app/src/main.rs
```diff
@@ -12,12 +12,13 @@ Abstract:
```
korran, line 12: Can you change this commit title from "[fix] Emulator should use external mailbox bus" to "[fix] sw-emulator app should use external mailbox bus"

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -723,6 +746,19 @@ impl SocRegistersImpl {
...
+
+        if self
+            .timer
+            .fired(&mut self.op_idevid_csr_read_complete_action)
+        {
+            // Download the Idevid CSR from the mailbox.
+            (self.download_idevid_csr_cb)(&mut self.mailbox, &self.log_dir);
+
+            // Clear the Idevid CSR requested bit.
+            self.cptra_dbg_manuf_service_reg
```
korran, line 758: I don't think the verilog does this. Shouldn't this be done by the callback (maybe by passing in this register as an argument)?

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -441,6 +443,9 @@ struct SocRegistersImpl {
     /// Timer
     timer: Timer,
 
+    /// Log Directory
+    log_dir: PathBuf,
```
korran, line 447: Why does this need to be here? Can't the callback just capture this in the app?

korran: Maybe this is used somewhere else in the emulator library?

sw-emulator/app/src/main.rs
rusty1968: This is already in main branch : https://github.com/chipsalliance/caliptra-sw/blob/main/sw-emulator/app/src/main.rs#L385

mhatrevi: Removed the commit under discussion.

sw-emulator/lib/periph/src/soc_reg.rs
mhatrevi: Updated.

sw-emulator/lib/periph/src/soc_reg.rs
mhatrevi: Fixed.

korran (approved):

PR #265, opened 2023-05-12
Fix emulator app callbacks

rusty1968:
Emulator app  callbacks and complete actions inside the library were not using MailboxExternal.


korran (changes requested):
I think https://github.com/chipsalliance/caliptra-sw/pull/266 is a better approach.

PR #267, opened 2023-05-12
Expose sw-emulator mbox regs via ureg. 

korran:
This reduces the amount of boilerplate code in emulator tests and glue
code.

### DIFF
--- a/sw-emulator/lib/periph/src/soc_reg.rs
+++ b/sw-emulator/lib/periph/src/soc_reg.rs
@@ -734,13 +734,13 @@ impl SocRegistersImpl {
 
         if self.timer.fired(&mut self.op_fw_read_complete_action) {
             // Receiver sets status as CMD_COMPLETE after reading the mailbox data.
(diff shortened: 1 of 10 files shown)


sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -734,9 +734,13 @@ impl SocRegistersImpl {
 
         if self.timer.fired(&mut self.op_fw_read_complete_action) {
             // Receiver sets status as CMD_COMPLETE after reading the mailbox data.
-            if !self.mailbox.is_status_cmd_busy() {
+            if !self.mailbox.regs().status().read().status().cmd_busy() {
```
rusty1968, line 737: Polling for busy is part of the sender protocol. We should change the comment and use soc_mailbox here.

korran: Added a followup commit to change this (I don't like to mix refactoring and functional changes in the same PR).

rusty1968 (approved):

PR #268, opened 2023-05-12
Add DPE as a submodule

sree-revoori1:


.gitmodules
```diff
@@ -5,3 +5,6 @@
 	path = hw-latest/caliptra-rtl
 	url = https://github.com/chipsalliance/caliptra-rtl
 	branch = main
+[submodule "caliptra-dpe"]
```
jhand2, line 8: Can you also add `branch = main`

.gitmodules
```diff
@@ -5,3 +5,6 @@
...
+	path = caliptra-dpe
```
jhand2, line 9: I think let's put this in the `runtime/` directory

jhand2 (commented):
Adding @korran as well. Is this fine, or would you rather we put this in the `runtime` directory

.gitmodules
korran: Either is fine with me. Can we also call the directory dpe instead of caliptra-dpe?

jhand2 (approved):

PR #276, opened 2023-05-16
Caliptra model for soc cmodel

### LINKED ISSUE #137: Caliptra model for SoC Cmodel
NVIDIA uses an SoC C-model to do code development and validation for fullchip collateral. We would want to integrate a Caliptra model of the external registers https://ereg.caliptra.org for the boot process and the SoC ROM code to develop against.

@nstewart-amd had proposed a similar problem statement to @korran and @vsonims during December meet-up. Unclear whether there was a follow-up, so filing the ticket.

fdamato:


rusty1968 (dismissed):

hw-model/c-binding/caliptra_model.h
```diff
@@ -26,9 +26,10 @@ enum {
   // more error enums here
 };
 
+
```
korran: Doesn't have to be in this PR, but what do you think of using [cbindgen](https://github.com/mozilla/cbindgen) to generate this file from lib.rs?

Kinda like [verilated/generate_rust_bindings.sh](https://github.com/chipsalliance/caliptra-sw/blob/main/hw-latest/verilated/generate_rust_bindings.sh) but going the other direction?

fdamato: done

hw-model/c-binding/examples/api/caliptra_fuses.h
```diff
@@ -0,0 +1,19 @@
+// Licensed under the Apache-2.0 license
+#ifndef CALIPTRA_FUSES_H
+#define CALIPTRA_FUSES_H
+
+#include "caliptra_api.h"
+
+#define CALIPTRA_FUSE_WRITE(model, offset, data) \
```
korran, line 7: Do these need to be macros, or could we get away with inline functions?

hw-model/c-binding/examples/api/caliptra_fuses.h
```diff
@@ -0,0 +1,19 @@
+// Licensed under the Apache-2.0 license
+#ifndef CALIPTRA_FUSES_H
+#define CALIPTRA_FUSES_H
+
+#include "caliptra_api.h"
+
```
korran, line 6: Can you add a big warning stating that these are for use in simulation only? "SoC firmware MUST NOT have read access to fuse values; a hardware state machine should be used to send fuse values to Caliptra over the APB bus."

hw-model/c-binding/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    // Generate Config
+    let config = cbindgen::Config {
+        header: Some(String::from_str("// Licensed under the Apache-2.0 license").unwrap()),
+        language: cbindgen::Language::C,
+        ..Default::default()
+    };
+
+    // Generate Output file
+    let out_file = PathBuf::from(&crate_dir)
```
korran, line 20: So far caliptra-sw has been following the pattern of checking generated code into the repo (like the caliptra-registers crate), and then enforcing in CI that the generated code is up to date. Three reasons:

- Code reviewers can easily tell during code review what the generated code looks like, and how a particular PR affects the generated code, which isn't always obvious.

- When discussing the generated code in Slack or email, we can paste links directly to the generated code in GitHub.

- The build is faster, as cbindgen and its dependencies don't normally need to be built.

We can leave this as-is for now, but I may revisit this in a future PR.

hw-model/c-binding/src/caliptra_model.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/// # Safety
+#[no_mangle]
+pub unsafe extern "C" fn caliptra_model_destroy(model: *mut caliptra_model) {
+    // Parameter check
+    assert!(!model.is_null());
+
+    // This will force model to be freed
+    let _ = Box::from_raw(model);
```
korran, line 62: This might be a bit clearer as `drop(Box::from_raw(model))`.

hw-model/c-binding/examples/api/caliptra_fuses.h
fdamato: done

hw-model/c-binding/examples/api/caliptra_fuses.h
fdamato: done

hw-model/c-binding/build.rs
fdamato: I can take care myself in a separate (and smaller PR).

hw-model/c-binding/src/caliptra_model.rs
fdamato: done

korran (approved):

PR #277, opened 2023-05-16
Give drivers exclusive control over their peripherals

korran:
**I strongly recommend reviewing each commit in this PR separately, reading the commit description for context.**

This is a major change affecting how drivers get access to their peripheral MMIO registers. Previously, any code in the project could call `caliptra_registers::mbox::RegisterBlock::mbox_csr()` and tweak the MMIO registers, without resorting to unsafe code.

In this PR, functions like `caliptra_registers::mbox::RegisterBlock::mbox_csr()` have been removed, and instead there is a new zero-size type `caliptra_registers::mbox::MboxCsr`, which is created by unsafe code near the top of main(). Ownership of this new type is passed to the driver, which can use the `MboxCsr::regs()` function to get read-only access to the registers, or `MboxCsr::regs_mut()` to get read/write access to the registers.


registers/src/csrng.rs
```diff
@@ -4,16 +4,48 @@
...
+    pub fn regs_mut(&mut self) -> RegisterBlock<ureg::RealMmioMut> {
+        RegisterBlock {
+            ptr: Self::PTR,
+            mmio: core::default::Default::default(),
+        }
+    }
+}
 #[derive(Clone, Copy)]
-pub struct RegisterBlock<TMmio: ureg::Mmio + core::borrow::Borrow<TMmio> = ureg::RealMmio> {
+pub struct RegisterBlock<TMmio: ureg::Mmio + core::borrow::Borrow<TMmio>> {
```
rkr35, line 44: So I get `CsrngReg` is exclusive, i.e., there's no way to have concurrent instances without dipping into `unsafe` multiple times. But what about the `RegisterBlock`s? They're `Copy`, which seems subversive to the lifetime tying `&self/&mut self` to `RealMmio/RealMmioMut`.

For example, this code rightly does _not_ compile:
```rust
fn test_multiple_register_blocks() {
    let mut csrng_reg = unsafe { CsrngReg::new() };
    let regs_mut1 = csrng_reg.regs_mut();
    let regs_mut2 = csrng_reg.regs_mut();
    println!("pop FIFO: {}", regs_mut2.genbits().read());
    println!("pop FIFO: {}", regs_mut1.genbits().read());
}
```
```
error[E0499]: cannot borrow `csrng_reg` as mutable more than once at a time
  --> drivers/test-fw/src/bin/csrng_tests.rs:93:21
   |
92 |     let regs_mut1 = csrng_reg.regs_mut();
   |                     -------------------- first mutable borrow occurs here
93 |     let regs_mut2 = csrng_reg.regs_mut();
   |                     ^^^^^^^^^^^^^^^^^^^^ second mutable borrow occurs here
94 |     println!("pop FIFO: {}", regs_mut2.genbits().read());
95 |     println!("pop FIFO: {}", regs_mut1.genbits().read());
   |                              ------------------- first borrow later used here

For more information about this error, try `rustc --explain E0499`.
```

But `Copy`ing `regs_mut1` compiles:
```rust
fn test_multiple_register_blocks() {
    let mut csrng_reg = unsafe { CsrngReg::new() };
    let regs_mut1 = csrng_reg.regs_mut();
    let regs_mut2 = regs_mut1;
    println!("pop FIFO: {}", regs_mut2.genbits().read());
    println!("pop FIFO: {}", regs_mut1.genbits().read());
}
```

drivers/src/soc_ifc.rs
```diff
@@ -15,6 +15,8 @@ Abstract:
 use caliptra_registers::soc_ifc;
 use caliptra_registers::soc_ifc::enums::DeviceLifecycleE;
 
+use crate::FuseBank;
+
 /// Device Life Cycle State
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum Lifecycle {
```
rusty1968: Perhaps not part of the scope of the PR, but some drivers are enum types and others are structs. Is there a reason for this?

registers/src/csrng.rs
```diff
@@ -4,16 +4,48 @@
...
     ptr: *mut u32,
     mmio: TMmio,
 }
-impl RegisterBlock<ureg::RealMmio> {
-    pub fn csrng_reg() -> Self {
-        unsafe { Self::new(0x20002000 as *mut u32) }
-    }
-}
+impl RegisterBlock<ureg::RealMmioMut<'_>> {}
```
rkr35, line 48: We don't need this empty `impl` anymore.

registers/src/csrng.rs
korran: This is on purpose, as there are situations where you need multiple RegRef references alive from the same register block (for example, to pass to utility functions). Because the RegisterBlock has to expose all the registers as member functions, the RegRef has to take the lifetime of the entire RegisterBlock.

Fortunately, the lifetime of all the register blocks (and RegRefs) is tied to the csrng_reg instance, and if you try to store those somewhere outside of test_multiple_register_blocks(), it will fail to compile.

jhand2 (approved):
Changes to runtime lgtm

rusty1968 (approved):

registers/src/csrng.rs
korran: To be clearer: RegisterBlock doesn't necessarily have to be Copy, but removing Copy doesn't make things much safer. Arrays and RegRefs all share the lifetime with RegisterBlock, where you can have multiple of them alive at once. Some of the drivers rely on this behavior, for example: https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/kv_access.rs#L130

korran: The main thing I care about is that whoever holds the CsrngReg can decide who gets RegisterBlocks, and if they try to pass the CsrngReg to somewhere else, those RegisterBlocks won't be able to outlive the move.

korran: If Rust member functions ever support partial borrows in the future (where mbox.status() can return a different lifetime than mbox.fifo()), we can revisit this decision.

drivers/src/soc_ifc.rs
korran: Lifecycle isn't a driver, just a enum with 4 variants.

fdamato (dismissed):

registers/src/csrng.rs
korran: Fixed in [58353e6](https://github.com/chipsalliance/caliptra-sw/pull/277/commits/58353e6bb190ea2e4ba9421525f3fa88463333c3)

rkr35 (approved):

fdamato (dismissed):

jhand2 (dismissed):

mhatrevi (dismissed):

jhand2 (approved):

mhatrevi (approved):

mhatrevi (approved):

PR #279, opened 2023-05-16
Document firmware error reporting

jhand2:
For any failed mailbox flows, document that errors are reported in the CPTRA_FW_ERROR_NON_FATAL register.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -73,10 +73,12 @@ Table: Mailbox command "result" codes:
 | `BAD_IMAGE`      | `0x4249_4D47` ("BIMG") | Malformed input image
 | `BAD_CHKSUM`     | `0x4243_484B` ("BCHK") | Checksum check failed on input arguments
 
-Relevant Mailbox Registers:
+Relevant Registers:
 
-* COMMAND: Command code to execute
-* DLEN: Number of bytes written to mailbox
+* mbox\_csr -> COMMAND: Command code to execute
+* mbox\_csr -> DLEN: Number of bytes written to mailbox
+* CPTRA\_FW\_ERROR\_NON\_FATAL: Status code of mailbox command. Any result
+  other than `SUCCESS` signifies a mailbox command failure.
 
 ### CALIPTRA\_FW\_LOAD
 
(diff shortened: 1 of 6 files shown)


runtime/README.md
```diff
@@ -159,7 +156,7 @@ Table: `ECDSA384_SIGNATURE_VERIFY` output arguments
 | **Name** | **Type** | **Description**
 | -------- | -------- | ---------------
 | chksum   | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
-| result   | u32      | Result code. Little endian.
+| result   | u32      | If 0, ECDSA verify failed. Else verify succeeded.
```
bluegate010, line 159: Doesn't 0 usually indicate success?

jhand2: I can change the "result" naming. This is supposed to just be a boolean (but isn't because of word alignment)

bluegate010: Why not treat it like `DISABLE_ATTESTATION` or `CALIPTRA_FW_LOAD` and return no output arguments?

jhand2: I wanted to differentiate between "Something bad happened in firmware" and "The ECDSA signature was bad". But I suppose that can just be another status code.

bluegate010: Yeah, separate status codes seem reasonable - would need the status register to report those anyways, right? Since this `result` field just conveys a bool.

jhand2: Done.

mhatrevi (dismissed):

runtime/README.md
```diff
@@ -155,14 +152,6 @@ Table: `ECDSA384_SIGNATURE_VERIFY` input arguments
 | signature\_r | u8[48]   | R portion of signature to verify
 | signature\_s | u8[48]   | S portion of signature to verify
```
bluegate010, line 154: Add "`ECDSA384_SIGNATURE_VERIFY` returns no output arguments." for consistency?

runtime/tests/integration_tests.rs
```diff
@@ -136,13 +136,10 @@ fn test_verify_cmd() {
         ],
     };
 
-    let resp = model
+    model
         .mailbox_execute(u32::from(CommandId::ECDSA384_VERIFY), cmd.as_bytes())
-        .unwrap()
         .unwrap();
-    let resp_bytes: &[u8] = resp.as_bytes();
```
bluegate010, line 143: Should we assert that no response is provided? Also fine if we decide it's not worth the hassle.

bluegate010 (dismissed):

runtime/tests/integration_tests.rs
jhand2: Done

runtime/README.md
jhand2: Done

bluegate010 (approved):

PR #283, opened 2023-05-17
Make hw-model sha acc interface more ergonomic

jhand2:
Abstract alignment, endianness, etc from callers of the hw-model sha512acc API. Instead, allow callers to just pass a byte slice.


hw-model/src/lib.rs
```diff
@@ -528,15 +529,11 @@ pub trait HwModel {
...
-    ///
     /// Returns the computed digest if the sha512acc lock can be acquired.
     /// Else, returns an error.
     fn compute_sha512_acc_digest(
         &mut self,
-        data: &[u32],
-        dlen: u32,
+        data: &[u8],
         sha384: bool,
     ) -> Result<Vec<u32>, ModelError> {
```
korran, line 538: If the input is bytes, maybe the output should be bytes too?

hw-model/src/lib.rs
```diff
@@ -557,8 +556,25 @@ pub trait HwModel {
             })
         });
 
-        for word in data {
-            self.soc_sha512_acc().datain().write(|_| word.swap_bytes());
+        // Unwrap cannot fail, count * sizeof(u32) is always smaller than data.len()
+        let data_words = LayoutVerified::<&[u8], [Unalign<u32>]>::new_slice_unaligned_from_prefix(
```
korran, line 560: Consider something like:

```
let (prefix_words, suffix_bytes) = LayoutVerified::<_, [Unalign<u32>]>::new_slice_unaligned_from_prefix(..)
```

jhand2: Done

hw-model/src/lib.rs
jhand2: Done

hw-model/src/lib.rs
```diff
@@ -528,24 +529,22 @@ pub trait HwModel {
...
-    /// data to be full words.
-    ///
     /// Returns the computed digest if the sha512acc lock can be acquired.
     /// Else, returns an error.
     fn compute_sha512_acc_digest(
         &mut self,
-        data: &[u32],
-        dlen: u32,
+        data: &[u8],
         sha384: bool,
```
korran, line 537: nit: boolean arguments in this style should be avoided in public APIs.

Consider using either an enum or just two different methods (compute_sha512_acc_digest, compute_sha384_acc_digest).

korran: If you use two different methods, you can make the result type a fixed-size array.

korran (dismissed):

hw-model/src/lib.rs
```diff
@@ -880,51 +896,58 @@ mod tests {
         })
         .unwrap();
 
-        // Note: Streamed `msg` data is big endian. `expected` digest values are little-endian
+        // Note: expected` digest values are little-endian
```
korran, line 899: I don't think this comment is relevant anymore.

hw-model/src/lib.rs
jhand2: Switched to an enum

korran (approved):

PR #286, opened 2023-05-18
Add initial CSRNG sw emulator support

rkr35:
This PR implements enough of the register logic to seed and query for random bits.

It emulates the [CTR_DRBG](https://doi.org/10.6028/NIST.SP.800-90Ar1) used in the CSRNG RTL, so seeds passed to either the emulator or the Verilated model should produce the same outputs (which is required for the driver unit tests to be agnostic to the backing hardware model).

Not implemented yet:
* Health checks
* `CONF`, `CTRL`, `MODULE_ENABLE` register logic
* Errors
* A way to pass entropy bits without hardcoding


sw-emulator/lib/periph/src/csrng.rs
```diff
@@ -0,0 +1,557 @@
...
+
+use rand_impl::{Block, CtrDrbg};
+
+use std::mem;
+
+#[derive(Bus)]
+pub struct Csrng {
+    // CSRNG registers
+    #[register(offset = 0x14)]
+    ctrl: ReadWriteRegister<u32>,
```
korran, line 16: nit: registers with read and write can just be a plain u32:

```
#[register(offset = 0x14)]
ctrl: u32,
```

korran (approved):

PR #288, opened 2023-05-18
Publish error codes

rusty1968:


attzonko:
I think there are some missing ones. I count 132 in this PR, but a quick grep on the code in repo, gives me 149. We should sync up before this goes in.

JohnTraverAmd (approved):

PR #289, opened 2023-05-18
Stub out Caliptra implementation of DPE Crypto trait

sree-revoori1:
This PR also stubs out Caliptra Platform implementation.


jhand2:
Can you move these changes to out to the `runtime/src` directory? That way they don't be a part of the submodule (which we want to just track the caliptra-dpe repo with nothing extra)

jhand2:
> Can you move these changes to out to the `runtime/src` directory? That way they don't be a part of the submodule (which we want to just track the caliptra-dpe repo with nothing extra)

Actually, you can just update https://github.com/chipsalliance/caliptra-dpe/pull/94 to make the Cargo.toml changes

jhand2 (commented):
Looks good. Once https://github.com/chipsalliance/caliptra-dpe/pull/95 is merged let's update this PR to also update the DPE submodule.

jhand2 (approved):

Cargo.lock
```diff
@@ -1181,9 +1411,9 @@ checksum = "b7e5500299e16ebb147ae15a00a942af264cf3688f47923b8fc2cd5858f23ad3"
 
 [[package]]
 name = "openssl"
-version = "0.10.48"
```
jhand2, line 1184: This file has more changes than I would have expected. In particular I think this change is causing the CI failure.

One thing to note is that if you delete the Cargo.lock file, it will get fully regenerated with new versions. To just diff based on new dependencies, you can checkout the Cargo.lock file from main and then run `cargo build` in the root.

jhand2 (changes requested):

Cargo.lock
jhand2: I think you'll want to copy the Cargo.lock file from main and then run `cargo build`. Instead of deleting it and starting from scratch.

runtime/src/caliptra_crypto.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl Crypto for CaliptraCrypto {
+    type Cdi = KeyId;
+    type Hasher = CaliptraHasher;
+    type PrivKey = KeyId;
+
+    #[cfg(feature = "deterministic_rand")]
+    fn rand_bytes(_dst: &mut [u8]) -> Result<(), CryptoError> {
+        Err(CryptoError::NotImplemented)
+    }
```
jhand2, line 28: We can leave this out. Only need a deterministic version in the DPE unit tests

jhand2 (approved):

.github/workflows/build-test.yml
```diff
@@ -73,6 +73,10 @@ jobs:
           cargo fmt --version
           cargo clippy --version
 
+      - name: Initialize submodules
```
korran, line 76: Please update the checkout action instead:

```
      - uses: actions/checkout@v3
        with:
          submodules: 'true'
```

korran: Actually, instead of that, can you just checkout the DPE submodule only? It's important to verify that this workflow passes when the RTL submodule is not populated.

.github/workflows/build-test-verilator.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          submodules: 'true'
```
jhand2, line 33: I think this one should still checkout all submodules since it spcifically depends on verilator

.github/workflows/build-test.yml
```diff
@@ -28,6 +28,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+        with:
+          repository: chipsalliance/caliptra-dpe
```
jhand2, line 32: I think this checks out caliptra-dpe instead of caliptra-sw. Instead you should be able to change the "Initialize submodules" step below to just checkout dpe. Something like

`git submodule update --init runtime/dpe`

.github/workflows/nightly-verilator.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          submodules: 'true'
+          repository: chipsalliance/caliptra-dpe
```
jhand2, line 25: Same as the other verilator workflow

jhand2 (approved):

PR #293, opened 2023-05-19
Add rt functions to retrieve and build certs produced by ROM

jhand2:
Get the LDevID and FmcAlias certs produced by ROM and stich them together to ensure runtime can properly produce these certificates.


jhand2:
Note: I can wait to merge this PR until https://github.com/chipsalliance/caliptra-sw/pull/277 is merged

runtime/src/dice.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Copy a certificate from `dccm_offset`, append signature, and write the
+/// output to `cert`.
+fn cert_from_dccm(cert: &mut [u8], cert_type: CertType) -> CaliptraResult<usize> {
+    let data_vault = DataVault::default();
+    let (tbs, sig) = match cert_type {
+        CertType::LDevId => {
+            let tbs = unsafe {
+                let ptr = &mut LDEVID_TBS_ORG as *mut u8;
+                core::slice::from_raw_parts_mut(ptr, LocalDevIdCertTbs::TBS_TEMPLATE_LEN)
```
bluegate010, line 39: It makes me a _little_ bit nervous that there's no length field here being written by ROM. Perhaps we need a feature-request to write that out.

runtime/test-fw/src/cert_tests.rs
```diff
@@ -0,0 +1,47 @@
...
+#![no_main]
+#![no_std]
+
+use caliptra_registers::mbox::enums::MboxStatusE;
+use caliptra_runtime::{dice, mailbox::Mailbox};
+use caliptra_test_harness::{runtime_handlers, test_suite};
+
+fn mbox_responder() {
+    let soc_ifc = caliptra_registers::soc_ifc::RegisterBlock::soc_ifc_reg();
+    soc_ifc.cptra_flow_status().write(|w| w.ready_for_fw(true));
```
bluegate010, line 14: Could this be `soc_ifc.flow_status_set_ready_for_firmware()`? But also, isn't that already done by ROM?

jhand2: Oh right. Removed.

bluegate010 (dismissed):

runtime/src/dice.rs
jhand2: Agreed this is a bit concerning. But I think there are competing considerations:

1. We could maybe read the value from the hand-off table. But an attacker which can control this value has a DCCM-read primitive. So from that perspective, a hard-coded value may be safer.

2. As you pointed out, the hard-coded value is a bit error prone. Especially if RT needs to work on multiple different ROM releases.

mhatrevi (dismissed):

runtime/src/dice.rs
jhand2: Closing this comment, we can track the enhancement in https://github.com/chipsalliance/caliptra-sw/pull/308

cpu/link.x
```diff
@@ -1,4 +1,4 @@
-SECTIONS
```
korran, line 1: Please move all the reformatting changes to a separate commit in this PR so it's easier to notice the important stuff.

runtime/src/dice.rs
```diff
@@ -0,0 +1,67 @@
...
+pub fn copy_fmc_alias_cert(dv: &DataVault, cert: &mut [u8]) -> CaliptraResult<usize> {
+    cert_from_dccm(dv, cert, CertType::FmcAlias)
+}
+
+/// Copy a certificate from `dccm_offset`, append signature, and write the
+/// output to `cert`.
+fn cert_from_dccm(dv: &DataVault, cert: &mut [u8], cert_type: CertType) -> CaliptraResult<usize> {
+    let (tbs, sig) = match cert_type {
+        CertType::LDevId => {
+            let tbs = unsafe {
```
korran, line 36: Consider doing something like this:

```diff
diff --git a/runtime/src/dice.rs b/runtime/src/dice.rs
index 64b7d981..77f0c424 100644
--- a/runtime/src/dice.rs
+++ b/runtime/src/dice.rs
@@ -5,8 +5,8 @@ use caliptra_drivers::{CaliptraResult, DataVault};
 use caliptra_x509::{Ecdsa384CertBuilder, Ecdsa384Signature, FmcAliasCertTbs, LocalDevIdCertTbs};
 
 extern "C" {
-    static mut LDEVID_TBS_ORG: u8;
-    static mut FMCALIAS_TBS_ORG: u8;
+    static LDEVID_TBS_ORG: [u8; LocalDevIdCertTbs::TBS_TEMPLATE_LEN];
+    static FMCALIAS_TBS_ORG: [u8; FmcAliasCertTbs::TBS_TEMPLATE_LEN];
 }
 
 enum CertType {
@@ -33,20 +33,10 @@ pub fn copy_fmc_alias_cert(dv: &DataVault, cert: &mut [u8]) -> CaliptraResult<us
 fn cert_from_dccm(dv: &DataVault, cert: &mut [u8], cert_type: CertType) -> CaliptraResult<usize> {
     let (tbs, sig) = match cert_type {
         CertType::LDevId => {
-            let tbs = unsafe {
-                let ptr = &mut LDEVID_TBS_ORG as *mut u8;
-                core::slice::from_raw_parts_mut(ptr, LocalDevIdCertTbs::TBS_TEMPLATE_LEN)
-            };
-
-            (tbs, dv.ldev_dice_signature())
+            (unsafe { &LDEVID_TBS_ORG[..] }, dv.ldev_dice_signature())
         }
         CertType::FmcAlias => {
-            let tbs = unsafe {
-                let ptr = &mut FMCALIAS_TBS_ORG as *mut u8;
-                core::slice::from_raw_parts_mut(ptr, FmcAliasCertTbs::TBS_TEMPLATE_LEN)
-            };
-
-            (tbs, dv.fmc_dice_signature())
+            (unsafe { &FMCALIAS_TBS_ORG[..] }, dv.fmc_dice_signature())
         }
     };
 
```

jhand2: Great idea. Done.

korran (approved):

korran (approved):

PR #299, opened 2023-05-21
Add LMS signatures to image

satyamsetti:
Reporting and revocation not yet done.


image/fake-keys/src/lib.rs
```diff
@@ -96,102 +96,102 @@ pub const VENDOR_KEY_3_PRIVATE: ImageEccPrivKey = [
     0xbc900c70, 0xb8cb36ac, 0x268b8079, 0xe3aeaaaf,
 ];
 pub const VENDOR_LMS_KEY0_PRIVATE: ImageLmsPrivKey = ImageLmsPrivKey {
-    tree_type: 0x0c000000,
-    otstype: 0x07000000,
+    tree_type: 0xc000000,
```
ericeilertson, line 99: Could these magic values be const values with human readable names?  or a comment next to them

image/gen/src/generator.rs
```diff
@@ -109,6 +117,10 @@ impl<Crypto: ImageGeneratorCrypto> ImageGenerator<Crypto> {
                 &config.vendor_config.pub_keys.ecc_pub_keys[ecc_key_idx as usize],
             )?;
             vendor_sigs.ecc_sig = sig;
+            let lms_sig =
```
korran, line 120: With this PR, the time to run the tests on my workstation increases from 30 seconds to 234 seconds. This is unacceptable.

We need to reduce the number of tests that do a full LMS sign, or we need to make this signing operation more than an order of magnitude faster.

drivers/src/lms.rs
```diff
@@ -272,37 +289,158 @@ const LMOTS_P: [LmotsParameter; 9] = [
...
+    let lmots_type = lookup_lmots_algorithm_type(slice_to_num(&public_string[pos..pos + 4]))
+        .ok_or(err_u32!(InvalidLmotsAlgorithmType))?;
+    pos += 4;
+
+    let (hash_width, _) = get_lms_parameters(&lms_type)?;
+    if hash_width as usize != N * 4 {
+        raise_err!(InvalidHashWidth);
+    }
+
+    let mut lms_identifier = [0u8; 16];
```
korran, line 341: prefer 

```
let lms_identifier: [u8; 16] = public_string[pos..][..16].try_into().unwrap();
```

drivers/src/lms.rs
```diff
@@ -272,37 +289,158 @@ const LMOTS_P: [LmotsParameter; 9] = [
...
+    lms_identifier.copy_from_slice(&public_string[pos..pos + 16]);
+    pos += 16;
+
+    let mut temp = [0u32; N];
+    //temp.copy_from_slice(&public_string[pos..pos + N]);
```
korran, line 346: remove commented-out-code

drivers/src/lms.rs
```diff
@@ -272,37 +289,158 @@ const LMOTS_P: [LmotsParameter; 9] = [
...
+
+    for t in temp.iter_mut().take(N) {
```
korran, line 348: I don't think you need .take(N), as temp is already length N.

image/verify/src/verifier.rs
```diff
@@ -193,6 +199,12 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
             &preamble.vendor_sigs.ecc_sig,
         );
 
+        let vendor_lms_pub_key_idx = preamble.vendor_lms_pub_key_idx;
+        let vendor_lms_info = (
+            &preamble.vendor_pub_keys.lms_pub_keys[vendor_lms_pub_key_idx as usize],
```
korran, line 204: You need to check that lms_pub_key_idx is in bounds; this can panic.

drivers/src/lms.rs
```diff
@@ -272,37 +289,156 @@ const LMOTS_P: [LmotsParameter; 9] = [
...
+        lms_identifier,
+        root_hash: public_hash,
+    };
+    Ok(pk)
+}
+
+pub fn parse_signature_contents<const N: usize, const P: usize, const H: usize>(
+    signature: &[u8],
+) -> CaliptraResult<LmsSignature<N, P, H>> {
+    if signature.len() < 8 {
```
korran, line 365: The compiler isn't smart enough to realize that pos doesn't overflow below. Can you make this

```
    if signature.len() != 8 + N * 4 + (P * N * 4) + 4 + N * H {
        raise_err!(InvalidSignatureLength);
    }
```

rom/dev/src/verifier.rs
```diff
@@ -69,6 +85,25 @@ impl<'a> ImageVerificationEnv for &mut RomImageVerificationEnv<'a> {
...
+        digest: &ImageDigest,
+        pub_key: &ImageLmsPublicKey,
+        sig: &ImageLmsSignature,
+    ) -> CaliptraResult<bool> {
+        let mut message = [0u8; SHA384_DIGEST_BYTE_SIZE];
+        for i in 0..digest.len() {
+            message[i * 4..][..4].copy_from_slice(&digest[i].to_be_bytes());
+        }
+        let lms_public_key = parse_public_contents::<SHA192_DIGEST_WORD_SIZE>(pub_key.as_bytes())?;
+        let lms_sig = parse_signature_contents::<
```
korran, line 99: Why do you need to parse this? Why aren't ImageLmsSignature and LmsSignature the same type?

Converting it to bytes and then back into a structure seems silly to me.

drivers/src/lms.rs
korran: The optimizer really hates the function. I was able to get rid of the panic reference with this disgusting patch, but I recommend getting rid of the need to parse altogether; just have the sign function take the struct directly, and validate the few parameters that need to be validated there.

```
diff --git a/drivers/src/lms.rs b/drivers/src/lms.rs
index 4f4ace3e..9396655e 100644
--- a/drivers/src/lms.rs
+++ b/drivers/src/lms.rs
@@ -393,14 +393,18 @@ pub fn parse_signature_contents<const N: usize, const P: usize, const H: usize>(
 
     let mut y = [HashValue::<N>::default(); P];
     for t in y.iter_mut() {
-        let mut tmp = [0u32; N];
-        for tt in tmp.iter_mut().take(N) {
-            *tt = slice_to_num(&signature[pos..pos + 4]);
+        for tt in t.0.iter_mut() {
+            if pos > signature.len() - 4 {
+                raise_err!(InvalidSignatureDepth);
+            }
+            *tt = slice_to_num(&signature[pos..][..4]);
             pos += 4;
         }
-        *t = HashValue::<N>::from(tmp);
     }
-    let lms_type = lookup_lms_algorithm_type(slice_to_num(&signature[pos..pos + 4]))
+    if pos > signature.len() - 4 || pos + 4 > signature.len() {
+        raise_err!(InvalidSignatureLength);
+    }
+    let lms_type = lookup_lms_algorithm_type(slice_to_num(&signature[pos..][..4]))
         .ok_or(err_u32!(InvalidLmsAlgorithmType))?;
     pos += 4;
 
@@ -420,13 +424,15 @@ pub fn parse_signature_contents<const N: usize, const P: usize, const H: usize>(
 
     let mut path = [HashValue::<N>::default(); H];
     for t in path.iter_mut() {
-        let mut tmp = [0u32; N];
-        for tt in tmp.iter_mut().take(N) {
-            *tt = slice_to_num(&signature[pos..pos + 4]);
+        for tt in t.0.iter_mut() {
+            if pos > signature.len() - 4 {
+                raise_err!(InvalidSignatureDepth);
+            }
+            *tt = slice_to_num(&signature[pos..][..4]);
             pos += 4;
         }
-        *t = HashValue::<N>::from(tmp);
     }
+
```

image/verify/src/verifier.rs
satyamsetti: Done.

drivers/src/lms.rs
satyamsetti: Done.

rom/dev/src/verifier.rs
satyamsetti: LMS driver was implemented by Eric as templatized struct while we need our signatures to be of deterministic size. Hence the conversion.

We will explore if we can change this later.

ericeilertson (dismissed):
Need to add owner keys, but that can be another PR.

image/verify/src/verifier.rs
```diff
@@ -348,6 +371,14 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
             Err(CaliptraError::IMAGE_VERIFIER_ERR_VENDOR_ECC_SIGNATURE_INVALID)?;
         }
```
ericeilertson, line 373: I had thought LMS was optional unless a fuse was set.  Is this the place to check for that fuse value and conditionally do the LMS verification?

ericeilertson (approved):
Looks good.  I had one comment about the verification code appears to require ECC and LMS.  I had thought LMS was not enabled by default and a fuse setting enabled the hybrid approach.

image/verify/src/verifier.rs
korran: I believe so; future PR?

korran (approved):

PR #302, opened 2023-05-22
Add intr_block_t registers

korran:


registers/src/soc_ifc.rs
```diff
@@ -1879,4 +3007,79 @@ pub mod meta {
         crate::soc_ifc::regs::InternalFwUpdateResetWaitCyclesWriteVal,
     >;
     pub type InternalNmiVector = ureg::ReadWriteReg32<0, u32, u32>;
+    pub type IntrBlockRfGlobalIntrEnR = ureg::ReadWriteReg32<
+        0,
+        crate::sha512_acc::regs::GlobalIntrEnTReadVal,
+        crate::sha512_acc::regs::GlobalIntrEnTWriteVal,
```
jlmahowa-amd, line 3013: The sha512 here was confusing since this is for the soc_ifc interrupt enable. It works, but the field definitions are located in the sha512 register file. It seems like identical register definitions got merged.

korran: Yeah, ureg tries to reuse these types across modules when they're identical. I should probably move them somewhere more generic though rather than the first place they were used.

PR #305, opened 2023-05-22
[feat] Add IV to the ECC384 key generation for SCA countermeasures.

mhatrevi:
This change uses the CSRNG driver to generate a nonce; the same is provided as an IV input to the key generation function of the ECC384 driver.

### DIFF
--- a/drivers/src/ecc384.rs
+++ b/drivers/src/ecc384.rs
@@ -12,9 +12,13 @@ Abstract:
 
 --*/
 
+use core::iter::zip;
+use core::num::NonZeroUsize;
+
 use crate::kv_access::{KvAccess, KvAccessErr};
 use crate::{
-    array_concat3, caliptra_err_def, wait, Array4x12, CaliptraResult, KeyReadArgs, KeyWriteArgs,
+    array_concat3, caliptra_err_def, wait, Array4x12, CaliptraResult, Csrng, KeyReadArgs,
+    KeyWriteArgs,
 };
 use caliptra_registers::ecc::EccReg;
 
@@ -161,6 +165,7 @@ impl Ecc384 {
     ///
     /// * `seed` - Seed for deterministic ECC Key Pair generation
     /// * `nonce` - Nonce for deterministic ECC Key Pair generation
+    /// * `csrng` - CSRNG driver instance
     /// * `priv_key` - Generate ECC-384 Private key
     ///
     /// # Returns
@@ -170,6 +175,7 @@ impl Ecc384 {
         &mut self,
         seed: Ecc384Seed,
         nonce: &Array4x12,
+        csrng: &mut Csrng,
         mut priv_key: Ecc384PrivKeyOut,
     ) -> CaliptraResult<Ecc384PubKey> {
         let ecc = self.ecc.regs_mut();
@@ -198,6 +204,14 @@ impl Ecc384 {
 
         // Copy nonce to the hardware
         KvAccess::copy_from_arr(nonce, ecc.nonce())?;
+
+        // Generate an IV.
+        let mut iv: [u32; 12] = [0u32; 12];
+        let num_words = NonZeroUsize::new(iv.len()).unwrap();
+        for (dst, src) in zip(&mut iv, csrng.generate(num_words)?) {
+            *dst = src;
+        }
+        KvAccess::copy_from_arr(&Array4x12::from(iv), ecc.iv())?;
 
         // Program the command register for key generation
         ecc.ctrl().write(|w| w.ctrl(|w| w.keygen()));
--- a/rom/dev/src/flow/cold_reset/crypto.rs
+++ b/rom/dev/src/flow/cold_reset/crypto.rs
@@ -136,7 +136,6 @@ impl Crypto {
         seed: KeyId,
         priv_key: KeyId,
     ) -> CaliptraResult<Ecc384KeyPair> {
-        // [TODO] Add Nonce to the ecc384_key_gen function
         let seed = Ecc384Seed::Key(KeyReadArgs::new(seed));
 
         let mut usage = KeyUsage::default();
@@ -146,7 +145,9 @@ impl Crypto {
 
         Ok(Ecc384KeyPair {
             priv_key,
-            pub_key: env.ecc384.key_pair(seed, &Array4x12::default(), key_out)?,
+            pub_key: env
+                .ecc384
+                .key_pair(seed, &Array4x12::default(), &mut env.csrng, key_out)?,
         })
     }
 
--- a/rom/dev/src/main.rs
+++ b/rom/dev/src/main.rs
@@ -54,7 +54,10 @@ Running Caliptra ROM ...
 pub extern "C" fn rom_entry() -> ! {
     cprintln!("{}", BANNER);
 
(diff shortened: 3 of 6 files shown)


rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -136,17 +137,26 @@ impl Crypto {
...
+        for (dst, src) in zip(&mut nonce, env.csrng.generate(num_words)?) {
+            *dst = src;
+        }
+
         Ok(Ecc384KeyPair {
             priv_key,
-            pub_key: env.ecc384.key_pair(seed, &Array4x12::default(), key_out)?,
+            pub_key: env
+                .ecc384
+                .key_pair(seed, &Array4x12::from(nonce), key_out)?,
```
bluegate010, line 159: Won't adding this nonce break the certificates issued for IDevID and AliasFMC? The CSRNG would be seeded differently on each boot, yeah?

korran: Yeah, I don't think we want a random nonce. Perhaps you mean to feed the entropy to the [IV register](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.ecc_reg.ECC_IV%5B0%5D)

From the [hw spec](https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/Caliptra_Hardware_Specification.pdf):

> Implementation of complete unified addition formula for the scalar multiplication avoids
information leakage and enhances architecture from security/mathematical perspective.
To protect the architecture against horizontal power/EM and differential power analysis (DPA)
attacks, two countermeasures are embedded to the design [9]. Since these countermeasures
require random inputs, HMAC-DRBG is fed by “IV” to generate these random values.
Since HMAC-DRBG generates random value in a deterministic way, firmware MUST feed
different IV to ECC engine for EACH keygen and signing operations.
The prior work shows that the countermeasure still shows enough resistance even though the
attack somehow finds a stronger attack mechanism.

rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -136,17 +137,26 @@ impl Crypto {
...
         let mut usage = KeyUsage::default();
         usage.set_ecc_private_key(true);
 
         let key_out = Ecc384PrivKeyOut::Key(KeyWriteArgs::new(priv_key, usage));
 
+        // Generate a nonce.
+        let mut nonce: [u32; 12] = [0u32; 12];
+        let num_words = NonZeroUsize::new(nonce.len()).unwrap();
+        let _ = env.csrng.generate(num_words)?;
+        for (dst, src) in zip(&mut nonce, env.csrng.generate(num_words)?) {
```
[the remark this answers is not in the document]
korran: We'll tune the driver API for code size once we see what folks want to do with it.

ajisaxena (approved):

rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -136,17 +137,26 @@ impl Crypto {
...
 
         let mut usage = KeyUsage::default();
         usage.set_ecc_private_key(true);
 
         let key_out = Ecc384PrivKeyOut::Key(KeyWriteArgs::new(priv_key, usage));
 
+        // Generate a nonce.
+        let mut nonce: [u32; 12] = [0u32; 12];
+        let num_words = NonZeroUsize::new(nonce.len()).unwrap();
+        let _ = env.csrng.generate(num_words)?;
```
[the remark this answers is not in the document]
mhatrevi: This was as per the unit test for the CSRNG driver.

rom/dev/src/flow/cold_reset/crypto.rs
mhatrevi: Correct, this should have been fed to the IV register. Updated.

rom/dev/src/flow/cold_reset/crypto.rs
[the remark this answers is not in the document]
rkr35: We can safely remove this first `generate()`.

This `generate()`-then-toss was likely copied over from this test, which requires two `generate()`s to match the CTR_DRBG KAT:
https://github.com/chipsalliance/caliptra-sw/blob/2e26d715dcbfa92c1e5e7eb3bcd2cda760db9792/drivers/test-fw/src/bin/csrng_tests.rs#L46

rom/dev/src/flow/cold_reset/crypto.rs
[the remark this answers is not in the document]
korran: rkr35: consider something like this:

```
fn generate<const N: usize>(&mut self) -> CaliptraResult<[u32; N]> {
  self.generate_start(N);
  Ok(core::array::from_fn(|_|, hw.get_a_word()))
}
#[inline(never)]
fn generate_start(n: usize) {
  // do stuff
}
```

rom/dev/src/flow/cold_reset/crypto.rs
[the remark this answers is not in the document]
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/crypto.rs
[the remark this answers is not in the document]
mhatrevi: @rkr35 will be updating the CSRNG driver as per the discussion; resolving the comment.

bluegate010 (dismissed):

drivers/src/ecc384.rs
```diff
@@ -170,6 +171,7 @@ impl Ecc384 {
         &mut self,
         seed: Ecc384Seed,
         nonce: &Array4x12,
+        iv: &Array4x12,
```
korran, line 174: Consider taking a &mut CsRng in as a parameter, and generating the iv here in the driver.

That way, the driver has a stronger guarantee it's actually using random data for the IV, which is necessary for security.

mhatrevi: Makes sense; updated.

ajisaxena (approved):

rom/dev/src/main.rs
```diff
@@ -54,7 +54,13 @@ Running Caliptra ROM ...
 pub extern "C" fn rom_entry() -> ! {
     cprintln!("{}", BANNER);
 
-    let mut env = unsafe { rom_env::RomEnv::new_from_registers() };
+    let mut env = unsafe {
+        let result = rom_env::RomEnv::new_from_registers();
+        if result.is_err() {
+            report_error(result.unwrap_err_unchecked().0.into());
```
korran, line 60: Please don't use unsafe function likes unwrap_err_unchecked. I recommend:

```diff
-    let mut env = unsafe {
-        let result = rom_env::RomEnv::new_from_registers();
-        if result.is_err() {
-            report_error(result.unwrap_err_unchecked().0.into());
-        }
-        result.unwrap()
+    let mut env = match unsafe { rom_env::RomEnv::new_from_registers() } {
+        Ok(env) => env,
+        Err(e) => report_error(e.into()),
     };
```

mhatrevi: Updated.

korran (approved):

PR #306, opened 2023-05-22
Update CI to use ubuntu 22.04

jhand2:
Update to a version of ubuntu with OpenSSL 3.x.x support for CI


korran:
We can't do this; many folks have workstations that don't have openssl 3.x.x.

PR #308, opened 2023-05-24
Place TBS size in front of TBS when writing out to DCCM.

bluegate010:
Will allow firmware to be more compatible across ROM changes.

### DIFF
--- a/rom/dev/src/flow/cold_reset/mod.rs
+++ b/rom/dev/src/flow/cold_reset/mod.rs
@@ -35,6 +35,14 @@ pub const KEY_ID_IDEVID_PRIV_KEY: KeyId = KeyId::KeyId7;
 pub const KEY_ID_LDEVID_PRIV_KEY: KeyId = KeyId::KeyId5;
 pub const KEY_ID_FMC_PRIV_KEY: KeyId = KeyId::KeyId7;
 
+caliptra_err_def! {
+    RomGlobal,
+    GlobalErr
+    {
+        TbsUnsupportedDataLength = 0x6,
+    }
+}
+
 extern "C" {
     static mut LDEVID_TBS_ORG: u8;
     static mut FMCALIAS_TBS_ORG: u8;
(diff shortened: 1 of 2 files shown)


rom/dev/src/flow/cold_reset/mod.rs
```diff
@@ -73,18 +81,33 @@ impl ColdResetFlow {
...
+            core::slice::from_raw_parts_mut(ptr, SIZE_LEN),
+            core::slice::from_raw_parts_mut(ptr.add(SIZE_LEN), tbs.len()),
+        )
     };
 
-    dst[..tbs.len()].copy_from_slice(tbs);
+    let len_bytes = tbs.len().to_le_bytes();
+
+    len_dst[..len_bytes.len()].copy_from_slice(&len_bytes);
+    tbs_dst[..tbs.len()].copy_from_slice(tbs);
```
jhand2, line 110: This can panic. Prefer `slice.get` instead.

rom/dev/src/flow/cold_reset/mod.rs
```diff
@@ -73,18 +81,33 @@ impl ColdResetFlow {
...
+        raise_err!(TbsUnsupportedDataLength);
+    }
+
+    let (len_dst, tbs_dst) = unsafe {
+        let ptr = match tbs_type {
+            TbsType::LdevidTbs => &mut LDEVID_TBS_ORG,
+            TbsType::FmcaliasTbs => &mut FMCALIAS_TBS_ORG,
+        } as *mut u8;
+
+        ptr.write_bytes(0, SIZE_LEN);
```
jhand2, line 99: I somewhat prefer using the hand-off table instead of adding structure to the cert portions of DCCM. It's easier to keep track of structure changes between different firmware stages if they're in the same place.

jhand2 (changes requested):

bluegate010:
Closing as this will be taken care of in a separate PR.

PR #310, opened 2023-05-24
[FEAT] Log the values from the fuse in to DCCM

ajisaxena:
This PR logs the values in the Fuse Logs in to DCCM. All the fuses whose values impact the image validation are logged. This includes :- 
1. Vendor public key index fuse
2. vendor public key revocation
3. FMC SVN From Manifest
4. FMC Min SVN
5. FMC SVN from Fuse
6. RT SVN From Manifest
7. RT Min SVN From Manifest


rom/dev/src/rom.ld
```diff
@@ -26,6 +26,7 @@ FHT_ORG          = 0x50002000;
 LDEVID_TBS_ORG   = 0x50003000;
 FMCALIAS_TBS_ORG = 0x50003400;
 PCR_LOG_ORG      = 0x50003800;
+FUSE_LOG_ORG     = 0x50003C00;
```
jhand2, line 29: This overlaps with the PCR log region which is 2K (see PCR_LOG_SIZE)

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -247,12 +251,63 @@ impl FmcAliasLayer {
...
         cprintln!(
             "[afmc] Image verified using Vendor ECC Key Index {}",
-            info.vendor_ecc_pub_key_idx
+            info.vendor_ecc_pub_key_idx,
         );
         report_boot_status(FmcAliasImageVerificationComplete.into());
         Ok(info)
     }
 
+    fn update_fuse_log(log_info: &ImageVerificationLogInfo) -> CaliptraResult<()> {
```
jhand2, line 260: Do we have an idea of what the fuse values will be used for by FMC/RT?

If it's just for mixing into RT Alias Key, ROM could just pass along a hash of all these fuse values.

ajisaxena: This is just to understand the failure of image validation. This is just for postmortem analysis of the failure. I do not think that we will be using these values for anything else.

jhand2: Ok, thanks

rom/dev/src/fuse.rs
```diff
@@ -0,0 +1,68 @@
...
+    if data.len() > 8 {
+        raise_err!(FuseLogUnsupportedDataLength);
+    }
+
+    // Create a FUSE log entry
+    let mut log_entry = FuseLogEntry {
+        entry_id: entry_id as u32,
+        ..Default::default()
+    };
+    log_entry.log_data.as_bytes_mut()[..data.len()].copy_from_slice(data);
```
jhand2, line 56: Prefer `.get` instead of slice indexing to prevent panic

rom/dev/src/fuse.rs
```diff
@@ -0,0 +1,68 @@
...
+/// # Arguments
+/// * `entry_id` - log entry ID
+/// * `data` -  data To log to the fuse log
+///
+/// # Return Value
+/// * `Ok(())` - Success
+/// * `Err(GlobalErr::FuseLogInvalidEntryId)` - Invalid Fuse log entry ID
+/// * `Err(GlobalErr::FuseLogUpsupportedDataLength)` - Unsupported data length
+///
+pub fn log_fuse_data(entry_id: FuseLogEntryId, data: &[u8]) -> CaliptraResult<()> {
```
jhand2, line 42: How should FMC/RT know the fuse log size?

rom/dev/src/flow/cold_reset/fmc_alias.rs
jhand2, line 260: Would it be better to have a fixed-size struct with the fuse log? Then it can be written to DCCM all at once.

rom/dev/src/rom.ld
jhand2, line 29: Can you also make these changes in `fmc/memory.x` and `runtime/memory.x`?

rom/dev/src/fuse.rs
ajisaxena: We have the entire fuse log as 1KB and this is always constant. Each entry is constant in size. If the FMC/RT needs to parse the log we can always use these constants to parse it. Don't think that this will change. Even if the sizes change, still the same methods can be applied to parse the logs. Pretty sure that we will never support variable size entries or the log size.

rom/dev/src/flow/cold_reset/fmc_alias.rs
ajisaxena: Right now we are logging all the fuses at once and the fixed-size struct can be used but at the same time that will take away the flexibility where we want to log a particular fuse values either multiple times or at different times in boot process.

rom/dev/src/rom.ld
ajisaxena: Sure will do.

rom/dev/src/flow/cold_reset/fmc_alias.rs
jhand2: I was thinking something like this:

```rust
#[repr(C)]
struct FuseLogEntry {
  ...
}

#[repr(C)]
struct FuseLog {
  count: u32, // Number of entries
  entries: [FuseLogEntry; MAX_FUSE_LOG_ENTRIES];
}
```

I think this would make serializing/parsing the log simpler. We could just use the `zerocopy` crate to move this to/from DCCM.

rom/dev/src/fuse.rs
jhand2: See suggestion below for how FMC/RT could figure out how many entries there are

rom/dev/src/rom.ld
ajisaxena: Created a separate issue.

https://github.com/chipsalliance/caliptra-sw/issues/322

Resolving this comment.

rom/dev/src/flow/cold_reset/fmc_alias.rs
ajisaxena: We did thought about having a count in place but decided not to for the following reason(s).
 
We have a entry id in each field and then entry_data. For each type of data there is separate entry ID. 0 is considered as invalid entry ID. Right now the parsing code is as simple as reading the entry and deserialize it in to 

```
pub struct FuseLogEntry {
     pub entry_id: u32,
     pub log_data: [u32; 1],
     pub reserved: [u32; 2],
}
```

Right now the log size is about 1k. Each entry is 16 bytes, which means that we can have at most 1024/16 = 64 entries. 
We can deserialize the entire 64 entries using zero copy in one shot and then parse them one by one till we do not hit a entry with a value of zero. In my opinion since the number of entries are fixed and little, we really do not need a count. 

Let me think/discuss this.

jhand2: Number of entries is fixed right now. But FMC/Runtime will need to support multiple different ROM images into the future. Depending on hardcoded values chosen by ROM is problematic because if ROM ever changes in future SoCs, higher level firmware gets very complicated.

ajisaxena: We are just following the PCR log pattern for this PR. Like the PCR, the number of fuses and the manifest data that we log is fixed and don't think that it will change. Having a count here would not help anyway. If we have a count in the structure we will still have to read the entire structure and then parse only the number of entries (still one by one). In my case where there is no count we still read the entire 64 entries and then parse the entries till we hit a invalid entry. Don't see much of a difference in two approaches.

jhand2: I don't understand the problem. `count` is only 4 bytes and then firmware doesn't have to do special-case parsing (fwiw, I would recommend the same thing for PCR log)

ajisaxena: Both the PCR and FUSE logs were basically designed to be no more than the snap-shot of the values. The entry-id of the log decides the location where the entry goes. This was done for the ease of parsing. But just to keep track of this discussion I have created a new issue and we will discuss this more. 

https://github.com/chipsalliance/caliptra-sw/issues/329

image/verify/src/lib.rs
```diff
@@ -21,6 +21,14 @@ use core::ops::Range;
 
 pub use verifier::ImageVerifier;
 
+/// Image Verifification Executable Info
```
mhatrevi, line 24: NIT: Verification

rom/dev/tests/test_fmcalias_derivation.rs
```diff
@@ -185,14 +185,146 @@ fn test_pcr_log() {
...
+
+#[test]
+fn test_fuse_log() {
+    const FMC_SVN: u32 = 4;
+    const FMC_MIN_SVN: u32 = 2;
+
+    let fuses = Fuses {
+        anti_rollback_disable: true,
+        fmc_key_manifest_svn: 0x0F,  // Value of FMC_SVN
+        runtime_svn: [0xF, 0, 0, 0], // Value of FMC_SVN
```
mhatrevi, line 208: Incorrect comment.

mhatrevi (dismissed):

rom/dev/src/fuse.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // Create a FUSE log entry
+    let mut log_entry = FuseLogEntry {
+        entry_id: entry_id as u32,
+        ..Default::default()
+    };
+    log_entry.log_data.as_bytes_mut()[..data.len()].copy_from_slice(data);
+
+    let dst = unsafe {
+        let offset = core::mem::size_of::<FuseLogEntry>() * (entry_id as usize - 1);
```
jhand2, line 59: This can overflow the log memory region

rom/dev/src/fuse.rs
```diff
@@ -0,0 +1,68 @@
...
+        let ptr = (&mut FUSE_LOG_ORG as *mut u8).add(offset);
```
jhand2, line 60: Instead of using this pattern, I would recommend something like

```
extern "C" {
    static mut FUSE_LOG_ORG: [FuseLogEntry; FUSE_LOG_SIZE / size_of::<FuseLogEntry>()];
}
```

That way you can write code that directly indexes that slice, do bounds checks, etc.

jhand2 (changes requested):

rom/dev/src/fuse.rs
ajisaxena: Updated the code.

rom/dev/src/fuse.rs
ajisaxena: Updated the code.

rom/dev/tests/test_fmcalias_derivation.rs
ajisaxena: updated.

image/verify/src/lib.rs
ajisaxena: updated

rom/dev/src/fuse.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    The file contains Fuse-related Implementations.
+
+--*/
+use caliptra_common::{FuseLogEntry, FuseLogEntryId};
+use caliptra_drivers::{CaliptraError, CaliptraResult};
+use core::mem::size_of;
+use zerocopy::AsBytes;
+
+const FUSE_LOG_SIZE: usize = 1024;
```
jhand2, line 18: Ideally we would this this value from the linker script. But I can't really figure out how to make that work. Will keep thinking about it, but doesn't need to block this PR.

jhand2 (dismissed):

mhatrevi (dismissed):

mhatrevi (approved):

PR #311, opened 2023-05-24
[feat] Move firmware download and validation into a separate layer

mhatrevi:
This change moves the firmware download and validation code out of the
FmcAlias layer into a new non-DICE layer.

### DIFF
--- a/rom/dev/src/flow/cold_reset/dice.rs
+++ b/rom/dev/src/flow/cold_reset/dice.rs
@@ -14,9 +14,9 @@ Abstract:
 --*/
 
 use crate::rom_env::RomEnv;
-use caliptra_drivers::{okref, Array4x12, CaliptraResult, Ecc384PubKey, KeyId};
+use caliptra_drivers::{Array4x12, CaliptraResult, Ecc384PubKey, KeyId};
 
-use super::crypto::Ecc384KeyPair;
+use super::{crypto::Ecc384KeyPair, fw_processor::FwProcInfo};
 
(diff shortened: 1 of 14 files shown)


rom/dev/src/flow/cold_reset/dice.rs
```diff
@@ -29,6 +29,8 @@ pub struct DiceInput<'a> {
 
     /// Authority Key Identifier
     pub auth_key_id: &'a [u8; 20],
+
+    pub fw_proc_info: FwProcInfo,
```
bluegate010, line 33: A problem with having this field here is, this struct is passed to every DICE layer, even though it's only actually needed by `FmcAliasLayer::derive()`. Makes the IDevID and LDevID layers harder to reason about.

The `DiceLayer` trait is forcing that all inputs for any layer be added to the `DiceInput` struct, even though they aren't actually needed by all layers. Would suggest deleting the `DiceLayer` trait, and modifying `FmcAliasLayer::derive()` to take a `DiceInput` and a `FwProcInfo` argument, separately.

Incidentally, could then also modify `InitDevIdLayer::derive()` to not take any `DiceInput`, and delete `DiceInput::default()`. The only reason `DiceInput::default()` exists is because all layers are required to take a `DiceInput`, even if `InitDevIdLayer` doesn't need it.

mhatrevi:
Created a new pull request for this: https://github.com/chipsalliance/caliptra-sw/pull/364

PR #330, opened 2023-05-30
Document error handling for Impactless Update

jhand2:
Document in both Runtime and ROM documentation what happens in impactless firmware update fails.

Additionally, provide more documentation on usage of DISABLE_ATTESTATION.

### DIFF
--- a/rom/dev/README.md
+++ b/rom/dev/README.md
@@ -687,3 +687,6 @@ The following are the pre-conditions that should be satisfied:
         - The TOC hash will NOT match, skip the TOC hash validation.
         - We still need to make sure that the hash of the FMC which was stored in the data vault register at cold boot
           still matches the FMC code. This is the hash of the FMC image portion.
+    - If validation fails during ROM boot, the new image will not be copied from
+      the mailbox. ROM will boot the existing FMC/Runtime images. Validation
+      errors will be reported via the CPTRA_FW_ERROR_NON_FATAL register.
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -11,16 +11,23 @@ The Runtime Firmware main function SHALL perform the following on Cold Boot rese
 
 For behavior during other types of reset, see "Runtime Firmware Updates".
 
+If `mailbox_flow_done` is not set during a "warm" "update" boot, it is assumed that Caliptra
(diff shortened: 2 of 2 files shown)


bluegate010 (approved):

rom/dev/README.md
```diff
@@ -687,3 +687,6 @@ The following are the pre-conditions that should be satisfied:
         - The TOC hash will NOT match, skip the TOC hash validation.
         - We still need to make sure that the hash of the FMC which was stored in the data vault register at cold boot
           still matches the FMC code. This is the hash of the FMC image portion.
+    - If validation fails during ROM boot, the new image will not be copied from
```
varuns-nvidia, line 690: This is an interesting tradeoff. I suppose FMC can take a different policy choice if we change our minds here. @vsonims to note this.

runtime/README.md
```diff
@@ -246,10 +252,6 @@ A Runtime Firmware update is triggered by the `CALIPTRA_FW_LOAD` command. Upon
 receiving this command, Runtime Firmware will:
 
 1. Write-lock mailbox
-1. Verify firmware image signature
-    1. Preserve hash for later steps
-1. Update the “Latest TCI” field of the TCI Node which contains the Runtime
-   Journey PCR (TYPE = RTJM, “Internal TCI” flag is set) with hash.
 1. Invoke “Impactless Reset”
 
 Once Impactless Reset has been invoked, FMC will load the hash of the image
```
varuns-nvidia, line 257: does this imply that the hash will not be copied if the Manifest does not pass verification?

jhand2: Hmm. So we have 2 PCRs here that get extended by FMC:

1. RT Journey
2. RT Current

I think the logic has to be:

* If verification passes, ROM resets (2), FMC extends both (1) and (2)
* If verification fails, ROM doesn't reset (2), FMC does nothing

cc @rusty1968

jhand2: Added these details

rom/dev/README.md
jhand2: The main motivation for this is to prioritize availability, which I think aligns with priorities I've heard in the past. Also I think it aligns with the spirit of Impactless Update.

mhatrevi (approved):

PR #332, opened 2023-05-31
FMC : Use error constants instead of placeholders

rusty1968:


bluegate010 (approved):

fdamato (approved):

error/src/lib.rs
```diff
@@ -252,6 +252,10 @@ impl CaliptraError {
 
     /// FMC Errors
     pub const FMC_RT_ALIAS_UNIMPLEMENTED: CaliptraError = CaliptraError::new_const(0x000F0001);
+    pub const FMC_HANDOFF_INVALID_PARAM: CaliptraError = CaliptraError::new_const(0x000F0002);
```
FerralCoder, line 255: going forward, I would like to see FMC_HANDOFF_INVALID_PARAM split into more specific error codes. If I am reading this right, we have 16 bits worth of codes just from FMC right?

error/src/lib.rs
```diff
@@ -252,6 +252,10 @@ impl CaliptraError {
 
     /// FMC Errors
     pub const FMC_RT_ALIAS_UNIMPLEMENTED: CaliptraError = CaliptraError::new_const(0x000F0001);
```
FerralCoder, line 254: move to the end of the list instead of the beginning so we don't leave a hole once we implement this code and remove this error

PR #337, opened 2023-06-01
Update to the latest top-of-tree RTL.

korran:
Also, fix the HwModelVerilated AHB trace behavior such that it doesn't fail with error "Unsupported hsize value 0b11" when the CPU emits a 64-bit AHB transaction. This was preventing the smoke_test from passing.


JohnTraverAmd:
Why are we pointing to a commit not on RTL repo not on main branch?

korran:
> Why are we pointing to a commit not on RTL repo not on main branch?

Because the main branch is not compatible with the firmware; it doesn't have the CSRNG enabled. That's why the nightly failed.

korran:
> Can we update RTL Main with this fix first?

I don't have the power to do that. Do fix the build we either need to add the cherry-pick or revert #305. The cherry-pick is easier.

JohnTraverAmd:
Can we disable the failing part and file an issue.  I'd prefer to not revert that commit.

korran:
#305 is now reverted, and I've updated the submodule in this PR to point to caliptra-rtl main HEAD.

fdamato (dismissed):

fdamato (approved):

PR #341, opened 2023-06-01
Create threat-model.md, opening PR for review/merge of ROM threat model

### LINKED ISSUE #295: Caliptra ROM Threat Model Analysis and Mitigation Recommendations
# Outline
The documentation’s purpose is to iterate threat that the Caliptra ROM will face from potential attackers, we will also look at the potential mitigations and the difficulty level of the attack. The goal for this document is to look at just the Caliptra ROM and what thread that it can potentially face.  There is a comprehensive [Caliptra Asset & Threats](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#caliptra-assets--threats) writing currently in the Caliptra specification that covers Caliptra IP as a whole. 
The main area of attack surface that we will look at includes the following,
-	ROM Confidentiality / Reverse Engineering Attack
-	Mailbox Interface Attack
-	Physical Attacks (Glitching)
-	DPA/DFA Attacks
-	Oracle Attacks

# ROM Confidentiality / Reverse Engineering Attack

## Attack Description
Usually, the ROM of a silicon’s ROT is considered confidential for many reasons, especially to prevent reverse engineering attempts that can potentially lead to more exploitable vulnerabilities from white box attacks.  Since Caliptra’s soul is an Open-Source Project and the ROM code is available for all to see.  There is no confidentiality in the ROM code at all and we need to consider all types of white box attacks.  Attackers are armed with full knowledge of the ROM code so any logical error might be able to be turned into an exploitable vulnerability.

## Mitigation
Mitigation for this will consist mostly typical code review and analysis methods.
- Rigorous code review 
  - The best defense is to ensure that the ROM code does not have exploitable vulnerability that can be exported.  Code review and finding coding error early on can help reduce vulnerability vectors that might be able to be exported by a malicious actor.
- Proactive static analysis 
  - Static analysis would be a great tool to employe in mitigating this.  Even though Rust does have compile-time error checks, it can still benefit from running rust focused static analysis tools to further reduce potential coding errors.
- ROM Patching
  - ROM Patching is one way to be able to patch the ROM without taping out silicon again if a bug in the masked ROM is found post tapeout.  Currently it is not supported by Caliptra, but it might be something we can consider in the future if the need arises.  ROM Patch does bring other security concern such as measuring and attesting to the ROM patch.

# Mailbox Interface Attacks
## Attack Description
One of the most common attacks that would be done on a secure processor such as Caliptra would be interface based attack on the external interface that is exposed to the external software conduct offensive fuzzing on the interface.  Any unexpected behavior not handled well could become a vector that results in an exploitable situation.  
Mitigation
Multiple mitigation can be done to alleviate such attack.  From enclosing Caliptra interface to another highly trusted entity such as SOC ROT/management process to proactively engaging in dynamic analysis and fuzzing of the ROM interface.
- Enclose Caliptra mailbox interface to SOC management processor.
  - SOC Management processor is usually of a higher trust level compared to application processor (CPU).  By enclosing the Caliptra mailbox interface to such entity, it reduces the attack surface of an attacker gaining the ability to attack Caliptra mailbox.  While this is not something Caliptra ROM can affect.  We can put this as a recommendation for Caliptra integration guidelines. 
- Proactive fuzzing
  - Conduct fuzzing on any Caliptra mailbox interface that the ROM provides.  This includes fuzzing the Caliptra mailbox commands and manipulate the Caliptra firmware image.  This shall be done for both Cold boot and Hitless update boot sequence.  
  - To reduce attack further, if we have decided that warm boot for Caliptra is no longer needed, we should take Warm boot code out of ROM to reduce attack surface.
- Strengthen error handling.
  - Another mitigation to block attacker from doing offensive interface attack is a robust error handling policy.  A fatal or unexpected error can cause Caliptra to go into a turtle like mechanism which can only be solved with a system reset will significantly increase the difficulty for the attacker to fuzz the interface offensively.
- Strong memory property protection
  - Many interface-based attack rely on out of bound access to memory region that should not be executable (stack / data).  By having strong memory property protection (R/W Only, NX), we can ensure that the risk of ROP type of attack can be decreased even if the attacker were able to compromise the interface initially. 

# Physical Attacks (Glitching / Fault Injection)
## Attack Description
Physical attack includes a wild range of attacks that includes but is not limited to the following.
- Voltage / Clock based glitch.
- Thermal attacks
- Electromagnetic fault attack
- Laser attack
While the attack methodology and expertise required differ significantly, the goal of the attack and the observable behavior from the attack can be summarized to be the unexpected behavior of code execution that does not follow the original intended flow.  The effect of a glitch attack is most likely randomized behavior in the system under attack.  The observable behavior that has security concerns can be the following.
- Instruction skipping or modification.
- Data corruption
- Control flow alteration
While the chance of success might be very low for an attacker to randomly attack Caliptra which is a very small IP within a larger silicon.  A more sophisticated attacker will be able to pinpoint the attack more precisely both spatially and temporally.  

## Mitigation
The most effective defense against physical glitch / fault injection attacks would be specialized circuits in HW that would detect or prevent such attacks.  However, from a coding perspective, there are well established glitch mitigation coding mechanics that can help with preventing glitched behavior to turn into an exploitable vulnerability.
- Initialize / Re-initialize status return to error prior to use.
  - By relying on compiler to initialize status registers is not reliable.  It is important to initialize all status registers to failure/error so that if a function call return that would change the status register is skipped or bypassed, the status check would result in error.  
  - It is also essential to re-initialize the status register to failure/error if it were to be used again later in the function for another call.
- Critical path redundancy checks on if/else/switch statements.
  - Majority of the glitch attack target critical path if/else statements where things such as debug enable check, authentication check is done and the test for result is critical to the security assurance of the system.  Running unauthenticated code for escalation of privilege is a known attack that has been demonstrated in multiple instances.  By applying multiple checks (check for failure again in success path for example) can help mitigate most of those attacks.  
- Increase hamming distance.
  - Glitch attack also depend heavily on bit flips.  Typical definition of a Boolean state would easily be switched from true to false with glitch attacks if it is presented with only one bit difference.  By increasing the hamming distance, it would mitigate the single bit flip failures that can be seen from glitches.
  - Hamming distance does not only apply to true/false usage, loop increment and termination condition can also be easily glitched without hamming distance implemented.  
- Randomized Delay
  - Since most successful glitch attack that result to exploitable vulnerability are not done randomly from a timing perspective, it is important to add randomized delay into the cold flow.  For example, the attacker could time off external bus access to storage device, x ms after would likely be the authentication call.  A timed attack with that knowledge is a lot more likely to succeed than a pure random attack.  By adding random delays in the critical path between externally measurable signal to internal critical security function call can mitigate this and make it harder for the attacker. 
- Control flow audit
  - Sometimes a large section of code execution can be skipped through glitching.  Adding counter track to ensure all the critical section is executed can catch and alleviate otherwise undetectable problems like that could have undesirable unknown effect for code executing later in the flow.

# DPA / DFA (Side Channel) Attacks
## Attack Description
DPA (Differential Power Analysis) and DFA (Differential Fault Analysis) attacks are common side-channel attacks targeting cryptographic systems in the secure element.  
- To launch a successful DPA attack on cryptographic system attacker must be able to control input of crypto operation either ciphertext of decryption or plaintext of encryption and measure power consumption of a target device with external means such as oscilloscope.
  - Caliptra's assets that can be attacked with DPA
    - UDS – if attacker is able to feed known IV value through glitching or probing it will allow an attacker to recover decrypted/deobfuscated UDS.
    - RTL key – if attacker able to deterministically glitch UDS seed known cipher text can be fed into deobfuscation engine and RTL key recovered with DPA attack.
- DPA on ECDSA
  - Side-channel on the elliptic curve exploits the fact that the secret scalar value is used one bit at a time in ECDSA point multiplication operation and if recovered can be used to derive private key. Successful attack requires multiple traces where secret scalar remains constant for each trace, but a different point is used for each trace however we don’t need a knowledge on what the point is.
    - Caliptra uses chip-class ECDSA-384 private key provisioned by a vendor to sign device identifier certificate.
- DFA attacks require feeding known plain/ciphertext, glitching cryptographic IP and observing the output of cryptographic operation. While DFA attacks require all 3 previously mentioned conditions, the number of traces required to perform such an attack is in the order of few thousand traces also often times DFA attacks can be performed purely from software if an attacker has ability to control SoC voltage via software interface.

## Mitigation
- IV
  - Treat IV as private and do not expose software interfaces that use IV as an input.
- Crypto IP
  - Use established coutermeasures such as adding dummy rounds, random data masking, timing randomization.

# Caliptra Integration Attacks
## Attack Description
Certain part of Caliptra’s security rely on the correct integration of Caliptra by the SOC Vendors.  For example, it is possible to integrate Caliptra without providing HW FUSE for certain elements and have HW to implement register interface that the SOC Management Processor can write to for pushing FUSE values into Caliptra.  An implementation like this can be invisible to the outside world as Caliptra can be only accessible by the SOC Management Processor.  A successful attack on the SOC Management Processor can also mean the security of the Caliptra is now lost.  
However, Caliptra is within SOC Vendor’s trust boundary as the Caliptra asset (keys, etc) belong to the SOC Vendor.  We can argue that a SOC Vendor attacking on Caliptra through integration would not be in scope of the threat model.  However correct integration should be enforced to lower other risks. 
Strictly from ROM perspective, the risk is the correct integration of the Caliptra ROM without modification.  This is not measured to be attestable, same applies to the RTL implementation of Caliptra overall.

## Mitigation
- Integration guidance
  - By giving detailed integration guidance, we can reduce the problems that might arise during integration.
- Caliptra certification
  - Having a proper certification process to ensure Caliptra integration is done correctly can also mitigate the risk here.

jameszhang-nvidia:
Threat model description of ROM based threat and mitigation suggestion.  Should be used as the base for ROM code dev and review.


rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Mitigation
+Mitigation for this will consist mostly typical code review and analysis methods.
+- Rigorous code review 
+  - The best defense is to ensure that the ROM code does not have exploitable vulnerability that can be exported.  Code review and finding coding error early on can help reduce vulnerability vectors that might be able to be exported by a malicious actor.
+- Proactive static analysis 
+  - Static analysis would be a great tool to employe in mitigating this.  Even though Rust does have compile-time error checks, it can still benefit from running rust focused static analysis tools to further reduce potential coding errors.
```
korran, line 20: We're currently using cargo clippy's defaults lints (all warnings are errors):

https://github.com/rust-lang/rust-clippy

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+- Proactive fuzzing
+  - Conduct fuzzing on any Caliptra mailbox interface that the ROM provides.  This includes fuzzing the Caliptra mailbox commands and manipulate the Caliptra firmware image.  This shall be done for both Cold boot and Hitless update boot sequence.  
+- Strengthen error handling.
+  - Another mitigation to block attacker from doing offensive interface attack is a robust error handling policy.  A fatal or unexpected error can cause Caliptra to go into a turtle like mechanism which can only be solved with a system reset will significantly increase the difficulty for the attacker to fuzz the interface offensively.
+- Strong memory property protection
+  - Many interface-based attack rely on out of bound access to memory region that should not be executable (stack / data).  By having strong memory property protection (R/W Only, NX), we can ensure that the risk of ROP type of attack can be decreased even if the attacker were able to compromise the interface initially.
```
korran, line 36: This is very primitive today; data in DCCM is NX, but all of ICCM, including .rodata, is executable.

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+- Enclose Caliptra mailbox interface to SOC management processor.
+  - SOC Management processor is usually of a higher trust level compared to application processor (CPU).  By enclosing the Caliptra mailbox interface to such entity, it reduces the attack surface of an attacker gaining the ability to attack Caliptra mailbox.  While this is not something Caliptra ROM can affect.  We can put this as a recommendation for Caliptra integration guidelines. 
+- Proactive fuzzing
+  - Conduct fuzzing on any Caliptra mailbox interface that the ROM provides.  This includes fuzzing the Caliptra mailbox commands and manipulate the Caliptra firmware image.  This shall be done for both Cold boot and Hitless update boot sequence.  
+- Strengthen error handling.
+  - Another mitigation to block attacker from doing offensive interface attack is a robust error handling policy.  A fatal or unexpected error can cause Caliptra to go into a turtle like mechanism which can only be solved with a system reset will significantly increase the difficulty for the attacker to fuzz the interface offensively.
```
korran, line 34: So are you suggesting that caliptra should add a exponentially-slower delay loop after every mailbox transation that ends in error?

korran: Since the firmware is fully open, I would expect most attackers to fuzz the mailbox interface in some sort of simulation environment, where they can remove these sorts of mitigations.

rom/dev/doc/threat-model.md
jameszhang-nvidia: Hi Kor, the goal for this is to list otu the thread and potential mitigations.  I think after we agree that this is the list of threat we care about we would indicate how they are addressed.  For example this threat can be somewhat addressed by using the werror equivalent for rust.  There are other static analysis strategy for RUST that might further reduce risk if we plan on using them.  However the goal for getting this PR merged is to just identify and list out the threats we care about

rom/dev/doc/threat-model.md
jameszhang-nvidia: RISC-V priv architecture does have a PMP (Physical Memory Protection) unit that can further isolate the memory of ICCM.  I do not believe that PMP is a unit that exist in the Caliptra RISC-V core today, but it can be something we try to get in for 2.0.  Again this is a threat model and maybe not all mitigation is feasible, or worth it (from ROI perspective) for 1.0.  

Another way to deal with this is to have ROM to de-priv itself and use MPU to protect it's own code from overflow type of errors.

rom/dev/doc/threat-model.md
jameszhang-nvidia: >So are you suggesting that caliptra should add a exponentially-slower delay loop after every mailbox transation that ends in error?
I don't think that we would need a delay loop for error handling.  The term turtle means to retreat back into a protective environment and stop all function.  For example if you detect a glitch attempt, and the glitch mitigation code has caught behavior that normal code execution should never happen.  At that time the ROT can shut down and refuse to service anymore.  

>Since the firmware is fully open, I would expect most attackers to fuzz the mailbox interface in some sort of simulation environment, where they can remove these sorts of mitigations.
You are right, similar to that we can not have confidentiality as the soul of this project is the open source nature. it does create more opportunity for attacker to do white box attacks on the system.  But I think strenghtening error handling is still useful as it can make glitch attack a lot more difficult to execute.

korran: I read your statement in the context of "unexpected input to the mailbox interface", is that what you intended? 

It definitely makes sense to cease all function if we have any indication of corrupt internal state, but I'm skeptical whether the outside world should be able to "turtle" us just by sending us badly formatted commands.

korran (dismissed):

rom/dev/doc/threat-model.md
jameszhang-nvidia: I don't think all input that fails sanitary input check should go into turtle.  However if an error does cause the code for handling the mailbox to throw an unexpected or un-handle-able error then that should cause caliptra to "turtle".

korran: CPU/bus traps are pretty much the only "unexpected" errors that can happen, and those definitely result in a [turtle state](https://github.com/chipsalliance/caliptra-sw/blob/966db5a0122eb5d98779cd0c021052dd99d49ae2/rom/dev/src/main.rs#L151-L161). We don't allow code that can panic (a test verifies the symbol has been optimized out).

Everything else is a normal error-code response from the various functions.

I do expect that yes-to-be-written CFI failures will jump to a "turtle state".

jameszhang-nvidia: I think most likely error that would result in turtle would definitely be some kind of abort exception that comes from CPU or bus errs.  However there can be software error too for checks that does not trigger hw exception, such as glitch mitigation detection which your CFI mention also falls under.  Overall I think it's a good idea for security to have a capability to shut off the ROT in presence of a perceived attack.

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+## Mitigation
+Mitigation for this will consist mostly typical code review and analysis methods.
+- Rigorous code review 
+  - The best defense is to ensure that the ROM code does not have exploitable vulnerability that can be exported.  Code review and finding coding error early on can help reduce vulnerability vectors that might be able to be exported by a malicious actor.
+- Proactive static analysis 
+  - Static analysis would be a great tool to employe in mitigating this.  Even though Rust does have compile-time error checks, it can still benefit from running rust focused static analysis tools to further reduce potential coding errors.
+- ROM Patching
+  - ROM Patching is one way to be able to patch the ROM without taping out silicon again if a bug in the masked ROM is found post tapeout.  Currently it is not supported by Caliptra, but it might be something we can consider in the future if the need arises.  ROM Patch does bring other security concern such as measuring and attesting to the ROM patch.
+
+# Mailbox Interface Attacks
```
jhand2, line 24: These same attacks probably also apply to all the registers in http://eregs.caliptra.org/

jhand2, line 24: Probably the biggest concern for ROM would be image load. It might be worth enumerating what an attacker can do to trick Caliptra into loading bad fw.

TOCTOU attacks come to mind since the image is validated while sitting in the mailbox.

rom/dev/doc/threat-model.md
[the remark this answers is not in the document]
jameszhang-nvidia: I was thinking something more drastic such as a HW (future version) where the HW itself will shut down and clear our asset/etc.
For current version I believe what we can do is to have a sw mechanism that clear out asset, clear out the code and just fill the ITCM with branch to self instruction (at least part of it and hang.

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+- Proactive fuzzing
+  - Conduct fuzzing on any Caliptra mailbox interface that the ROM provides.  This includes fuzzing the Caliptra mailbox commands and manipulate the Caliptra firmware image.  This shall be done for both Cold boot and Hitless update boot sequence.  
+- Strengthen error handling.
+  - Another mitigation to block attacker from doing offensive interface attack is a robust error handling policy.  A fatal or unexpected error can cause Caliptra to go into a turtle like mechanism which can only be solved with a system reset will significantly increase the difficulty for the attacker to fuzz the interface offensively.
+- Strong memory property protection
+  - Many interface-based attack rely on out of bound access to memory region that should not be executable (stack / data).  By having strong memory property protection (R/W Only, NX), we can ensure that the risk of ROP type of attack can be decreased even if the attacker were able to compromise the interface initially. 
+
+# Physical Attacks (Glitching / Fault Injection)
+## Attack Description
+Physical attack includes a wild range of attacks that includes but is not limited to the following.
```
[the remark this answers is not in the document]
jameszhang-nvidia: Thanks, will fix

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,94 @@
...
+- Voltage / Clock based glitch.
+- Thermal attacks
+- Electromagnetic fault attack
+- Laser attack
+While the attack methodology and expertise required differ significantly, the goal of the attack and the observable behavior from the attack can be summarized to be the unexpected behavior of code execution that does not follow the original intended flow.  The effect of a glitch attack is most likely randomized behavior in the system under attack.  The observable behavior that has security concerns can be the following.
```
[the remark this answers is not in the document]
jameszhang-nvidia: will fix

rom/dev/doc/threat-model.md
jameszhang-nvidia: added

jameszhang-nvidia: Will add

jameszhang-nvidia: Pushed the changes that addresses your comment Jordan, please take a look.  thanks

rom/dev/doc/threat-model.md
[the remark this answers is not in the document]
jhand2: Ok. No objections from me.

jhand2 (dismissed):

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Mitigation
+
+Mitigation for this will consist mostly typical code review and analysis methods.
+
+- Rigorous code review 
+  - The best defense is to ensure that the ROM code does not have exploitable vulnerability that can be exported.  Code review and finding coding error early on can help reduce vulnerability vectors that might be able to be exported by a malicious actor.
+- Proactive static analysis 
+  - Static analysis would be a great tool to employe in mitigating this.  Even though Rust does have compile-time error checks, it can still benefit from running rust focused static analysis tools to further reduce potential coding errors.
+- ROM Patching
```
steven-bellock, line 27: Is it a security requirement that the Caliptra bootrom can only be realized as a mask ROM, and not, for instance, injected into OTP during manufacturing? If so then the main specification should state that clearly. However if folks are OK with storing the bootrom in OTP then, at least for newly manufactured devices, the bootrom can be patched during manufacturing. Whether devices in the field can be patched is another question.

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,127 @@
+# Outline
+
+The documentation’s purpose is to iterate threat that the Caliptra ROM will face from potential attackers, we will also look at the potential mitigations and the difficulty level of the attack. The goal for this document is to look at just the Caliptra ROM and what thread that it can potentially face.  There is a comprehensive [Caliptra Asset & Threats](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#caliptra-assets--threats) writing currently in the Caliptra specification that covers Caliptra IP as a whole.
```
steven-bellock, line 3: `thread` -> `threat`

rom/dev/doc/threat-model.md
```diff
@@ -0,0 +1,127 @@
...
+
+Time of Check Time of Use is a typical attack where the attacker try to time the authenticator/checker if both the attacker and authenticator has write access to the data that the authentircator has access to at the same time.  The attacker can time the attack in a way that it lets the authenticator verify an authentic copy of the data then change the data underneath the authenticator so that a malicious copy can be used. 
+For Caliptra ROM, this is especially something we need to worry about for ROM image load.
+
+## Mitigation
+
+Mitigation will consist of either block the potential attacker (sending in image for authentication) from modifing the Caliptra-fw image prior to Caliptra ROM is to authenticate it, or have Caliptra ROM to load Caliptra-FW image into internal TCM prior to authentication.
+
+- Disable external entity ability to modify Caliptra mailbox content post mailbox command that triggers Caliptra ROM to act on mailbox data.
+- Have Caliptra ROM move mailbox content into TCM prior to authentication
```
steven-bellock, line 57: `TCM` -> `CCM`.

rom/dev/doc/threat-model.md
jameszhang-nvidia: That would cause even more problem such as manufacture security.  It's also not easily detectable if something malicious happens unless there's HW engine that's measuring and extending the measurement of ROM

rom/dev/doc/threat-model.md
jameszhang-nvidia: Fixed

mschilder123 (approved):

jhand2 (approved):

mhatrevi (approved):

mhatrevi (approved):

JohnTraverAmd (approved):

PR #347, opened 2023-06-02
[feat] Store ldevid/fmclias TBS sizes and PCR/Fuse log entry counts in FHT

### LINKED ISSUE #343: Add PCR and Fuse log entry counts, and LDEVID and FMCALIAS TBS sizes to the Firmware Handoff Table

mhatrevi:


rom/dev/src/flow/cold_reset/mod.rs
```diff
@@ -73,16 +73,31 @@ impl ColdResetFlow {
...
+        TbsType::FmcaliasTbs => {
+            env.fht_data_store.fmcalias_tbs_size = tbs.len() as u16;
+            unsafe {
+                let ptr = &mut FMCALIAS_TBS_ORG as *mut u8;
+                core::slice::from_raw_parts_mut(ptr, tbs.len())
+            }
+        }
     };
 
     dst[..tbs.len()].copy_from_slice(tbs);
```
bluegate010, line 103: Shall we error out if tbs.len() is greater than the TBS region size?

common/src/hand_off.rs
```diff
@@ -261,6 +261,18 @@ pub struct FirmwareHandoffTable {
     /// Index of RT SVN value in the Data Vault
     pub rt_svn_dv_hdl: HandOffDataHandle,
 
+    /// LdevId TBS Size.
```
bluegate010, line 264: @jhand2 had a suggestion to also propagate addresses, rather than rely on the linker script. Thoughts?

mhatrevi: Sounds good; I'll add (TBS and LOG) addresses to the FHT and will remove them from the linker script.

rom/dev/src/flow/cold_reset/mod.rs
jhand2: I would recommend doing something like this (can be done in a follow up PR if we want)

```rust
extern "C" {
    static mut LDEVID_TBS_ORG: [u8; LDEVID_TBS_SIZE];
    static mut FMCALIAS_TBS_ORG: [u8; FMCALIAS_TBS_SIZE];
}

...

let dst = match tbs_type {
    TbsType::LdevidTbs => {
            env.fht_data_store.ldevid_tbs_size = tbs.len() as u16;
            unsafe { &LDEVID_TBS_ORG[..] }
        }
        TbsType::FmcaliasTbs => {
            env.fht_data_store.fmcalias_tbs_size = tbs.len() as u16;
            unsafe { &FMCALIAS_TBS_ORG[..] }
        }
};
```

This makes the bounds checking easier and ensures the compiler helps you out (by inserting a panic which our test should check).

mhatrevi: Will fix it as part of https://github.com/chipsalliance/caliptra-sw/issues/354

common/src/hand_off.rs
```diff
@@ -261,6 +261,18 @@ pub struct FirmwareHandoffTable {
...
+    pub ldevid_tbs_size: u16,
+
+    /// FmcAlias TBS Size.
+    pub fmcalias_tbs_size: u16,
+
+    /// PCR log entry count.
+    pub pcr_log_entry_count: u16,
+
+    ///Fuse log entry count.
+    pub fuse_log_entry_count: u16,
```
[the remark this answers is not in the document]
mhatrevi: This was the conversation with @ajisaxena about not really needing the log counts since the log array is fixed size and is indexed by the entry id. If this is no more needed, I'll remove the entry counts.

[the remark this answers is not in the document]
jhand2: IMO it would make more sense to fill the fuse log sequentially and then have a count. Having a sparsely filled out array seems unnecessary. But I could be missing something.

So my recommendation would be:

* If log is filled sequentially, include fuse log count in the FHT
* If not, fuse log count can be omitted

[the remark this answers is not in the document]
mhatrevi: Removed the fuse and pcr log counts.

jhand2 (approved):

ajisaxena (approved):

rom/dev/src/fht.rs
```diff
@@ -28,7 +28,22 @@ use crate::{
 const FHT_MAJOR_VERSION: u16 = 1;
 const FHT_MINOR_VERSION: u16 = 0;
 
-struct FhtDataStore {}
+extern "C" {
+    static mut LDEVID_TBS_ORG: u8;
```
korran, line 32: Please make these u32 so to avoid the casts below.

Issue #350, opened 2023-06-04
FMC and Runtime address overlap, and image signing code doesn't complain 


korran:
At top of tree:

```
  1,885,675 UART: [afmc] Loading FMC at address 0x40000000 len 9176
  1,900,372 UART: [afmc] Loading Runtime at address 0x40002000 len 5008
```

Before closing this bug, three things must be done:

- [ ] The ROM should abort if the FMC and runtime addresses overlap
- [ ] The image generation code should complain if the FMC and runtime addresses overlap
- [ ] The linker script for the runtime firmware should be adjusted to make room for the larger FMC, or the FMC should get some size optimizations.
- [ ] Add an integration test to check for failure in the ROM when the FMC and runtime addresses overlap.

PR #352, opened 2023-06-05
[fix] [issue #350] FMC and RT overlap address detection

rusty1968:
* Detects FMC/RT overlap in ROM.
* Increases size of FMC to 16k.


rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -313,6 +313,27 @@ impl FmcAliasLayer {
...
+        cprintln!(
+            "[afmc] Loading Runtime at address 0x{:08x} len {}",
+            manifest.runtime.load_addr,
+            manifest.runtime.size
+        );
+
+        // Detect if FMC and RT address ranges overlap
+        // Load address of FMC <= End address of RT
+        // and End address of FMC >= Load address of RT
+        if manifest.fmc.load_addr <= manifest.runtime.load_addr + manifest.runtime.size
```
fdamato, line 325: is this first check really required and if so, shall we also check the dual condition fmc after rt fw

fdamato (approved):

rom/dev/src/flow/cold_reset/fmc_alias.rs
fdamato: sorry I did misread the code here...

korran, line 325: Can we add an integration test?

korran: Coordinate with @ajisaxena so you don't duplicate work

korran: Actually, let's add the integration test in a future PR. Updated #350 with the new task.

fmc/test-fw/test-rt/memory.x
```diff
@@ -13,7 +13,7 @@ NSTACK_ORG  = 0x5001FC00;
 
 ROM_SIZE    = 32K;
 MBOX_SIZE   = 128K;
-ICCM_SIZE   = 128K;
+ICCM_SIZE   = 112K;
```
FerralCoder, line 16: What does this actually represent here? Unused size?

image/types/src/lib.rs
```diff
@@ -392,4 +392,13 @@ impl ImageTocEntry {
     pub fn image_size(&self) -> u32 {
         self.size
     }
+
+    pub fn overlaps(&self, other: &ImageTocEntry) -> bool {
```
FerralCoder, line 396: The check here is incomplete. This only works if the param passed in has a higher load address than "self". For completeness, it would be best to check both directions.

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -313,6 +313,27 @@ impl FmcAliasLayer {
...
+            && manifest.fmc.load_addr + manifest.fmc.size >= manifest.runtime.load_addr
```
FerralCoder, line 326: again, this check is only valid if we always load FMC at a lower address than RT. I would suggest checking for overlap the other way too, in case RT is ever loaded lower than FMC,

korran: This is the standard overlap detection logic. Equivalent to:

```
fmc.start_addr <= runtime.end_add &&
fmc.end_addr >= runtime.start_addr
```

(although though I would argue for use of < rather than <=)

korran (changes requested):

mhatrevi (changes requested):
The fix needs additional changes to image generation; this is being discussed between Ajitabh, Anthony and myself over Slack.

image/types/src/lib.rs
korran: This is the standard overlap detection logic. Equivalent to:

```
a.start <= b.end && a.end >= b.start
```

(but I would argue it should be > instead of >=, and < instead of <=)

rusty1968:
@mhatrevi , if you already have a PR ready  with all the spots I can close mine.

rusty1968:
@mhatrevi  i closed the PR, but this is impacting scheduled tasks. @fdamato was debugging this over the weekend. if you can address this problem in several prs instead of one big pr, that would be better.

PR #362, opened 2023-06-07
Use PCR0 instead of PCR1 for Alias FMC CDI derivation.

bluegate010:
ROM does not place measurements in PCR1.


mhatrevi (approved):

jhand2 (commented):
Feel free to do in a follow up, but I also see some comments in rom/dev/src/pcr.rs that need to be updated

bluegate010:
> Feel free to do in a follow up, but I also see some comments in rom/dev/src/pcr.rs that need to be updated

Ack, will do in a follow-up.

PR #364, opened 2023-06-07
[feat] Move firmware download and validation into a separate layer

mhatrevi:
This change moves the firmware download and validation code out of the FmcAlias layer into a new layer.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fmc_alias.rs
+++ b/rom/dev/src/flow/cold_reset/fmc_alias.rs
@@ -14,40 +14,33 @@ Abstract:
 --*/
 
 use super::crypto::{Crypto, Ecc384KeyPair};
-use super::dice::{DiceInput, DiceLayer, DiceOutput};
+use super::dice::{DiceInput, DiceOutput};
+use super::fw_processor::FwProcInfo;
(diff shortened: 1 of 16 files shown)


rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -121,272 +77,14 @@ impl DiceLayer for FmcAliasLayer {
...
-        }
-
         // Generate Local Device ID Certificate
-        Self::generate_cert_sig(env, info, input, &output, &nb.not_before, &nf.not_after)?;
+        Self::generate_cert_sig(env, input, &output, fw_proc_info)?;
 
         report_boot_status(FmcAliasDerivationComplete.into());
         cprintln!("[afmc] --");
 
         Ok(output)
```
bluegate010, line 86: Do we need to return `output` here?

bluegate010 (dismissed):

rom/dev/src/flow/cold_reset/fmc_alias.rs
mhatrevi: Not really needed; removed.

korran (approved):

PR #367, opened 2023-06-08
Update rust toolchain from 1.68 to 1.70 (-2076 bytes)

korran:
Set the desired toolchain in rust-toolchain.toml, so everyone uses the same toolchain by default.

[Size analysis](https://github.com/chipsalliance/caliptra-sw/actions/runs/5206880511)

The improved risc-v codegen in this release reduces the size of the ROM
by 2076 bytes.

As part of this change, we can remove the cargo index cache, as in 1.70
cargo [switched to](https://github.com/rust-lang/rust/blob/642c92e63008ffb49f6ad8344e07bfa7d5b0d9bb/RELEASES.md?plain=1#L101) the [sparse index protocol](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html#background), which is much faster.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,7 @@
+# Licensed under the Apache-2.0 license
+
+[toolchain]
+channel = "1.70"
+targets = ["riscv32imc-unknown-none-elf"]
+profile = "minimal"
+components = ["rustfmt", "clippy"]
--- a/.github/workflows/build-test-verilator.yml
+++ b/.github/workflows/build-test-verilator.yml
@@ -22,7 +22,7 @@ jobs:
       SCCACHE_GHA_CACHE_TO: sccache-verilator-10000
       SCCACHE_GHA_CACHE_FROM: sccache-verilator-
       # Change this to a new random value if you suspect the cache is corrupted
-      SCCACHE_C_CUSTOM_CACHE_BUSTER: f3e6951f0c1d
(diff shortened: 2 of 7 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+# Licensed under the Apache-2.0 license
+
+[toolchain]
+channel = "1.70"
+targets = ["riscv32imc-unknown-none-elf"]
```
rkr35, line 5: This is neat. I wasn't aware of `rust-toolchain.toml`.

Do we still need the "Install risv32imc target" build step in CI?

rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
...
+profile= "default"
```
rkr35, line 6: The `minimal` profile skips downloading the `rust-docs`. I wonder if something like this config is faster for CI:
```toml
[toolchain]
channel = "1.70"
targets = ["riscv32imc-unknown-none-elf"]
profile= "minimal"
components = ["rustfmt", "clippy"]
```

rust-toolchain.toml
korran: Good idea. Removed.

mhatrevi (approved):

rust-toolchain.toml
korran: Changed. Let's see if CI passes.

PR #373, opened 2023-06-09
[feat] Zeroize the crypto hardware blocks

mhatrevi:
This change updates the crypto drivers to zeroize a crypto block
on completion of an crypto operation. Additionally, in the event of an error condition,
the ROM has been updated to zeroize all the crypto hardware blocks.

### DIFF
--- a/drivers/src/ecc384.rs
+++ b/drivers/src/ecc384.rs
@@ -193,6 +193,8 @@ impl Ecc384 {
             y: Array4x12::read_from_reg(ecc.pubkey_y()),
         };
 
+        self.zeroize();
+
         Ok(pub_key)
     }
 
@@ -240,6 +242,8 @@ impl Ecc384 {
             s: Array4x12::read_from_reg(ecc.sign_s()),
         };
 
+        self.zeroize();
+
         Ok(signature)
     }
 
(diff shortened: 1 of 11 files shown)


JohnTraverAmd (approved):

drivers/src/ecc384.rs
```diff
@@ -286,11 +290,16 @@ impl Ecc384 {
...
+        let result = verify_r == signature.r;
+
+        self.zeroize();
+
+        Ok(result)
+    }
+
+    /// Zeroize the hardware registers.
+    pub fn zeroize(&mut self) {
+        self.ecc.regs_mut().ctrl().write(|w| w.zeroize(true));
```
korran, line 302: Does the zeroize operation really happen in a single cycle? Do we need to wait for this bit to clear before we can use the ECC peripheral again?

korran: As best as I can tell from the RTL, this doesn't seem to be a state machine.

mhatrevi: It's a one cycle operation; this has been confirmed with the hardware team.

korran (approved):

rom/dev/src/main.rs
```diff
@@ -115,7 +115,8 @@ extern "C" fn exception_handler(exception: &exception::ExceptionRecord) {
 
     // TODO: Signal non-fatal error to SOC
 
-    report_error(CaliptraError::ROM_GLOBAL_EXCEPTION.into());
+    let mut env = unsafe { rom_env::RomEnv::new_from_registers() };
```
korran, line 118: We shouldn't call this here, as some drivers (in particular, CSRNG), will need to do some initialization, which could cause some serious problems in a trap.

These zeroization calls are probably going to need to be unsafe static functions, similar to `Mailbox::abort_pending_soc_to_uc_transactions()`.

PR #382, opened 2023-06-13
[feat] RT Alias Layer must update  FHT data structure fields

rusty1968:
- Add FHT field to store the RT Alias key pair.
- Update the newly created FHT field during FMC's RT Alias DICE flow.
- Update FMC's README.md with new FHT field.

### DIFF
--- a/common/src/hand_off.rs
+++ b/common/src/hand_off.rs
@@ -252,12 +252,6 @@ pub struct FirmwareHandoffTable {
     /// Index of RT Public Alias Key Y Coordinate in the Data Vault.
     pub rt_pub_key_y_dv_hdl: HandOffDataHandle,
 
-    /// Index of RT Certificate Signature R Component in the Data Vault.
-    pub rt_cert_sig_r_dv_hdl: HandOffDataHandle,
-
-    /// Index of RT Certificate Signature S Component in the Data Vault.
-    pub rt_cert_sig_s_dv_hdl: HandOffDataHandle,
-
     /// Index of RT SVN value in the Data Vault
     pub rt_svn_dv_hdl: HandOffDataHandle,
 
@@ -279,8 +273,10 @@ pub struct FirmwareHandoffTable {
     /// Fuse log Address
     pub fuse_log_addr: u32,
 
+    pub rt_dice_sign: [u8; core::mem::size_of::<caliptra_drivers::Ecc384Signature>()],
+
     /// Reserved for future use.
-    pub reserved: [u8; 32],
+    pub reserved: [u8; 60],
 }
 
 impl Default for FirmwareHandoffTable {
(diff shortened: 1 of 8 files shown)


common/src/hand_off.rs
```diff
@@ -279,8 +279,10 @@ pub struct FirmwareHandoffTable {
     /// Fuse log Address
     pub fuse_log_addr: u32,
 
+    pub rt_dice_sign: [u8; core::mem::size_of::<caliptra_drivers::Ecc384Signature>()],
```
jhand2, line 282: Could the type here just be the Ecc384Signature structure directly? (Although it might require adding the zerocopy trait to a couple types in the driver crate).

fmc/src/hand_off.rs
```diff
@@ -176,10 +176,15 @@ impl HandOff {
...
-        // Where should this be stored?
-        // No DV slots avaiable anymore.
+    /// Store runtime Dice Signature
+    pub fn set_rt_dice_signature(&mut self, sig: &Ecc384Signature) {
+        let data = unsafe {
+            core::slice::from_raw_parts(
+                sig as *const Ecc384Signature as *const u8,
+                core::mem::size_of::<Ecc384Signature>(),
+            )
+        };
```
jhand2, line 186: Recommend using zerocopy instead

common/src/hand_off.rs
rusty1968: Correct. This  requires Ecc384Signature, Array4x12 to implement zerocopy as well.

jhand2: Is that something we can do, or is there some concern with doing this?

rusty1968: I think it is doable. Let me take a stab at it.

rusty1968: AsBytes is not supported with generics.
[11:46](https://caliptraworkspace.slack.com/archives/D04RK3XSKV4/p1686768401422639)
error: unsupported on types with type parameters
  --> drivers/src/array.rs:21:17
   |
21 | #[derive(Debug, AsBytes, Clone, Copy, PartialEq, Eq)]
   |                 ^^^^^^^
   |
   = note: this error originates in the derive macro `AsBytes` (in Nightly builds, run with -Z macro-backtrace for more info)
[11:47](https://caliptraworkspace.slack.com/archives/D04RK3XSKV4/p1686768427178189)
#[repr(transparent)]
#[derive(Debug, AsBytes, Clone, Copy, PartialEq, Eq)]
pub struct Array4xN<const W: usize, const B: usize>(pub [u32; W]);
impl<const W: usize, const B: usize> Array4xN<W, B> {
    pub const fn new(val: [u32; W]) -> Self {
        Self(val)
    }
}
[11:47](https://caliptraworkspace.slack.com/archives/D04RK3XSKV4/p1686768469919159)
/// ECC-384 Coordinate
pub type Ecc384Scalar = Array4x12;

fmc/src/hand_off.rs
rusty1968: done

fdamato (dismissed):

jlmahowa-amd (dismissed):
FWIW tested and passes on FPGA.

jhand2 (dismissed):
Thanks! This looks great.

jhand2 (approved):

PR #383, opened 2023-06-14
Make sw-emulator DOE peripiheral consistent with RTL, and add bit-exact DOE test assertions

korran:
Reviewers: For context, I strongly recommend looking at each commit in this PR separately.

This PR does everything necessary to implement a DOE driver test that works on both verilator and sw-emulator and can deterministically validate that key-vault values written by DOE match the expected bits exactly, with both debug_locked=false and debug_locked=true.

As part of generating test vectors, I've included a standalone implementation (`DoeTestVectors::generate()`) of the cryptographic operations necessary to generate the expected DOE test's HMAC results from fuse values and silicon secrets, using only openssl.


drivers/tests/integration_tests.rs
```diff
@@ -29,9 +38,271 @@ fn run_driver_test(test_bin_name: &'static str) {
...
+        // data).
+        result.field_entropy = [input.keyvault_initial_word_value; 12];
+        result.field_entropy[0..8]
+            .as_bytes_mut()
+            .copy_from_slice(&aes256_decrypt_blocks(
+                swap_word_bytes(&input.doe_obf_key).as_bytes(),
+                &input.doe_iv,
+                swap_word_bytes(&input.field_entropy_seed).as_bytes(),
+            ));
+        swap_word_bytes_inplace(&mut result.field_entropy);
```
bluegate010, line 143: Might be a bit cleaner if the word swapping was taken care of in `aes256_decrypt_blocks()` and `hmac384()`.

korran: I was trying to make it clear to the reader where things differ from the standard byte order, which those aes256_decrypt_blocks() and hmac384() use.

korran: Also, doe_iv doesn't have its words swapped (to be consistent with the doe driver), and neither does the buffer data passed to hmac.

bluegate010 (approved):

mhatrevi (approved):

PR #388, opened 2023-06-14
Update README for new DICE key derivation paths.

bluegate010:


bluegate010:
For now this needs to wait until https://github.com/chipsalliance/caliptra-sw/pull/383 and https://github.com/chipsalliance/caliptra-sw/pull/372 are merged, before it can actually be implemented.

PR #389, opened 2023-06-14
Expand the smoke_test to validate key derivation and certificate signatures

korran:
As usual, I recommend reviewing each of the 14 commits independently.


sw-emulator/lib/periph/src/doe.rs
```diff
@@ -410,4 +413,26 @@ mod tests {
         assert_eq!(soc_reg.doe_key(), expected_doe_key);
         assert_eq!(soc_reg.field_entropy(), expected_fe);
     }
+
+    #[test]
+    fn test_cbc() {
```
jhand2, line 418: How is this related to cert testing?

test/tests/smoke_test.rs
```diff
@@ -50,4 +99,53 @@ fn smoke_test() {
...
+    let ldev_cert = openssl::x509::X509::from_der(&ldev_cert_der).unwrap();
+    let ldev_cert_txt = String::from_utf8(ldev_cert.to_text().unwrap()).unwrap();
+
+    // To update the ldev cert testdata:
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.txt", &ldev_cert_txt).unwrap();
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.der", &ldev_cert_der).unwrap();
+
+    assert_eq!(
+        ldev_cert_txt.as_str(),
+        include_str!("smoke_testdata/ldevid_cert.txt")
```
jhand2, line 119: What's the rationale for comparing against both the der and text?

sw-emulator/lib/periph/src/doe.rs
jhand2: Oh, I just realized this is a PR into your own branch. Never mind :)

test/tests/smoke_test.rs
```diff
@@ -50,4 +99,53 @@ fn smoke_test() {
...
+    let ldev_cert_der = hw
+        .mailbox_execute(TEST_ONLY_GET_LDEV_CERT, &[])
+        .unwrap()
+        .unwrap();
+    let ldev_cert = openssl::x509::X509::from_der(&ldev_cert_der).unwrap();
+    let ldev_cert_txt = String::from_utf8(ldev_cert.to_text().unwrap()).unwrap();
+
+    // To update the ldev cert testdata:
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.txt", &ldev_cert_txt).unwrap();
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.der", &ldev_cert_der).unwrap();
```
fdamato, line 115: is this test basically validating the DUT against  n-1 DUT ?

test/tests/smoke_test.rs
korran: I'm not sure whether openssl includes every bit of the cert in the text, so this way we have both a human-readable diff in any PR that changes this logic, plus the actual bits for exhaustive assertion.

test/tests/smoke_test.rs
korran: Yes, where n is a git commit. These certs and CSRs are generally not expected to change, so we want to make sure the test makes some noise if they do unexpectedly.

sw-emulator/lib/periph/src/doe.rs
korran: I've since rebased onto main (since that other PR has been submitted).

test/tests/smoke_test.rs
jhand2: SGTM

test/tests/smoke_test.rs
```diff
@@ -50,4 +99,53 @@ fn smoke_test() {
...
+
+    println!("ldev-cert: {}", ldev_cert_txt);
+
+    let fmc_alias_cert = hw
+        .mailbox_execute(TEST_ONLY_GET_FMC_ALIAS_CERT, &[])
+        .unwrap()
+        .unwrap();
+    let fmc_alias_cert = openssl::x509::X509::from_der(&fmc_alias_cert).unwrap();
+
+    // TODO: (DO NOT SUBMIT) Validate the contents / measurements in the fmc_alias_cert.
```
jhand2, line 141: The rest LGTM. I'll come back when this is implemented.

I also think it would be fine to open an issue and take care of this in a follow-up

jhand2 (approved):

test/src/lib.rs
```diff
@@ -1 +1,432 @@
...
+    pub layer: Option<u32>,
+    pub index: Option<u32>,
+    pub fwids: Vec<DiceFwid>,
+    pub flags: Option<u32>,
+    pub vendor_info: Option<Vec<u8>>,
+    pub ty: Option<Vec<u8>>,
+}
+
+impl DiceTcbInfo {
+    fn parse(d: &mut asn1::Parser) -> Result<Self, asn1::ParseError> {
```
jhand2, line 265: It would probably be easier to use the asn1 crate's parsing trait. See https://github.com/chipsalliance/caliptra-dpe/blob/7ad2c990bf9e06a1374bb2a02a8289761ee8e97f/dpe/src/x509.rs#LL965C1-L965C1

korran: I started down that path originally, but didn't like that I couldn't use ownership types (string, vec, etc) as the field types.

test/src/derive.rs
```diff
@@ -1,5 +1,11 @@
 // Licensed under the Apache-2.0 license
 
+// The IV fed to the DOE when the ROM deobfuscates the UDS seed (as passed to doe registers)
+pub const DOE_UDS_IV: [u32; 4] = [0xfb10365b, 0xa1179741, 0xfba193a1, 0x0f406d7e];
+
+// The IV fed to the DOE when the ROM deobfuscates the field entropy seed (as passed to doe registers)
```
bluegate010, line 25: Uber nit: no need to refer to this as "field entropy seed". UDS seed was chosen for DICE compliance reasons IIRC.

test/src/x509.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub vendor: Option<String>,
+    pub model: Option<String>,
+    pub version: Option<String>,
+    pub svn: Option<u32>,
+    pub layer: Option<u32>,
+    pub index: Option<u32>,
+    pub fwids: Vec<DiceFwid>,
+    pub flags: Option<u32>,
+    pub vendor_info: Option<Vec<u8>>,
+    pub ty: Option<Vec<u8>>,
```
bluegate010, line 30: I presume this is shortened because `type` is a keyword - could this be `type_`?

bluegate010 (commented):
Fantastic stuff.

test/src/derive.rs
korran: This code deals with both the obfuscated field entropy and the deobfuscated field entropy. I've been calling the deobfuscated value the "field entropy" and the obfuscated the "field entropy seed", to match the UDS stuff. What are the correct terms to use to distinctly refer to these two values?

test/src/x509.rs
korran: Yes. type_ might be a better option, but I've been using ty for this in quite a few other places in the codebase (mostly in ureg and systemrdl libraries).

bluegate010: Guess we can change it in a follow-up PR.

test/src/derive.rs
bluegate010: Ah, I suppose "field entropy seed" is alright then.

bluegate010 (dismissed):

test/tests/smoke_test.rs
```diff
@@ -12,18 +20,133 @@ fn assert_output_contains(haystack: &str, needle: &str) {
...
+    let csr = openssl::x509::X509Req::from_der(&csr_der).unwrap();
+    let csr_txt = String::from_utf8(csr.to_text().unwrap()).unwrap();
+
+    // To update the CSR testdata:
+    // std::fs::write("tests/smoke_testdata/idevid_csr.txt", &csr_txt).unwrap();
+    // std::fs::write("tests/smoke_testdata/idevid_csr.der", &csr_der).unwrap();
+
+    println!("csr: {}", csr_txt);
+
+    assert_eq!(
```
bluegate010, line 51: Would it make sense to put some helpful panic message here, like "If your change is expected to modify generated certs or public keys, please re-generate the testdata (see comment in this file)"?

test/tests/smoke_test.rs
```diff
@@ -50,4 +173,115 @@ fn smoke_test() {
...
+        .unwrap()
+        .unwrap();
+    let ldev_cert = openssl::x509::X509::from_der(&ldev_cert_der).unwrap();
+    let ldev_cert_txt = String::from_utf8(ldev_cert.to_text().unwrap()).unwrap();
+
+    // To update the ldev cert testdata:
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.txt", &ldev_cert_txt).unwrap();
+    // std::fs::write("tests/smoke_testdata/ldevid_cert.der", &ldev_cert_der).unwrap();
+
+    assert_eq!(
```
bluegate010, line 191: Ditto

bluegate010 (approved):

test/tests/smoke_test.rs
```diff
@@ -50,4 +157,78 @@ fn smoke_test() {
...
+        "fmc-alias cert: {}",
+        String::from_utf8_lossy(&fmc_alias_cert.to_text().unwrap())
+    );
+
+    let expected_fmc_alias_key = FmcAliasKey::derive(
+        &Pcr0::derive(&Pcr0Input {
+            security_state,
+            fuse_anti_rollback_disable: false,
+            vendor_pub_key_hash: Default::default(),
+            // TODO: Is this right? Should this really be mixed in even if the fuses aren't set?
```
jhand2: I don't see any harm in doing so. Seems less error prone to unconditionally measure it, even if the value is all 0 or all 0xff.

test/src/lib.rs
jhand2: Could just implement Asn1Writable and Asn1Readable for the types you want to use, right? And can use the asn1 functions to implement them, so I'd think they would be small implementations. But I could be wrong.

If it's not too much work, I think it will make things a lot easier/more readable to be able to use the annotations for optional (implicit/explicit) fields.

jhand2 (approved):

test/src/lib.rs
korran: Rust doesn't let you implement an external trait (like AsnReadable) for an external type (like String, Vec). Either the trait or type need to be in my crate.

test/tests/smoke_test.rs
korran: How will we know if the owner fuses are set correctly if we're not measuring the fuse value?

When the owner fuses aren't set, we don't check the signature.

test/tests/smoke_test.rs
korran: Yes. Will do this in a followup PR.

test/tests/smoke_test.rs
korran: Filed #415

PR #399, opened 2023-06-20
[feat] Enable Watchdog timer in ROM

mhatrevi:


drivers/src/soc_ifc.rs
```diff
@@ -131,6 +132,85 @@ impl SocIfc {
...
+    /// * `enable` - Enable or disable WDT1
+    ///
+    pub fn configure_wdt1(&mut self, enable: bool) {
+        let soc_ifc_regs = self.soc_ifc.regs_mut();
+        soc_ifc_regs
+            .cptra_wdt_timer1_en()
+            .write(|w| w.timer1_en(enable));
+    }
+
+    fn get_cycle_count(&self, seconds: u32) -> CaliptraResult<u64> {
```
korran, line 148: Why do we care about seconds vs cycles? AFAICT caliptra takes N cycles to accomplish a particular task regardless of clock rate.

drivers/src/soc_ifc.rs
```diff
@@ -131,6 +132,85 @@ impl SocIfc {
...
+        const PICO_UNIT: u64 = 0xE8D4A51000;
+        let clock_period_picosecs = self.soc_ifc.regs().cptra_timer_config().read() as u64;
+        if clock_period_picosecs == 0 {
+            Err(CaliptraError::DRIVER_SOC_IFC_INVALID_TIMER_CONFIG)
+        } else {
+            Ok((seconds as u64) * (PICO_UNIT / clock_period_picosecs))
```
korran, line 154: This division is bloating the ROM size due to compiler_builtins::int::specialized_div_rem::u64_div_rem

drivers/src/soc_ifc.rs
```diff
@@ -131,6 +132,85 @@ impl SocIfc {
         let flags: MfgFlags = soc_ifc_regs.cptra_dbg_manuf_service_reg().read().into();
         flags.contains(MfgFlags::GENERATE_IDEVID_CSR)
     }
+
+    /// Enbale or disable WDT1
+    ///
+    /// # Arguments
+    /// * `enable` - Enable or disable WDT1
+    ///
+    pub fn configure_wdt1(&mut self, enable: bool) {
```
korran, line 141: Don't you need to setup the interrupt handling somewhere?

korran: Looking at the docs, it looks like an NMI will eventually be generated if we don't respond to the original interrupt; maybe that's good enough.

korran: I'd still like to see this thing actually fire in verilator to make sure we're doing this right.

drivers/src/soc_ifc.rs
mhatrevi: Removed cycle count computation.

drivers/src/soc_ifc.rs
mhatrevi: Not handling WDT1 expiration interrupt causes the WDT2 timer to be "enabled'. Subsequently, not handling WDT2 expiration interrupt triggers an NMI.

drivers/src/soc_ifc.rs
mhatrevi: Switched to a 32-bit division.

rom/dev/src/main.rs
```diff
@@ -89,6 +94,10 @@ pub extern "C" fn rom_entry() -> ! {
         Err(err) => report_error(err.into()),
     }
 
+    // Stop the watchdog timer.
+    // [TODO] Reset the watchdog timer and let FMC take ownership of it.
```
korran, line 98: FWIW, I think you should keep the watchdog timer going; the FMC can extend the delay if it wants to, but generally I would expect us to get all the way to runtime, where it turns off the watchdog right before it goes to sleep waiting for mailbox commands.

korran (dismissed):

rom/dev/src/main.rs
mhatrevi: That was the idea; I'll file an issue on runtime to add the WDT functionality and will remove the stop operation in the ROM.

korran (approved):

PR #400, opened 2023-06-20
FPGA Support

jlmahowa-amd:
- FPGA wrapper implementation
- HW Model for fpga_realtime
- FPGA specific workarounds


jlmahowa-amd:
fpga_realtime feature check required for:
- Using start.S version that disables ROM caching, pipelining, and branch prediction. This is a workaround for the ROM/FMC execution issues.
- Performing handshake with test SW for serial communication. Verilator/SW emulator method was not viable on FPGA. This should be resolved when the FW supports the real UART.

hw-model/src/model_fpga_realtime.rs
```diff
@@ -0,0 +1,210 @@
...
+
+// TODO: Make this configurable
+const SOC_PAUSER: u32 = 0xffff_ffff;
+
+fn fmt_uio_error(err: UioError) -> String {
+    format!("{err:?}")
+}
+
+pub struct ModelFpgaRealtime {
+    _dev: UioDevice,
```
rusty1968, line 25: Is this field being left unused? What is the point of having it in the struct?

hw-model/src/model_fpga_realtime.rs
```diff
@@ -0,0 +1,210 @@
...
+    gpio: *mut u32,
+    mbox: *mut u32,
+    soc_ifc: *mut u32,
+    output: Output,
+    start_time: time::Instant,
+}
+
+impl ModelFpgaRealtime {
+    fn is_ready_for_fuses(&self) -> bool {
+        unsafe { (self.gpio.offset(2).read_volatile() & 0x4000_0000) != 0 }
```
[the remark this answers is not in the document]
korran: I think it's pretty clear here that 0x4000_0000 is the mask for the is_ready_for_fuses gpio bit, which is already in the name of the function. I don't think additional indirection will make this code any clearer.

test/tests/smoke_test.rs
```diff
@@ -14,13 +14,25 @@ fn assert_output_contains(haystack: &str, needle: &str) {
...
+    #[cfg(not(feature = "fpga_realtime"))]
+    let (rom, fmc, app) = { (ROM_WITH_UART, FMC_WITH_UART, APP_WITH_UART) };
+
+    #[cfg(feature = "fpga_realtime")]
+    let (rom, fmc, app) = {
+        let mut rom_copy = ROM_WITH_UART;
+        let mut fmc_copy = FMC_WITH_UART;
+        let mut app_copy = APP_WITH_UART;
+
+        rom_copy.features = &["emu fpga_realtime"];
```
korran, line 26: Shouldn't this be &["emu", "fpga_realtime"]?

test/tests/smoke_test.rs
```diff
@@ -14,13 +14,25 @@ fn assert_output_contains(haystack: &str, needle: &str) {
...
-    let image = caliptra_builder::build_and_sign_image(
-        &FMC_WITH_UART,
-        &APP_WITH_UART,
-        ImageOptions::default(),
-    )
-    .unwrap();
+    #[cfg(not(feature = "fpga_realtime"))]
+    let (rom, fmc, app) = { (ROM_WITH_UART, FMC_WITH_UART, APP_WITH_UART) };
+
+    #[cfg(feature = "fpga_realtime")]
```
korran, line 20: This isn't the only test folks want to run on the FPGA. Maybe we should do this globally in caliptra_builder?

https://github.com/chipsalliance/caliptra-sw/blob/main/builder/src/lib.rs#L118

```
if cfg!(feature = "fpga_realtime") {
  features_csv.push_str("fpga_realtime");
}
```

rom/dev/src/start_fpga.S
```diff
@@ -0,0 +1,354 @@
+/*++
```
korran, line 1: This is too much copy-and paste for my liking.

I think we should use the preprocessor, like in https://github.com/chipsalliance/caliptra-sw/pull/428

test/tests/smoke_test.rs
jlmahowa-amd: It should have been, I don't know why this wasn't a problem until pulling in your preprocessor patch.
This goes away with moving to the builder.

test/tests/smoke_test.rs
jlmahowa-amd: Will do.

hw-model/src/model_fpga_realtime.rs
jlmahowa-amd: Removed. I thought this would be RAII but not retaining dev doesn't impact the result.

hw-model/src/model_fpga_realtime.rs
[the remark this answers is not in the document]
jlmahowa-amd: I'll use bitfields to clean this up.

rusty1968 (dismissed):

rom/dev/src/start.S
```diff
@@ -77,6 +77,17 @@ _start:
     li x1, 0xAAAAA0A9
     csrw 0x7C0, x1
 
+#ifdef CARGO_FEATURE_FPGA_REALTIME
```
korran, line 80: Is this a GitHub issue for this? Maybe reference it here?

jlmahowa-amd: Added new GitHub issue to track ROM/ICCM execution issues. Bharat the previously filed one.

hw-latest/fpga/fpga_configuration.tcl
```diff
@@ -0,0 +1,247 @@
...
+#     Make the VEER core be optimized for FPGA (no clock gating)
+#     Define VEER TEC_RV_ICG to allow beh_lib to synthesise without error
+set VERILOG_OPTIONS {TECH_SPECIFIC_ICG USER_ICG=fpga_fake_icg RV_FPGA_OPTIMIZE TEC_RV_ICG=clockhdr}
+if {FALSE} {
+  # Add option to use Caliptra's internal TRNG instead of ETRNG
+  lappend VERILOG_OPTIONS CALIPTRA_INTERNAL_TRNG
+}
+
+# Start the Vivado GUI for interactive debug
+start_gui
```
korran, line 25: Is this part necessary? I'd really like to be able to run this as part of CI, which doesn't have an x server.

hw-latest/fpga/FPGA_Instructions.md
```diff
@@ -0,0 +1,91 @@
...
+   - `ZCU104 Development Board`
+   - https://www.xilinx.com/products/boards-and-kits/zcu104.html
+ - [Not fully enabled] JTAG debugger
+   - `Olimex ARM-USB-TINY-H`
+
+### ZCU104 Processing System One Time Setup: ###
+1. Install ZCU104 SD card image
+   - https://ubuntu.com/download/amd-xilinx
+1. Configure SW6 to boot from SD1.
+   - Mode SW6[4:1]: OFF, OFF, OFF, ON
```
korran, line 31: Can we include this picture?

PXL_20230419_183359704

hw-latest/fpga/FPGA_Instructions.md
```diff
@@ -0,0 +1,91 @@
...
+ - Data bits: 8
+ - Stop bits: 1
+ - Parity: None
+ - Flow control: None
+
+### FPGA Build Steps: ###
+1. Launch Vivado with tcl script defining FPGA build
+    - `vivado -mode batch -source fpga_configuration.tcl`
+1. Run Synthesis
+    - Use GUI or: `launch_runs synth_1`
```
korran, line 47: I think we need to make it clearer how to script this, and run it with parallelization

```
launch_runs synth_1 -jobs 10
wait_on_runs synth_1
launch_runs impl_1 -jobs 10
wait_on_runs impl_1
```

drivers/src/uart.rs
```diff
@@ -46,13 +46,30 @@ impl Uart {
...
     pub fn write_byte(&mut self, byte: u8) {
         // TODO: cleanup after final UART RTL definition is in place
         const STDOUT: *mut u32 = 0x3003_00C8 as *mut u32;
         unsafe {
             ptr::write_volatile(STDOUT, byte as u32);
         }
     }
+
+    #[cfg(feature = "fpga_realtime")]
+    pub fn write_byte(&mut self, byte: u8) {
```
korran, line 59: This is fine for now, but I do have an idea for future unification:

1. Change write_byte() to toggle a "char_clk" pin in generic_output_wires, which is edge-triggered to signify that a new byte  has been written.
2. Add a character FIFO (4k?) to the FPGA that buffers up the written bytes for later access by the A53 cpu.

hw-latest/fpga/src/caliptra_veer_sram_export.sv
```diff
@@ -0,0 +1,654 @@
...
+	PIC_REGION             : 8'h06         ,
+	PIC_SIZE               : 13'h0020       ,
+	PIC_TOTAL_INT          : 12'h01F        ,
+	PIC_TOTAL_INT_PLUS1    : 13'h0020       ,
+	RET_STACK_SIZE         : 8'h08         ,
+	SB_BUS_ID              : 5'h01         ,
+	SB_BUS_PRTY            : 6'h02         ,
+	SB_BUS_TAG             : 8'h01         ,
+	TIMER_LEGAL_EN         : 5'h01
+}
```
korran, line 191: This seems a bit brittle. Can we use `include "el2_param.vh" here instead?

korran: And if we do that, this is identical to `hw-latest/caliptra-rtl-fpga/src/integration/tb/caliptra_veer_sram_export.sv`, no?

hw-latest/fpga/io_module/io_module.c
```diff
@@ -0,0 +1,64 @@
+// Licensed under the Apache-2.0 license
```
korran, line 1: Are you sure these drivers are necessary? It seems like I can access this memory via /dev/mem from userspace.

hw-latest/fpga/FPGA_Instructions.md
```diff
@@ -0,0 +1,91 @@
...
+1. Set Up Debug signals on Synthesized Design
+1. Run Implementation and generate bitstream
+    - Use GUI or: `launch_runs impl_1`
+1. FPGA image location:
```
korran, line 51: Are we missing a `write_bitstream` step here? I don't see the bitstream file after impl_1 finishes.

hw-latest/fpga/src/caliptra_veer_sram_export.sv
jlmahowa-amd: Trying out using fpga/src/integration/tb/caliptra_veer_sram_export.sv.
On older RTL el2_param.vh caused issues and embedding this definition was my workaround. Should no longer be required.

hw-latest/fpga/fpga_configuration.tcl
jlmahowa-amd: No, my tentative plan is to have a common tcl file doing all the setup, then separate GUI and automation tcl files.

hw-latest/fpga/io_module/io_module.c
jlmahowa-amd: You _can_ access the memory through /dev/mem, but is it a good idea? I thought that UIO would be preferred.
I could see changing the rom backdoor to be part of the UIO driver, which would simplify setup.

korran: If we don't need interrupts, I think it's preferable to avoid dependencies on kernel modules, since that complicates deployment.

This is fine for now, however.

hw-latest/fpga/fpga_configuration.tcl
jlmahowa-amd: Making this configurable.

hw-latest/fpga/FPGA_Instructions.md
jlmahowa-amd: Including picture in doc.

hw-latest/fpga/FPGA_Instructions.md
jlmahowa-amd: Adding the build flow steps to the script, including this, and making the GUI options more clear.

hw-latest/fpga/src/caliptra_veer_sram_export.sv
jlmahowa-amd: Passes synthesis so I think it is good. Still waiting on image to finish though.

hw-latest/fpga/io_module/io_module.c
jlmahowa-amd: Added to my TODO list. Marking as resolved to move on with merge.

drivers/src/uart.rs
jlmahowa-amd: Added to my TODO list. Marking as resolved to move on with merge.

korran (approved):

PR #402, opened 2023-06-20
Packet abstraction and checksum utilities

rusty1968:
This commit implements checksum utilities for mailbox commands as defined by the runtime specification.

### DIFF
--- a/runtime/src/verify.rs
+++ b/runtime/src/verify.rs
@@ -6,13 +6,16 @@ use caliptra_drivers::{
 };
 use zerocopy::FromBytes;
 
+/// Start of payload (skips checksum field).
+const PAYLOAD_OFFSET: usize = 4;
+
 /// Handle the `ECDSA384_SIGNATURE_VERIFY` mailbox command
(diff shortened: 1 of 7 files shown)


runtime/src/verify.rs
```diff
@@ -9,6 +9,13 @@ use zerocopy::FromBytes;
 /// Handle the `ECDSA384_SIGNATURE_VERIFY` mailbox command
 pub(crate) fn handle_ecdsa_verify(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<()> {
     if let Some(cmd) = EcdsaVerifyCmd::read_from(cmd_args) {
+        if !caliptra_common::checksum::verify_checksum(
+            cmd.chksum,
+            CommandId::ECDSA384_VERIFY.into(),
+            &cmd_args[4..],
```
jhand2, line 15: Would be good to add a comment explaining that this offset is due to skipping the checksum

jhand2 (approved):
Looks great, thanks!

runtime/src/verify.rs
rusty1968: Done.

jhand2 (approved):

PR #407, opened 2023-06-21
[feat] Zeroize local variables on stack

### LINKED ISSUE #324: All Temp Data Must be Zeroized on Completion of Operation
Requirement: All Temp Data Must be Zeroized on Completion of Operation

Details:  Looking into rust, if we can prove that all data is zeroized upon garbage collection, then this will meet the requirement.  We will have to prove that this isn't disabled or able to work around.
Otherwise, before we leave scope for any intermediate data related to crypto operations, must be zeroized.

mhatrevi:
These changes are in response to the following FIPs requirement (https://github.com/chipsalliance/caliptra-sw/issues/324)

### DIFF
--- a/drivers/src/ecc384.rs
+++ b/drivers/src/ecc384.rs
@@ -110,6 +110,11 @@ impl Ecc384PubKey {
     pub fn to_der(&self) -> [u8; 97] {
         array_concat3([0x04], (&self.x).into(), (&self.y).into())
     }
+
+    pub fn zeroize(&mut self) {
+        self.x.0.fill(0);
+        self.y.0.fill(0);
+    }
 }
 
 /// ECC-384 Signature
@@ -121,6 +126,13 @@ pub struct Ecc384Signature {
 
     /// Proof
     pub s: Ecc384Scalar,
+}
+
+impl Ecc384Signature {
+    pub fn zeroize(&mut self) {
+        self.r.0.fill(0);
+        self.s.0.fill(0);
+    }
 }
 
 /// Elliptic Curve P-384 API
@@ -146,7 +158,7 @@ impl Ecc384 {
     /// * `Ecc384PubKey` - Generated ECC-384 Public Key
     pub fn key_pair(
         &mut self,
-        seed: Ecc384Seed,
+        seed: &Ecc384Seed,
         nonce: &Array4x12,
         trng: &mut Trng,
         mut priv_key: Ecc384PrivKeyOut,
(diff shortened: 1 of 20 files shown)


JohnTraverAmd:
https://github.com/chipsalliance/caliptra-sw/pull/373

drivers/src/ecc384.rs
```diff
@@ -109,6 +109,11 @@ impl Ecc384PubKey {
     pub fn to_der(&self) -> [u8; 97] {
         array_concat3([0x04], (&self.x).into(), (&self.y).into())
     }
+
+    pub fn zeroize(&mut self) {
```
korran, line 113: Have you looked at the machine code to see if this actually happens? I would expect the optimizer to remove this.

drivers/src/lms.rs
```diff
@@ -366,6 +366,7 @@ impl Lms {
         }
         hasher.finalize()?;
         let result = HashValue::<N>::from(digest);
+        digest.0.fill(0);
```
korran, line 369: Why does it matter here? none of these values are secret....

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -70,14 +72,15 @@ impl FmcAliasLayer {
...
-        let output = DiceOutput {
+        let mut output = DiceOutput {
             subj_key_pair: key_pair,
             subj_sn,
             subj_key_id,
         };
 
         // Generate Local Device ID Certificate
         Self::generate_cert_sig(env, input, &output, fw_proc_info)?;
+        output.zeroize();
```
korran, line 83: Are these really "temporaries"? Considering that they're being stored in the certificate, which isn't being zeroized, this seems silly to me.

mhatrevi: Discussed in the Firmware standup; this is as per FIPS requirement (https://github.com/chipsalliance/caliptra-sw/issues/324)

drivers/src/lms.rs
```diff
@@ -453,11 +454,14 @@ impl Lms {
             temp = HashValue::<N>::from(digest);
             node_num /= 2;
             i += 1;
+            digest.0.fill(0);
         }
-        let candidate_key = temp;
+        let mut candidate_key = temp;
```
korran, line 459: I think you probably want to clear temp here too, no?

I suggest:

```
let candidate_key = &temp;
if *candidate_key != HashValue::from(lms_public_key.digest) {
    return Ok(false);
}
digest.0.fill(0);
temp.fill(0);
```

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -95,9 +98,11 @@ impl FmcAliasLayer {
     fn derive_cdi(env: &mut RomEnv, measurements: Array4x12, cdi: KeyId) -> CaliptraResult<()> {
```
korran, line 98: Should measurements be a reference and zeroized by the caller? (measurements cleared below is a separate shadowed variable)

drivers/src/lms.rs
mhatrevi: Updated.

mhatrevi: Updated.

korran (approved):

PR #409, opened 2023-06-22
Add nightly release GitHub workflow

amd-isaac:
Add two GitHub workflows. The first will update the git submodule, run CI tests if there are changes, and if successful commit the updated submodule changes to the main branch. The second will perform the first, then run the full Verilator suite and generate a tagged nightly release for the repo.


.github/workflows/build-test-verilator.yml
```diff
@@ -28,9 +33,25 @@ jobs:
       EXTRA_CARGO_CONFIG: "target.'cfg(all())'.rustflags = [\"-Dwarnings\"]"
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Setup ref to use
+        id: ref
+        run: |
+          REF_REQ=${{ inputs.ref }}
+          echo "ref=${REF_REQ:-"Blah"}" >> $GITHUB_OUTPUT
```
amd-isaac, line 40: Make something other than "Blah"

.github/workflows/rtl-repo-sync.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+          ref: 'main'
+
+      - name: Update the Caliptra RTL submodule and rebuild registers
+        run: |
+          git submodule update --remote hw-latest/caliptra-rtl
+          cargo run --release --manifest-path registers/bin/generator/Cargo.toml -- hw-latest/caliptra-rtl registers/src/
```
amd-isaac, line 27: Use existing shell script

.github/workflows/nightly-verilator.yml
```diff
@@ -1,5 +1,10 @@
 name: Verilator Nightly
 on:
+  workflow_call:
```
amd-isaac, line 3: Check if workflow_call ref is auto-picked up

.github/workflows/rtl-repo-sync.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          submodules: 'true'
+          ref: 'main'
+          fetch-depth: 0
+
+      - name: Merge RTL changes to main branch
+        run: |
+          git config --global user.name "GitHub CI"
+          git config --global user.email "username@users.noreply.github.com"
+          git merge origin/${{ needs.update-rtl.outputs.branch_name }}
```
amd-isaac, line 85: May fail here if there's a commit that comes in during the CI run before this

amd-isaac: Consider consolidating CI checks into a single workflow, so when somebody adds a new one, they add it to the single CI workflow, and then we can call it here. Make sure all CI checks show up independently in the PR.

.github/workflows/nightly-release.yml
```diff
@@ -0,0 +1,100 @@
...
+          submodules: 'true'
+          ref: 'main'
+          fetch-depth: 0
+
+      - name: Find latest release
+        id: find
+        run: |
+          git config --global user.name "GitHub CI"
+          git config --global user.email "username@users.noreply.github.com"
+          TAG_BASE=release_v
```
amd-isaac, line 35: Create a release tag with a date stamp rather than a version number

.github/workflows/nightly-release.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+          ref: ${{ needs.find-latest-release.outputs.release_ref }}
+
+      - name: Generate release zip
+        run: |
+          mkdir release_${{ steps.date.outputs.date }}
+          cargo run --manifest-path=builder/Cargo.toml --bin image -- --rom caliptra_rom.bin --fw image_bundle.bin
```
amd-isaac, line 83: We should have a single script that we call here that generates the release (to let us strip out large files that are not needed, etc)

.github/workflows/rtl-repo-sync.yml
amd-isaac: Also check if we can just create/manage/close a PR

mhatrevi (changes requested):

release/build_release.sh
```diff
@@ -0,0 +1,10 @@
+# Licensed under the Apache-2.0 license
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+mkdir release_${{ steps.date.outputs.date }}
+cargo run --manifest-path=../builder/Cargo.toml --bin image -- --rom caliptra_rom.bin --fw image_bundle.bin
```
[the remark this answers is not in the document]
amd-isaac: Done

.github/workflows/rtl-repo-sync.yml
amd-isaac: > May fail here if there's a commit that comes in during the CI run before this

Fixed this

.github/workflows/nightly-verilator.yml
amd-isaac: actions/checkout@v3 may not need us to specify an explicit git ref

.github/workflows/nightly-release.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  verilator-full-suite:
+    name: Temp
+    needs: find-latest-release
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Get current date
+        run: echo "date=$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
+
+  #verilator-full-suite:
```
amd-isaac, line 66: Note: Uncomment this and remove the temporary verilator-full-suite above before merging!

.github/workflows/rtl-repo-sync.yml
```diff
@@ -0,0 +1,112 @@
...
+  build-test-verilator:
+    name: Temp
+    needs: update-rtl
+    if: needs.update-rtl.outputs.ci_checks_needed
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Get current date
+        run: echo "date=$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
+
+  #build-test:
```
amd-isaac, line 75: Before merge: uncomment this and remove the above placeholders

.github/workflows/nightly-verilator.yml
fdamato: I couldn`t find from the documentation a clear answer to this...in the meantime, I`d propose to keep this logic as it is more explicit...

korran: From https://github.com/actions/checkout:

> ref: The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.

korran: Also, from https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow:

> {ref} can be a SHA, a release tag, or a branch name. Using the commit SHA is the safest for stability and security. F

rusty1968 (dismissed):

korran (approved):

mhatrevi (approved):

PR #414, opened 2023-06-22
[feat] Enable SCA countermeasures.

mhatrevi:
This change uses the CSRNG driver to generate a random IV and an LFSR seed; the IV is provided as an input to the key generation and signing functions of the ECC384 driver, the LFSR seed is provided as an input to the MAC generation functions of HMAC384 driver.

### DIFF
--- a/rom/dev/src/rom_env.rs
+++ b/rom/dev/src/rom_env.rs
@@ -18,11 +18,13 @@ Abstract:
 use crate::fht::FhtDataStore;
 use caliptra_drivers::{
     DataVault, DeobfuscationEngine, Ecc384, Hmac384, KeyVault, Lms, Mailbox, PcrBank, Sha1, Sha256,
-    Sha384, Sha384Acc, SocIfc,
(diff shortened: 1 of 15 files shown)


rom/dev/src/rom_env.rs
```diff
@@ -72,6 +74,9 @@ pub struct RomEnv {
 
     /// FHT Data Store
     pub fht_data_store: FhtDataStore,
+
+    /// Cryptographically Secure Random Number Generator
+    pub csrng: Csrng,
```
korran, line 79: This won't work with the default verilator; you need to use the Trng type here.

korran: Then it's just a call to trng.generate() to get a 384-bit value, and it will come from either the etrng or the itrng depending on how the hardware was built.

mhatrevi: Updated

korran (approved):

PR #421, opened 2023-06-23
Pass IDevID pub from ROM to RT

jhand2:
The IDevID public key is needed by the SoC to populate an IDevID cert template. Pass this from ROM so that runtime can expose it.

### DIFF
--- a/runtime/src/main.rs
+++ b/runtime/src/main.rs
@@ -35,9 +35,8 @@ const BANNER: &str = r#"
 #[no_mangle]
 pub extern "C" fn entry_point() -> ! {
     cprintln!("{}", BANNER);
-    let mut drivers = unsafe { Drivers::new_from_registers() };
-
-    if let Some(_fht) = caliptra_common::FirmwareHandoffTable::try_load() {
+    if let Some(mut fht) = caliptra_common::FirmwareHandoffTable::try_load() {
+        let mut drivers = unsafe { Drivers::new_from_registers(&mut fht) };
         cprintln!("Caliptra RT listening for mailbox commands...");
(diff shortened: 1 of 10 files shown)


runtime/src/main.rs
```diff
@@ -35,9 +35,8 @@ const BANNER: &str = r#"
 #[no_mangle]
 pub extern "C" fn entry_point() -> ! {
     cprintln!("{}", BANNER);
-    let mut drivers = unsafe { Drivers::new_from_registers() };
-
-    if let Some(_fht) = caliptra_common::FirmwareHandoffTable::try_load() {
+    if let Some(mut fht) = caliptra_common::FirmwareHandoffTable::try_load() {
+        let mut drivers = unsafe { Drivers::new_from_registers(&mut fht) };
```
rusty1968, line 39: We should report error to  Soc via non fatal error register if we cant load FHT.

jhand2: I wonder if we should use the fatal register here since firmware won't continue to execute. But also we use the non-fatal register for all other errors, so maybe that's better for consistency

rusty1968: The error policy is still unclear to me. Have we published it yet?

rusty1968 (approved):

runtime/src/main.rs
jhand2: I don't think it's published. It looks like you added a call to `report_handoff_error_and_halt` to resolve this in a different PR, so closing this comment and rebasing on your changes.

rusty1968: Sounds good. Feel free to tweak it later if you prefer to make it fatal.

PR #423, opened 2023-06-23
Use diversification labels when performing CDI derivation or keygen

bluegate010:


mschilder123 (approved):

korran:
Do we want to merge this after #389 ?

bluegate010:
> Do we want to merge this after #389 ?

Sure

bluegate010:
> > Do we want to merge this after #389 ?
> 
> Sure

Ready for review.

rusty1968 (approved):

fmc/src/flow/crypto.rs
```diff
@@ -59,66 +60,74 @@ impl Crypto {
         env.sha384.digest(data)
     }
 
-    /// Calculate HMAC-384
+    /// Calculate HMAC-348 KDF
```
rusty1968, line 63: NIT : 384.

bluegate010: Thanks, fixed.

fmc/src/flow/rt_alias.rs
```diff
@@ -44,26 +44,25 @@ impl RtAliasLayer {
         hand_off: &mut HandOff,
         input: &DiceInput,
     ) -> CaliptraResult<DiceOutput> {
+        if Self::kv_slot_collides(input.cdi) {
```
rusty1968, line 47: Can collision detection be a compile time assertion?

bluegate010: Yes, but I think that only works if FMC ignores the key handle number it observes in the handoff table and instead uses a hard-coded value that was compile-time-asserted to be unique from the values used by ROM.

rusty1968 (approved):

fmc/src/flow/rt_alias.rs
bluegate010: We could make it harder to accidentally introduce collisions though, if we move the key IDs defined [in ROM](https://github.com/chipsalliance/caliptra-sw/blob/2be0e16e6832400053e67069f9f89989a530958c/rom/dev/src/flow/cold_reset/mod.rs#L34-L39) and [in FMC](https://github.com/chipsalliance/caliptra-sw/blob/2be0e16e6832400053e67069f9f89989a530958c/fmc/src/flow/mod.rs#L29-L30) into one file in the common directory. Could take care of this in a follow-up.

rusty1968: I like the idea.

rusty1968 (approved):

mhatrevi (approved):

Issue #425, opened 2023-06-24
Don't call RomEnv::new_from_registers() in trap handlers.


korran:
https://github.com/chipsalliance/caliptra-sw/blob/99b5a3a0dec6b31b8cdb8b9b652c49a3b19a877c/rom/dev/src/main.rs#L108-L118

We shouldn't call RomEnv::new_from_registers() here, as some drivers (in particular, CSRNG), will need to do some initialization, which could cause some serious problems in a trap.

These zeroization calls are probably going to need to be unsafe static functions, similar to `Mailbox::abort_pending_soc_to_uc_transactions()`.

_Originally posted by @korran in https://github.com/chipsalliance/caliptra-sw/pull/373#discussion_r1240570799_

PR #427, opened 2023-06-26
[fix] Convert zeroize methods on crypto drivers to unsafe static functions.

mhatrevi:
This is done to avoid creating a new ROM instance for calling the report_error function.
Some crypto drivers (eg CSRNG) cannot be instantiated in a Trap.

Fix for issue: https://github.com/chipsalliance/caliptra-sw/issues/425


drivers/src/hmac384.rs
```diff
@@ -212,17 +212,30 @@ impl Hmac384 {
...
         self.hmac.regs_mut().ctrl().write(|w| w.zeroize(true));
     }
+
+    /// Zeroize the hardware registers.
+    ///
+    /// This is useful to call from a fatal-error-handling routine.
+    ///
+    /// # Safety
+    ///
+    /// This function is safe to call from a trap handler.
```
korran, line 231: Can you add "Caller must be certain that the results of any pending cryptographic operations will not be used after this function is called" here and elsewhere, to make it clear that this should only be used in "fatal-error" type scenarios.

korran: actually, I'll do this in a PR

korran (approved):

PR #428, opened 2023-06-26
ROM: Use preprocessor for start.S.

korran:
All of the CARGO_* environment variables are available as preprocessor variables:

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

`CARGO_FEATURE_<name>` would be particularly useful for conditional compilation.

### DIFF
--- a/rom/dev/build.rs
+++ b/rom/dev/build.rs
@@ -12,21 +12,49 @@ Abstract:
 
 --*/
 
+use std::process::Command;
+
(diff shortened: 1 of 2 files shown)


vsonims (approved):

mhatrevi (approved):

jlmahowa-amd (approved):
Works perfectly. I was able to get rid of the copied start.S using this.

PR #429, opened 2023-06-27
[RT Spec]: Add mailbox commands for PCR quoting

jhand2:
* Remove strong link between DPE and PCRs. These will instead operate as separate subsystems in Caliptra. Note: Caliptra's own PCRs will still factor into DPE initialization.
* Add mailbox commands for PCR quote, extend, and reset counter.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -237,7 +237,9 @@ Table: `ECDSA384_SIGNATURE_VERIFY` output arguments
 Make a measurement into the DPE default context. This command is intendend for
 callers who update infrequently and cannot tolerate a changing DPE API surface.
 
-Internally, this will call the DPE DeriveChild command.
+* Call the DPE DeriveChild command with the DefaultContext in the locality of
+  the PL0 PAUSER.
+* Extend the measurement into PCR31 (`PCR_ID_STASH_MEASUREMENT`).
 
 Command Code: `0x4D45_4153` ("MEAS")
 
@@ -292,6 +294,8 @@ Table: `DISABLE_ATTESTATION` output arguments
 
 ### INVOKE\_DPE\_COMMAND
 
+Invoke a serialized DPE command.
+
 Command Code: `0x4450_4543` ("DPEC")
 
 Table: `INVOKE_DPE_COMMAND` input arguments
@@ -311,6 +315,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
 | fips_status | u32           | Indicates if the command is FIPS approved or an error
 | data_size   | u32           | Length in bytes of the valid data in the data field
 | data        | u8[...]       | DPE response structure as defined in the DPE iRoT profile.
+
+### QUOTE\_PCRS
+
+Generate a signed quote over all Caliptra hardware PCRs using the Caliptra PCR quoting key.
+All PCR values are hashed together with the nonce to produce the quote.
+
+Command Code: `0x5043_5251` ("PCRQ")
+
+Table: `QUOTE_PCRS` input arguments
+
+| **Name**     | **Type**      | **Description**
+| --------     | --------      | ---------------
+| chksum       | u32           | Checksum over other input arguments, computed by the caller. Little endian.
+| nonce        | u8[32]        | Caller-supplied nonce to be included in signed data
+
+Table: `QUOTE_PCRS` output arguments
+
+PcrValue is defined as u8[48]
+
+| **Name**     | **Type**     | **Description**
+| --------     | --------     | ---------------
+| chksum       | u32          | Checksum over other output arguments, computed by Caliptra. Little endian.
+| PCRs         | PcrValue[32] | Values of all PCRs
+| reset\_ctrs  | u32[32]      | Reset counters for all PCRs
+| signature\_r | u8[48]       | R portion of the signature over the PCR quote.
+| signature\_s | u8[48]       | S portion of the signature over the PCR quote.
+
+### EXTEND\_PCR
+
+Extend a Caliptra hardware PCR
+
+Command Code: `0x5043_5245` ("PCRE")
+
+Table: `EXTEND_PCR` input arguments
+
+| **Name**     | **Type**      | **Description**
+| --------     | --------      | ---------------
+| chksum       | u32           | Checksum over other input arguments, computed by the caller. Little endian.
+| index        | u32           | Index of the PCR to extend
+| value        | u8[..]        | Value to extend into the PCR at `index`
+
+
+`EXTEND_PCR` returns no output arguments.
+
+### INCREMENT\_PCR\_RESET\_COUNTER
+
+Increment the reset counter for a PCR
+
+Command Code: `0x5043_5252` ("PCRR")
+
+Table: `INCREMENT_PCR_RESET_COUNTER` input arguments
+
+| **Name**     | **Type**      | **Description**
+| --------     | --------      | ---------------
+| chksum       | u32           | Checksum over other input arguments, computed by the caller. Little endian.
+| index        | u32           | Index of the PCR for which to increment the reset counter
+
+`INCREMENT_PCR_RESET_COUNTER` returns no output arguments.
 
 ## Checksum
 
@@ -455,11 +517,6 @@ Caliptra Runtime firmware is responsible for initializing DPE’s Default Contex
 *Note: the Runtime CDI (from KeyVault) will be used as-needed and will not be
 accessed during initialization.*
 
-### TCI Storage
-
-Caliptra SHALL set `MAX_TCI_NODES` to 24. DPE will hold these TCI nodes in DCCM
-which persists on warm/impactless updates.
-
 ### CDI Derivation
 
 The DPE Sign and CertifyKey commands derive an asymmetric key for that handle.
@@ -468,7 +525,7 @@ DPE will first collect measurements and concatenate them in a byte buffer
 `MEASUREMENT_DATA`:
 
 * LABEL parameter passed to Sign or CertifyKey.
-* The TCI values for each TCI node in the path from the current TCI node to the
+* The `TCI_NODE_DATA` structures in the path from the current TCI node to the
   root, inclusive, starting with the current node.
 
 To derive a CDI for a given context, DPE shall use KeyVault hardware with the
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -227,6 +229,57 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+| **Name**     | **Type**      | **Description**
+| --------     | --------      | ---------------
+| chksum       | u32           | Checksum over other input arguments, computed by the caller. Little endian.
+| index        | u32           | Index of the PCR to extend
+| value        | u8[..]        | Value to extend into the PCR at `index`
+
+
+`EXTEND_PCR` returns no output arguments.
+
+### INCREMENT\_PCR\_RESET\_COUNTER
```
varuns-nvidia, line 268: Does this mean that DPE cannot attest to journey TCIs as a function of resets, i.e., journey TCIs are only a function of initial boot + hitless updates?

jhand2: It would say it is left up to the caller how to handle resets. The caller may choose to

* Treat a reset as a hitless update (and optionally record the reset in a log somewhere)
* Not remeasure (leave the existing pre-reset measurements) if code doesn't change
* Treat a reset as a cold reset for that component and remove the tree

I think the right behavior depends on the context and DPE should be capable of accommodating all of the above scenarios.

runtime/README.md
```diff
@@ -227,6 +229,57 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
 | data        | u8[...]       | DPE response structure as defined in the DPE iRoT profile.
 
+### QUOTE\_PCRS
+
+Generate a signed quote over all Caliptra hardware PCRs using the Caliptra PCR quoting key.
+All PCR values are hashed together for the quote
+
+Command Code: `0x5043_5251` ("PCRQ")
+
+`QUOTE_PCRs` takes no input arguments.
```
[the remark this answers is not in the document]
jhand2: Added a nonce here, but I'm still not super clear how the hardware takes the nonce (I think I remember hearing that it does take one though).

cc @bharatpillilli

runtime/README.md
```diff
@@ -227,6 +229,57 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+
+Table: `QUOTE_PCRS` output arguments
```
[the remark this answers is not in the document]
jhand2: Done. Also removed the digest since it can/should be computed when verifying the signature.

runtime/README.md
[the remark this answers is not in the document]
varuns-nvidia: Here's the register for it https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.sha512_reg.SHA512_GEN_PCR_HASH_NONCE

runtime/README.md
varuns-nvidia: Got it, but ultimately DPE is unaware of reset counters, it's just tracked on a per-PCR basis.

runtime/README.md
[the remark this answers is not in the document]
varuns-nvidia: Looks like just needs to be 32-bits instead of 48 bytes

runtime/README.md
jhand2: Correct

runtime/README.md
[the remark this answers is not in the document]
jhand2: Changed this to 32 bytes based on discussion in https://github.com/chipsalliance/caliptra-rtl/issues/139

runtime/README.md
```diff
@@ -227,6 +229,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+| nonce        | u8[32]        | Caller-supplied nonce to be included in signed data
+
+Table: `QUOTE_PCRS` output arguments
+
+PcrValue is defined as u8[48]
+
+| **Name**     | **Type**     | **Description**
+| --------     | --------     | ---------------
+| chksum       | u32          | Checksum over other output arguments, computed by Caliptra. Little endian.
+| PCRs         | PcrValue[24] | Digest of all PCRs
```
varuns-nvidia, line 253: nit: Description should be PCR values

jhand2: Done.

varuns-nvidia (approved):

ArthurHeymans (commented):
Just a few question to try understanding this better.

runtime/README.md
```diff
@@ -227,6 +229,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
 | chksum      | u32           | Checksum over other output arguments, computed by Caliptra. Little endian.
 | data        | u8[...]       | DPE response structure as defined in the DPE iRoT profile.
 
+### QUOTE\_PCRS
+
+Generate a signed quote over all Caliptra hardware PCRs using the Caliptra PCR quoting key.
```
ArthurHeymans, line 234: How does one obtain this PCR quoting (public) key when using the mailbox?

runtime/README.md
```diff
@@ -227,6 +229,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+All PCR values are hashed together with the nonce to produce the quote.
+
+Command Code: `0x5043_5251` ("PCRQ")
+
+Table: `QUOTE_PCRS` input arguments
+
+| **Name**     | **Type**      | **Description**
+| --------     | --------      | ---------------
+| chksum       | u32           | Checksum over other input arguments, computed by the caller. Little endian.
+| nonce        | u8[32]        | Caller-supplied nonce to be included in signed data
```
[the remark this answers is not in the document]
jhand2: There is hardware support for quoting the PCRs. The flow looks something like this:

* Program nonce registers: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.sha512_reg.SHA512_GEN_PCR_HASH_NONCE%5B0%5D
* Run PCR hash: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.sha512_reg.SHA512_GEN_PCR_HASH_CTRL
* Run PCR sign (which will sign the digest from the SHA512 block): https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.ecc_reg.ECC_CTRL#PCR_SIGN.desc

From what I can tell, the drivers don't currently implement the PCR signing flow. @mhatrevi to confirm.

runtime/README.md
jhand2: I believe this will be the FMC Alias Key. @ericeilertson @mhatrevi to confirm

runtime/README.md
```diff
@@ -227,6 +229,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+
+Table: `QUOTE_PCRS` output arguments
+
+PcrValue is defined as u8[48]
+
+| **Name**     | **Type**     | **Description**
+| --------     | --------     | ---------------
+| chksum       | u32          | Checksum over other output arguments, computed by Caliptra. Little endian.
+| PCRs         | PcrValue[24] | Values of all PCRs
+| reset\_ctrs  | u32[24]      | Reset counters for all PCRs
```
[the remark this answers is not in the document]
jhand2: Correct, this is to be implemented in software. This spec change documents how the reset counters work: https://github.com/chipsalliance/Caliptra/pull/54

In general they are informational to allow the verifier to make sense of the measurement log.

runtime/README.md
```diff
@@ -227,6 +229,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+| nonce        | u8[32]        | Caller-supplied nonce to be included in signed data
+
+Table: `QUOTE_PCRS` output arguments
+
+PcrValue is defined as u8[48]
+
+| **Name**     | **Type**     | **Description**
+| --------     | --------     | ---------------
+| chksum       | u32          | Checksum over other output arguments, computed by Caliptra. Little endian.
+| PCRs         | PcrValue[24] | Values of all PCRs
```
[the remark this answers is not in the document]
jhand2: You're right. I'll make that change.

runtime/README.md
```diff
@@ -489,6 +544,7 @@ To derive an asymmetric key for Sign and CertifyKey, RT will
 * For Sign: Sign passed data
 * Erase KeyVault slots 0 and 1
 
+<<<<<<< HEAD
```
varuns-nvidia, line 547: merge artifact

runtime/README.md
```diff
@@ -312,6 +314,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
...
+
+| **Name**     | **Type**     | **Description**
+| --------     | --------     | ---------------
+| chksum       | u32          | Checksum over other output arguments, computed by Caliptra. Little endian.
+| PCRs         | PcrValue[32] | Values of all PCRs
+| reset\_ctrs  | u32[32]      | Reset counters for all PCRs
+| signature\_r | u8[48]       | R portion of the signature over the PCR quote.
+| signature\_s | u8[48]       | S portion of the signature over the PCR quote.
+
+### EXTEND\_PCR
```
varuns-nvidia, line 343: should we note in STASH_MEASUREMENT that it also extends PCRs?

jhand2: Done

runtime/README.md
jhand2: Thanks, fixed

varuns-nvidia (approved):

runtime/README.md
```diff
@@ -312,6 +316,64 @@ Table: `INVOKE_DPE_COMMAND` output arguments
 | data_size   | u32           | Length in bytes of the valid data in the data field
 | data        | u8[...]       | DPE response structure as defined in the DPE iRoT profile.
 
+### QUOTE\_PCRS
+
+Generate a signed quote over all Caliptra hardware PCRs using the Caliptra PCR quoting key.
+All PCR values are hashed together with the nonce to produce the quote.
```
bluegate010, line 322: Would eventually be good to update this with the exact formula for validating the signature.

bluegate010 (approved):

runtime/README.md
jhand2: Ack. I'll update this once the implementation is done and I can verify how it's done.

PR #432, opened 2023-06-27
Minor changes to FMC DICE stack

bluegate010:
* ROM has retired the DiceLayer trait. Removed from FMC for consistency.
* DiceInput's subj_priv_key and uds_key fields are unused.
* Check for KV slot collision.
* Clarified name of RT CDI KV slot.


fdamato (approved):
Thanks Jeff. Minor, but very thoughtful to check for slot collisions.

Issue #438, opened 2023-06-29
Integration Test Cases For Update-Reset [Hitless Update]


ajisaxena:
Please update the integration test suite to cover the update-reset test case.

jhand2:
FYI im working on adding the hitless update mailbox command to RT that should make it easier to test end to end. I should have a PR out tomorrow.

PR #440, opened 2023-06-29
[Bug Fix] Fixes for Image Validation Failure on hitless updates

### LINKED ISSUE #422: ROM should continue transferring control to FMC in the event of an update reset flow failure

### LINKED ISSUE #441: Manifest mirroring copying more data than necessary
I think the manifest mirroring code during update reset in ROM looks incorrect: https://github.com/chipsalliance/caliptra-sw/blob/3fb8f24356e116a0e3aba9561f3301c4c02c9b5b/rom/dev/src/flow/update_reset.rs#L104-L128

The size to be copied takes into account the size of the FMC and RT code. But this mirroring should only apply to the actual manifest. I think this should be something like

```
let dst = unsafe {
    let ptr = &mut MAN1_ORG as *mut u32;
    core::slice::from_raw_parts_mut(
        ptr,
        core::mem::size_of::<ImageManifest>() / 4,
    )
};

let src = unsafe {
    let ptr = &mut MAN2_ORG as *mut u32;

    core::slice::from_raw_parts_mut(
        ptr,
        core::mem::size_of::<ImageManifest>() / 4,
    )
};
```

ajisaxena:
This PR resolves the following:
1. When the update reset fails for some reason [like invalid image], the ROM was not transferring the control to the FMC. 
2. The Manifest regions were smaller than the actual manifest size. 
3. The registers were not getting locked in Update_Reset case

**NOTE**: 
We realized that the integration test coverage for the update_reset is not there in our integration test suite. This PR is just a Bug-Fix  and we did not want have integration test cases clubbed with this PR. Expect another PR which will contain additional integration tests to cover the update-reset test case. 

[https://github.com/chipsalliance/caliptra-sw/issues/438]


common/src/hand_off.rs
```diff
@@ -374,11 +375,30 @@ pub fn print_fht(fht: &FirmwareHandoffTable) {
 }
 
 impl FirmwareHandoffTable {
+    fn reset_reason() -> ResetReason {
```
rusty1968, line 378: I don't see this as part of the set of operations for the FHT type. We can make it a free funtion.

rom/dev/src/fht.rs
```diff
@@ -183,14 +183,14 @@ pub fn make_fht(env: &RomEnv) -> FirmwareHandoffTable {
     }
 }
 
-pub fn load_fht(fht: FirmwareHandoffTable) {
+pub fn store_fht(fht: FirmwareHandoffTable) {
```
rusty1968, line 186: Can we use this ? https://github.com/chipsalliance/caliptra-sw/blob/main/common/src/hand_off.rs#L423

ajisaxena: I still want to use "load" and "store". Easy to understand. I will change the name to "store".

jhand2 (dismissed):

common/src/hand_off.rs
```diff
@@ -377,8 +378,12 @@ impl FirmwareHandoffTable {
     /// Perform valdity check of the table's data.
     /// The fields below should have been populated by ROM with
     /// valid data before it transfers control to mutable code.
+    /// This function can only be called from non test case environment
+    /// as this function accesses the registers to get the reset_reason.
     pub fn is_valid(&self) -> bool {
```
rusty1968, line 383: We can pass reset reason as an argument and keep the FHT hardware agnostic.

ajisaxena: I had it that way but then I realized that is_valid is also getting called from try_load, which again gets called from some other tests. So one way or other we will end up putting this limitation. If we keep the reset_reason local to this function we just need modification in one test case. This is a small thing and we can revisit this later, if needed.

mhatrevi (dismissed):

rusty1968 (dismissed):

rom/dev/src/main.rs
```diff
@@ -154,6 +168,20 @@ fn rom_panic(_: &core::panic::PanicInfo) -> ! {
     report_error(CaliptraError::ROM_GLOBAL_PANIC.into());
 }
 
+fn report_error_update_reset(code: u32) {
+    cprintln!("ROM Error: 0x{:08X}", code);
+    report_fw_error_non_fatal(code);
+
+    // Zeroize the crypto blocks.
+    unsafe {
```
korran, line 176: @mhatrevi Is this actually necessary? I would prefer not to go behind the back of the driver instances in RomEnv to change peripheral state, since they're going to be used later on.

These unsafe methods are really only intended for use in fatal trap handlers, where we can't use the safe interface because we don't have access to it.

I think there's two ways forward:

- Don't zeroize here, since these aren't fatal FIPS errors, and the drivers should already zeroize their contents.

- If zeroization here is a requirement, use a safe API on the drivers in the RomEnv.

ajisaxena: I agree. I do not think that it is a requirement to zeroize here. We are able to transfer the control to FMC exactly like cold boot without zeroizing. We are still transferring the control here, don't see why we should need to zeroize.

mhatrevi: I would suggest going with the first option since this is a soft failure.

ajisaxena: Updated!!

mhatrevi (dismissed):

mhatrevi (approved):

korran (approved):

PR #442, opened 2023-06-29
[feat] Fatal vs. Non-fatal error attribution

### LINKED ISSUE #155: ROM flow error attribution review
Classify errors into fatal vs non-blocking

mhatrevi:
This change documents the fatal (cold reset) and non-fatal errors (update-reset) raised by the ROM. Additionally, it fixes few error reporting discrepancies.


rom/dev/doc/error-attribution.md
```diff
@@ -0,0 +1,232 @@
+# **Fatal Errors**
```
korran, line 1: Is there any information here that's not already in `error/src/lib.rs`? Manually maintaining a duplicate seems like a waste of time to me...

hw-model/src/lib.rs
```diff
@@ -569,7 +571,11 @@ pub trait HwModel {
             self.soc_mbox().execute().write(|w| w.execute(false));
             let soc_ifc = self.soc_ifc();
             return Err(ModelError::MailboxCmdFailed(
-                soc_ifc.cptra_fw_error_non_fatal().read(),
+                if cmd == MBOX_DOWNLOAD_FIRMWARE_CMD_ID {
```
korran, line 574: Eh, I don't really like this sort of hard-coded logic. Can we put the error in both fw_error_fatal and fw_error_non_fatal?

rom/dev/doc/error-attribution.md
mhatrevi: Errors in lib.rs don't have an attribution of fatal vs. non-fatal. Eventually, this document can get published as official.

hw-model/src/lib.rs
mhatrevi: Discussed offline; filed https://github.com/chipsalliance/caliptra-sw/issues/453 to add fatal error to non-fatal registers for keeping test functions generic.

rom/dev/doc/error-attribution.md
mhatrevi: Discussed offline; we'll keep this for now and auto-generate it later. At that time, this file should be removed.

korran (approved):

PR #444, opened 2023-06-29
Add impactless update command to RT

jhand2:
1. Add impactless update mailbox command
2. Fix a keyvault emulation bug which was clearing usage bits
3. Make ROM use the old FHT on update reset

### DIFF
--- a/drivers/src/mailbox.rs
+++ b/drivers/src/mailbox.rs
@@ -347,6 +347,19 @@ impl MailboxRecvTxn<'_> {
         Ok(())
     }
 
+    /// Pulls at most `count` words from the mailbox and throws them away
+    pub fn drop_words(&mut self, count: usize) -> CaliptraResult<()> {
+        let mbox = self.mbox.regs_mut();
+        let dlen_bytes = mbox.dlen().read() as usize;
+        let dlen_words = (dlen_bytes + 3) / 4;
+        let words_to_read = min(count, dlen_words);
+        for _ in 0..words_to_read {
+            _ = mbox.dataout().read();
+        }
+
+        Ok(())
+    }
+
     /// Pulls at most `data.len()` words from the mailbox FIFO without performing state transition.
     ///
     /// # Arguments
@@ -356,7 +369,7 @@ impl MailboxRecvTxn<'_> {
     /// # Returns
     ///
     /// Status of Operation
-    ///   
+    ///
     pub fn copy_request(&mut self, data: &mut [u32]) -> CaliptraResult<()> {
         if self.state != MailboxOpState::Execute {
             return Err(CaliptraError::DRIVER_MAILBOX_INVALID_STATE);
@@ -374,7 +387,7 @@ impl MailboxRecvTxn<'_> {
     /// # Returns
     ///
     /// Status of Operation
-    ///   
+    ///
     pub fn recv_request(&mut self, data: &mut [u32]) -> CaliptraResult<()> {
         self.copy_request(data)?;
         self.complete(true)?;
--- a/rom/dev/src/flow/update_reset.rs
+++ b/rom/dev/src/flow/update_reset.rs
@@ -11,7 +11,7 @@ Abstract:
     File contains the implementation of update reset flow.
 
 --*/
-use crate::{cprintln, fht, rom_env::RomEnv, verifier::RomImageVerificationEnv, wdt};
+use crate::{cprintln, rom_env::RomEnv, verifier::RomImageVerificationEnv, wdt};
 
 use caliptra_common::memory_layout::{MAN1_ORG, MAN2_ORG};
 use caliptra_common::FirmwareHandoffTable;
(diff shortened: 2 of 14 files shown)


jlmahowa-amd (approved):

rom/dev/src/fht.rs
```diff
@@ -176,3 +178,13 @@ pub fn store(fht: FirmwareHandoffTable) {
     caliptra_common::print_fht(&fht);
     slice.copy_from_slice(fht.as_bytes());
 }
+
+pub fn load_from_dccm() -> CaliptraResult<FirmwareHandoffTable> {
+    let slice = unsafe {
```
rusty1968, line 183: cant we use the function try_load in common?

jhand2: Done.

drivers/src/mailbox.rs
```diff
@@ -347,6 +347,19 @@ impl MailboxRecvTxn<'_> {
         Ok(())
     }
 
+    /// Pulls at most `count` words from the mailbox and throws them away
+    pub fn drop_words(&mut self, count: usize) -> CaliptraResult<()> {
```
rusty1968, line 351: create an issue to add a unit test for this later?

rom/dev/src/rom.ld
```diff
@@ -19,7 +19,8 @@ ENTRY(_start)
 ROM_ORG          = 0x00000000;
 ICCM_ORG         = 0x40000000;
 DCCM_ORG         = 0x50000000;
-DATA_ORG         = 0x50004000;
+/* Avoid overlap with other long-lived DCCM structures */
+DATA_ORG         = 0x50004400;
```
rusty1968, line 23: not related to this PR, but create an issue to generate rom.ld from memory_layout.rs information.

rom/dev/src/rom.ld
```diff
@@ -19,7 +19,8 @@ ENTRY(_start)
 ROM_ORG          = 0x00000000;
 ICCM_ORG         = 0x40000000;
 DCCM_ORG         = 0x50000000;
-DATA_ORG         = 0x50004000;
+/* Avoid overlap with other long-lived DCCM structures */
```
mhatrevi, line 22: Please update memory_layout.rs in common\src with this change.

rom/dev/src/flow/update_reset.rs
```diff
@@ -81,9 +81,14 @@ impl UpdateResetFlow {
         Self::copy_regions();
         report_boot_status(UpdateResetOverwriteManifestComplete.into());
 
+        // Return the old FHT from DCCM
+        let fht = FirmwareHandoffTable::try_load()
```
mhatrevi, line 85: Can the load and entry point addresses of RT be different than the previous in the new FW bundle? If yes, the values of these in FHT would be different from the actual load address of the new RT in ICCM.

mhatrevi (changes requested):

rom/dev/src/flow/update_reset.rs
jhand2: Hmm. The ICCM address could change, but I don't think the DataVault handle can change. So we probably don't need to update the FHT. I'll update this.

However, we should be overwriting the load address in DataVault. But when I try to call `write_warm_reset_entry4(WarmResetEntry4::RtLoadAddr, manifest.runtime.load_addr)` during update reset, I get a Load/Store fault. I would expect these datavault entries to be unlocked. Maybe a simulator issue?

rom/dev/src/rom.ld
jhand2: Actually I don't think this change is needed anymore. I don't see any overlap. I'll remove this.

rom/dev/src/rom.ld
jhand2: This change is no longer needed. Removing.

drivers/src/mailbox.rs
jhand2: Added a unit test

rom/dev/src/flow/update_reset.rs
jhand2: Changed to just not change FHT.

For the second part of the comment, opened https://github.com/chipsalliance/caliptra-sw/issues/518

rom/dev/src/rom.ld
```diff
@@ -33,7 +33,7 @@ ROM_RELAXATION_PADDING = 4k;
 ROM_SIZE          = 32K;
 ICCM_SIZE         = 128K;
 DCCM_SIZE         = 128K;
-DATA_SIZE         = 96K;
+DATA_SIZE         = 95K;
```
mhatrevi, line 36: This change is no longer needed, right?

rom/dev/src/flow/warm_reset.rs
```diff
@@ -25,11 +25,11 @@ impl WarmResetFlow {
     ///
     /// * `env` - ROM Environment
     #[inline(never)]
-    pub fn run(env: &mut RomEnv) -> CaliptraResult<FirmwareHandoffTable> {
+    pub fn run(_env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
         cprintln!("[warm-reset] ++");
 
         cprintln!("[warm-reset] --");
```
mhatrevi, line 32: There will be a conflict here due to https://github.com/chipsalliance/caliptra-sw/pull/516. Suggest keeping the workaround and not returning the FHT (this change).

mhatrevi (changes requested):

rom/dev/src/rom.ld
jhand2: Done.

rom/dev/src/flow/warm_reset.rs
jhand2: Done.

mhatrevi (approved):

PR #462, opened 2023-07-07
[feat] Zeroize stack and CPU registers before transferring to FMC

mhatrevi:
This change zeroes out the stack before transferring to the FMC.

### DIFF
--- a/rom/dev/src/start.S
+++ b/rom/dev/src/start.S
@@ -342,7 +342,31 @@ _nmi_handler:
 .global exit_rom
 exit_rom:
     .cfi_startproc
-    jr a0
+
+    //
+    // Clear the stack
+    //
+    
+    // Save the FMC address
+    addi a3, a0, 0
+    la a0, STACK_ORG         // dest
+    la a1, STACK_SIZE        // len 
+    call _zero_mem256
+
+   // Clear all registers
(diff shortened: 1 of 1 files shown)


rom/dev/src/start.S
```diff
@@ -342,6 +342,25 @@ _nmi_handler:
 .global exit_rom
 exit_rom:
     .cfi_startproc
+
+    // Clear the stack
```
korran, line 346: Can we just call _zero_mem256? It doesn't touch the stack.

vsonims (dismissed):

rom/dev/src/start.S
```diff
@@ -342,7 +342,16 @@ _nmi_handler:
...
 exit_rom:
     .cfi_startproc
-    jr a0
+
+    // Clear the stack
+    // Save the FMC address
+    addi a3, a0, 0
+    la a0, STACK_ORG         // dest
+    la a1, STACK_SIZE        // len 
+    call _zero_mem256
```
korran, line 351: If we clearing the stack, we should probably also clear all the CPU registers, no?

mhatrevi: Updated.

korran (approved):

Issue #391, opened 2023-06-15
HW Release: Caliptra C API package provided as part of the supporting ZIP File


fdamato:
The Caliptra C API Library should be distributed in the collateral supporting ZIP File, as part of the HW Release.
The provided APIs should allow the SOC to interact with Caliptra, through (included but not limited to) mailbox communication.
The library would need to be compiled by the integrator in their environment, they would need to:
         - Implement SOC interface functions which will perform the SOC-specific Caliptra operations
               `void apb_write(uint32_t address, uint32_t data);`
                `uint32_t apb_read(uint32_t address);`
         - Provide PATH to their cross compiler

wmaroneAMD:
Reassigned to myself. I've linked my branch on my fork of caliptra-sw.

wmaroneAMD:
I think this is done.

PR #465, opened 2023-07-10
RE: #391 C API Package (libcaliptra)

wmaroneAMD:
Derived from the api example in hw-model/c-binding/examples/api, this is the initial skeleton of a more generic client API for Caliptra, intended to abstract away the details of the implementation.

Client applications will include caliptra_api.h and implementations will provide definitions to the functions declared in caliptra_if.h, check the README.md for more information.

### DIFF
--- a/libcaliptra/Makefile
+++ b/libcaliptra/Makefile
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2020 Western Digital Corporation or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Set Q= on the command line for verbosity
+Q=@
+
+# Set build targets, flags, and a defaultr
+ifeq ($(findstring "debug",$(MAKECMDGOALS)),"debug")
+CARGO_OUT = $(abspath ../target/debug)
+CARGO_FLAG =
+else ifeq ($(findstring "release,$(MAKECMDGOALS)),"release")
+CARGO_OUT = $(abspath ../target/release)
+CARGO_FLAG = --release
+else
+CARGO_OUT = $(abspath ../target/debug)
+endif
+
+RTL_SOC_IFC_INCLUDE_PATH = ../hw-latest/caliptra-rtl/src/soc_ifc/rtl
+CFLAGS += -I$(RTL_SOC_IFC_INCLUDE_PATH) -I$(CARGO_OUT) -I$(abspath inc)
+
+SOURCE += src/caliptra_api.c
+
+OBJS := $(patsubst %.c,%.o, $(filter %.c,$(SOURCE)))
+
+TESTS = $(sort $(notdir $(wildcard test/*)))
+CLEAN_TESTS = $(addprefix clean-, $(TESTS))
+
+.PHONY:%_test %_run clean $(CLEAN_TESTS)
+
+debug: all
+release: all
+tests: $(TESTS)
+all: $(OBJS) $(TESTS)
+
+$(OBJS): $(SOURCE)
+	@echo [CC] $@
+	$(Q)$(CC) ${CFLAGS} -c $< -o $@
+
+$(CLEAN_TESTS): FORCE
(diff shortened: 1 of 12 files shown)


rusty1968 (dismissed):

libcaliptra/src/caliptra_api.c
```diff
@@ -0,0 +1,211 @@
...
+        return -EIO;
+    } else if(status == CALIPTRA_MBOX_STATUS_CMD_COMPLETE) {
+        caliptra_mbox_write_execute(false);
+        return 0;
+    } else if (status != CALIPTRA_MBOX_STATUS_DATA_READY) {
+        return -EIO;
+    }
+
+    // Read Mbox out Data Len
+    uint32_t dlen = caliptra_mbox_read_dlen();
```
rusty1968, line 184: Are we discarding dlen?

wmaroneAMD: We are. The same thing happens in the c-binding example API. We don't have any real build flags to require we do something with it, so then the question is if reading dlen has any side-effects within the caliptra hardware/rom. If not then we can remove this.

rusty1968: Reading dlen does not change state.

wmaroneAMD: Ok. Then I'll double check and remove it.

libcaliptra/test/if_test/if_test.c
```diff
@@ -0,0 +1,21 @@
+//Licensed under the Apache-2.0 license
+
+o#include <stdint.h>
```
korran, line 3: This doesn't compile due to this extra character

libcaliptra/Makefile
```diff
@@ -0,0 +1,70 @@
...
+clean: $(CLEAN_TESTS)
+	@echo [CLEAN] $(CARGO_OUT) $(OBJS)
+	$(Q)$(RM) -r $(CARGO_OUT)
+	$(Q)$(RM) $(OBJS)
+
+export CARGO_OUT
+export MODEL_HEADER
+export CFLAGS
+
+%_test: $(MODEL_HEADER)
```
korran, line 66: Can you test these makefile targets as part of our CI?

https://github.com/chipsalliance/caliptra-sw/blob/main/.github/workflows/build-test.yml

libcaliptra/test/if_test/if_test.c
wmaroneAMD: oops, sorry about that.

libcaliptra/test/if_test/if_test_impl.c
```diff
@@ -0,0 +1,31 @@
+//Licensed under the Apache-2.0 license
```
korran, line 1: We probably need an implementation of this that uses the hw-model; are you planning on adding this in a future PR?

wmaroneAMD: Yes, I'm working with Fabrizio to move this in that direction.

libcaliptra/Makefile
wmaroneAMD: Added.

libcaliptra/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This is an early check-in to ensure its presence upstream and to generate feedback going forward.
+
+* Expand available APIs
+* Interrupts?
+** Handling thereof?
+* Add a demonstration interface connecting to the hardware model
+* Add a demonstration interface to the software simulator (?)
+* Add support for building the caliptra interface implementation if present?
+** Produce an .a file for later link?
```
fdamato, line 45: I would remove this line...I don`t think we can provide in binary format as it might be very complex if not impossible to keep it ABI compatible with the SOC integrator

wmaroneAMD: Fair, if the goal is to have the consumer handle compile as well then fancy build products don't matter here.

libcaliptra/inc/caliptra_if.h
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * caliptra_ready_for_fuses
+ *
+ * Reports if the Caliptra hardware is ready for fuse data
+ *
+ * @return bool True if ready, false otherwise
+ */
+bool caliptra_ready_for_fuses(void);
+
+// Firmware
```
fdamato, line 45: can we move those 2 functions into the caliptra_api.c and have a generic caliptra_wait() function which is implemented by the user.
Probably we can rename those as caliptra_wait_for_fuses(..), caliptra_wait_for_firmware(..)

wmaroneAMD: Right, and the underlying wait function would be defined within caliptra_if.h

libcaliptra/README.md
korran: Yeah, source only please.

wmaroneAMD: Removed

PR #466, opened 2023-07-11
[fix] Update reset flow overwriting data

mhatrevi:
Update reset flow is copying four times the size of the ImageManifest, overwriting the TBS and PCR/Fuse log regions.


korran:
Ugh, I'd really like to get rid of all this unsafe code and just treat the entire "persistent" part of DCCM as a slice of bytes.

korran (approved):

PR #467, opened 2023-07-11
RT FIPS Infrastructure

rusty1968:
This commit adds scaffold  for FIPs command handling in RT firmware.

### DIFF
--- a/runtime/src/fips.rs
+++ b/runtime/src/fips.rs
@@ -1,52 +1,27 @@
 // Licensed under the Apache-2.0 license
-
+use caliptra_common::cprintln;
 use caliptra_drivers::CaliptraError;
 use caliptra_drivers::CaliptraResult;
-use core::convert::TryFrom;
-
-use caliptra_drivers::KeyVault;
 use caliptra_registers::mbox::enums::MboxStatusE;
-/// FIPS module environment.
-pub struct FipsEnv<'a> {
-    pub key_vault: &'a mut KeyVault,
-}
 
-#[derive(PartialEq, Eq)]
-pub struct FipsModuleApi(pub u32);
+use crate::Drivers;
 
-/// FIPS module commands.
-impl FipsModuleApi {
-    /// The status command.
-    pub const VERSION: Self = Self(0x4650_5652); // "FPVR"
-    /// The self-test command.
-    pub const SELF_TEST: Self = Self(0x4650_4C54); // "FPST"
-    /// The shutdown command.
-    pub const SHUTDOWN: Self = Self(0x4650_5344); // "FPSD"
-}
+/// Fips command handler.
+pub struct FipsModule;
 
-impl TryFrom<u32> for FipsModuleApi {
-    type Error = CaliptraError;
-
-    fn try_from(value: u32) -> Result<Self, Self::Error> {
-        match value {
-            0x4650_5652 => Ok(Self::VERSION),   // "STAT"
-            0x4650_4C54 => Ok(Self::SELF_TEST), // "SELF"
-            0x4650_5344 => Ok(Self::SHUTDOWN),  // "SHDN"
-            _ => Err(CaliptraError::RUNTIME_UNIMPLEMENTED_COMMAND),
-        }
+impl FipsModule {
+    pub fn version(_env: &Drivers) -> CaliptraResult<MboxStatusE> {
+        cprintln!("[rt] FIPS Version");
+        Err(CaliptraError::RUNTIME_FIPS_UNIMPLEMENTED)
     }
-}
 
-/// Convert the FipsModuleApi to a u32.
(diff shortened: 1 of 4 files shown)


runtime/src/fips.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// FIPS module commands.
+impl FipsModuleApi {
+    /// The status command.
+    pub const VERSION: Self = Self(0x4650_5652); // "FPVR"
+    /// The self-test command.
+    pub const SELF_TEST: Self = Self(0x4650_4C54); // "FPST"
+    /// The shutdown command.
+    pub const SHUTDOWN: Self = Self(0x4650_5344); // "FPSD"
+}
```
jhand2, line 25: Could we just group these with the regular mailbox commands in lib.rs? They can still be implemented in fips.rs

runtime/src/fips.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Convert the FipsModuleApi to a u32.
+impl From<FipsModuleApi> for u32 {
+    fn from(api: FipsModuleApi) -> Self {
+        api.0
+    }
+}
+
+/// FIPS module trait.
+pub trait FipsManagement {
```
jhand2, line 48: Do we need a trait for this, or can it just be a struct with member functions? I don't think we need multiple implementations of these functions right?

runtime/src/lib.rs
```diff
@@ -164,3 +169,45 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) {
         }
     }
 }
+
+#[derive(Default)]
+struct FipsModule;
+
+impl FipsManagement for FipsModule {
```
jhand2, line 176: Can probably put this implementation in fips.rs just for readability.

runtime/src/lib.rs
```diff
@@ -164,3 +169,45 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) {
...
+    }
+
+    fn shutdown(&self, _env: &FipsEnv) -> CaliptraResult<MboxStatusE> {
+        cprintln!("[rt] FIPS shutdown");
+        Err(CaliptraError::RUNTIME_FIPS_UNIMPLEMENTED)
+    }
+}
+
+impl FipsModule {
+    pub fn handle_command(
```
jhand2, line 194: Similar comment to above, can this just be part of the main `handle_command` function

runtime/src/fips.rs
rusty1968: Sure.

runtime/src/fips.rs
rusty1968: I was hoping this could be a common trait for caliptra but for the moment we don't need an interface.

runtime/src/lib.rs
rusty1968: Done

runtime/src/lib.rs
rusty1968: Done

runtime/src/fips.rs
rusty1968: Done.

runtime/src/fips.rs
```diff
@@ -0,0 +1,26 @@
...
+use caliptra_drivers::CaliptraError;
+use caliptra_drivers::CaliptraResult;
+use caliptra_registers::mbox::enums::MboxStatusE;
+
+use crate::Drivers;
+
+pub struct FipsModule;
+
+impl FipsModule {
+    pub fn status(_env: &Drivers) -> CaliptraResult<MboxStatusE> {
```
JohnTraverAmd, line 12: Can we change this to Version similar to the command name?

rusty1968: Sure.

rusty1968: Done

runtime/tests/integration_tests.rs
```diff
@@ -217,6 +217,30 @@ fn test_verify_cmd() {
...
+
+    let resp = model.mailbox_execute(u32::from(CommandId::SHUTDOWN), &cmd);
+    assert_eq!(resp, expected_err);
+
+    let resp = model.mailbox_execute(u32::from(CommandId::SELF_TEST), &cmd);
+    assert_eq!(resp, expected_err);
+
+    let expected_err = Err(ModelError::MailboxCmdFailed(0xe0002));
+    // Send something that is not a valid RT command.
+    let resp = model.mailbox_execute(0xBABEFACE, &cmd);
```
jhand2, line 240: nit: Would recommend changing this to something like 0xAABBCCDD. This aligns with a similar effort in the RTL to remove potentially non-inclusive constants/langauge.

JohnTraverAmd (dismissed):

jhand2 (approved):

PR #469, opened 2023-07-12
Update DPE to latest commit

jhand2:
In particular, this adds license headers to several files which makes the file header check run cleanly on the DPE code.

It also contains several other DPE features.


jhand2:
No longer relevant. Submodule is updated well past this point.

PR #470, opened 2023-07-12
Enable CSRNG health tests

rkr35:
I recommend reviewing commit-by-commit.

This PR makes the following changes:

- Enables and sets thresholds for the Repetition Count and Adaptive Proportion health tests specified in [NIST SP 800-90B]( https://doi.org/10.6028/NIST.SP.800-90B).
- Enables FIPS mode for the `entropy_src` peripheral, which:
  - uses a larger window for health tests (2048 bits instead of 384 bits).
  - SHA-3 conditions 2048 raw entropy bits into 384 bits, which seeds the CSRNG.
- Adds unit tests for boot-time health checks.
- Adds sw emulation of health checks.


korran (approved):
Great work!

mhatrevi (approved):

PR #472, opened 2023-07-13
[feat] Store boot status in the boot status register only on an error

### LINKED ISSUE #489: [ROM] Store boot status in the boot status register only on an error

mhatrevi:
Originally, boot statuses were logged to the boot status register as the ROM progressed with its execution. This potentially opens up a surface for glitching attacks as the exact point of ROM execution is known. With this change, the boot status is updated in the DCCM and is stored in the boot status register only in the case of a failure.

### DIFF
--- a/drivers/src/soc_ifc.rs
+++ b/drivers/src/soc_ifc.rs
@@ -20,9 +20,23 @@ use crate::FuseBank;
 
 pub type Lifecycle = DeviceLifecycleE;
 
+// [TODO] Move memory layout out of the common crate
+// to avoid circular dependency with the drivers crate.
+pub const BOOT_STATUS_ORG: u32 = 0x500047FC;
+
 pub fn report_boot_status(val: u32) {
     let mut soc_ifc = unsafe { soc_ifc::SocIfcReg::new() };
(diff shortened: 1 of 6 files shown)


drivers/src/soc_ifc.rs
```diff
@@ -20,9 +20,23 @@ use crate::FuseBank;
 
 pub type Lifecycle = DeviceLifecycleE;
 
+// [TODO] Move memory layout out of the common crate
```
korran, line 23: to work around the circular dep issues, I think [memory_layout.rs](http://memory_layout.rs/) should just be in drivers/src/

korran: The problem with these sorts of TODOs is that they often get missed, and then we end up with hard-to-debug memory issues when the addresses change.

vsonims (approved):

ajisaxena (approved):

drivers/src/soc_ifc.rs
mhatrevi: I'll be addressing this TODO as part of https://github.com/chipsalliance/caliptra-sw/issues/609

PR #477, opened 2023-07-15
[fix] FMC: Address PCR handling issues.

rusty1968:
This commit addresses the  problem of  FMC "Journey PCR", on a warm reset,  being extended with an additional measurement of the same RT module that was there for the previous boot. Additionally it is  fixes the bug of FMC not locking journey and current PCRs and not erasing the current PCR before extending it.

### DIFF
--- a/fmc/src/flow/rt_alias.rs
+++ b/fmc/src/flow/rt_alias.rs
@@ -23,7 +23,9 @@ use caliptra_common::cprintln;
 use caliptra_common::crypto::Ecc384KeyPair;
 use caliptra_common::keyids::{KEY_ID_RT_CDI, KEY_ID_RT_PRIV_KEY, KEY_ID_TMP};
 use caliptra_common::HexBytes;
-use caliptra_drivers::{okref, report_boot_status, CaliptraError, CaliptraResult, KeyId};
+use caliptra_drivers::{
+    okref, report_boot_status, CaliptraError, CaliptraResult, KeyId, ResetReason,
+};
 use caliptra_x509::{NotAfter, NotBefore, RtAliasCertTbs, RtAliasCertTbsParams};
 
 const SHA384_HASH_SIZE: usize = 48;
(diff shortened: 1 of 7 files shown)


fmc/src/flow/rt_alias.rs
```diff
@@ -104,6 +106,11 @@ impl RtAliasLayer {
         Self::extend_pcrs(env, hand_off)?;
         cprintln!("[alias rt] Extend RT PCRs Done");
 
+        cprintln!("[alias rt] Lock RT PCRs");
+        env.pcr_bank.set_pcr_lock(caliptra_common::FMC_CURRENT_PCR);
```
bluegate010, line 110: The comment says we're locking RT PCRs, but the ID is FMC's PCR.

rusty1968: renamed it.

fmc/src/flow/rt_alias.rs
```diff
@@ -147,7 +156,11 @@ impl RtAliasLayer {
     /// * `hand_off` - HandOff
     pub fn extend_pcrs(env: &mut FmcEnv, hand_off: &HandOff) -> CaliptraResult<()> {
         extend_current_pcr(env, hand_off)?;
-        extend_journey_pcr(env, hand_off)?;
+        match env.soc_ifc.reset_reason() {
+            // Cold Reset Flow
+            ResetReason::ColdReset | ResetReason::UpdateReset => extend_journey_pcr(env, hand_off)?,
+            _ => cprintln!("[alias rt : skip journey pcr extension"),
```
bluegate010, line 162: Nit: mismatched brace.

fmc/src/flow/rt_alias.rs
```diff
@@ -104,6 +106,13 @@ impl RtAliasLayer {
         Self::extend_pcrs(env, hand_off)?;
         cprintln!("[alias rt] Extend RT PCRs Done");
 
+        cprintln!("[alias rt] Lock RT PCRs");
+        env.pcr_bank
+            .set_pcr_lock(caliptra_common::RTALIAS_CURRENT_PCR);
```
bluegate010, line 111: Nit: "alias" is a DICE concept, but these PCRs are a different thing. Suggest calling it "RT_FW_CURRENT_PCR" or something.

fmc/src/flow/rt_alias.rs
```diff
@@ -147,7 +156,11 @@ impl RtAliasLayer {
     /// * `hand_off` - HandOff
     pub fn extend_pcrs(env: &mut FmcEnv, hand_off: &HandOff) -> CaliptraResult<()> {
         extend_current_pcr(env, hand_off)?;
-        extend_journey_pcr(env, hand_off)?;
+        match env.soc_ifc.reset_reason() {
```
bluegate010, line 159: We should clear `RTALIAS_CURRENT_PCR` on update reset before extending into it, yeah?

fmc/src/flow/rt_alias.rs
rusty1968: done

fmc/src/flow/rt_alias.rs
rusty1968: done

bluegate010 (approved):

rusty1968:
Follow up needed : Sppend to the PCR log in DCCM, and  test all the extensions performed by FMC

PR #478, opened 2023-07-15
Do not write LDevID or FMC alias public keys to DV slots.

bluegate010:
These are unnecessary, and this change frees up DV slots for future use by FMC.

This change also removes these entries in FHT, removes the explicit ECDSA verify from RT alias cert generation (will be done anyways as part of the driver), and removes a test that only checks to ensure the keys in DV match those in the cert.

### DIFF
--- a/rom/dev/README.md
+++ b/rom/dev/README.md
@@ -360,32 +360,25 @@ Local Device ID Layer derives the Owner CDI & ECC Keys. This layer represents th
     `LDevIdPubKey = ecc384_keygen(KvSlot3, KvSlot5)`
     `kv_clear(KvSlot3)`
 
-4.	Store and lock (for write) the LDevID Public Key in Data Vault (48 bytes) Slot 0 & Slot 1
-
-    `dv48_store(LDevIdPubKey.X, Dv48Slot0)`
-    `dv48_lock_wr(Dv48Slot0)`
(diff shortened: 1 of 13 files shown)


rusty1968:
This PR does quite a bit more than what its summary describes.

bluegate010:
> This PR does quite a bit more than what its summary describes.

Ack, have amended the description with the knock-on changes.

rom/dev/README.md
```diff
@@ -397,10 +390,8 @@ Local Device ID Layer derives the Owner CDI & ECC Keys. This layer represents th
 
 | Slot | Key Vault | PCR Bank | Data Vault 48 Byte (Sticky) | Data Vault 4 Byte (Sticky) |
 |------|-----------|----------|-----------------------------|----------------------------|
-| 0 | | | 🔒LDevID Pub Key X |
-| 1 | | | 🔒LDevID Pub Key Y |
-| 2 | | | 🔒LDevID Cert Signature R |
```
vsonims, line 402: How do we generate cert chains without pub key. I want to make sure all content generated by ROM that is critical for the Caliptra IP to function correctly without depending on DCCM. This change violates this design principle

vsonims (changes requested):

bluegate010:
Closing as not necessarily needed for now.

PR #479, opened 2023-07-15
Do not write FMC or RT load addresses to DV slots.

bluegate010:
These are unnecessary and the slots could be used for other purposes.


common/src/hand_off.rs
```diff
@@ -451,7 +442,7 @@ pub fn report_handoff_error_and_halt(msg: &str, code: u32) -> ! {
 mod tests {
     use super::*;
     use core::mem;
-    const FHT_SIZE: usize = 512;
+    const FHT_SIZE: usize = 508;
```
rusty1968, line 445: Can we keep a power of 2?

bluegate010: Done

rusty1968 (dismissed):

mhatrevi (approved):

korran (approved):

PR #497, opened 2023-07-17
[fix] Change DeobfuscationEngine.clear_secrets return type to void.

### LINKED ISSUE #486: [ROM] DeobfuscationEngine.clear_secrets method cannot fail; change return type.

mhatrevi:
Since DeobfuscationEngine.clear_secrets cannot fail, there isn't any value of having a return type of CalitpraResult on the clear_secrets API.

### DIFF
--- a/drivers/src/doe.rs
+++ b/drivers/src/doe.rs
@@ -79,7 +79,7 @@ impl DeobfuscationEngine {
     /// * Deobfuscation Key
     /// * Encrypted UDS
     /// * Encrypted Field entropy
-    pub fn clear_secrets(&mut self) -> CaliptraResult<()> {
+    pub fn clear_secrets(&mut self) {
         let doe = self.doe.regs_mut();
 
         // Wait for hardware ready
@@ -90,7 +90,5 @@ impl DeobfuscationEngine {
 
         // Wait for command to complete
         wait::until(|| doe.status().read().valid());
-
-        Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


drivers/src/doe.rs
```diff
@@ -79,7 +79,7 @@ impl DeobfuscationEngine {
     /// * Deobfuscation Key
     /// * Encrypted UDS
     /// * Encrypted Field entropy
-    pub fn clear_secrets(&mut self) -> CaliptraResult<()> {
+    pub fn clear_secrets(&mut self) {
```
jhand2, line 82: Should this be checking if the DOE status bit [DEOBF_SECRETS_CLEARED](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.doe_reg.DOE_STATUS#DEOBF_SECRETS_CLEARED.desc) ever gets set?

I'm not sure what it would imply if that bit doesn't get set, but it makes me nervous that the bit exists but we aren't checking it.

jhand2: Unless this is just best effort for FIPS zeroization, in which case it's probably fine.

nvpikhur: Best practice is to check instead of making it void. Since we actually have a bit in hardware that gets set if secrets were cleared. Assumption is that if there is fault injection this bit won't be set and can be caught by simple check in the ROM to act on it. e.g. reset or hang.

mhatrevi:
Closing this PR in lieu of a new upcoming PR with updated register programming sequences.

PR #504, opened 2023-07-17
[fix] Don't update FHT on a Warm Reset

### LINKED ISSUE #445: ROM warm reset flow not implemented
Because the warm reset flow skips FW loading and ldevid/fmcalias key derivation some fields of the FHT are not populated:
- manifest_load_addr
- ldevid_tbs_size
- fmcalias_tbs_size
Because FHT fields are not populated the FMC sees it as invalid and fails.

mhatrevi:


rom/dev/src/flow/warm_reset.rs
```diff
@@ -25,11 +25,12 @@ impl WarmResetFlow {
...
     /// * `env` - ROM Environment
     #[inline(never)]
-    pub fn run(env: &mut RomEnv) -> CaliptraResult<FirmwareHandoffTable> {
+    pub fn run(_env: &mut RomEnv) -> CaliptraResult<FirmwareHandoffTable> {
         cprintln!("[warm-reset] ++");
 
         cprintln!("[warm-reset] --");
 
-        Ok(fht::make_fht(env))
+        // Caller does not expect an FHT in the case of a warm reset.
```
korran, line 33: If this statement is true, perhaps this function should return CaliptraResult<Option<FirmwareHandoffTable>> rather than a bogus FHT?

rom/dev/src/main.rs
```diff
@@ -87,7 +87,10 @@ pub extern "C" fn rom_entry() -> ! {
     let result = flow::run(&mut env);
     match result {
         Ok(fht) => {
-            fht::store(fht);
+            // Leave the FHT untouched in case of a warm reset.
+            if reset_reason != ResetReason::WarmReset {
```
korran, line 91: I recommend:

Ok(Some(fht)) => {
  fht::store(fht);
}
Ok(None) => {
  // Do nothing; some flows (like WarmReset) don't generic an FHT
}
}

PR #505, opened 2023-07-18
ROM Integrity test

korran:
As usual, I recommend reviewing each commit independently.

### DIFF
--- a/drivers/src/sha256.rs
+++ b/drivers/src/sha256.rs
@@ -97,6 +97,31 @@ impl Sha256 {
         Ok(digest)
     }
 
+    /// Take a raw sha256 digest of 0 or more 64-byte blocks of memory. Unlike
+    /// digest(), the each word is passed to the sha256 peripheral without
+    /// byte-swapping to reverse the peripheral's big-endian words. This means the
+    /// hash will be measured with the byte-swapped value of each word.
+    ///
+    /// # Safety
+    ///
+    /// The caller is responsible for ensuring that the safety requirements of
+    /// [`core::ptr::read`] are valid for every value between `ptr.add(0)` and
+    /// `ptr.add(n_blocks - 1)`.
+    #[inline(always)]
+    pub unsafe fn digest_blocks_raw(
+        &mut self,
+        mut ptr: *const [u32; 16],
+        n_blocks: usize,
+    ) -> CaliptraResult<Array4x8> {
+        for i in 0..n_blocks {
+            self.sha256.regs_mut().block().write_ptr(ptr);
+            self.digest_op(i == 0)?;
+            ptr = ptr.wrapping_add(1);
+        }
+        self.digest_partial_block(&[], n_blocks == 0, n_blocks * 64)?;
+        Ok(Array4x8::read_from_reg(self.sha256.regs_mut().digest()))
+    }
+
     /// Zeroize the hardware registers.
     fn zeroize_internal(&mut self) {
         self.sha256.regs_mut().ctrl().write(|w| w.zeroize(true));
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,6 +191,8 @@ dependencies = [
  "elf",
  "hex",
  "once_cell",
+ "openssl",
+ "zerocopy",
 ]
 
 [[package]]
--- a/builder/Cargo.toml
+++ b/builder/Cargo.toml
@@ -18,6 +18,8 @@ clap.workspace = true
 elf.workspace = true
 hex.workspace = true
 once_cell.workspace = true
+openssl.workspace = true
+zerocopy.workspace = true
 
 [[bin]]
 name = "image"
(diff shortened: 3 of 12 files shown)


JohnTraverAmd (dismissed):

drivers/src/sha256.rs
```diff
@@ -215,9 +244,6 @@ impl Sha256 {
             sha256.ctrl().write(|w| w.mode(true).init(false).next(true));
         }
```
mhatrevi, line 245: Why is this removed?

mhatrevi (changes requested):

rusty1968:
@korran , is this still going into ROM? @JohnTraverAmd  made this a requirement for runtime so perhaps the entry point for invoking the test could have an environment parameter that is decoupled from ROM.

This is a pre-requisite for the self-test pr for runtime.

korran:
> @korran , is this still going into ROM? @JohnTraverAmd made this a requirement for runtime so perhaps the entry point for invoking the test could have an environment parameter that is decoupled from ROM.
> 
> This is a pre-requisite for the self-test pr for runtime.

Yes, this needs to go into the ROM. I assume the entry point for "run all the FIPS tests" will include running this test, which the runtime firmware can call on-demand.

There's no need to do an integrity test of the FMC or runtime firmware at boot because that's done as part of the ROM validating the image. I suppose there is still a need for re-validating the FMC or runtime firmware on request, but I don't think that is related to this ROM self-test; that logic should be running the full FMC / runtime verification routine that includes parsing and validating the image headers.

drivers/src/sha256.rs
mhatrevi: Since this change is unrelated to the ROM integrity testing, can we make this as a separate change?

mhatrevi: For a multi-block case, how do we ensure that the crypto engine is not processing the data in the block registers? digest_block simply copies the data to the block registers and then calls digest_op, which then checks for status.ready. Previously, digest_op would have only returned when the processing was complete.

rusty1968:
> > @korran , is this still going into ROM? @JohnTraverAmd made this a requirement for runtime so perhaps the entry point for invoking the test could have an environment parameter that is decoupled from ROM.
> > This is a pre-requisite for the self-test pr for runtime.
> 
> Yes, this needs to go into the ROM. I assume the entry point for "run all the ROM FIPS tests" will include running this test, which the runtime firmware can call on-demand.
> 
> There's no need to do an integrity test of the FMC or runtime firmware at boot because that's done as part of the ROM validating the image. I suppose there is still a need for re-validating the FMC or runtime firmware on request, but I don't think that is related to this ROM self-test; that logic should be running the full FMC / runtime verification routine that includes parsing and validating the image headers. Unfortunately because that requires doing the SHA checksum against ICCM, it won't be the exact same logic as the ROM uses (which uses the mailbox sha384acc hardware).

@JohnTraverAmd had specified that RT would invoke ROM integrity tests as part of command handling : i.e. linking the ROM integrity test code and invoking it in context of RT.

drivers/src/sha256.rs
korran: I asked Michael:

> Kor: I'm trying to stream data to the sha256 peripheral as fast as possible, do I need to wait for the ready bit before writing to the SHA256_BLOCK register, or do I only have to check that before writing the command? (edited) 
> 
> Michael: I believe init/next commands will latch the data in BLOCK. You can begin writing the next block while the command is running setting next again only after ready
> 
> Kor: ok, that's my interpretation of the RTL as well
> 
> Michael: That's how I do it for sha accelerator and any tests we've written. I only check ready prior to starting the next command

You can see it grabbing the entire register immediately after the command is received:

https://github.com/chipsalliance/caliptra-rtl/blob/main/src/sha256/rtl/sha256_w_mem.v#L209

vsonims: Is this really needed? Can we continue to use the old interface. How much does this improve the time by?

vsonims (dismissed):

drivers/src/sha256.rs
korran: digest_blocks_raw() is definitely required, as it is undefined behavior to create a slice reference to address zero. No way around that one.

Removing the wait for status::valid saves 45,483 cycles, due to the additional pipelining.

vsonims: Is ~45000 a big win that warrents such large changes this late?

korran: I honestly don't know, since we don't have much guidance on performance. This does delay how long it takes the ROM to be ready for the first measurements.

I did the optimization because I wanted to collect perf numbers with the optimal implementation when I wrote this 2 months ago.

If you're not comfortable, I can revert.

korran: Ok, in the interest of getting this PR in sooner, I've reverted the validity check optimization.

mhatrevi (approved):

PR #509, opened 2023-07-18
Generate a PR when new RTL Pointer gets updated and abort the releae process

fdamato:


.github/workflows/nightly-release.yml
```diff
@@ -39,11 +39,13 @@ jobs:
           while git tag | grep ${TAG_BASE}${INDEX}; do
               ((INDEX+=1))
           done
+          git submodule update --remote hw-latest/caliptra-rtl
+          CHECK_RELEASE_SYNC=$(git status -s | head -1)"
```
korran, line 43: git status -s is for humans, not scripts (and may change to always print something in the future). Use --porcelain instead

korran (dismissed):

.github/workflows/rtl-repo-sync.yml
```diff
@@ -3,6 +3,8 @@ name: RTL Repo Sync
 on:
   workflow_call:
   workflow_dispatch:
+permissions:
+  contents: write
```
korran, line 7: Does the PR creation logic work? I assume you need `pull-requests: write` here too, no?

attzonko (approved):

JohnTraverAmd (approved):

PR #513, opened 2023-07-19
[fix] Update RT SVN fuse to 128-bits

### LINKED ISSUE #481: RT SVN fuse is four words wide, but only two words are used by the driver
The FuseBank [runtime_svn()](https://github.com/chipsalliance/caliptra-sw/blob/26fb69901e49b2fe50fe1547e75558e4051edbff/drivers/src/fuse_bank.rs#L235) accessor is using two words, but there are [four](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.fuse_runtime_svn%5B0%5D) words in this fuse's array. Should the driver count the zeroes across all four words?

mhatrevi:


bluegate010 (dismissed):

drivers/src/fuse_bank.rs
```diff
@@ -234,9 +234,11 @@ impl FuseBank<'_> {
     ///
     pub fn runtime_svn(&self) -> u32 {
         let soc_ifc_regs = self.soc_ifc.regs();
-        64 - ((soc_ifc_regs.fuse_runtime_svn().at(1).read() as u64) << 32
-            | soc_ifc_regs.fuse_runtime_svn().at(0).read() as u64)
-            .leading_zeros()
+        let fuse: u128 = ((soc_ifc_regs.fuse_runtime_svn().at(3).read() as u128) << 96)
```
korran, line 237: Can we extract this logic into a function? And add unit tests?

korran: fuse_runtime_svn().read() will return a [u32; 4]

korran: A function that converts that to u128 would be good.

mhatrevi: Added.

bluegate010 (approved):

ajisaxena (approved):

PR #516, opened 2023-07-19
[fix] Workaround for ROM Warm Reset

mhatrevi:
Currently, RTL unlocks and clears the following warm reset registers: (NONSTICKY_DATA_VAULT_ENTRY/NonStickyLockableScratchReg/NonStickyGenericScratchReg) instead of only unlocking them. As a result, FMC cannot get the RT load address and offset on a Warm Reset. Following issue is filed to fix the RTL: https://github.com/chipsalliance/caliptra-rtl/issues/161 Once the RTL fix is in, this commit needs to be reverted.


jlmahowa-amd (approved):
Tested with the warm reset test I am working at and passes. Thanks for the fix Vishal.

vsonims (approved):

PR #517, opened 2023-07-19
Warm reset test

jlmahowa-amd:
HW-model support for warm reset.
Set ready_for_runtime in RT FW to check if boot is complete without parsing log.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -295,7 +295,10 @@ fn mbox_read_fifo(mbox: mbox::RegisterBlock<impl MmioMut>) -> Vec<u8> {
     result
 }
 
-fn mbox_write_fifo(mbox: &mbox::RegisterBlock<impl MmioMut>, buf: &[u8]) -> Result<(), ModelError> {
+pub fn mbox_write_fifo(
+    mbox: &mbox::RegisterBlock<impl MmioMut>,
+    buf: &[u8],
+) -> Result<(), ModelError> {
     const MAILBOX_SIZE: u32 = 128 * 1024;
 
     let Ok(input_len) = u32::try_from(buf.len()) else {
@@ -376,6 +379,14 @@ pub trait HwModel {
         Ok(hw)
     }
 
+    /// Trigger a warm reset and advance the boot
+    fn warm_reset_flow(&mut self, fuses: &Fuses) {
+        self.warm_reset();
+
+        self.init_fuses(fuses);
+        self.soc_ifc().cptra_bootfsm_go().write(|w| w.go(true));
+    }
+
     /// The APB bus from the SoC to Caliptra
     ///
     /// WARNING: Reading or writing to this bus may involve the Caliptra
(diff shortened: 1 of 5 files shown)


jlmahowa-amd:
@korran I am open to ideas to make warm_reset_flow and warm_reset more clear.

fdamato (approved):

hw-model/src/lib.rs
```diff
@@ -395,6 +406,11 @@ pub trait HwModel {
         }
     }
 
+    /// Toggle reset pins and wait for ready_for_fuses
+    fn warm_reset(&mut self) {
+        panic!("warm_reset unimplemented");
```
jhand2, line 411: Can we add a tracking Github issue to implement this on the sw emulator?

One thing that concerns me is we may not catch any regressions to code that interacts with warm reset, since the test automation only runs against sw-emulator.

hw-model/src/lib.rs
```diff
@@ -376,6 +379,14 @@ pub trait HwModel {
         Ok(hw)
     }
 
+    /// Trigger a warm reset and advance the boot
+    fn warm_reset_flow(&mut self) {
+        self.warm_reset();
+
+        self.soc_ifc().cptra_fuse_wr_done().write(|w| w.done(true));
```
jhand2, line 386: Will this keep all the fuse values that were set on cold boot?

test/tests/warm_reset_during_fw_load.rs
```diff
@@ -0,0 +1,79 @@
...
+use zerocopy::AsBytes;
+
+fn bytes_to_be_words_48(buf: &[u8; 48]) -> [u32; 12] {
+    let mut result: [u32; 12] = zerocopy::transmute!(*buf);
+    swap_word_bytes_inplace(&mut result);
+    result
+}
+
+#[test]
+fn warm_reset_during_fw_load() {
```
jhand2, line 19: nit: Would probably make sense to just put this test in `warm_reset.rs`. This would also make it easier to deduplicate some of this setup code.

runtime/src/main.rs
```diff
@@ -37,6 +37,14 @@ pub extern "C" fn entry_point() -> ! {
...
     if let Some(mut fht) = caliptra_common::FirmwareHandoffTable::try_load() {
         let mut drivers = unsafe { Drivers::new_from_registers(&mut fht) };
+
+        // Indicator to SOC that RT firmware is ready
+        drivers
+            .soc_ifc
+            .regs_mut()
+            .cptra_flow_status()
+            .write(|w| w.ready_for_runtime(true));
+
```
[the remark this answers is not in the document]
jlmahowa-amd: RDL description for that bit is not correct. Here is the issue tracking that: https://github.com/chipsalliance/caliptra-rtl/issues/160
From the integration spec this is described as: "Indicates that Caliptra FW is
ready for RT flows"

test/tests/warm_reset_during_fw_load.rs
jlmahowa-amd: Will do

runtime/src/main.rs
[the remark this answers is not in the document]
jhand2: Sounds good, thanks for the clarification.

jhand2:
I think there was a merge issue. The commit list contains a lot of commits that I think are already in main.

jhand2 (approved):

fdamato (approved):

Issue #186, opened 2023-04-20
Status Command Needs to have FIPS Related Information


JohnTraverAmd:
FIPS requires the following information:

- Module Name ASCII (Calitpra RTM)
- Version
- Mode of operation

This may require a new HW register or an expansion of the status command.

Details for this will be completed when we get FIPS consultation.

andreslagarcavilla:
Can you file the caliptra-hw issue as well. Thanks

JohnTraverAmd:
I can, however, this could possibly be done entirely in the Mailbox command.  I was stating that I'm not sure yet on this.

I'll file the HW issue and if we get to reject it then Yay!

JohnTraverAmd:
For Notes related to the requirement:

- The cryptographic module shall output the name or module identifier and the versioning information that can be correlated with a validation record (e.g. hardware, software, and/or firmware versioning information).
- Versioning information (AS04.13) shall be provided for 
  - the physical single-chip including any excluded functional subsystem firmware (shall be specified in the OE field of the validation)
  - the sub-chip cryptographic subsystem soft and hard circuitry cores, and 
  - the associated firmware.

JohnTraverAmd:
Because we are looking into having the ROM perform crypto functionality for the user, the ROM will have to perform this command.

JohnTraverAmd:
Return Values

- Module Name ASCII = (Calitpra RTM)
- FIPS Version (HW Version | ROM Version | FW Version)
  - HW Version comes from https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_HW_REV_ID
  - ROM Version comes from https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_FW_REV_ID%5B0%5D
  - FW Version comes from https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_FW_REV_ID%5B0%5D
- Mode of operation
  - Boolean FIPS Approved/Non-Approved (Defaulted to FIPS Approved)

rusty1968:
The ROM version information should be written to the first word of the array by ROM  and runtime should populate the second word. Any objections?

JohnTraverAmd:
I think that works.

PR #526, opened 2023-07-21
Runtime [feat] FIPS Version Command

rusty1968:
This commit implements FIPS Version command handling in the runtime module and adds fw-rev-id and hw-rev-id registers to the emulated model (sw-emulator).

### DIFF
--- a/libcaliptra/src/caliptra_api.c
+++ b/libcaliptra/src/caliptra_api.c
@@ -17,7 +17,7 @@ static inline uint32_t caliptra_read_status(void)
     uint32_t status;
 
     caliptra_read_u32(CALIPTRA_TOP_REG_GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS, &status);
-   
+
     return status;
 }
 
@@ -34,14 +34,13 @@ bool caliptra_ready_for_fuses(void)
 
     status = caliptra_read_status();
 
-    if ((status & CALIPTRA_TOP_REG_GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS)
-                == GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS_READY_FOR_FUSES_MASK)
+    if ((status & CALIPTRA_TOP_REG_GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS) == GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS_READY_FOR_FUSES_MASK)
     {
         return true;
     }
 
     return false;
- }
+}
 
 /**
  * caliptra_init_fuses
@@ -55,16 +54,17 @@ bool caliptra_ready_for_fuses(void)
 int caliptra_init_fuses(struct caliptra_fuses *fuses)
 {
     // Parameter check
-    if (!fuses) {
+    if (!fuses)
+    {
         return -EINVAL;
     }
 
     // Check whether caliptra is ready for fuses
-    // 
+    //
     // This check is disabled until the below ticket is resolved.
     // https://github.com/chipsalliance/caliptra-sw/issues/508
     //
-    //if (!caliptra_ready_for_fuses())
+    // if (!caliptra_ready_for_fuses())
     //    return -EPERM;
 
     // Write Fuses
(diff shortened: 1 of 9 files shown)


runtime/src/fips.rs
```diff
@@ -1,18 +1,61 @@
...
+    pub name: [u8; 12],
+    pub fips_rev: [u32; 3],
+    pub mode: u32,
+}
+
+impl VersionResponse {
+    pub const NAME: &str = "Caliptra RTM";
+
+    pub fn new() -> Self {
+        //let hw_rev: u32 = env.soc_ifc.regs().cptra_hw_rev_id().read();
```
rusty1968, line 26: Accesses to the hardware and firmware revision ID are hanging the emulator.

rusty1968: Fxed : they were not implemented.

runtime/tests/integration_tests.rs
```diff
@@ -191,12 +191,37 @@ fn test_fips_cmd_api() {
...
+    pub const NAME: &str = "Caliptra RTM";
+
+    // Check command size
+    let fips_version: &[u8] = fips_version_resp.as_bytes();
+    assert_eq!(fips_version.len(), 28);
+
+    // Check module name
+    let name = &fips_version[16..28];
+    let result = name.cmp(NAME.as_bytes());
+    assert!(result.is_eq());
```
jhand2, line 224: Might be easier to do `let fips_version = VersionResponse::read_from(fips_version_resp.as_bytes());`. Then you can read individual fields.

rusty1968: Done!

runtime/src/fips.rs
```diff
@@ -1,18 +1,61 @@
...
+#[repr(C)]
+#[derive(Clone, Debug, Default, AsBytes, FromBytes)]
+pub struct VersionResponse {
+    pub mode: u32,
+    pub fips_rev: [u32; 3],
+    pub name: [u8; 12],
+}
+
+impl VersionResponse {
+    pub const NAME: &str = "Caliptra RTM";
```
jhand2, line 22: You should be able to do

```rust
pub const NAME: [u8; 12] = *b"Caliptra RTM";
```

And then you should be able to avoid the copy and can just return

```rust
Self{
    ...
    name: NAME,
    ...
}
```

runtime/tests/integration_tests.rs
```diff
@@ -191,12 +191,40 @@ fn test_fips_cmd_api() {
...
+        .cptra_fw_rev_id()
+        .at(0)
+        .write(|_| b'R'.into());
+
+    model
+        .soc_ifc()
+        .cptra_fw_rev_id()
+        .at(1)
+        .write(|_| b'A'.into());
+
```
jhand2, line 205: The REV_ID registers should be read-only from the SoC side, so this won't work on FPGA/verilator.

runtime/tests/integration_tests.rs
```diff
@@ -191,12 +191,40 @@ fn test_fips_cmd_api() {
...
+        .mailbox_execute(u32::from(CommandId::VERSION), &cmd)
+        .unwrap()
+        .unwrap();
+
+    // Check command size
+    let fips_version_bytes: &[u8] = fips_version_resp.as_bytes();
+    assert_eq!(
+        fips_version_bytes.len(),
+        core::mem::size_of::<VersionResponse>()
+    );
```
jhand2, line 220: nit: `VersionResponse::read_from` will do this size check for you.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -438,6 +444,12 @@ struct SocRegistersImpl {
     #[register_array(offset = 0x00c8, write_fn = on_write_generic_output_wires)]
     cptra_generic_output_wires: [u32; CPTRA_GENERIC_OUTPUT_WIRES_SIZE / 4],
 
+    #[register(offset = 0x00d0)]
+    cptra_hw_rev_id: u32,
+
+    #[register_array(offset = 0x00d4)]
+    cptra_fw_rev_id: [u32; CPTRA_FW_REV_ID_SIZE / 4],
```
jhand2, line 451: As mentioned above, we should make these read-only for the SoC

jhand2 (changes requested):

sw-emulator/lib/periph/src/soc_reg.rs
rusty1968: I am not sure if we have that kind of support on the sw-emulator.

runtime/tests/integration_tests.rs
rusty1968: I will add conditional compilator for emulated model only then,

sw-emulator/lib/periph/src/soc_reg.rs
jhand2: Huh, you're right. We should probably add that at some point. But we can skip it for this PR.

runtime/tests/integration_tests.rs
jhand2: I would recommend just checking that these values are 0 in this PR. Then once we implement the version registers, we can update the test to check them.

runtime/src/fips.rs
```diff
@@ -1,18 +1,61 @@
...
+}
+
+impl VersionResponse {
+    pub const NAME: [u8; 12] = *b"Caliptra RTM";
+    pub const MODE: u32 = 0x46495053;
+
+    pub fn new(env: &Drivers) -> Self {
+        let hw_rev: u32 = env.soc_ifc.regs().cptra_hw_rev_id().read();
+        let fw_rev = env.soc_ifc.regs().cptra_fw_rev_id().read();
+        let mut name: [u8; Self::NAME.len()] = [0_u8; Self::NAME.len()];
```
jhand2, line 28: Can remove this

jhand2 (approved):

jhand2 (dismissed):

rusty1968:
https://github.com/chipsalliance/caliptra-sw/issues/186

libcaliptra/src/caliptra_api.c
```diff
@@ -127,6 +127,8 @@ static int caliptra_mailbox_write_fifo(struct caliptra_buffer *buffer)
 
     // Write DLEN
     caliptra_mbox_write_dlen(buffer->len);
+    if (buffer->data == NULL)
```
wmaroneAMD, line 130: Should this be up a few lines and return non-zero?

wmaroneAMD (approved):
Looks good.

jhand2 (dismissed):

libcaliptra/src/caliptra_api.c
rusty1968: This covers the case for a request that does not have a payload.

fdamato (dismissed):

jhand2 (approved):

wmaroneAMD (approved):

jhand2 (approved):

PR #533, opened 2023-07-23
Allow test-fmc to be interactive.

bluegate010:
Future tests will send a series of host commands.


rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -195,32 +195,44 @@ fn get_pcr_entry(entry_index: usize) -> PcrLogEntry {
...
+
+    #[cfg(feature = "interactive_test_fmc")]
+    loop {
+        if mbox.status().read().mbox_fsm_ps().mbox_execute_uc() {
+            process_mailbox_command(&mbox);
+        }
+    }
+
+    #[cfg(not(feature = "interactive_test_fmc"))]
+    process_mailbox_command(&mbox);
```
jhand2, line 233: Is this necessary? It should be fine to just always use the interactive version since we aren't booting an RT image anyway.

The test can just run 1 or more mailbox commands and then exit.

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -195,32 +195,44 @@ fn get_pcr_entry(entry_index: usize) -> PcrLogEntry {
...
         }
         0x1000_0004 => {
-            trigger_update_reset(&mbox);
+            trigger_update_reset(mbox);
         }
         _ => {}
     }
 }
 
+fn process_mailbox_commands() {
```
jhand2, line 221: Another option would be to run these tests from RT. Not sure if the tests you need to run have to be done in FMC or not.

bluegate010: Yes, that's an option, would need to add handoff code to this FMC, since it doesn't know how to boot an RT image. I think the idea behind this test-fmc was to just be a minimal image, and while we package RT FW with it in tests, I don't think that RT fw ever actually runs.

But yeah we could just take all the host commands in this image and move them out to RT FW.

rom/dev/tools/test-fmc/src/main.rs
bluegate010: I initially tried this but then had to touch several other tests that expect test-fmc to run and quit without interaction.

jhand2: Ya, other tests would need to be changed, just think it ends up with a cleaner final product. But I don't feel too strongly about this.

rom/dev/tools/test-fmc/src/main.rs
jhand2: The RT tests use the regular FMC instead of the test FMC. Would that be a problem?

I'm not super convinced test-fmc needs to exist at all now that real FMC exists (but would be fine leaving it for the tests that already depend on it).

rom/dev/tools/test-fmc/src/main.rs
bluegate010: Not just the integration tests, there's also a GitHub presubmit that invokes ROM makefile's `run` target, which assembles test-fmc with test-rt and doesn't provide any handy way to invoke a host command to get the image to actually exit.

rom/dev/tools/test-fmc/src/main.rs
bluegate010: Yeah I think test-fmc and test-rt could go away.

However, that migration looks fairly involved. Mind if we just extend test-fmc in the short term to unblock test coverage for new PCR functionality?

mhatrevi (approved):

PR #534, opened 2023-07-24
Adding FIPS status and standardizing mailbox responses

nquarton:
WIP - testing and documentation incomplete:Wq
- Adds FIPS status field to all mailbox responses
- Adds chksum to all mailbox requests and respones
- Adds mailbox_api.rs for single-point-of-truth for all mailbox message IDs and formats
- Standardizes above, common mailbox behaviors to packet.rs


runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,147 @@
...
+        value.0
+    }
+}
+
+// Helpers
+// Assumes given byte array is >= size_of T
+// Will return CaliptraError::RUNTIME_INTERNAL otherwise
+pub fn cast_bytes_to_struct<T: AsBytes + FromBytes>(bytes: &[u8]) -> CaliptraResult<&T> {
+    Ok(
+        LayoutVerified::<&[u8], T>::new(&bytes[..core::mem::size_of::<T>()])
```
rusty1968, line 44: can't just use zerocopy api?

nquarton: This is a good question, I am hoping there is a better way to do this. So LayoutVerified is part of the zerocopy crate. But most of what I think you would call the actual API for the crate seems to only support creating _copies_ from struct to buffer or vice-versa instead of working on the same buffer in memory.

So this LayoutVerified::new was the only way I found to handle that. Then it was a bit messy so I just wrapped it in this function.

nquarton: One small correction on my previous comment: It is supported to get a reference to a byte array _from_ a struct using as_bytes() or as_bytes_mut(). It's just the other direction I could not find a cleaner solution for in zerocopy or elsewhere.

runtime/src/lib.rs
```diff
@@ -173,12 +113,47 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+
+    cprintln!(
+        "[rt] Received command=0x{:x}, len={}",
+        req_packet.cmd,
+        req_packet.len
+    );
+
+    // Create a response packet
+    // TODO: OPEN: How big is our stack? We are throwing 8k on it between these 2 packets
+    let mut resp_packet = Packet::default();
```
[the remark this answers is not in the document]
rusty1968: Do u mean copying responses directly to mailbox data memory bypassing the register interface?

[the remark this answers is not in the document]
jhand2: No, I mean taking a response struct, calling as_bytes() and writing it to the mailbox instead of first copying it to a Packet structure.

runtime/src/mailbox_api.rs
nquarton: No problem, I can get rid of this

runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+pub const FIPS_STATUS_APPROVED: u32 = 0;
+
+// CALIPTRA_FW_LOAD
+#[repr(C)]
+#[derive(AsBytes, FromBytes)]
+pub struct CaliptraFwLoadReq {
+    pub common: MailboxReqCommon,
+    pub data: [u8; 0], // variable length
```
jhand2, line 78: I would recommend just choosing a static value (probably 1024). Especially since these are just test commands.

Declaring variable arrays like this doesn't work in rust, since it would be unsafe.

runtime/src/lib.rs
```diff
@@ -115,56 +78,33 @@ fn wait_for_cmd(_mbox: &mut Mailbox) {
     //}
 }
 
-fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
```
jhand2, line 118: Not sure I understand why this function needed to be split. Will probably make reviewing easier to not change this for now.

runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct MailboxReqCommon {
+    pub chksum: i32,
+}
+
+#[repr(C)]
+#[derive(AsBytes, FromBytes)]
+pub struct MailboxRespCommon {
+    pub chksum: i32,
+    pub fips_status: u32,
+}
```
jhand2, line 69: Not sure it's necessary to have common structs for this few fields. But if we keep them, I would recommend naming  it `Mailbox{Req, Resp}Header`

runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,147 @@
...
+
+pub const FIPS_STATUS_APPROVED: u32 = 0;
```
jhand2, line 71: nit: To make it more clear what this is for, you could do

```rust
impl MailboxRespCommon {
  pub const FIPS_STATUS_APPROVED: u32 = 0;
}
```

Then refer to it with `MailboxRespCommon::FIPS_STATUS_APPROVED`

runtime/src/mailbox_api.rs
nquarton: Thanks, this was something I was meaning to ask about. Should 1024 be safe for all the messages with variable length fields? (CALIPTRA_FW_LOAD request, GET_IDEV_CSR response, GET_LDEV_CERT response, INVOKE_DPE_COMMAND request and response)

runtime/src/lib.rs
nquarton: That's fine I can re-combine this

runtime/src/mailbox_api.rs
nquarton: I went back and forth on this. The reason I did it was so we can grab the common/header fields on a message regardless of its type. Like when we are adding or verifying the checksum

runtime/src/mailbox_api.rs
nquarton: This is definitely how I should have done it, thanks!

runtime/src/mailbox_api.rs
jhand2: `CALIPTRA_FW_LOAD` shouldn't be copied from the mailbox at all. We don't have enough stack to hold the full firmware image. On reset, ROM will read the data from the mailbox into ICCM.

For GET_IDEV_CSR response and GET_LDEV_CERT response, 1K should be fine.

INVOKE_DPE_COMMAND, we're still working it out. I think the response buffer may need to be 4KiB. But we can figure it out later when that command is added.

runtime/src/mailbox_api.rs
jhand2: Ok, that's fine to leave this then

runtime/src/lib.rs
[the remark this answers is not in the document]
rusty1968: perhaps add some scratch buffers in the environment (Drivers) with static life time.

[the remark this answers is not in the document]
nquarton: That sounds like it could be a good long-term solution. I'm open to working on that but I may need a little hand-holding to do it right. For now, Jordan pointed me to a similar example using an enum of structs. It doesn't fully solve the stack allocation issue, but it at least limits it to the smallest size possible. I'll push a mostly complete example of that soon.

runtime/src/mailbox_api.rs
nquarton: That makes more sense, I was a little unclear on that for a while.

Sounds good for the rest.

runtime/src/dice.rs
```diff
@@ -52,3 +53,31 @@ fn cert_from_dccm(dv: &DataVault, cert: &mut [u8], cert_type: CertType) -> Calip
...
+/// Returns a TestGetCertResp and optionally the response size
+pub fn handle_get_fmc_alias_cert(dv: &DataVault) -> CaliptraResult<(TestGetCertResp, Option<usize>)> {
+    let mut resp = TestGetCertResp {
+        hdr: MailboxRespHeader::default(),
+        data: [0u8; 1024],
+    };
+
+    let cert_size = copy_fmc_alias_cert(dv, &mut resp.data)?;
+
+    Ok((resp, Some(cert_size + core::mem::size_of::<MailboxRespHeader>())))
```
jhand2, line 82: A couple small comments:

1. Could add size as a field to TestGetCertResp
2. Don't think the size needs to be optional

runtime/src/lib.rs
```diff
@@ -115,7 +75,7 @@ fn wait_for_cmd(_mbox: &mut Mailbox) {
...
+fn handle_and_respond(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
```
jhand2, line 78: Maybe add a comment to this function the explain what handle_and_respond means. Maybe something like

```rust
/// Handles any pending mailbox command and writes response to the mailbox
```

rusty1968:
time to rebase :-)

runtime/src/lib.rs
```diff
@@ -126,59 +86,73 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+        req_packet.cmd,
+        req_packet.len
     );
 
-    match CommandId::from(packet.cmd) {
-        // FIRMWARE_LOAD expected to already be handled
-        CommandId::FIRMWARE_LOAD => Err(CaliptraError::RUNTIME_UNEXPECTED_UPDATE_RETURN),
+    // Some responses have a variable size
+    // Leave as None otherwise
+    let resp_size_ovrd: Option<usize> = None;
```
jhand2, line 100: Based on comment above, if the response contains the size, then the whole response can just be copied to the mailbox (at least for now)

nquarton:
> time to rebase :-)

Ha, yeah I will! Figured you guys can at least give me opinions on how this approach looks while I rebase, update testing, and clean up

runtime/src/lib.rs
```diff
@@ -126,59 +86,73 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
-        CommandId::SHUTDOWN => FipsModule::shutdown(drivers),
         _ => Err(CaliptraError::RUNTIME_UNIMPLEMENTED_COMMAND),
-    }
+    }?;
+
+    // If there was no reponse payload from the handler, still send back a header
+    let mut resp = match resp {
+        Some(resp) => resp,
+        None => MailboxResp::Header(MailboxRespHeader::default()),
+    };
```
jhand2, line 142: Could just return `MailboxResp::Header(MailboxRespHeader::default())` from commands that only return the header. That way the code is easier to follow.

jhand2 (commented):
Thanks for reorganizing this! It looks a lot cleaner to have all the cmds/responses in the mailbox_api file.

runtime/src/dice.rs
nquarton: Thanks, adding this size field let me clean things up a lot.

runtime/src/lib.rs
nquarton: Changed this behavior and made a note in the code for now. Would be pretty easy go back to just sending what's needed later with the new size field in the response messages

runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub data: [u8; GetLdevCertResp::DATA_MAX_SIZE], // variable length
+}
+impl GetLdevCertResp {
+    pub const DATA_MAX_SIZE: usize = 1024;
+}
+
+// ECDSA384_SIGNATURE_VERIFY
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct EcdsaVerifyCmdReq {
```
jhand2, line 158: nit: Can just name this EcdsaVerifyReq

runtime/src/mailbox_api.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub hdr: MailboxReqHeader,
+    pub pub_key_x: [u8; 48],
+    pub pub_key_y: [u8; 48],
+    pub signature_r: [u8; 48],
+    pub signature_s: [u8; 48],
+}
+// No command-specific output args
```
jhand2, line 165: Shouldn't this return a FIPS status?

runtime/src/mailbox_api.rs
nquarton: No problem

runtime/src/mailbox_api.rs
nquarton: This comment may need to change. What I was trying to convey is that it has not additional fields specific to the command. I could either
1. Make the comment clearer, like "Resp header is the only output" or any other suggestion
2. (I don't think we want to do this) create req and resp structs for these cases that only contain the header

jhand2: Oh right, I forgot we just return the header. For consistency, maybe we should have request/response structures for every command, even if they only field is the header. wdyt?

nquarton: Yeah that would be fine.  The minor downside to that is then we do need to actually put an entry in the resp enum (and in as_bytes) for each type instead of just using generic "header" in there if we want to actually use them in the runtime code.

Or we can add them for clarity but continue to use the generic "header" in the enum, but that may be more confusing.

Just a tradeoff of consistency for a little redundancy. Whichever you prefer is good with me.

jhand2: Hmm. Ok, go ahead and leave it as you have it for now.

runtime/src/fips.rs
```diff
@@ -5,55 +5,39 @@ use caliptra_drivers::CaliptraError;
...
-    pub fips_rev: [u32; 3],
-    pub name: [u8; 12],
-}
-
-impl VersionResponse {
-    pub const NAME: [u8; 12] = *b"Caliptra RTM";
-    pub const MODE: u32 = 0x46495053;
-
-    pub fn new(_env: &Drivers) -> Self {
+impl FipsVersionResp {
```
rusty1968, line 11: Consider placing struct and impl in the same file.

nquarton: I can change this. I was on the fence about which approach made the most sense here. This was an attempt to separate the message formats and implementation. But I can see how this may be less intuitive.

jhand2: I don't feel super strongly about this. But I think in the future a nice pattern here could be

```rust
// <mailbox_cmd_impl>.rs
struct Cmd {
   ...
}

impl Cmd {
    fn execute(&self) -> CaliptraResult<Resp> {}
}
```

Then you can call it like this

```rust
let cmd = Cmd{};
cmd.execute();
```

So that would probably mean putting each command struct in the corresponding implementation file. But again, not a big deal either way.

rusty1968: +1 split

runtime/src/verify.rs
```diff
@@ -1,25 +1,15 @@
...
-pub(crate) fn handle_ecdsa_verify(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<()> {
-    if let Some(cmd) = EcdsaVerifyCmd::read_from(cmd_args) {
-        if !caliptra_common::checksum::verify_checksum(
-            cmd.chksum,
-            CommandId::ECDSA384_VERIFY.into(),
-            &cmd_args[PAYLOAD_OFFSET..],
-        ) {
-            return Err(CaliptraError::RUNTIME_INVALID_CHECKSUM);
-        }
+pub(crate) fn handle_ecdsa_verify(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
```
rusty1968, line 11: The handler has  low cohesion with  this file.

runtime/src/dice.rs
```diff
@@ -52,3 +53,33 @@ fn cert_from_dccm(dv: &DataVault, cert: &mut [u8], cert_type: CertType) -> Calip
 
     Ok(size)
 }
+
```
rusty1968, line 56: low cohesion

nquarton:
Let me know if this latest change addresses the most recent feedback at all. I've left the resp and req structs in the common file for now, but I can separate them all out if we want. That will cause some circular dependencies with the resp enum though.

runtime/src/fips.rs
nquarton: I just pushed a patch that I think is what you were talking about here. I followed the existing example in the info implementation.

runtime/src/verify.rs
nquarton: Let me know if the new format addresses this at all. I didn't catch you on IM before I made this most recent change.

runtime/src/dice.rs
nquarton: Let me know if the new format addresses this at all. I didn't catch you on IM before I made this most recent change.

jhand2 (approved):

runtime/src/fips.rs
jhand2: Let's leave it how it is now. I have some other ideas for refactoring, but this PR is already very big. I think it would be good to get these changes in as-is.

jhand2:
I'll merge once:

* CI failures are fixed
* @rusty1968 approves

jhand2 (approved):

runtime/src/dice.rs
rusty1968: It can be a follow up PR. Note you had to import several definitions - the mailbox API can reside in a different module rather than being included in several places. Handlers can be all in a single module.

nquarton: Ah okay I get what you are saying. I do like the idea of consolidating it. Maybe the three of us should do a quick sync on follow up changes this week. Stuff like that will be much easier to tweak since it won't really affect testing or anything as much compared to this change.

rusty1968 (dismissed):

wmaroneAMD (approved):

jhand2 (approved):

wmaroneAMD (approved):

PR #536, opened 2023-07-24
[fix] Set the ready_for_fuses bit in the CPTRA_FLOW_STATUS register in the Emulator

### LINKED ISSUE #508: sw-emulator does not properly set fuse ready bit in status
The sw-emulator does not properly set GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS_READY_FOR_FUSES_MASK in CALIPTRA_TOP_REG_GENERIC_AND_FUSE_REG_CPTRA_FLOW_STATUS. This was confirmed to be an emulator issue by enabling verilator and polling the same bit, which happens immediately.

This test was done using the libcaliptra hwmodel example (currently in PR), which defaults to the sw-emulator. Rebuilding the model and binding with --features verilator and rebuilding the example (and enabling the currently disabled initial poll of the fuse bit) we poll the bit and immediately pass the check. In the software simulator it fails.

Of note, the following line appears on the verilated output:

`4,100 ready_for_fuses is high`

This line is never seen in the sw-simulator, I do not know if it should be.

mhatrevi:
This fix contains the following:

1. Set the ready_for_fuses bit in the CPTRA_FLOW_STATUS register when Emulator starts.
2. Reset the ready_for_fuses bit when the 'done' bit in CPTRA_FUSE_WR_DONE register is set.

### DIFF
--- a/sw-emulator/lib/periph/src/soc_reg.rs
+++ b/sw-emulator/lib/periph/src/soc_reg.rs
@@ -588,6 +588,9 @@ impl SocRegistersImpl {
         iccm: Iccm,
         mut args: CaliptraRootBusArgs,
     ) -> Self {
+        let flow_status = InMemoryRegister::<u32, FlowStatus::Register>::new(0);
+        flow_status.write(FlowStatus::READY_FOR_FUSES.val(1));
+
         let regs = Self {
             cptra_hw_error_fatal: ReadWriteRegister::new(0),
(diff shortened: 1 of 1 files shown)


jhand2 (approved):

wmaroneAMD (approved):

wmaroneAMD:
Will follow with changes in the C API to reflect this.

ajisaxena (approved):

PR #546, opened 2023-07-26
runtime/test: Implement wycheproof vectors test

ArthurHeymans:
Use the wycheproof crate to run test vectors. The design of the mailbox (fixed length input for key and signature) limits the amount of tests that can meaningfully be ran on in the mailbox so those tests are skipped.

The openssl crate has some troubles with parsing invalid DER. Those tests are skipped for now with a whitelist.
This has no impact as an invalid DER signature cannot be parsed into valid R, S ranges that the mailbox accepts.


runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
-                0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
-            ],
-        },
-    ];
-
-    for c in cases {
-        // Stream to SHA ACC
-        model
-            .compute_sha512_acc_digest(c.msg, ShaAccMode::Sha384Stream)
+    #[allow(dead_code)]
```
jhand2, line 23: Not clear to me why this is dead code

runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
-            pub_key_y: c.pub_y,
-            signature_r: c.sig_r,
-            signature_s: c.sig_s,
-        };
+            fn report_test(err: &mut Vec<WycheproofResults>, id: usize, comment: &str) {
+                err.push(WycheproofResults {
+                    id,
+                    comment: comment.to_owned(),
+                });
+            }
```
jhand2, line 76: This functions replaces one line of code with one line of code. Can probably just call this inline where needed.

runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
+                &cmd.as_bytes()[4..],
+            );
+            let cmd = EcdsaVerifyCmd {
+                chksum: checksum,
+                ..cmd
+            };
+            match model.mailbox_execute(u32::from(CommandId::ECDSA384_VERIFY), cmd.as_bytes()) {
+                Ok(resp) => match test.result {
+                    wycheproof::TestResult::Valid | wycheproof::TestResult::Acceptable => {
+                        if resp.is_some() || model.soc_ifc().cptra_fw_error_non_fatal().read() != 0
```
jhand2, line 114: If model.soc_ifc().cptra_fw_error_non_fatal().read() != 0, `mailbox_execute` will return an Err. So it's fine to not check this.

runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
+            let checksum = caliptra_common::checksum::calc_checksum(
+                u32::from(CommandId::ECDSA384_VERIFY),
+                &cmd.as_bytes()[4..],
+            );
+            let cmd = EcdsaVerifyCmd {
+                chksum: checksum,
+                ..cmd
+            };
+            match model.mailbox_execute(u32::from(CommandId::ECDSA384_VERIFY), cmd.as_bytes()) {
+                Ok(resp) => match test.result {
```
jhand2, line 112: Instead of checking `if resp.is_some()`, you can match `Ok(Some(resp))`. if `Ok(None)` matches, we should probably panic since that is unexpected.

runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
+                    wycheproof::TestResult::Valid | wycheproof::TestResult::Acceptable => {
+                        if resp.is_some() || model.soc_ifc().cptra_fw_error_non_fatal().read() != 0
+                        {
+                            report_test(&mut wyche_err, test.tc_id, &test.comment);
+                            model.soc_ifc().cptra_fw_error_non_fatal().write(|_| 0);
```
jhand2, line 117: Discussed offline, this should be written by the runtime firmware.

But also I think the caller should only check the error register if the [mailbox status](https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.mbox_csr.mbox_status#status.desc) is `CMD_FAILURE`.

runtime/tests/ecdsa.rs
ArthurHeymans: It silences the following warning:

warning: fields `id` and `comment` are never read
  --> runtime/tests/ecdsa.rs:25:9
   |
24 |     struct WycheproofResults {
   |            ----------------- fields in this struct
25 |         id: usize,
   |         ^^
26 |         comment: String,
   |         ^^^^^^^
   |
   = note: `WycheproofResults` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` on by default

runtime/tests/ecdsa.rs
```diff
@@ -12,140 +11,138 @@ use zerocopy::AsBytes;
...
+                "appending unused 0's to sequence [r, s]",
+                "indefinite length",
+                "indefinite length with garbage",
+                "length of r uses long form encoding",
+                "length of r contains a leading 0",
+                "prepending 0's to r",
+                "length of s uses long form encoding",
+                "length of s contains a leading 0",
+                "prepending 0's to s",
+            ]
```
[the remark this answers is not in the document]
ArthurHeymans: With C string literals would be in the .rodata elf section and not pushed on the stack as variables. I would have expected &str to behave similarly in rust?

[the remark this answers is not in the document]
jhand2: Ya, that makes sense. Fine to leave this

runtime/tests/ecdsa.rs
jhand2: Ah
> note: WycheproofResults has a derived impl for the trait Debug, but this is intentionally ignored during dead code analysis".

That's weird. Ok, fine to leave this then.

jhand2 (approved):

ArthurHeymans:
Deleted in favor of https://github.com/chipsalliance/caliptra-sw/pull/579 so maintainers can make edits which is not possible from the 9e fork.

PR #548, opened 2023-07-26
Clear fw error register on successful mailbox command

### LINKED ISSUE #545: CPTRA_FW_ERROR_NON_FATAL not being cleared on mbox success
When a mailbox command fails, we set the error in CPTRA_FW_ERROR_NON_FATAL. But when a mailbox command succeeds, we don't properly clear it. So mailbox command failures are sticky.

jhand2:
Fixes #545

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -282,6 +282,9 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
             match handle_command(drivers) {
                 Ok(status) => {
                     drivers.mbox.set_status(status);
+
+                    // Clear any error code set by prior commands
+                    caliptra_drivers::report_fw_error_non_fatal(0);
                 }
                 Err(e) => {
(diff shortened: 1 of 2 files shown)


rusty1968 (approved):

fdamato (approved):

runtime/src/lib.rs
```diff
@@ -282,6 +282,9 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
             match handle_command(drivers) {
                 Ok(status) => {
                     drivers.mbox.set_status(status);
+
+                    // Clear any error code set by prior commands
+                    caliptra_drivers::report_fw_error_non_fatal(0);
```
korran, line 287: To avoid a potential race condition, this needs to be done before you call set_status.

jhand2:
This is fixed by https://github.com/chipsalliance/caliptra-sw/pull/667

PR #550, opened 2023-07-27
Update DPE spec

jhand2:
* Remove FIPS-mode derivation. This is not necessary given updated guidance on deterministic key derivation.
* Add list of supported DPE commands.
* Remove reliance on PCRs.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -322,6 +322,28 @@ details specific requirements for the Caliptra DPE implementation.
 | Supports Rotate Context    | Yes                            | Whether Caliptra supports the optional RotateContextHandle command.
 | CertifyKey Alias Key       | Caliptra Runtime Alias Key     | The key that will be used to sign certificates produced by the DPE CertifyKey command.
 
+### Supported DPE Commands
+
+Caliptra DPE supports the following commands
+
+* GetProfile
+* InitializeContext
+* DeriveChild
+* CertifyKey
+  * Caliptra DPE supports two formats for CertifyKey: X.509 and PKCS#10 CSR.
+    X.509 is only available to PL0 PAUSERs.
+* Sign
+* RotateContextHandle
+* DestroyContext
+* GetCertificateChain
+
+In addition, Caliptra supports the following profile-defined commands:
+
+* ExtendTci: Extend a TCI measurement made by DeriveChild to provide additional
+             measurement data.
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -334,38 +356,13 @@ Caliptra Runtime firmware is responsible for initializing DPE’s Default Contex
...
-These PCRs will store only the cumulative journey of the PCRs.
-
-#### PCR Properties
-
-* 48 bytes
-* Initial value of all zeros
-* Extend operation: `NEW_VALUE = SHA384_HASH(OLD_VALUE || MEASUREMENT)`
-* Clear operation: Resettable to initial value by Runtime Firmware
-
-#### DPE PCR Usage
```
pkwidzin-amd, line 354: There are some other references to PCR (eg. table below) that should be rephrase - can scrub the rest of spec for those?

runtime/README.md
```diff
@@ -381,69 +378,21 @@ DPE will first collect measurements and concatenate them in a byte buffer
...
-  * Derive an ECC P384 keypair from KV slot 0 CDI into the next available Runtime
-    KeyVault Slot
-  * If no slots are available, evict the oldest Runtime slot and use that slot.
-* For CertifyKey: Request the public key
-* For Sign: Sign passed data
-* Erase KeyVault slot 0
-
-### FIPS Mode Caching
-
-Caliptra has 32 KeyVault slots and a max of 24 DPE contexts. To accomodate
```
pkwidzin-amd, line 423: Is that part FIPS specific? I thought regardless of FIPS we still have those upper limits?

pkwidzin-amd: I guess this be derived implicitly from max TCI nodes so maybe add a statement in that section that max contexts == max TCI nodes

jhand2: In the deterministic (regular) mode, we can derive a key, use it, and then clear the slot. So number of hw KeyVault slots isn't a limitation.

runtime/README.md
jhand2: Removed (note there are still some PCR references which refer to PCRs that hold measurements of Caliptra firmware)

pkwidzin-amd: Thanks

rusty1968 (approved):

PR #554, opened 2023-07-28
[fuzzing] Implement initial support for Image Verifier fuzzing

benjamindoron:
First implementation of fuzz testing the Image Verifier library, forming part of the fuzzing work. Instructions and notes are in the readme.md files, these can be removed or migrated to actual documentation if required.

An open question: How or where do we want to run this? It didn't seem like it could be added to `cargo test`, because the fuzzers need to modify the compiler's command line arguments to add their instrumentation. Maybe fuzz testing should be another action in the CI.


korran:
> An open question: How or where do we want to run this? It didn't seem like it could be added to `cargo test`, because the fuzzers need to modify the compiler's command line arguments to add their instrumentation. Maybe fuzz testing should be another action in the CI.

For doing the actual fuzzing, let's setup a nightly CI workflow that runs for ~hours and tries its best to find whatever it can. Hopefully, eventually those of us with large compute farms (Google, Microsoft) can donate some cloud resources to help us fuzz more intensely.

It may also be reasonable to setup a presubmit CI workflow that makes sure everything builds with the fuzzer and can run a small number of iterations.

benjamindoron:
Will reopen PR from personal repo so maintainers can edit the PR branch.

PR #556, opened 2023-07-28
Add libcaliptra (SOC C Integration API Library) into hw release archieve

fdamato:


wmaroneAMD:
Do we expect the libcaliptra examples to build from within the release directory?

fdamato:
> Do we expect the libcaliptra examples to build from within the release directory?

Not necessarily

wmaroneAMD:
Then we may want to exclude the libcaliptra/examples/ subdirectory from the release as it doesn't have required inputs to do so.

fdamato:
I don`t think it is necessary...It will only create complexity IMHO

wmaroneAMD (approved):

attzonko (approved):

jlmahowa-amd (approved):

PR #559, opened 2023-07-29
[feat] Measurement command handling in ROM

mhatrevi:


ericeilertson (dismissed):

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -237,6 +242,13 @@ fn process_mailbox_commands() {
     process_mailbox_command(&mbox);
 }
 
+fn read_pcr1(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
+    let pcr_bank = unsafe { PcrBank::new(PvReg::new()) };
+    // Note: read_pcr returns PCR in big-endiean; convert to little-endian before sending to mailbox.
```
varuns-nvidia, line 247: nit: s/endiean/endian

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -385,4 +427,31 @@ impl FirmwareProcessor {
...
+            core::mem::size_of::<StashMeasurement>() / core::mem::size_of::<u32>();
+        let mut slice: [u32; MEASUREMENT_WORDS] = [0u32; MEASUREMENT_WORDS];
+        txn.copy_request(&mut slice)?;
+
+        let measurement = StashMeasurement::read_from(slice.as_bytes())
+            .ok_or(CaliptraError::FW_PROC_STASH_MEASUREMENT_READ_FAILURE)?;
+
+        // Extend measurement into PCR1.
+        pcr_bank.extend_pcr(PcrId::PcrId1, sha384, measurement.measurement.as_bytes())?;
+
```
varuns-nvidia, line 454: will the event log come in a separate PR?

varuns-nvidia (dismissed):

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -102,52 +114,82 @@ impl FirmwareProcessor {
...
-                    txn.start_txn().complete(false)?;
-                    continue;
+                match txn.cmd() {
+                    Self::MBOX_STASH_MEASUREMENT_CMD_ID => {
+                        if measurement_count == Self::MEASUREMENT_MAX_COUNT {
+                            cprintln!(
+                                "[afmc] Maximum supported number of measurements already received, ignoring."
+                            );
+                            txn.start_txn().complete(false)?;
+                            continue;
```
jhand2, line 153: Shouldn't we fail the mailbox command (by writing to the proper status registers) instead of ignoring it?

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -102,52 +114,82 @@ impl FirmwareProcessor {
...
+                        }
+
+                        // Re-borrow mailbox to work around https://github.com/rust-lang/rust/issues/54663
+                        let txn = mbox
+                            .peek_recv()
+                            .ok_or(CaliptraError::FW_PROC_MAILBOX_STATE_INCONSISTENT)?;
+
+                        // Extend the measurement into PCR1.
```
jhand2, line 161: Why PCR1? This is for extending measurements of SoC firmware into DPE. I don't think they should impact the PCR containing Caliptra's own firmware measurements.

Can we pick a different PCR to separate them from Caliptra RT measurements?

jhand2: I also think we should also pass these measurements up in the FHT so RT can use them to populate the TcbInfos in the cert that DPE generates.

jhand2 (changes requested):

rom/dev/src/flow/cold_reset/fw_processor.rs
vsonims: Will be part of this PR.

jhand2:
@mhatrevi Can we also add https://github.com/chipsalliance/caliptra-sw/issues/606? We can use the `reserved1` field in the `PcrLogEntry` to store it. No need to measure the SVN into the PCR if we don't want.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: Added.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: > Why PCR1? This is for extending measurements of SoC firmware into DPE. I don't think they should impact the PCR containing Caliptra's own firmware measurements.
> 
> Can we pick a different PCR to separate them from Caliptra RT measurements?

This is now being extended to PCR31.

mhatrevi: > I also think we should also pass these measurements up in the FHT so RT can use them to populate the TcbInfos in the cert that DPE generates.

Can you file a separate issue for this?

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: Completing the mailbox transaction with a value of 'false' writes a failure status (MboxStatusE::CmdFailure) to the mailbox status register. Are you asking for something additional to this?

rom/dev/src/flow/cold_reset/fw_processor.rs
jhand2: > This is now being extended to PCR31.

👍 

> Can you file a separate issue for this?

Actually, I think it should be fine for RT to parse these values from the PCR log.

rom/dev/src/flow/cold_reset/fw_processor.rs
jhand2: Ok, that should be fine.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -106,52 +121,81 @@ impl FirmwareProcessor {
...
+
+                        // Re-borrow mailbox to work around https://github.com/rust-lang/rust/issues/54663
+                        let txn = mbox
+                            .peek_recv()
+                            .ok_or(CaliptraError::FW_PROC_MAILBOX_STATE_INCONSISTENT)?;
+
+                        let mut txn = ManuallyDrop::new(txn.start_txn());
+                        Self::extend_measurement(pcr_bank, sha384, &mut txn)?;
+                        measurement_count += 1;
+                        txn.complete(true)?;
```
jhand2, line 171: We should be writing the response as well right? (See readme at https://github.com/chipsalliance/caliptra-sw/tree/main/runtime). dpe_status and fips_status can always return success I think.

common/src/measurement.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    Mailbox measurement related types.
+
+--*/
+use zerocopy::{AsBytes, FromBytes};
+
+/// Caliptra DPE Measurement    
+#[repr(C)]
+#[derive(AsBytes, FromBytes, Default, Debug)]
+pub struct StashMeasurement {
```
jhand2, line 18: See comment about adding an svn field. We can also do this in a follow-up PR if you'd prefer.

rom/dev/src/flow/cold_reset/fw_processor.rs
nquarton: Just a note here I am planning on adding the FIPS status to the response once this change is in. I'm not familiar with the DPE status but if it's simple I can probably handle that in the same PR.

Also the mailbox driver used in ROM currently does not support sending a response. I am working with Anthony on adding that separately right now.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -410,4 +461,105 @@ impl FirmwareProcessor {
...
+            return Err(CaliptraError::ROM_GLOBAL_PCR_LOG_UNSUPPORTED_DATA_LENGTH);
+        }
+
+        if (log_index * core::mem::size_of::<PcrLogEntry>()) > MEASUREMENT_LOG_SIZE {
+            return Err(CaliptraError::ROM_GLOBAL_MEASUREMENT_LOG_EXHAUSTED);
+        }
+
+        // Create a PCR log entry
+        let mut pcr_log_entry = PcrLogEntry {
+            id: PcrLogEntryId::StashMeasurement as u16,
```
varuns-nvidia, line 548: should the log entry also contain the metadata?

mhatrevi: I changed the reserved1 field in the PcrLogEntry to 'metadata'. The metadata from the measurement is stored in this field in the PCR log; it won't be extended to PCR31 though. Will this suffice?

common/src/measurement.rs
bluegate010: Would propose adding a [u32; 12] `svn_context` field and a u32 `svn` field. The context field can represent the hash of the key that bound `svn` to the code being measured.

If the code being measured has no SVN, then `svn_context` and `svn` can be all zeroes.

drivers/src/memory_layout.rs
```diff
@@ -32,15 +32,16 @@ pub const LDEVID_TBS_ORG: u32 = 0x50003800;
...
-pub const CFI_VAL_ORG: u32 = 0x50004BE8;
-pub const CFI_MASK_ORG: u32 = 0x50004BEC;
-pub const CFI_XO_S0_ORG: u32 = 0x50004BF0;
-pub const CFI_XO_S1_ORG: u32 = 0x50004BF4;
-pub const CFI_XO_S2_ORG: u32 = 0x50004BF8;
-pub const CFI_XO_S3_ORG: u32 = 0x50004BFC;
-pub const DATA_ORG: u32 = 0x50004C00;
+pub const MEASUREMENT_LOG_ORG: u32 = 0x50004800;
+pub const FUSE_LOG_ORG: u32 = 0x50004C00;
+pub const BOOT_STATUS_ORG: u32 = 0x50005000;
```
korran, line 37: (note for a future PR)

The runtime still needs to store DPE state in PersistentData. We should consider moving these non-persistent values far away (ideally before) the PersistentData struct.

Because these values are accessed out-of-band by unsafe code, they cannot be "in the middle" of PersistentData without violating the LLVM memory model. If we don't move them we'll be forced to split PersistentData into two separate structs.

drivers/src/measurement.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct StashMeasurement {
+    /// Checksum
+    pub checksum: u32,
+
+    /// Metadata
+    pub metadata: [u8; 4],
+
+    /// Measurement
+    pub measurement: [u32; 12],
+}
```
jhand2, line 29: This should rely on the structures in `mailbox_api.rs` to make sure we're using the same values: https://github.com/chipsalliance/caliptra-sw/blob/9cb9465c343da8c5c937d7269b32469644839735/common/src/mailbox_api.rs#L240

drivers/src/pcr_log.rs
```diff
@@ -71,7 +74,8 @@ pub struct PcrLogEntry {
     // PCR data
     pub pcr_data: [u32; 12],
 
-    pub reserved1: [u8; 4],
+    // PCR Metadata
+    pub metadata: [u8; 4],
```
jhand2, line 78: We want to put the SVN field from `StashMeasurementReq` here

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -211,7 +232,49 @@ impl FirmwareProcessor {
...
+                    CommandId::STASH_MEASUREMENT => {
+                        if measurement_count == MEASUREMENT_MAX_COUNT {
+                            cprintln!(
+                                "[afmc] Maximum supported number of measurements already received, ignoring."
+                            );
+                            txn.start_txn().complete(false)?;
+                            continue;
+                        }
+
+                        // Re-borrow mailbox to work around https://github.com/rust-lang/rust/issues/54663
```
korran, line 245: This isn't necessary; txn isn't long-lived here.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -211,7 +232,49 @@ impl FirmwareProcessor {
...
+                                ..Default::default()
+                            };
+                            Self::extend_measurement(
+                                pcr_bank,
+                                env.sha384,
+                                measurement_log,
+                                &fake_measurement,
+                                0_usize,
+                            )?;
+                        }
```
jhand2, line 276: I assume this is here to defend against STASH_MEASUREMENT commands being dropped by an attacker?

Can we unconditionally do this? It cleanly separates whether measurements were made during ROM or RT execution and makes it a bit easier for verifiers to interpret.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -502,4 +569,108 @@ impl FirmwareProcessor {
...
+        measurement_log: &mut StashMeasurementArray,
+        txn: &mut MailboxRecvTxn,
+        log_index: u32,
+    ) -> CaliptraResult<()> {
+        let mut slice: [u8; core::mem::size_of::<StashMeasurement>()] =
+            [0u8; core::mem::size_of::<StashMeasurement>()];
+        txn.copy_request(&mut slice)?;
+
+        let measurement = StashMeasurement::read_from(slice.as_bytes())
+            .ok_or(CaliptraError::FW_PROC_STASH_MEASUREMENT_READ_FAILURE)?;
```
korran, line 596: Failure is impossible here; slice is always the size of StashMeasurement.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -502,4 +569,108 @@ impl FirmwareProcessor {
...
+    /// * `()` - Ok
+    ///     Err - StashMeasurementReadFailure
+    fn stash_measurement(
+        pcr_bank: &mut PcrBank,
+        sha384: &mut Sha384,
+        measurement_log: &mut StashMeasurementArray,
+        txn: &mut MailboxRecvTxn,
+        log_index: u32,
+    ) -> CaliptraResult<()> {
+        let mut slice: [u8; core::mem::size_of::<StashMeasurement>()] =
```
korran, line 591: This is more complex than necessary. I suggest:

```rs
  let mut measurement = StashMeasurement::default();
  txn.copy_request(&mut measurement.as_bytes_mut())?;
```

It may also be worth checking that they sent you the correct number of bytes, which you're not currently doing:

```rs
        let mut measurement = StashMeasurement::default();
        if txn.dlen() as usize != measurement.as_bytes().len() {
            return Err(CaliptraError::FW_PROC_STASH_MEASUREMENT_READ_FAILURE);
        }
        txn.copy_request(&mut measurement.as_bytes_mut())?;
```

bluegate010:
Have just sent out https://github.com/chipsalliance/caliptra-sw/pull/748/.

drivers/src/pcr_log.rs
jhand2: Actually, thinking more about this. I think we want all of the additional fields from `StashMeasurementReq` in the log:

* metadata (4 bytes)
* svn (4 bytes)
* context (48 bytes)

If we need to expand the log that's ok. Also if you want us to do a follow up PR for the additional log items, that's fine.

drivers/src/persistent.rs
```diff
@@ -31,8 +35,11 @@ pub struct PersistentData {
     pub pcr_log: PcrLogArray,
     reserved3: [u8; 4],
 
+    pub measurement_log: StashMeasurementArray,
```
jhand2, line 38: We can just put this in the PcrLogArray, right?

drivers/src/memory_layout.rs
korran: Will resolve in a future PR.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: https://github.com/chipsalliance/caliptra-sw/pull/734 will address the ROM response .

drivers/src/measurement.rs
mhatrevi: Updated.

drivers/src/pcr_log.rs
mhatrevi: Let's do a follow-up PR.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: Updated.

drivers/src/persistent.rs
mhatrevi: In case of an update reset, since there are no measurements stashed, resulting in a PCR log that might have stale entries from cold reset flow. Additional logic will be needed to clear the cold-reset PCR log and then start a new one. To keep things simple, I have separated these logs.

jhand2: Sounds good

common/src/measurement.rs
jhand2: Opened https://github.com/chipsalliance/caliptra-sw/issues/759 to track. We can add in a follow-up PR.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: Updated

rom/dev/src/flow/cold_reset/fw_processor.rs
jhand2: PR mostly LGTM, only open for me is this comment

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -268,6 +289,26 @@ fn read_pcr_log(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
...
 }
 
+fn read_measurement_log(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
+    let mut measurement_entry_count = 0;
+    loop {
+        let measurement_entry = get_pcr_entry(measurement_entry_count, true);
+        measurement_entry_count += 1;
+        send_to_mailbox(mbox, measurement_entry.as_bytes(), false);
+
+        if measurement_entry_count == 8 {
```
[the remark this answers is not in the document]
mhatrevi: Fixed.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: This was discussed with Varun/VishalS of having the fake measurement unconditionally and the consensus was to be either/or. We can re-visit this and fix it as a bug fix.

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -218,6 +233,13 @@ fn process_mailbox_commands() {
     process_mailbox_command(&mbox);
 }
 
+fn read_pcr31(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
+    let pcr_bank = unsafe { PcrBank::new(PvReg::new()) };
+    // Note: read_pcr returns PCR in big-endian; convert to little-endian before sending it to mailbox.
+    let pcr1: [u8; 48] = pcr_bank.read_pcr(PCR_ID_STASH_MEASUREMENT).into();
```
bluegate010, line 239: pcr31?

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -218,6 +233,13 @@ fn process_mailbox_commands() {
     process_mailbox_command(&mbox);
 }
 
+fn read_pcr31(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
+    let pcr_bank = unsafe { PcrBank::new(PvReg::new()) };
+    // Note: read_pcr returns PCR in big-endian; convert to little-endian before sending it to mailbox.
```
bluegate010, line 238: It doesn't look like this comment is accurate? I don't see endian conversion here.

rom/dev/tools/test-fmc/src/main.rs
```diff
@@ -268,6 +291,26 @@ fn read_pcr_log(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
...
 }
 
+fn read_measurement_log(mbox: &caliptra_registers::mbox::RegisterBlock<RealMmioMut>) {
+    let mut measurement_entry_count = 0;
+    loop {
+        let measurement_entry = get_pcr_entry(measurement_entry_count, true);
+        measurement_entry_count += 1;
+        send_to_mailbox(mbox, measurement_entry.as_bytes(), false);
+
+        if measurement_entry_count == MEASUREMENT_MAX_COUNT {
```
bluegate010, line 301: The other log reading function tests whether `PcrLogEntryId::from(pcr_entry.id) == PcrLogEntryId::Invalid`. Any reason not to do the same here?

mhatrevi: The other log can contain up to 17 entries; currently only 11 PCRs are stored, so trying to read the 12th (invalid) entry does not result in an exception. This is not the case for the measurement log which can contain only 8 entries, hence the limit check.

rom/dev/tools/test-fmc/src/main.rs
mhatrevi: Fixed

rom/dev/tools/test-fmc/src/main.rs
mhatrevi: Array4X12 changes the endianess.

jhand2 (dismissed):

rom/dev/tools/test-fmc/src/main.rs
bluegate010: Ack. I think both routines should do both checks. A limit check to ensure we don't panic by reading past the end of the array, and an is-valid check to ensure we don't return more log entries than are actually populated (since as-is it looks like this will always return eight entries even if only a couple measurements were stashed).

bluegate010: But will leave this as optional future work since this is just test code.

korran (approved):

PR #562, opened 2023-07-31
Wycheproof: hmac

ArthurHeymans:
This implements:
- The mailbox command to verify HMAC
- running hmac sha384 tests over the HMAC mailbox

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -65,6 +65,7 @@ impl CommandId {
 
     pub const TEST_ONLY_GET_LDEV_CERT: Self = Self(0x4345524c); // "CERL"
     pub const TEST_ONLY_GET_FMC_ALIAS_CERT: Self = Self(0x43455246); // "CERF"
+    pub const TEST_ONLY_HMAC384_VERIFY: Self = Self(0x484D4143); // "HMAC"
 
     /// FIPS module commands.
     /// The status command.
@@ -89,6 +90,7 @@ pub struct Drivers<'a> {
     pub mbox: Mailbox,
     pub sha_acc: Sha512AccCsr,
     pub ecdsa: Ecc384,
+    pub hmac: Hmac384,
     pub data_vault: DataVault,
     pub soc_ifc: SocIfcReg,
     pub regions: MemoryRegions,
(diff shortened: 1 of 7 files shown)


runtime/tests/hmac.rs
```diff
@@ -0,0 +1,86 @@
...
+                comment: test.comment.to_string(),
+            });
+            let mut cmd = HmacVerifyCmd {
+                chksum: 0,
+                key: test.key[..].try_into().unwrap(),
+                tag: test.tag[..].try_into().unwrap(),
+                len: u32::try_from(test.msg.len()).unwrap(),
+                msg: [0u8; 256],
+            };
+            cmd.msg[..test.msg.len()].clone_from_slice(test.msg.as_slice());
```
jhand2, line 47: Just to confirm, I assume none of the messages are larger than 256 bytes? (Otherwise I the test would probably panic.

runtime/src/lib.rs
```diff
@@ -211,6 +236,10 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
             verify::handle_ecdsa_verify(drivers, cmd_bytes)?;
             Ok(MboxStatusE::CmdComplete)
         }
+        CommandId::HMAC384_VERIFY => {
```
jhand2, line 239: Can you mark this with ` #[cfg(feature = "test_only_commands")]`? I don't think we have any production use-cases for exposing this command.

Also rename the command ID to `TEST_ONLY_HMAC384_VERIFY` and group it with the other `TEST_ONLY` commands in the list of command ID constants

runtime/tests/integration_tests.rs
```diff
@@ -199,6 +201,74 @@ fn test_verify_cmd() {
     );
 }
 
+#[test]
+fn test_hmac_verify_cmd() {
```
jhand2, line 205: should this go in hmac.rs?

Or if you want could just omit it. I think running the wycheproof vectors is sufficient. I assume you added to initially test with some known data.

runtime/tests/hmac.rs
ArthurHeymans: I confirm

runtime/src/lib.rs
ArthurHeymans: Done

runtime/tests/integration_tests.rs
ArthurHeymans: Done

ArthurHeymans:
Removed in favor of https://github.com/chipsalliance/caliptra-sw/pull/578 as creating a PR from 9e organisation does not allow for maintainers to edit.

PR #569, opened 2023-07-31
Introducing the  "no-kats"  feature and the FAST_ROM target 

rusty1968:
Enable integrators to produce a test ROM that skips the FIPS KATS testing. Add asserts to the smoke test to ensure that KATs tests are not skipped in the production ROM target. Additionally , created a test to ensure the new "FAST ROM" target skips KAT tests.

### DIFF
--- a/common/src/boot_status.rs
+++ b/common/src/boot_status.rs
@@ -18,6 +18,7 @@ const FWPROCESSOR_BOOT_STATUS_BASE: u32 = 129;
 const FMCALIAS_BOOT_STATUS_BASE: u32 = 193;
 const COLD_RESET_BOOT_STATUS_BASE: u32 = 257;
 const UPDATE_RESET_BOOT_STATUS_BASE: u32 = 321;
+const ROM_GLOBAL_BOOT_STATUS_BASE: u32 = 385;
 
 /// Statuses used by ROM to log dice derivation progress.
 #[repr(u32)]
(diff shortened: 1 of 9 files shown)


nquarton (approved):

rom/dev/src/main.rs
```diff
@@ -77,10 +78,8 @@ pub extern "C" fn rom_entry() -> ! {
     // Start the watchdog timer
     wdt::start_wdt(&mut env.soc_ifc);
 
-    let result = kat::execute_kat(&mut env);
-    if let Err(err) = result {
-        handle_fatal_error(err.into());
-    }
+    #[cfg(not(feature = "no-kats"))]
+    execute_kats(&mut env);
```
korran, line 82: Rather than using multiple #[cfg...] attributes, consider this:


```rs
    if !cfg!(feature = "no-kats") {
      let result = kat::execute_kat(&mut env);
      if let Err(err) = result {
          handle_fatal_error(err.into());
      }
    }
```

rusty1968: Done

common/src/boot_status.rs
```diff
@@ -62,8 +62,9 @@ pub enum RomBootStatus {
     FmcAliasDerivationComplete = FMCALIAS_BOOT_STATUS_BASE + 5,
 
     // Cold Reset Statuses
-    ColdResetStarted = COLD_RESET_BOOT_STATUS_BASE,
-    ColdResetComplete = COLD_RESET_BOOT_STATUS_BASE + 1,
+    KatStarted = COLD_RESET_BOOT_STATUS_BASE,
```
mhatrevi, line 65: KATs are not specific to Cold Boot. Can you add the following global:
const ROM_GLOBAL_BOOT_STATUS_BASE: u32 = 385;

And in RomBootStatus num:
    // Rom Global Statuses
    KatStarted = ROM_GLOBAL_BOOT_STATUS_BASE,
    KatComplete = ROM_GLOBAL_BOOT_STATUS_BASE + 1,

mhatrevi (changes requested):

common/src/boot_status.rs
rusty1968: The enumeration is using a value greater than the ColdResetStarted.  As per current implementation, we are using monotonically increasing boot status code to indicate order of events.

rusty1968: Done

mhatrevi (approved):

korran (approved):

PR #571, opened 2023-08-01
Runtime : handle fatal errors

rusty1968:
Runtime reports panic, exception and NMI events as fatal errors and handles them accordingly.

### DIFF
--- a/runtime/src/main.rs
+++ b/runtime/src/main.rs
@@ -16,7 +16,8 @@ Abstract:
 
 use caliptra_common::cprintln;
 use caliptra_cpu::TrapRecord;
-use caliptra_drivers::{report_fw_error_non_fatal, Mailbox};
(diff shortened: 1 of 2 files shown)


jlmahowa-amd (dismissed):

runtime/src/main.rs
```diff
@@ -94,13 +95,43 @@ fn runtime_panic(_: &core::panic::PanicInfo) -> ! {
...
-    report_error(0xdead);
+    handle_fatal_error(caliptra_drivers::CaliptraError::RUNTIME_GLOBAL_PANIC.into());
+}
+
+/// Report fatal F/W error
+///
+/// # Arguments
+///
+/// * `val` - F/W error code.
+pub fn report_fw_error_fatal(val: u32) {
```
jhand2, line 106: nit: Why does this need to be public?

jhand2 (approved):

runtime/src/main.rs
rusty1968: Done

jhand2 (approved):

PR #572, opened 2023-08-01
FMC : Watchdog timer

rusty1968:
FMC starts the watchdog upon start of the reset flow and stops it prior to transitioning to the runtime module.


common/src/wdt.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// * `soc_ifc` - SOC Interface
+///
+///
+pub fn start_wdt(soc_ifc: &mut SocIfc, timeout: WdtTimeout) {
+    cprintln!("[state] Starting the Watchdog Timer");
+
+    let wdt1_timeout_cycles: u64 = match soc_ifc.get_cycle_count(timeout.0) {
+        Ok(cycle_count) => max(EXPECTED_CALIPTRA_BOOT_TIME_IN_CYCLES * 10, cycle_count),
+        Err(_) => EXPECTED_CALIPTRA_BOOT_TIME_IN_CYCLES * 10,
```
jhand2, line 48: If get_cycle_count fails, should we fail this function? I assume that would mean that `timeout` is an invalid value (too large or something)?

common/src/wdt.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/// Restart the Watchdog Timer
+///
+/// # Arguments
+///
+/// * `env` - ROM Environment
+///
+pub fn restart_wdt(soc_ifc: &mut SocIfc) {
+    cprintln!("[state] Restarting the Watchdog Timer");
```
jhand2, line 68: I wonder if we should remove these printlns. For a long-running system, I worry these will pollute the UART logs.

runtime/src/lib.rs
```diff
@@ -189,16 +189,12 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
-        .regs_mut()
-        .cptra_flow_status()
-        .write(|w| w.ready_for_runtime(true));
+    drivers.soc_ifc.assert_ready_for_runtime();
     caliptra_drivers::report_boot_status(RtBootStatus::RtReadyForCommands.into());
     loop {
         wait_for_cmd(&mut drivers.mbox);
-
         if drivers.mbox.is_cmd_ready() {
+            caliptra_common::wdt::start_wdt(&mut drivers.soc_ifc, WdtTimeout::default());
```
jhand2, line 197: Could put the start/stop in `wait_for_cmd`. This more closely aligns with John's request that the WDT is handled directly before/after wfi.

runtime/src/main.rs
```diff
@@ -46,6 +46,7 @@ pub extern "C" fn entry_point() -> ! {
                 );
             }
         };
+        caliptra_common::wdt::stop_wdt(&mut drivers.soc_ifc);
```
jhand2, line 49: If you handle the wdt in wait_for_cmd, you can probably remove this line.

common/src/wdt.rs
rusty1968: Good point.

common/src/wdt.rs
rusty1968: This is , as per Vishal , logic to assign a default value if timer period is not configured by the Soc.

runtime/src/lib.rs
rusty1968: we don't have wfi yet - we would keep hammering the register with no wait.

runtime/src/main.rs
rusty1968: When/if wfi is implemented , it will make sense.

runtime/src/lib.rs
jhand2: Ack. Can you add a TODO comment here?

rusty1968: Sure. Done.

runtime/src/wdt.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+impl From<WdtTimeout> for u32 {
+    fn from(val: WdtTimeout) -> Self {
+        core::num::NonZeroU32::from(val).get()
+    }
+}
+
+const EXPECTED_CALIPTRA_BOOT_TIME_IN_CYCLES: u64 = 20_000_000; // 20 million cycles
+const WDT1_DEFAULT_TIMEOUT_SECS: u32 = 5;
```
jhand2, line 55: I heard we want to use cycle counts instead of relying on second -> cycle conversions. Especially since we want to be sure we have the same behavior on different emulation platforms.

Should we remove the wall-clock based stuff?

jhand2 (approved):

runtime/src/wdt.rs
rusty1968: Done.

mhatrevi (approved):

fdamato (approved):

PR #575, opened 2023-08-02
Export caliptra_rom_run_fips_test() from the ROM to FMC/RT.

korran:
This PR creates a 6-entry jump table in the ROM. When mutable firmware tries to call a unimplemented exported function, it will return the CaliptraError::ROM_GLOBAL_UNIMPLEMENTED_ENTRY_POINT error.

Only the first entry is implemented, with this symbol: 

```
pub fn caliptra_rom_run_fips_tests() -> u32
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,12 +519,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "caliptra-rom-exports"
+version = "0.1.0"
+dependencies = [
+ "caliptra-error",
+]
+
+[[package]]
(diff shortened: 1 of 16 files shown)


vsonims (changes requested):
I am thinking of a different approach. It seems the challenge is the test vector size. Can we expose just the vectors so the tests are in RT and data can come from ROM?

korran:
It's not just the test vector size. For example, we have no need for an LMS implementation in the runtime; it'd be a shame to have to waste ~4k of code on it just for an on-demand fips test.

PR #577, opened 2023-08-02
[fuzzing] Implement initial support for Image Verifier fuzzing

benjamindoron:
First implementation of fuzz testing the Image Verifier library, forming part of the fuzzing work. Instructions and notes are in the readme.md files, these can be removed or migrated to actual documentation if required.

As discussed on https://github.com/chipsalliance/caliptra-sw/pull/554, maybe fuzz testing should be another action in the CI, both presubmission and nightly (hopefully also with more extensive fuzzing using larger compute farms). I'll create a workflow sometime soon.


benjamindoron:
I've looked into writing a fuzzing dictionary, and I don't think it would help. there are few fixed constants that need to be present. I think the same applies to DPE, probably, having a seed corpus with examples of each command is sufficient, but I can write one anyway. At least there's more than one entry there.

benjamindoron:
I didn't want to have to compare them at the time, and `cargo-fuzz` is easier to start, but another question is whether we run libFuzzer or AFL on the CI, or cloud, if that happens.

jhand2:
I have a lot of the same comments on these other PRs as I did on the DPE fuzzer PR. Once we get through that PR, I'll have a closer look at these ones.

image/verify/fuzz/src/fuzz_target_1.rs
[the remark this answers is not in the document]
benjamindoron: Now that the CI is formally reusable, with inputs, no need for a common target name. Done.

image/types/src/lib.rs
```diff
@@ -50,6 +50,7 @@ pub type ImageEccPrivKey = ImageScalar;
 
 #[repr(C)]
 #[derive(AsBytes, FromBytes, Default, Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
```
jhand2, line 53: Maybe this is a dumb question, but doesn't the `arbitrary` feature need to be defined in image/types/Cargo.toml?

benjamindoron: As an optional dependency, the feature is automatically defined.

jhand2: Ack, thanks for clarifying.

image/verify/fuzz/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = { version = "0.4.6", optional = true }
+afl = { version = "0.13.3", optional = true }
+
+[patch.crates-io]
+byteorder = { git = "https://github.com/benjamindoron/byteorder.git", branch = "struct_aware-1.4.3" }
+zerocopy = { git = "https://github.com/benjamindoron/zerocopy.git", branch = "struct_aware-v0.6.3" }
```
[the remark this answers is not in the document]
jhand2: I'm good with merging this as-is, but it would be good if you can send upstream PRs so we can eventually depend on upstream.

jhand2 (dismissed):

image/verify/fuzz/Cargo.toml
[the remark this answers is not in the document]
benjamindoron: Okay, thanks. The list of changes aren't long, as long as they don't mind being a dependency in fuzzing targets, it should be fine - std is required for cfg(test) or cfg(fuzzing).

image/types/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 doctest = false
 
 [dependencies]
+arbitrary = { version = "1.3.0", optional = true, features = ["derive"] }
```
korran, line 12: To make it easier to keep all the versions the same, can we please add this as a workspace dependency?

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

korran: You should still be able to use optional with workspace deps. For example:

https://github.com/chipsalliance/caliptra-sw/blob/2ea7949c437c37541b3dc41cd046cfd60f54c701/hw-model/Cargo.toml#L24

image/verify/fuzz/Cargo.toml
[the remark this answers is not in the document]
korran: Does this actually work? Isn't the image crate still using the other version of these crates defined in the workspace?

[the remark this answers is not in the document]
korran: Oh I see, image/verify/fuzz is not part of the workspace, so it can patch anything it wants. Probably worth a comment.

image/verify/fuzz/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+[dependencies.caliptra-drivers]
+path = "../../../drivers"
+
+[dependencies.caliptra-image-verify]
+path = ".."
+
+[features]
+struct-aware = []
+
+# Prevent this from interfering with workspaces
```
korran, line 37: As this is not part of the workspace, we need some sort of CI step that will make sure this project builds / is formatted correctly.

image/verify/fuzz/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+byteorder = { git = "https://github.com/benjamindoron/byteorder.git", branch = "struct_aware-1.4.3" }
+zerocopy = { git = "https://github.com/benjamindoron/zerocopy.git", branch = "struct_aware-v0.6.3" }
+
+[dependencies.caliptra-lms-types]
+path = "../../../lms-types"
+features = ["arbitrary"]
+
+[dependencies.caliptra-image-types]
+path = "../../types"
+features = ["arbitrary"]
```
korran, line 26: AFAICT, caliptra-image-types doesn't have an arbitrary feature:

```
[features]
default = ["std"]
std = []
```

korran: Hmm, apparently cargo automatically adds features for optional deps. That's news to me:

https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

korran (approved):
Approval for image/

image/verify/fuzz/Cargo.toml
benjamindoron: Optional dependencies have implicitly defined features to enable them.

image/verify/fuzz/Cargo.toml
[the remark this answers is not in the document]
benjamindoron: https://github.com/chipsalliance/caliptra-sw/issues/681; and commenting done, will be in the next push.

image/types/Cargo.toml
benjamindoron: This doesn't work (defined `workspace.dependencies` in fuzz/Cargo.toml). It seems this crate's dependencies are still used from inside the actual root workspace, not the local workspace.

image/verify/fuzz/Cargo.toml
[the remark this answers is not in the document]
benjamindoron: I'm not certain it's because this crate is it's own workspace that it can patch anything, because I couldn't define `arbitrary`'s version, as we discussed below, for the root workspace by defining it here. It might just be that any crate build can override dependencies.

Or, I've misunderstood the differences between dependencies in the workspace, and external dependencies.

image/verify/fuzz/Cargo.toml
benjamindoron: https://github.com/chipsalliance/caliptra-sw/pull/631 will run nightly, but that's not really soon enough. Maybe a for-loop (in the main build CI workflow) over each target to run these common steps will do.

jhand2: Thanks @benjamindoron, a for loop over these targets to run cargo build/fmt/clippy works for us

image/types/Cargo.toml
korran: Yeah, I was just referring to the image/types Cargo.toml (and any other crates in the main workspace). For the fuzz workspace you can do whatever you need.

image/verify/fuzz/Cargo.toml
benjamindoron: Okay, I'll push that to a separate PR, since it's not specific to any existing at the moment.

Note that clippy is broken until https://github.com/chipsalliance/caliptra-sw/issues/681 is done. The patch mechanism does not support features, so, I derived from `arbitrary` without checking. But that depends on `std`, which is usually disabled in those crates we used.

benjamindoron: Done, see https://github.com/chipsalliance/caliptra-sw/pull/700.

image/types/Cargo.toml
benjamindoron: Ah, I understand now. So, we'll define the version once for the root workspace, and also for the targets. Okay, then this is done now.

jhand2 (dismissed):

image/verify/fuzz/Cargo.toml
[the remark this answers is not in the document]
jhand2: Not sure I understand the question above. I think dependencies here are completely separate from the root workspace because this crate has its own workspace.

Anyway, as long as this works (which it looks like it does) we're good with it.

jhand2 (approved):
LGTM. @korran can you sign off again for the `image` directory?

korran (approved):

PR #583, opened 2023-08-02
Updated hw-latest/caliptra-rtl to d792ce264da10b5303b2e5831768fce97a78b210

jlmahowa-amd:


jlmahowa-amd:
One issue: uart.rs uses raw address for generic output wires, which did change address

jlmahowa-amd:
Right now seeing two issues:
Store access fault when setting a data vault entry. Specifically WarmResetEntry48::RtTci
FMC is not logging (emulator and FPGA). ROM and RT do log.

drivers/src/data_vault.rs
```diff
@@ -518,7 +518,7 @@ impl DataVault {
     ///
     pub fn read_warm_reset_entry48(&self, entry: WarmResetEntry48) -> Array4x12 {
         let dv = self.dv.regs();
-        Array4x12::read_from_reg(dv.nonsticky_data_vault_entry().at(entry.into()))
+        Array4x12::read_from_reg(dv.sticky_data_vault_entry().at(entry.into()))
```
korran, line 521: This is wrong. I think nonsticky_data_vault_entry was renamed data_vault_entry().

korran: Same elsewhere

korran (approved):

fdamato (approved):

PR #585, opened 2023-08-03
Val ROM

nquarton:
- Change no-kats feature name to fast-rom
- Added check to prohibit fast mode in production lifecycle

### DIFF
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -40,7 +40,7 @@ pub const ROM_WITH_UART: FwId = FwId {
 pub const ROM_FAST_WITH_UART: FwId = FwId {
     crate_name: "caliptra-rom",
     bin_name: "caliptra-rom",
-    features: &["emu", "no-kats"],
+    features: &["emu", "fast-rom"],
     workspace_dir: None,
 };
 
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -37,10 +37,10 @@ pub const ROM_WITH_UART: FwId = FwId {
     workspace_dir: None,
 };
 
-pub const ROM_FAST_WITH_UART: FwId = FwId {
+pub const ROM_VAL_WITH_UART: FwId = FwId {
     crate_name: "caliptra-rom",
     bin_name: "caliptra-rom",
-    features: &["emu", "fast-rom"],
+    features: &["emu", "val-rom"],
     workspace_dir: None,
 };
 
--- a/rom/dev/src/flow_val_rom.rs
+++ b/rom/dev/src/flow_val_rom.rs
@@ -8,7 +8,7 @@ File Name:
 
 Abstract:
 
-    File contains the implementation of the fast ROM reset flows
+    File contains the implementation of the val ROM reset flows
 
 --*/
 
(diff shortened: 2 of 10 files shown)


rom/dev/tests/test_fast_rom.rs
```diff
@@ -21,3 +22,29 @@ fn test_skip_kats() {
...
+            rom: &rom,
+            security_state,
+            ..Default::default()
+        },
+        ..Default::default()
+    })
+    .unwrap();
+
+    // Let it run until a fatal error (should fail very early)
+    hw.step_until(|m| m.soc_ifc().cptra_fw_error_fatal().read() != 0);
```
nquarton, line 43: This will hang if no fatal error is reported instead of the test failing quickly. Is there a better way to do this or is that an acceptable way to fail?

rusty1968: Consider checking for the actual error code.

nquarton: I do check for the specific error code (ROM_GLOBAL_FAST_ROM_IN_PRODUCTION) below with the assert. I just needed a condition to "step until" to know that was valid to check

builder/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub const ROM_WITH_UART: FwId = FwId {
 pub const ROM_FAST_WITH_UART: FwId = FwId {
     crate_name: "caliptra-rom",
     bin_name: "caliptra-rom",
-    features: &["emu", "no-kats"],
+    features: &["emu", "fast-rom"],
```
fdamato, line 43: shall we call it val-rom, as per today discussion ?

nquarton: I agree, changed now

rom/dev/src/flow_val_rom.rs
```diff
@@ -0,0 +1,63 @@
...
+use caliptra_drivers::{report_boot_status, CaliptraResult, ResetReason};
+use caliptra_error::CaliptraError;
+use fw_processor::FirmwareProcessor;
+
+/// Execute ROM Flows based on reset reason
+///
+/// # Arguments
+///
+/// * `env` - ROM Environment
+pub fn run(env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
```
mhatrevi, line 32: We had discussed having a different reason other than Cold/Update/Warm Reset to trigger this validation flow. Is that not in plan?

mhatrevi (changes requested):

rom/dev/src/flow_val_rom.rs
fdamato: You are right, that was the initial plan, and I was personally the one to propose it. However while implementing Nick realized that it would require either a much more "invasive" change (or a lot of duplication) . That`s because the Mailbox APIs are private to the coldreset flow.
Do you see any issues with this "new" approach ?

nquarton: The issue of the lower-level code being private is largely resolved by the introduction of the val-specific verifier environment (we no longer had to skip the verify call as we originally thought, it's bypassed in a different way).

So an approach that adds a new flow is technically possible. But that would require touching and adding this feature switch in more of the existing ROM code files. This current approach let us eliminate changes to any rom code outside of main.rs as well as limit what needed to be touched in that file.

nquarton:
Replacing with https://github.com/chipsalliance/caliptra-sw/pull/620 so Vishal M can push additional changes

PR #587, opened 2023-08-03
Add ctx-based sha384 driver impl

jhand2:
Add a set of APIs to the sha384 driver to avoid needing a persistent reference to the driver.

These APIs hold all state in a context struct which is then passed the the sha384 driver to operate on it. This greatly simplifies lifetime/borrow management of long-running hash operations. It is also safe to have multiple active contexts at the same time (assuming single-threaded execution).

This PR leaves the existing set of op-based APIs in place to avoid changing ROM. This new set of APIs is primarily intended to be used in runtime firmware.

### DIFF
--- a/drivers/src/sha384.rs
+++ b/drivers/src/sha384.rs
@@ -32,6 +32,37 @@ pub struct Sha384 {
     sha512: Sha512Reg,
 }
 
+pub struct Sha384Ctx {
+    buf: [u8; SHA384_BLOCK_BYTE_SIZE],
+    buf_idx: usize,
+    data_size: usize,
+    state: Sha384DigestState,
+}
+
+impl Default for Sha384Ctx {
+    fn default() -> Self {
+        Self {
+            buf: [0u8; SHA384_BLOCK_BYTE_SIZE],
+            buf_idx: 0,
+            data_size: 0,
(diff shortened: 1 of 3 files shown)


drivers/src/sha384.rs
```diff
@@ -57,6 +86,63 @@ impl Sha384 {
         Ok(op)
     }
 
+    /// Update the digest with data
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - Data to used to update the digest
+    pub fn update_ctx(&mut self, ctx: &mut Sha384Ctx, data: &[u8]) -> CaliptraResult<()> {
```
jhand2, line 94: These implementation are mostly copied from the Op-API. If we want I can update that API to use these. I just didn't want to mess too much with driver functions that ROM is using.

korran: For code-size in binaries that use both APIs, it makes sense to combine. But like you say there may be some pushback there....

korran:
FYI, I'm hoping that this PR isn't necessary, as it should be possible to use the ownership API.

jhand2:
Dropping this PR, I don't think it really works as I've described anyway. I'll try to find another way to make this work.

PR #593, opened 2023-08-04
FMC : Panic, exceptions and NMIs are handled as fatal errors

rusty1968:
FMC : Panic, exceptions and NMIs are handled as fatal errors and reported to the SoC via fatal error register.


fmc/src/main.rs
```diff
@@ -109,6 +110,48 @@ fn report_error(code: u32) -> ! {
...
+/// # Arguments
+///
+/// * `val` - F/W error code.
+fn report_fw_error_fatal(val: u32) {
+    let mut soc_ifc = unsafe { SocIfcReg::new() };
+    soc_ifc.regs_mut().cptra_fw_error_fatal().write(|_| val);
+}
+
+#[allow(clippy::empty_loop)]
+fn handle_fatal_error(code: u32) -> ! {
```
bluegate010, line 124: If we think this might be useful for RT FW, could put it in common/. Or could move it there once we get to that point in RT FW development.

bluegate010 (approved):

fmc/src/main.rs
rusty1968: Sure. I can make this common on a follow up.

Issue #581, opened 2023-08-02
Findings from fuzzing x509


benjamindoron:
Attempting to fuzz x509 shows that `Ecdsa384CertBuilder::build()` returns certs that OpenSSL's `X509::from_der()` will throw errors trying to parse. I'm trying to determine if I'm even fuzzing correctly here.

[...]

PR #595, opened 2023-08-04
[fuzzing] Implement initial support for Caliptra x509 fuzzing

benjamindoron:
The intention of this fuzz target is to test that caliptra-x509 correctly parses TBS blobs - "to-be-signed" - into valid x509. Currently, it only produces x509 that OpenSSL cannot parse.

Instructions for running the supported fuzzers can be found at https://gist.github.com/benjamindoron/59313b863b079f6f5af56af83d32648e.


jhand2 (approved):

jhand2:
This LGTM, but as Benjamin pointed out in https://github.com/chipsalliance/caliptra-sw/issues/581, this will trigger lots of panics due to invalid X.509 being generated. This isn't running in CI yet, so I'm going to merge this. I think it will make it easier to fix those issues if others can run it.

PR #596, opened 2023-08-04
Fixes for linker script generation

jhand2:
1. RTALIAS_TBS_ORG was missing from memory_layout.rs. This takes 1K away
   from the DATA section.
2. Generate the FMC and RT linker scripts entirely from the constants in
   memory_layout.rs
3. Shrink ROM data section by 1K to account for RTALIAS_TBS_ORG. This
   prevents ROM from overwriting the RT Alias Cert on resets.
4. Increase RT and FMC stack size to 22K (from 14). This is necessary to
   account for the lager DPE command/response buffers. This takes
   another 8K away from data.

Total change:

* Stack increases by 8K
* RT Alias takes up 1K
* Data shrinks by 9K


common/src/memory_layout.rs
```diff
@@ -30,11 +30,12 @@ pub const MAN2_ORG: u32 = 0x50001800;
...
 pub const FMCALIAS_TBS_ORG: u32 = 0x50003C00;
-pub const PCR_LOG_ORG: u32 = 0x50004000;
-pub const FUSE_LOG_ORG: u32 = 0x50004400;
-pub const BOOT_STATUS_ORG: u32 = 0x500047FC;
-pub const DATA_ORG: u32 = 0x50004800;
-pub const STACK_ORG: u32 = 0x5001C000;
+pub const RTALIAS_TBS_ORG: u32 = 0x50004000;
+pub const PCR_LOG_ORG: u32 = 0x50004400;
+pub const FUSE_LOG_ORG: u32 = 0x50004800;
+pub const BOOT_STATUS_ORG: u32 = 0x50004BFC;
```
korran, line 36: You need to update this in drivers/src/soc_ifc.rs

common/src/memory_layout.rs
```diff
@@ -30,11 +30,12 @@ pub const MAN2_ORG: u32 = 0x50001800;
...
+pub const DATA_ORG: u32 = 0x50004C00;
```
korran, line 37: There's a reference to this in fmc/test-fw/test-rt/memory.x

korran: Also in rom/dev/tools/test-fmc/src/fmc.ld and rom/dev/tools/test-rt/src/rt.ld

common/src/memory_layout.rs
jhand2: Fixed by #617

common/src/memory_layout.rs
jhand2: Done.

korran (approved):

bluegate010 (approved):

cpu/gen/src/lib.rs
```diff
@@ -1,30 +1,30 @@
...
-              
-        ICCM_SIZE   = 0x{:08X};
-        DCCM_SIZE   = 128K;
-        DATA_SIZE   = 94K;
-        STACK_SIZE  = 14K;
-        ESTACK_SIZE = 1K;
-        NSTACK_SIZE = 1K;
-        
-        
+        ICCM_ORG  		 = 0x{iccm_org:08X};
```
bluegate010, line 6: Nit: any reason this isn't capitalized like the others? Same for `iccm_size` below.

jhand2: Because it's passed as a parameter to this function instead of a constant. I'm not sure why, that's just how it was before. Happy to change this in a follow-up PR.

bluegate010: Ah I missed that.

korran: Leave it as-is to follow the rust style rules; constants should be upper case, variables lower-case.

korran: Oh; maybe you mean you can get rid of the function parameters and replace them with constants...

PR #482, opened 2023-07-17
Extend both a current and journey PCR in ROM.

bluegate010:
For consistency with FMC. Also adds tests to validate the log against the PCR values in hardware.

PR #598, opened 2023-08-04
FMC : Log PCR extensions to the PCR_Log

rusty1968:
FMC appends to the PCR log in DCCM, to enhance testability.

### DIFF
--- a/fmc/src/flow/pcr.rs
+++ b/fmc/src/flow/pcr.rs
@@ -23,9 +23,30 @@ Note:
 use crate::flow::tci::Tci;
 use crate::fmc_env::FmcEnv;
 use crate::HandOff;
-use caliptra_drivers::{okref, CaliptraResult, PcrId};
+use caliptra_drivers::{okref, CaliptraError, CaliptraResult, PcrBank, PcrId, Sha384};
 
-use caliptra_common::{RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR};
+use caliptra_common::{
+    memory_layout::{PCR_LOG_ORG, PCR_LOG_SIZE},
+    PcrLogEntry, PcrLogEntryId, RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR,
+};
+use zerocopy::AsBytes;
+
+struct PcrExtender<'a> {
+    pcr_bank: &'a mut PcrBank,
+    sha384: &'a mut Sha384,
+}
+impl PcrExtender<'_> {
+    fn extend(
+        &mut self,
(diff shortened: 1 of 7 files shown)


fmc/src/flow/pcr.rs
```diff
@@ -23,9 +23,30 @@ Note:
...
+impl PcrExtender<'_> {
+    fn extend(
+        &mut self,
+        pcr_id: PcrId,
+        bytes: &[u8],
+        pcr_entry_id: PcrLogEntryId,
+    ) -> CaliptraResult<()> {
+        self.pcr_bank
+            .extend_pcr(PcrId::PcrId0, self.sha384, bytes)?;
+        log_pcr(pcr_entry_id, pcr_id, bytes)
```
bluegate010, line 47: With https://github.com/chipsalliance/caliptra-sw/pull/600, we could adjust this logic a bit such that we only create one log when extending into both PCRs. (This wouldn't be strictly necessary, we could keep creating one log for each extend operation, but it's more space efficient if we can group them together).

fmc/src/flow/pcr.rs
```diff
@@ -56,16 +77,62 @@ pub fn extend_journey_pcr(env: &mut FmcEnv, hand_off: &HandOff) -> CaliptraResul
...
+    };
+    pcr_log_entry.pcr_data.as_bytes_mut()[..data.len()].copy_from_slice(data);
+
+    let dst: &mut [PcrLogEntry] = unsafe {
+        let ptr = PCR_LOG_ORG as *mut PcrLogEntry;
+        core::slice::from_raw_parts_mut(ptr, PCR_LOG_SIZE / core::mem::size_of::<PcrLogEntry>())
+    };
+
+    // Store the log entry.
+    dst[pcr_entry_id as usize - 1] = pcr_log_entry;
```
bluegate010, line 135: This current logic will leave holes in the log buffer if ROM and FMC do not agree on how many extensions should have occurred. https://github.com/chipsalliance/caliptra-sw/pull/482 brings in a log entry counter, which ROM initializes to zero, and which FMC can initialize by walking the log space until it finds an invalid entry. Or, could have ROM put the entry count in FHT.

fmc/src/flow/pcr.rs
rusty1968: Can you make that code common?

fmc/src/flow/pcr.rs
FerralCoder: I think it would make the most sense for ROM to fill in this count, and for FMC to update the count

PR #610, opened 2023-08-08
Adding support to send a response for mailbox commands in ROM

nquarton:


nquarton:
Replaced with https://github.com/chipsalliance/caliptra-sw/pull/627 to make collaboration with Anthony easier for additional changes

PR #611, opened 2023-08-08
Smoke test : Do not use UART output to check boot success

rusty1968:
This commit retires usage of UART output to in the smoke test.


FerralCoder:
is this PR still relevant? Can it be closed?

PR #612, opened 2023-08-08
Call derive child in runtime and implement invoke_dpe mbox cmd

sree-revoori1:


runtime/dpe
jhand2, line 1: Should probably avoid updating DPE in the PR since it's unrelated.

runtime/src/disable.rs
```diff
@@ -0,0 +1,56 @@
...
+            .erase_key(KEY_ID_RT_PRIV_KEY)
+            .map_err(|_| CaliptraError::RUNTIME_DISABLE_ATTESTATION_FAILED)?;
+        Self::generate_dice_key(drivers)?;
+        Ok(MailboxResp::default())
+    }
+
+    fn generate_dice_key(drivers: &mut Drivers) -> CaliptraResult<()> {
+        hmac384_kdf(
+            &mut drivers.hmac384,
+            KeyReadArgs::new(KEY_ID_RT_CDI).into(),
```
jhand2, line 28: I'm not completely sure what happens  when you erase a keyvault slot and then use it uninitialized. The SW emulator may allow it, but I'm not sure the real hardware will.

Could you make a rom test and then run it against the verilator model to be sure? Ideally we would explicitly fill the CDI keyvault slot with zeros, but I'm not sure if there's a way to do that.

A few tips on verilator:

* Instructions for using verilator: https://github.com/chipsalliance/caliptra-sw#testing-against-verilator. You'll need to install it and the build will take a while
* Verilator is pretty slow, so the test may take longer than expected to run
* You can use the `test_ecdsa_verify_cmd` test as an example of something that can run against verilator but still run mailbox commands

runtime/src/invoke_dpe.rs
```diff
@@ -0,0 +1,42 @@
...
+        if let Some(cmd) = InvokeDpeReq::read_from(cmd_args) {
+            // Won't panic, full_digest is always larger than digest
+            let mut response_buf = [0u8; InvokeDpeResp::DATA_MAX_SIZE];
+            let mut env = DpeEnv::<CptraDpeTypes> {
+                crypto: DpeCrypto::new(&mut drivers.sha384),
+                platform: DpePlatform,
+            };
+            match drivers
+                .dpe
+                .execute_serialized_command(&mut env, INVOKE_DPE_LOCALITY, &cmd.data)
```
jhand2, line 24: I think we should set the locality to the mailbox pauser: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.mbox_csr.mbox_user. I think you'll need to add support to the mailbox driver to read this register.

runtime/src/lib.rs
```diff
@@ -153,8 +160,24 @@ impl<'a> Drivers<'a> {
             crypto: DpeCrypto::new(&mut sha384),
             platform: DpePlatform,
         };
-        let dpe = DpeInstance::new(&mut env, DPE_SUPPORT)
+        let mut dpe = DpeInstance::new(&mut env, DPE_SUPPORT)
             .map_err(|_| CaliptraError::RUNTIME_INVOKE_DPE_FAILED)?;
+        let pcr_bank = PcrBank::new(PvReg::new());
+        // TODO: Set target_locality to SoC's initial locality
```
jhand2, line 166: Can you open a github issue to track this?

runtime/src/lib.rs
```diff
@@ -153,8 +160,24 @@ impl<'a> Drivers<'a> {
             crypto: DpeCrypto::new(&mut sha384),
             platform: DpePlatform,
         };
-        let dpe = DpeInstance::new(&mut env, DPE_SUPPORT)
+        let mut dpe = DpeInstance::new(&mut env, DPE_SUPPORT)
```
jhand2, line 163: I think it might be worth moving this into a helper function which creates the DPE instance, initializes it, and then returns it.

runtime/src/lib.rs
```diff
@@ -211,7 +235,18 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+            let resp = InvokeDpeCmd::execute(drivers, cmd_bytes)?;
+            match resp {
+                MailboxResp::InvokeDpeCommand(InvokeDpeResp {
+                    data_size, data, ..
+                }) => {
+                    drivers.mbox.write_response(&data[..data_size as usize])?;
+                    Ok(resp)
+                }
+                _ => Err(CaliptraError::RUNTIME_INVOKE_DPE_FAILED),
+            }
```
jhand2, line 248: Could you make this a function on the `InvokeDpeResp`?

I'm thinking something like

```rust
impl InvokeDpeResp {
  fn write_to_mbox(&self, mbox: &mut Mailbox) -> CaliptraResult<()> {}
}
```

And then we can have a `write_to_mbox` function on MailboxResponse, which calls `write_to_mbox` on the individual response types. That way down on line 233 we can just call `resp.write_to_mbox`.

For all the non-dpe commands, you can just call `Packet::copy_to_mbox(drivers, &mut resp)` for now

runtime/tests/integration_tests.rs
```diff
@@ -132,6 +140,57 @@ fn test_fw_info() {
     assert_eq!(info.pl0_pauser, 0xFFFF0000);
 }
 
+#[test]
+fn test_invoke_dpe_get_profile_cmd() {
+    let mut model = run_rom_test("mbox");
```
jhand2, line 145: I think we should make this an RT test: `run_rt_test(None)`. A lot of the DPE functionality depends on things being set up by ROM/FMC (like the DICE keys)

runtime/tests/integration_tests.rs
```diff
@@ -132,6 +140,57 @@ fn test_fw_info() {
...
+
+    model.step_until(|m| {
+        m.soc_mbox().status().read().mbox_fsm_ps().mbox_idle()
+            && m.soc_ifc().cptra_boot_status().read() == 1
+    });
+
+    let mut data = [0u8; InvokeDpeReq::DATA_MAX_SIZE];
```
jhand2, line 152: I think this is unused

runtime/tests/integration_tests.rs
```diff
@@ -132,6 +140,57 @@ fn test_fw_info() {
...
+        .mailbox_execute(u32::from(CommandId::INVOKE_DPE), cmd.as_bytes())
+        .unwrap()
+        .expect("We should have received a response");
+
+    let resp_hdr: &InvokeDpeResp = LayoutVerified::<&[u8], InvokeDpeResp>::new(resp.as_bytes())
+        .unwrap()
+        .into_ref();
+
+    assert_eq!(resp_hdr.data_size as usize, size_of::<GetProfileResp>());
+    let data = resp_hdr.data;
```
jhand2, line 182: I think you can use zerocopy here: `GetProfileResp::read_from(resp.as_bytes())`

runtime/tests/integration_tests.rs
```diff
@@ -132,6 +140,57 @@ fn test_fw_info() {
...
+    // check dpe profile
+    assert_eq!(data[8..12], (DPE_PROFILE as u32).to_le_bytes());
+    // check vendor id
+    assert_eq!(&data[16..20], VENDOR_ID.to_le_bytes());
+    // check vendor sku
+    assert_eq!(data[20..24], VENDOR_SKU.to_le_bytes());
+    // check flags
+    assert_eq!(data[28..32], DPE_SUPPORT.get_flags().to_le_bytes());
+    assert_eq!(model.soc_ifc().cptra_fw_error_non_fatal().read(), 0);
```
jhand2, line 191: mailbox_execute will return an error if this is non-zero, so we can omit this check.

runtime/src/invoke_dpe.rs
jhand2: That is the address of that register. The SoC will write to that address to set the caller whenever a mailbox command is sent. So on each mailbox command we'll need to read from that register to get the value. You can use look at the [mailbox driver](https://github.com/chipsalliance/caliptra-sw/blob/7e9dc76e9083c09027c476d1776f1b96b3caea72/runtime/src/mailbox.rs#L24) to see how we read from other registers.

runtime/tests/integration_tests.rs
jhand2: Oh, it's probably because when DPE gets build as a dependency, it's not being built in test mode. We should probably just change the DPE structures to always derive AsBytes/FromBytes, not conditionally. The extra code only gets linked if it's actually used anyway.

jhand2 (approved):

PR #614, opened 2023-08-08
[fix] Fail ROM boot flow on receiving an invalid Mailbox command

mhatrevi:


vsonims (dismissed):

hw-model/src/lib.rs
```diff
@@ -554,6 +554,37 @@ pub trait HwModel {
         }
     }
 
+    fn step_until_fatal_error(&mut self, expected_error: u32) {
+        // Since the boot takes less than 30M cycles, we know something is wrong if
```
fdamato, line 558: it takes ~28M cycles now, but I`m not sure we can rely on this number for long term. As we saw in the last few weeks, this number can vary, and this can potentially hide a bug.
Can this logic go outside of the model, and probably on the test case, where we can pass the MAX_WAIT_CYCLE as a parameter ?

bluegate010:
It seems like Caliptra should have some way to report which commands it supports, or indicate whether it supports a given command. Otherwise, client code that wants to be agnostic to different Caliptra vintages might risk triggering a fatal error if it issues a command that some Caliptras support but not others.

hw-model/src/lib.rs
mhatrevi: Updated.

mhatrevi:
> It seems like Caliptra should have some way to report which commands it supports, or indicate whether it supports a given command. Otherwise, client code that wants to be agnostic to different Caliptra vintages might risk triggering a fatal error if it issues a command that some Caliptras support but not others.

For V1, we can document the mailbox commands supported as part of ROM documentation.

korran (approved):

PR #617, opened 2023-08-08
Move memory_layout.rs (and deps) to caliptra_drivers

korran:
This is necessary to allow the drivers to reference memory-layout
constants without introducing a circular dependency.

Also, take advantage of this change to remove a copy-pasted constant from report_boot_status().

Also, delete the rom printing infrastructure, and use the one in caliptra-drivers instead.

Also, don't attempt to print the FHT from the ROM, as it will actually do something now (the rom enables caliptra-drivers' emu feature, unlike caliptra-common).


mhatrevi (approved):

drivers/src/printer.rs
```diff
@@ -25,7 +25,7 @@ impl uWrite for MutablePrinter {
     #[inline(never)]
     fn write_str(&mut self, _str: &str) -> Result<(), Self::Error> {
         #[cfg(feature = "emu")]
-        caliptra_drivers::Uart::default().write(_str);
```
rusty1968, line 28: nit : Instead of MutablePrinter, let's call it just Printer.

korran: That's the name it had before; I didn't change it.

rusty1968 (approved):

drivers/src/printer.rs
rusty1968: I know. I regret naming it like that.

mhatrevi (approved):

jhand2 (approved):

drivers/src/printer.rs
korran: Renamed.

jhand2 (approved):

rusty1968 (approved):

PR #618, opened 2023-08-09
[fuzzing] Implementing initial support for fuzzing LMS

benjamindoron:


benjamindoron:
Will address feedback on other PRs first before marking this as ready for review, expecting similar changes to apply.

drivers/Cargo.toml
```diff
@@ -16,6 +16,7 @@ caliptra-image-types.workspace = true
 caliptra-lms-types.workspace = true
 caliptra-registers.workspace = true
 cfg-if.workspace = true
+sha2.workspace = true
```
jhand2, line 19: Can this be a dev dependency? (Or see suggestion below to move sw impl outside of the driver lib)

drivers/fuzz/src/fuzz_target_lms.rs
```diff
@@ -0,0 +1,56 @@
...
+    if data.len() < (size_of::<LmsPublicKey<SHA256_DIGEST_WORD_SIZE>>() + size_of::<LmsSignature<SHA256_DIGEST_WORD_SIZE, LOCAL_P, LOCAL_H>>()) {
+        return;
+    }
+
+    // TODO: Alternatively, use structure-aware fuzzing, input comprising arguments
+    unsafe {
+        input = &data[(size_of::<LmsPublicKey<SHA256_DIGEST_WORD_SIZE>>() + size_of::<LmsSignature<SHA256_DIGEST_WORD_SIZE, LOCAL_P, LOCAL_H>>())..];
+        pub_key = &*(data.as_ptr() as *const LmsPublicKey<SHA256_DIGEST_WORD_SIZE>);
+        sig = &*(data[size_of::<LmsPublicKey<SHA256_DIGEST_WORD_SIZE>>()] as *const LmsSignature<SHA256_DIGEST_WORD_SIZE, LOCAL_P, LOCAL_H>);
+    }
```
jhand2, line 39: nit: Could use zerocopy to copy the data into a rust struct. Would make this code easier to read (and non-unsafe). But this may not be relevant if you switch to structure-aware

drivers/src/sha256.rs
```diff
@@ -216,6 +245,53 @@ impl Sha256 {
     }
 }
 
+pub struct Sha256SoftwareDriver {
```
jhand2, line 248: I think this could be implemented in the fuzzer library instead of in the driver itself, as long as the Sha256 trait is public.

benjamindoron:
I don't understand why clippy throws a 'no Default' error for `Sha256SoftwareDriver` but not the hardware implementation. Is there an implied default via `Sha256Reg`?

benjamindoron:
TODO: Consider generating a seed corpus from tests around the tree.

drivers/fuzz/src/fuzz_target_lms.rs
```diff
@@ -0,0 +1,83 @@
...
+fn harness_unstructured(data: &[u8]) {
+    let pub_key: &LmsPublicKey<SHA256_DIGEST_WORD_SIZE>;
+    let sig: &LmsSignature<SHA256_DIGEST_WORD_SIZE, LOCAL_P, LOCAL_H>;
+    let input: &[u8];
+
+    if data.len() < (size_of::<LmsPublicKey<SHA256_DIGEST_WORD_SIZE>>() + size_of::<LmsSignature<SHA256_DIGEST_WORD_SIZE, LOCAL_P, LOCAL_H>>()) {
+        return;
+    }
+
+    // TODO: The corpus may be seeded with (pub_key, sig), so parse the data as these first
```
benjamindoron, line 48: TODO: Consider generating a seed corpus from test data around the tree.

drivers/fuzz/src/fuzz_target_lms.rs
```diff
@@ -0,0 +1,83 @@
...
+use afl::fuzz;
+
+#[cfg(not(feature = "struct-aware"))]
+use std::mem::size_of;
+
+use caliptra_drivers::{Lms, Sha256SoftwareDriver};
+use caliptra_lms_types::{LmsPublicKey, LmsSignature};
+
+pub const SHA256_DIGEST_WORD_SIZE: usize = 8;
+// TODO: When using SHA256_DIGEST_WORD_SIZE, these tend to accompany?
```
benjamindoron, line 21: I wonder what happens if we break this expectation. It will make it too difficult to provide a seed corpus though.

drivers/src/sha256.rs
jhand2: Let's do that, at least for now. I want to keep the drivers crate to just the actual hardware drivers if we can.

jhand2:
> I don't understand why clippy throws a 'no Default' error for `Sha256SoftwareDriver` but not the hardware implementation. Is there an implied default via `Sha256Reg`?

I think it's because Sha256Reg doesn't implement Default, so the hardware driver doesn't have a trivial Default implementation. But the software driver can just derive default because all the fields can be default-initialized.

drivers/fuzz/src/fuzz_target_lms.rs
jhand2: Since these are specified at compile time, I think it's fine to not fuzz different combinations. But for consistency with what ROM is doing, I think we should depend directly on these P and H values: https://github.com/chipsalliance/caliptra-sw/blob/8cb6e40db5a79148fe2956d78380b46198174848/image/types/src/lib.rs#L37

drivers/fuzz/src/fuzz_target_lms.rs
benjamindoron: Done

drivers/Cargo.toml
benjamindoron: I've migrated it into the target, so this is addressed by the next push.

drivers/src/sha256.rs
benjamindoron: Okay, I'll be pushing soon.

drivers/fuzz/src/fuzz_target_lms.rs
benjamindoron: Okay, done. I'll be remaking the seed corpus soon too

drivers/fuzz/src/fuzz_target_lms.rs
benjamindoron: The `input` argument to `verify_lms_signature()` is itself a byte slice, and this doesn't implement FromBytes - possibly odd, since the implementation is trivial, I believe? I'm not sure if it would be possible to zerocopy the `data` argument here to the struct we need.

jhand2: I mean something like this:

```rust
let mut offset = 0;
let pub_key = ImagePublicKey::read_from_prefix(data);
offset += sizeof::<ImagePublicKey>();
let sig = ImageLmsSignature::read_from_prefix(&data[offset..];
offset += sizeof::<ImageLmsSignature>();
let input = &data[offset..];
```

This would require deriving FromBytes for those types, but I think that's ok.

benjamindoron: Ah, I see. `read_from_prefix()` will work.

jhand2 (dismissed):
LGTM. @korran to sign off for changes to `drivers` and `image`

@rusty1968 can you take a look for changes to `fmc`?

jhand2 (dismissed):

drivers/fuzz/src/sha256.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    fn copy_digest_to_buf(&mut self, buf: &mut Array4x8) -> CaliptraResult<()> {
+        *buf = Array4x8::from(self.hash);
+        Ok(())
+    }
+
+    fn digest_block(
+        &mut self,
+        block: &[u8; SHA256_BLOCK_BYTE_SIZE],
+        _first: bool,
```
korran, line 33: Don't you need to reset the hash when _first is true?

There's no way this implementation is going to generate the correct output if it mixes ALL operations together.

drivers/src/sha256.rs
```diff
@@ -137,21 +113,14 @@ impl Sha256 {
...
     ///
     /// # Arguments
     ///
     /// * `buf` - Digest buffer
-    fn copy_digest_to_buf(&mut self, buf: &mut Array4x8) -> CaliptraResult<()> {
-        let sha256 = self.sha256.regs();
-        *buf = Array4x8::read_from_reg(sha256.digest());
-        Ok(())
-    }
+    fn copy_digest_to_buf(&mut self, buf: &mut Array4x8) -> CaliptraResult<()>;
```
korran, line 123: I don't like how this formerly private function is now exposed to the world.

As an experiemnt, I tried making this refactor as minimal as possible:
https://github.com/chipsalliance/caliptra-sw/pull/731

The main downside to my approach is that the fuzz tests don't cover any of the additional logic in the sha256 driver.

drivers/fuzz/src/sha256.rs
benjamindoron: Oh. Possibly. Because while new Sha256DigestOp objects are created, there is only one driver passed to each. They'd have to share internal state. I'll probably need it to call zeroize(), like digest() does.

korran:
I've refactored the Sha256 driver changes to be the minimal possible.

drivers/fuzz/src/sha256.rs
korran: The new abstraction just calls the high-level sha256 methods directly, so there's no need for this change.

drivers/src/sha256.rs
korran: The experiment has been merge into this PR.

korran:
Note: The ROM binary is identical after this change (the only difference is the git commit in RomInfo):

```
git checkout pr618 && cargo run -p caliptra-builder -- --rom /tmp/pr618 --fw /dev/null && git checkout main && cargo run -p caliptra-builder -- --rom /tmp/main --fw /dev/null && diff --unified <(hexdump -C /tmp/main) <(hexdump -C /tmp/pr618)
@@ -2542,7 +2542,7 @@
 *
 00009fc0  38 4e 59 06 98 d1 34 56  d5 a8 8b d3 21 e0 a4 de  |8NY...4V....!...|
 00009fd0  6d d6 79 c7 3e 7e 39 96  41 dc a7 9e 35 87 23 09  |m.y.>~9.A...5.#.|
-00009fe0  bf 1a 7f 78 e9 07 5b 1f  9c 78 d1 47 d1 47 d1 47  |...x..[..x.G.G.G|
+00009fe0  5c 8a aa 12 7d 1c 51 5d  46 d4 d1 47 d1 47 d1 47  |\...}.Q]F..G.G.G|
 00009ff0  d1 47 d1 47 00 00 00 00  00 00 00 00 00 00 00 00  |.G.G............|
 0000a000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
 *
```

mhatrevi (approved):

korran (approved):

bluegate010 (approved):

PR #620, opened 2023-08-09
Val ROM

nquarton:
- Changing no-kats feature to val-rom
- Adding check to prohibit fast mode in production lifecycle
- Bypassing ECC and LMS verify with a mock verifier env for val rom
- Adding new flow implementation to bypass cert generation


fdamato (approved):
Thanks for the collaboration Vishal/Nick. 
We do need an extra PR to include update/warm reset flows in the val_rom.

JohnTraverAmd (approved):

PR #622, opened 2023-08-09
Update to validation ROM flow.

mhatrevi:


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -187,6 +188,13 @@ impl FirmwareProcessor {
         manifest: &ImageManifest,
         img_bundle_sz: u32,
     ) -> CaliptraResult<ImageVerificationInfo> {
+        #[cfg(feature = "val-rom")]
```
nquarton, line 191: I don't think there is much of a point for the added complexity of the val verify environment if we can add the cfg feature switches in the lower level files. You could just cut out the actual verify call instead.

The reason I introduced that was because I was under the impression changes were to be limited to main.rs

nquarton (commented):
Was I mistaken that we were supposed to be limiting changes to main.rs? It looks like this approach should accomplish the same overall goal, but it's more integrated into the lower-level code. I was under the impression that was to be avoided to clearly separate the validation code from the real code to mitigate the possibility of hidden errors.

PR #625, opened 2023-08-10
svn mailbox command

sphinxc0re:


runtime/src/mailbox_api.rs
```diff
@@ -264,3 +268,11 @@ pub struct FwInfoResp {
...
     // status.
 }
+
+/// SVN response
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetSvnResp {
+    pub runtime_svn: u32,
+    pub fmc_manifest_svn: u32,
+}
```
jhand2, line 278: Instead of making a new command, I think we should add these to the FW_INFO command. Thanks!

sphinxc0re: Done!

error/src/lib.rs
```diff
@@ -20,410 +20,365 @@ use core::num::NonZeroU32;
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct CaliptraError(pub NonZeroU32);
 impl CaliptraError {
-    const fn new_const(val: u32) -> Self {
```
jhand2, line 23: Are the changes in this file required? I would prefer to avoid including this change in this PR if we don't have to.

sphinxc0re: Sure, I will put it in a separate PR then.

PR #626, opened 2023-08-10
In sha drivers, take digest during finalize instead of init

jhand2:
Currently, the sha drivers take a digest reference during init, and then don't use it until finalize is called. This can be complicated for managing the lifetime of the digest structure. RT would like to be able to hold a sha context which contains a SHA op and a digest; but the lifetime management for doing this is very tricky.

Instead, take the digest as input to finalize and write the state to it then.

This does not change any functionality of the SHA drivers, only the interface.

### DIFF
--- a/drivers/src/sha1.rs
+++ b/drivers/src/sha1.rs
@@ -31,17 +31,13 @@ impl Sha1 {
     /// # Returns
     ///
     /// * `Sha1Digest` - Object representing the digest operation
-    pub fn digest_init<'a>(
-        &'a mut self,
-        digest: Sha1Digest<'a>,
-    ) -> CaliptraResult<Sha1DigestOp<'a>> {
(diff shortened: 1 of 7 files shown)


drivers/src/sha1.rs
```diff
@@ -244,7 +237,7 @@ impl<'a> Sha1DigestOp<'a> {
     }
 
     /// Finalize the digest operations
-    pub fn finalize(&mut self) -> CaliptraResult<()> {
+    pub fn finalize(&mut self, digest: &mut Array4x5) -> CaliptraResult<()> {
```
korran, line 240: Can you also have this take `self` or `mut self` instead of `&mut self`? We shouldn't allow people to call finalize multiple times.

jhand2: Done.

mhatrevi (approved):

PR #627, opened 2023-08-10
Mailbox driver response support

nquarton:
New ROM features require the ability to send a response to incoming (SoC->uC) commands. This PR adds support for sending responses to the mailbox driver, and refactors the populate-fifo code to be reused between sending responses and sending requests.


drivers/src/mailbox.rs
```diff
@@ -314,6 +287,49 @@ impl<'a> MailboxRecvPeek<'a> {
     }
 }
 
+/// Mailbox Fifo abstraction
+pub struct MailboxTxFifo<'a> {
+    mbox: &'a mut MboxCsr,
+}
+
+impl<'a> MailboxTxFifo<'a> {
+    pub fn try_enqueue(mbox: &'a mut MboxCsr, buf: &[u8]) -> CaliptraResult<()> {
```
korran, line 296: The signature here is a bit inconsistent this function takes mbox, but the other 2 take &self.

It's probably simplest just to make these top-level private functions:

fn try_fifo_enqueue(mbox: &mut MboxCsr, buf: &[u8]) -> CaliptraResult<()>
fn fifo_enqueue_words(mbox: &mut MboxCsr, buf: &[Unalign<u32>]);
fn fifo_enqueue_bytes(mbox: &mut MboxCsr, buf: &[u8])

drivers/src/mailbox.rs
```diff
@@ -360,6 +376,27 @@ impl MailboxRecvTxn<'_> {
         Ok(())
     }
 
+    /// Writes buf.len() bytes to the mailbox datain reg as dwords
+    fn enqueue(&mut self, buf: &[u8]) -> CaliptraResult<()> {
+        MailboxTxFifo::try_enqueue(self.mbox, buf)
+    }
+
+    /// Writes number of bytes to data length register.
+    pub fn write_dlen(&mut self, dlen: u32) -> CaliptraResult<()> {
```
korran, line 385: Why is this public? We don't anyone from the outside calling it.

drivers/src/mailbox.rs
```diff
@@ -394,15 +431,60 @@ impl MailboxRecvTxn<'_> {
...
+    /// # Arguments
     ///
-    pub fn complete(&mut self, success: bool) -> CaliptraResult<()> {
+    /// * `data` - data buffer.
+    ///
+    /// # Returns
+    ///
+    /// Status of Operation
+    ///
+    pub fn copy_response(&mut self, data: &[u8]) -> CaliptraResult<()> {
```
korran, line 445: Can we make this function private? Nobody should be calling it from the outside.

drivers/src/mailbox.rs
```diff
@@ -394,15 +431,60 @@ impl MailboxRecvTxn<'_> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - data buffer.
+    ///
+    /// # Returns
+    ///
+    /// Status of Operation
+    ///
+    pub fn send_response(&mut self, data: &[u8]) -> CaliptraResult<()> {
```
korran, line 470: Can this function take `mut self` rather than `&mut self`? The transaction should not be used for anything after the response has been sent.

drivers/src/mailbox.rs
```diff
@@ -394,15 +431,60 @@ impl MailboxRecvTxn<'_> {
...
+        self.copy_response(data)?;
+        self.complete(true)
+    }
+
+    ///
+    /// Transitions from Execute or RdyForData-> Idle
+    ///
+    pub fn complete(&mut self, success: bool) -> CaliptraResult<()> {
```
korran, line 478: This can be `mut self` instead of `&mut self`.

drivers/test-fw/src/bin/mailbox_driver_responder.rs
```diff
@@ -68,6 +68,24 @@ extern "C" fn main() {
...
+            0xB000_0000 => {
+                let mut buf = [0u32; 2];
+                let dlen = txn.dlen() as usize;
+                let dlen_words = (dlen + 3) / 4;
+                println!("dlen: {dlen}");
+                for _ in 0..((dlen_words + (buf.len() - 1)) / buf.len()) {
+                    txn.copy_request(&mut buf).unwrap();
+                    println!("buf: {:08x?}", buf);
+                }
+                txn.send_response(&[0x98, 0x76]).unwrap();
```
korran, line 87: Can you add some more tests for the 0-byte case, the 4-byte case, and maybe a 9-byte case?

korran: Zerocopy change saves 74 bytes. nice!

korran:
Zerocopy change saves 74 bytes. nice!

drivers/src/mailbox.rs
nquarton: I'll need Anthony to weigh in on this but I think the "mod fifo" suggestion makes sense. If Anthony agrees I can make that change.

drivers/src/mailbox.rs
nquarton: Thanks! This was an error on my part.

drivers/src/mailbox.rs
nquarton: This is a good point. It made sense for copy_request to allow the caller to separate the copy and the complete call. But I agree. I don't see a reason why the same option would be needed for the response.

drivers/src/mailbox.rs
nquarton: Makes sense, I can change that

drivers/src/mailbox.rs
nquarton: No problem

drivers/test-fw/src/bin/mailbox_driver_responder.rs
nquarton: Yeah I can expand this to cover those cases as well

nquarton (commented):
I'll make these changes. Just need input from @rusty1968 on the one about the MailboxTxFifo struct.

drivers/src/mailbox.rs
rusty1968: try_enqueue instantiates the MailboxTxFifo. That's why I made it an associated function. It is aligned with what we do in the Mailbox abstraction as we instantiate the txn objects. That's why the other functions are methods. We change it to free functions or organizing them in a module. If that's the approach we should move the dequeue functions there too.

drivers/src/mailbox.rs
```diff
@@ -394,15 +431,60 @@ impl MailboxRecvTxn<'_> {
...
+    pub fn copy_response(&mut self, data: &[u8]) -> CaliptraResult<()> {
         if self.state != MailboxOpState::Execute {
             return Err(CaliptraError::DRIVER_MAILBOX_INVALID_STATE);
         }
+
+        self.state = MailboxOpState::RdyForDlen;
+        // Set dlen
+        self.write_dlen(data.len() as u32)?;
+
+        self.state = MailboxOpState::RdyForData;
```
rusty1968, line 454: Should we retire the software state and use mailbox state information in hardware instead? This implementation was created before we had those registers in place.

drivers/src/mailbox.rs
nquarton: The issue I'm hitting with this is the drop trait does not take ownership and it needs to call complete. The two options I see are to leave this as-is or to add a private function for complete that doesn't take ownership that Drop() and the public complete function (now a wrapper) can call. 

If we go with the latter, Is there a naming convention for anything like that?

fn complete_private(&mut self, success: bool) // <- naming convention for this one
pub fn complete(mut self, success: bool) {
    self.complete_private()
}

drivers/src/mailbox.rs
nquarton: Since this is still open, I only updated the naming to use enqueue_words and corrected that to be a private function for now. Does that address the concern with the mismatched signatures since only try_enqueue was supposed to be public?

nquarton: This is updated to be a module now

drivers/src/mailbox.rs
korran: Ah, makes sense. I'm good either way.

FWIW, elsewhere in caliptra this is complete_internal().

drivers/tests/integration_tests.rs
```diff
@@ -450,6 +450,42 @@ fn test_mailbox_soc_to_uc() {
...
+        let resp = model.mailbox_execute(0xC000_0000, &[]).unwrap().unwrap();
+        assert_eq!(model.output().take(usize::MAX), "cmd: 0xc0000000\n");
+        assert_eq!(resp, [0x0A, 0x0B, 0x0C, 0x0D, 0x05, 0x04, 0x03, 0x02, 0x01]);
+    }
+
+    // Test reponse with 0 bytes (still calls copy_response)
+    {
+        let resp = model.mailbox_execute(0xD000_0000, &[]).unwrap();
+        assert_eq!(model.output().take(usize::MAX), "cmd: 0xd0000000\n");
+        assert_eq!(resp, None);
```
korran, line 487: Ooh...this is surprising. To the hardware, a zero-byte response is actually different than a None response (DataReady vs Complete; see  https://github.com/chipsalliance/caliptra-sw/blob/24d9fce148155391e0625e98635d8ae74f93ff03/hw-model/src/lib.rs#L991-L995).

Not sure whether we care enough to make this distinction in the driver.

drivers/src/mailbox.rs
```diff
@@ -390,19 +406,68 @@ impl MailboxRecvTxn<'_> {
...
+    /// # Returns
+    ///
+    /// Status of Operation
+    ///
+    fn copy_response(&mut self, data: &[u8]) -> CaliptraResult<()> {
         if self.state != MailboxOpState::Execute {
             return Err(CaliptraError::DRIVER_MAILBOX_INVALID_STATE);
         }
+
+        if !data.is_empty() {
```
korran, line 428: Is this actually necessary? To the hardware, sending a 0-byte response is different than sending a regular completed response.

nquarton: Oh sure if we'd rather that be the case I can absolutely drop this check. I wasn't sure if "DataReady" with zero size was a case we considered valid or if it was better to prevent that from happening.

korran: I think we should be able to remove this, and then add a separate test-case for the 0-byte response vs the completed (no data) response.

drivers/src/mailbox.rs
nquarton: Cool I will add this as "complete_private"

drivers/tests/integration_tests.rs
nquarton: I will update the driver to allow this case and make sure we test both

drivers/src/mailbox.rs
nquarton: I entirely meant to type "complete_internal" here...

nquarton: I think it makes sense to leave this out for now. The use within cold reset flow in ROM seems to only have this as a mutable reference so there's an issue with calling these terminal functions. I think leaving this and not touching the ROM code is probably the safer bet at this point.

drivers/tests/integration_tests.rs
nquarton: This is done in the latest commit

drivers/src/mailbox.rs
korran: Ok, I'll followup in a future PR.

korran (approved):

mhatrevi (approved):

PR #631, opened 2023-08-11
.github/workflows: Enable fuzz testing as a nightly CI task

benjamindoron:
TODO:
- We should finalise max length to fuzz per target.
- Parameterise `runs-on` to consider using GCP?

### DIFF
--- a/.github/workflows/nightly-fuzzing.yml
+++ b/.github/workflows/nightly-fuzzing.yml
@@ -1,0 +1,161 @@
+name: Fuzz Testing Nightly
+on:
+  schedule:
+    # 3:00 AM PST monday-saturday
+    - cron: '00 11 * * 1-6'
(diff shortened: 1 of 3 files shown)


korran (dismissed):

jhand2 (approved):

jhand2 (approved):

benjamindoron:
Sure, I think the max length values are at least good start now.

benjamindoron:
Necessary follow-ups:
- libFuzzer: GH ignore return value (target crash is not CI error). `max_total_time` does not work as expected, jobs continue to be spawned. image/verify corpus unavailable, though the job is dependent, it ran on another system.
- AFL: Fixup shell script syntax error. Consider disabling for now, or look into exit codes to identify error.

PR #632, opened 2023-08-11
Allocate reserved command codes.

bluegate010:
Unlike unknown commands, these do not trigger a fatal error. Instead, they simply return failures. In the future, these codes can be repurposed if needed. The code values were chosen at random.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor.rs
@@ -19,7 +19,9 @@ use crate::{cprintln, verifier::RomImageVerificationEnv};
 use crate::{pcr, wdt};
 use caliptra_cfi_derive::cfi_impl_fn;
 use caliptra_cfi_lib::{cfi_assert, cfi_assert_eq, cfi_launder};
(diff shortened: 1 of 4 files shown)


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -129,30 +127,40 @@ impl FirmwareProcessor {
...
+                        }
+
+                        cprintln!("");
+                        cprintln!("[afmc] Received Image of size {} bytes" txn.dlen());
+                        report_boot_status(FwProcessorDownloadImageComplete.into());
+                        return Ok(txn);
+                    }
+                    MBOX_CMD_ID_RESERVED_O | MBOX_CMD_ID_RESERVED_1 | MBOX_CMD_ID_RESERVED_2 => {
+                        cprintln!("Reserved command 0x{:08x} received", txn.cmd());
+                        txn.start_txn().complete(false)?;
```
korran, line 155: I assume we want to add a call to report_fw_error_non_fatal(CaliptraError::ROM_UPDATE_RESET_FLOW_INVALID_FIRMWARE_COMMAND.into()) here?

bluegate010:
Closing in favor of https://github.com/chipsalliance/caliptra-sw/pull/680

PR #645, opened 2023-08-15
Add disable attestation mailbox command

sree-revoori1:


runtime/src/disable.rs
```diff
@@ -0,0 +1,56 @@
...
+            .map_err(|_| CaliptraError::RUNTIME_DISABLE_ATTESTATION_FAILED)?;
+        drivers
+            .key_vault
+            .erase_key(KEY_ID_RT_PRIV_KEY)
+            .map_err(|_| CaliptraError::RUNTIME_DISABLE_ATTESTATION_FAILED)?;
+        Self::generate_dice_key(drivers)?;
+        Ok(MailboxResp::default())
+    }
+
+    fn generate_dice_key(drivers: &mut Drivers) -> CaliptraResult<()> {
```
jhand2, line 25: Can you add a comment to this function explaining that this key is derived from an empty CDI slot, so it will not match the key that was certified in the Runtime Alias certificate?

runtime/src/disable.rs
```diff
@@ -0,0 +1,56 @@
...
+            KeyUsage::default().set_ecc_private_key_en(),
+        ));
+
+        drivers.ecc384.key_pair(
+            &KeyReadArgs::new(KEY_ID_TMP).into(),
+            &Array4x12::default(),
+            &mut drivers.trng,
+            key_out,
+        )?;
+        drivers.key_vault.erase_key(KEY_ID_TMP)?;
```
jhand2, line 52: I don't think this is needed. This just derives the public key, but we don't need it for anything.

runtime/src/disable.rs
```diff
@@ -0,0 +1,56 @@
...
+            KeyWriteArgs::new(
+                KEY_ID_RT_PRIV_KEY,
+                KeyUsage::default()
+                    .set_hmac_key_en()
+                    .set_ecc_key_gen_seed_en(),
+            )
+            .into(),
+        )?;
+
+        let key_out = Ecc384PrivKeyOut::Key(KeyWriteArgs::new(
```
jhand2, line 41: nit: Since the result isn't used, you can do `let _ = ...`

runtime/tests/integration_tests.rs
```diff
@@ -127,6 +134,39 @@ fn test_fw_info() {
...
+            .unwrap()
+            .into_ref();
+
+    assert_eq!(
+        resp_hdr.fips_status,
+        MailboxRespHeader::FIPS_STATUS_APPROVED
+    );
+    // Checksum is just going to be 0 because FIPS_STATUS_APPROVED is 0
+    assert_eq!(resp_hdr.chksum, 0);
+    assert_eq!(model.soc_ifc().cptra_fw_error_non_fatal().read(), 0);
```
jhand2, line 167: nit: Can be removed

runtime/tests/integration_tests.rs
```diff
@@ -127,6 +134,39 @@ fn test_fw_info() {
...
+    let resp = model
+        .mailbox_execute(
+            u32::from(CommandId::DISABLE_ATTESTATION),
+            payload.as_bytes(),
+        )
+        .unwrap()
+        .unwrap();
+
+    let resp_hdr: &MailboxRespHeader =
+        LayoutVerified::<&[u8], MailboxRespHeader>::new(resp.as_bytes())
```
jhand2, line 157: nit: Might look a little leaner to use read_from instead of LayoutVerified

runtime/tests/integration_tests.rs
```diff
@@ -127,6 +134,39 @@ fn test_fw_info() {
...
+    let mut model = run_rt_test(None);
+
+    let payload = MailboxReqHeader {
+        chksum: caliptra_common::checksum::calc_checksum(
+            u32::from(CommandId::DISABLE_ATTESTATION),
+            &[],
+        ),
+    };
+
+    let resp = model
```
jhand2, line 148: nit: Can you add a comment that once DPE APIs are enabled, we should make sure the RT alias key in the cert is different from the key that signs DPE keys (i.e. make sure the signature check fails)

runtime/test-fw/src/keyvault_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+        &mut trng,
+        KeyWriteArgs::new(
+            KEY_ID_RT_PRIV_KEY,
+            KeyUsage::default()
+                .set_hmac_key_en()
+                .set_ecc_key_gen_seed_en(),
+        )
+        .into(),
+    )
+    .is_ok());
```
jhand2, line 53: Same. Can just call unwrap instead of assert is_ok

runtime/src/disable.rs
```diff
@@ -0,0 +1,44 @@
...
+        drivers
+            .key_vault
+            .erase_key(KEY_ID_RT_CDI)
+            .map_err(|_| CaliptraError::RUNTIME_DISABLE_ATTESTATION_FAILED)?;
+        drivers
+            .key_vault
+            .erase_key(KEY_ID_RT_PRIV_KEY)
+            .map_err(|_| CaliptraError::RUNTIME_DISABLE_ATTESTATION_FAILED)?;
+
+        // dice key is derived from an empty CDI slot so it will not match the key that was certified in the rt_alias cert
```
jhand2, line 21: nit: I would recommend putting thing comment as a doc comment on the generate_dice_key definition.

jhand2 (approved):

PR #646, opened 2023-08-15
[fix] CFI fixes

mhatrevi:
This change contains the following fixes:

	1. CFI Counter increment is moved to the start of the original function from the wrapper function to make the increment/decrement calls asymmetrical.

	2. Adds a delay to the launder assert checks.

	3. Addresses feedback from original CFI PR: 
	        a. Renamed cfi_ctr.rs to cfi_counter.rs 
                b. Effective launder function use.


mschilder123 (approved):
LGTM

korran (approved):

PR #647, opened 2023-08-16
Rename constructor for error type

sphinxc0re:
Since there is no need to differentiate between a const-time constructor and a runtime constructor in this case. The `_const` suffix seems redundant.

### DIFF
--- a/cfi/lib/src/cfi_ctr.rs
+++ b/cfi/lib/src/cfi_ctr.rs
@@ -97,7 +97,7 @@ impl CfiCounter {
         Self::write(CfiInt::default());
     }
 
-    // Zero the counter
+    /// Zero the counter
     pub fn corrupt() {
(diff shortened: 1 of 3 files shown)


jhand2 (dismissed):
LGTM. Will merge when conflicts are resolved.

jhand2 (approved):

korran (changes requested):
The problem is that CaliptraError::new() will panic if the value is zero. I really want to discourage people from calling it from firmware; it should only be used from const contexts.

Instead, code that wants to create a CaliptraError from a u32 should use TryFrom (with unwrap if it doesn't care about panics). Apprently that commit has not gone in yet; here's the PR:

https://github.com/chipsalliance/caliptra-sw/pull/655

sphinxc0re:
Fair enough. Will close this in favor of #655

PR #651, opened 2023-08-16
Activate authentication via libcaliptra

wmaroneAMD:
This adds a demonstration of signature setup in caliptra fuses of firmware hashes. Additionally, it reorganizes some of the headers and adds types and structs for the caliptra firmware image to ease access to the public key object within.

On startup, the demo will load the firmware and extract the vendor and owner public keys. It will then generate a sha384 digest and program them into the fuses - right as ROM is loaded.

This replicates behavior seen in the smoke test (test/tests/smoke_test.rs).

* caliptra_if_init - Required to disconnect main firmware image load into the test from the first register write. This behavior is unchanged for the ROM.
* Added fixed register values from the smoke test to enable auth.
* Added support for OpenSSL within the generic example.
* Relocated enums and types to unique files.
* Added a header that includes types for the caliptra image.
  > Would prefer this to be generated via cbindgen from the image/types/ rust library, but cbindgen wasn't having it.
* Fixed an issue in caliptra_fuse_array_write that indexed the array incorrectly, causing invalid data to be written to the right registers.
  > Fixed this in the c-binding's implementation as well, just for good measure.


wmaroneAMD:
With regards to cbindgen, it doesn't like the setup in lms-types and complains of an Unsupported call expression and stops processing anything that follows.

rusty1968 (approved):

hw-model/c-binding/examples/api/caliptra_fuses.h
```diff
@@ -4,6 +4,8 @@
 
 #include "caliptra_api.h"
 
+#define AS_U32(x) (x / sizeof(uint32_t))
```
nquarton, line 7: nit: I was a little confused by the naming here at first thinking it was converting/casting as a u32. This seems specific to convert a size or offset. But it's possible I'm understanding it incorrectly and this is clear to others.

This certainly doesn't require a change if you prefer it this way.

nquarton (approved):

hw-model/c-binding/examples/api/caliptra_fuses.h
wmaroneAMD: Yeah it's about just converting the size from bytes to dwords.

Maybe "AS_DWORDS" would be better?

nquarton: That would definitely make me less likely to misunderstand it. But again, this might be unique to me. So feel free to do whatever you see fit here.

korran: Isn't this just ARRAY_SIZE?

naive implementation:

#define ARRAY_SIZE(array) \
    (sizeof(array) / sizeof(*array))

wmaroneAMD: That works too.

nquarton (dismissed):

libcaliptra/examples/generic/main.c
```diff
@@ -5,22 +5,98 @@
...
+    EVP_DigestFinal_ex(ctx, (unsigned char*)md_vendor_pubkey, &mdlen);
+    EVP_MD_CTX_destroy(ctx);
+
+    if (mdlen != SHA384_DIGEST_BYTE_SIZE)
+    {
+        printf("SHA384 digest from OpenSSL is not the correct size! e: %u a: %u", SHA384_DIGEST_BYTE_SIZE, mdlen);
+    }
+
+    ctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(ctx, md, NULL);
```
korran, line 57: openssl provides a utility function for all this:

https://github.com/openssl/openssl/blob/39ed7636e0d8a90512e7ccb811cd0bfcb7a79650/include/openssl/sha.h#L132

korran (dismissed):

libcaliptra/examples/generic/main.c
wmaroneAMD: When I went looking for that, their own documentation said "Applications should use the higher level functions [EVP_DigestInit(3)](https://www.openssl.org/docs/man1.0.2/man3/EVP_DigestInit.html) etc. instead of calling the hash functions directly."

https://www.openssl.org/docs/man1.0.2/man3/SHA384.html

I can swap over to calling them directly if that's preferred.

korran: No, I didn't realize that SHA384() doesn't just call the same implementation as EVP_DigestInit, sigh... Leave it as is.

korran (dismissed):

nquarton (approved):

Issue #606, opened 2023-08-07
Amend STASH_MEASUREMENT command to accept an optional SVN


bluegate010:
Will allow SVNs to be attested later.

PR #653, opened 2023-08-16
Implement STASH_MEASUREMENT mbox command

### LINKED ISSUE #640: Set derive child's target locality to SoC's initial locality in RT
When we call derive child in Caliptra RT after initializing the DPE Instance, we need to make the target locality the SoC's initial locality instead of the current value of 0.

sree-revoori1:
Also adds an svn field into StashMeasurementReq. This change also implements the rand_bytes function in dpe_crypto in order to test derive_child properly.

Fixes https://github.com/chipsalliance/caliptra-sw/issues/606
Fixes #640

### DIFF
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -1,15 +1,18 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_drivers::{Array4x12, KeyId, Sha384, Sha384DigestOp};
+use core::cmp::min;
+
+use caliptra_drivers::{Array4x12, KeyId, Sha384, Sha384DigestOp, Trng};
 use crypto::{AlgLen, Crypto, CryptoError, Digest, EcdsaPub, EcdsaSig, Hasher, HmacSig};
 
 pub struct DpeCrypto<'a> {
     sha384: &'a mut Sha384,
+    trng: &'a mut Trng,
 }
 
 impl<'a> DpeCrypto<'a> {
-    pub fn new(sha384: &'a mut Sha384) -> Self {
-        Self { sha384 }
+    pub fn new(sha384: &'a mut Sha384, trng: &'a mut Trng) -> Self {
+        Self { sha384, trng }
     }
 }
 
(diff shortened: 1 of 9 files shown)


runtime/src/mailbox_api.rs
```diff
@@ -108,6 +108,7 @@ impl MailboxResp {
     pub fn write_to_mbox(&mut self, drivers: &mut Drivers) -> CaliptraResult<()> {
         match self {
             MailboxResp::InvokeDpeCommand(resp) => resp.write_to_mbox(&mut drivers.mbox),
+            MailboxResp::StashMeasurement(resp) => resp.write_to_mbox(&mut drivers.mbox),
```
varuns-nvidia, line 111: Is our general strategy to keep the runtime mailbox implementation as separate from the ROM ala https://github.com/chipsalliance/caliptra-sw/pull/559 ?

jhand2: Yes. The behavior of this mailbox command is a bit different in runtime. At runtime DPE is initialized so we can just call DeriveChild directly. In ROM we have to stash the measurement so that Runtime can do the necessary DPE operations later.

varuns-nvidia (dismissed):
Admittedly haven't learned Rust but lgtm

runtime/src/stash_measurement.rs
```diff
@@ -0,0 +1,49 @@
...
+            };
+            let derive_child_resp = DeriveChildCmd {
+                handle: ContextHandle::default(),
+                data: cmd.measurement,
+                flags: DeriveChildCmd::MAKE_DEFAULT
+                    | DeriveChildCmd::CHANGE_LOCALITY
+                    | DeriveChildCmd::INPUT_ALLOW_CA
+                    | DeriveChildCmd::INPUT_ALLOW_X509,
+                tci_type: u32::from_be_bytes(cmd.metadata),
+                target_locality: TARGET_LOCALITY,
```
jhand2, line 32: We should set this to the user of the mailbox caller

runtime/src/stash_measurement.rs
```diff
@@ -0,0 +1,49 @@
...
+            }
+            .execute(&mut drivers.dpe, &mut env, DPE_LOCALITY);
```
jhand2, line 34: Same with the locality here

runtime/src/mailbox_api.rs
```diff
@@ -203,6 +204,7 @@ pub struct StashMeasurementReq {
     pub hdr: MailboxReqHeader,
     pub metadata: [u8; 4],
     pub measurement: [u8; 48],
+    pub svn: u32,
```
wmaroneAMD, line 207: Please update runtime/README.md to reflect this.

runtime/src/dpe_crypto.rs
```diff
@@ -45,8 +48,21 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
-        Err(CryptoError::NotImplemented)
+    fn rand_bytes(&mut self, dst: &mut [u8]) -> Result<(), CryptoError> {
+        let mut curr_idx = 0;
+        while curr_idx < dst.len() {
+            let trng_bytes = <[u8; 48]>::from(
+                self.trng
+                    .generate()
+                    .map_err(|_| CryptoError::CryptoLibError)?,
+            );
+            let bytes_to_write = min(dst.len() - curr_idx, 48);
```
jhand2, line 59: nit: Instead of hard-coding 48, can we use trng_bytes.len()?

runtime/src/dpe_crypto.rs
```diff
@@ -45,8 +48,21 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+            dst.get_mut(curr_idx..curr_idx + bytes_to_write)
+                .ok_or(CryptoError::Size)?
+                .copy_from_slice(&trng_bytes[..bytes_to_write]);
```
jhand2, line 62: nit: I wonder if doing a normal subslice (`dst[curr_idx..curr_idx + bytes_to_write].copy_from_slice(&trng_bytes[..bytes_to_write]`) doesn't panic now. Not a big deal either way though. If it panics, what you have here is fine.

jhand2 (approved):

wmaroneAMD (approved):

PR #656, opened 2023-08-17
Rework how the DPE response is written to the mailbox

jhand2:
Update how DPE is sliced to prevent needing the special write_to_mbox implementation. This maintains the property that extra bytes aren't written to the mailbox if they aren't needed.


runtime/tests/integration_tests.rs
```diff
@@ -174,17 +174,16 @@ fn test_invoke_dpe_get_profile_cmd() {
...
         .mailbox_execute(u32::from(CommandId::INVOKE_DPE), cmd.as_bytes())
         .unwrap()
         .expect("We should have received a response");
+    assert!(resp.len() <= std::mem::size_of::<InvokeDpeResp>());
+    resp.resize(std::mem::size_of::<InvokeDpeResp>(), 0);
 
-    let resp_hdr: &InvokeDpeResp = LayoutVerified::<&[u8], InvokeDpeResp>::new(resp.as_bytes())
-        .unwrap()
-        .into_ref();
+    let resp_hdr = InvokeDpeResp::read_from(resp.as_bytes()).unwrap();
```
jhand2, line 184: @korran I'm interested if you have any suggestions for a cleaner way to do this.

InvokeDpeResp has a data field with a max size, and we only want to write the bytes to the mailbox that are actually populated (so that we don't have to send/receive unnecessary null bytes over the mailbox). On the Caliptra side, this is fine.

On the caller side it would be size if the caller could cleanly use zerocopy to say "Populate the header and the first N bytes of `data` with the response buffer. For the remaining bytes, fill it with zeros." I couldn't find a nice way to do this, so I just resized the response Vec to be larger.

korran: You should be able to copy directly into the InvokeDpeResp struct:

```
let mut resp_hdr = InvokeDpeResp::default();
resp_hdr.as_bytes_mut()[..resp.len()].copy_from_slice(&resp)
```

You probably still want to enforce that resp.len() be an expected value, no?

runtime/src/mailbox_api.rs
```diff
@@ -65,7 +63,7 @@ impl MailboxResp {
             MailboxResp::GetIdevCsr(resp) => resp.as_bytes(),
             MailboxResp::GetLdevCert(resp) => resp.as_bytes(),
             MailboxResp::StashMeasurement(resp) => resp.as_bytes(),
-            MailboxResp::InvokeDpeCommand(resp) => resp.as_bytes(),
+            MailboxResp::InvokeDpeCommand(resp) => resp.as_bytes_partial(),
```
nquarton, line 66: Should we cover the as_bytes_mut case too? I think it's only currently used in the populate_chksum function in this file and it doesn't matter there technically.

runtime/tests/integration_tests.rs
korran: FWIW, you probably want some sort of helper function:

```
trait MailboxCommand {
  const ID: u32;
  type Request: zerocopy::AsBytes;
  type Response: zerocopy::AsBytes;
}

fn mailbox_invoke<T: MailboxCommand>(req: &T::Request) -> T::Response;
```

jhand2: Ah, ya that's much better :) Thanks!

runtime/src/mailbox_api.rs
jhand2: Ah, ya I'll update it there as well just to be sure. It is interesting though because the checksum will be the same as long as all the trailing bytes are 0 😄

nquarton: Yeah I can't think of a scenario where it would actually break anything since we use "as_bytes" for the mailbox send call. Just a question of covering it for completeness. I wish there were a cleaner way to implement functions like this that only differ in mutability.

runtime/tests/integration_tests.rs
jhand2: Done. I will follow up with a helper and reoeganize all the callers.

rusty1968 (approved):

nquarton (approved):

PR #657, opened 2023-08-18
libcaliptra: Stash Management plus cleanup

wmaroneAMD:
This adds support for the Stash Management RT FW command.

* Added structs for stash management request and completions.
  * Including svn field on Sree's branch
* Resolved an issue with mailbox request status handling.
* Resolved an issue with checksum calculation on completion.
* Unified generic command response handling.
* Added test in generic flow that asserts STASH_MEASUREMENT command will fail, must be updated once merged.

### DIFF
--- a/libcaliptra/examples/generic/main.c
+++ b/libcaliptra/examples/generic/main.c
@@ -69,15 +69,41 @@ int main(int argc, char *argv[])
         status = caliptra_get_fips_version(&version);
         if (status)
         {
-            printf("Caliptra C API Integration Test Failed: %x\n", status);
-            return status;
+            printf("Get FIPS Version failed!\n");
+            break;
+        }
+        else
+        {
+            printf("FIPS_VERSION = mode: 0x%x, fips_rev (0x%x, 0x%x, 0x%x), name %s \n", version.mode,
+                version.fips_rev[0], version.fips_rev[1], version.fips_rev[2], version.name);
+        }
+
+        // Need some representative values for these, see below.
+        struct caliptra_stash_measurement_req r = {0};
+        struct caliptra_stash_measurement_resp c = {0};
+
+        status = caliptra_stash_measurement(&r, &c);
+
+        if (status)
+        {
+            printf("Stash measurement failed!\n");
+        }
+        else
+        {
+            printf("Stash measurement: OK\n");
         }
 
         break;
     }
-    printf("Caliptra C API Integration Test Passed: \n\tFIPS_VERSION = mode: 0x%x, fips_rev (0x%x, 0x%x, 0x%x), name %s \n", version.mode,
-                version.fips_rev[0], version.fips_rev[1], version.fips_rev[2], version.name);
+
+    if (status)
+    {
+        printf("Caliptra C API Integration Test Failed: %x\n", status);
+    }
+    else
+    {
+        printf("Caliptr a C API Integration Test Passed!\n");
+    }
+
     return 0;
 }
-
-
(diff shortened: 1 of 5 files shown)


libcaliptra/examples/generic/main.c
```diff
@@ -31,15 +31,42 @@ int main(int argc, char *argv[])
...
+        status = caliptra_stash_measurement(&r, &c);
+
+        if (!status)
+        {
+            printf("Stash Measurement command not ready, returned success!");
+            status = -EINVAL;
+        }
+        else
+        {
+            printf("STASH_MEASUREMENT not yet implemented, see https://github.com/chipsalliance/caliptra-sw/pull/653\n");
```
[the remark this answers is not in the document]
wmaroneAMD: Absolutely.

libcaliptra/inc/caliptra_types.h
```diff
@@ -52,3 +52,14 @@ struct caliptra_fips_version {
...
 };
 
+struct caliptra_stash_measurement_req {
+    caliptra_checksum checksum;
+    uint8_t           metadata[4];
+    uint8_t           measurement[48];
+    uint32_t          svn;
+};
+
+struct caliptra_stash_measurement_cpl {
```
jhand2, line 62: nit: I would recommend aligned with https://github.com/chipsalliance/caliptra-sw/blob/main/common/src/mailbox_api.rs and using `resp` instead of `cpl` as a prefix

libcaliptra/inc/caliptra_api.h
```diff
@@ -22,5 +22,8 @@ int caliptra_upload_fw(struct caliptra_buffer *fw_buffer);
 // Read Caliptra FIPS Version
 int caliptra_get_fips_version(struct caliptra_fips_version *version);
 
+// Stash Measurement command
+int caliptra_stash_measurement(struct caliptra_stash_measurement_req *req, struct caliptra_stash_measurement_cpl *cpl);
```
jhand2, line 26: (maybe for a different PR)

The error codes returned by this library are a bit confusing. It seems like they might be a Caliptra error code (from https://github.com/chipsalliance/caliptra-sw/blob/main/error/src/lib.rs) or a linux errno. Maybe we should align on just using caliptra error codes and then define a range of error codes that are client-side failures (e.g. an invalid checksum or invalid input arguements)

wmaroneAMD: At the top level they should (or have been) only unix-style error codes. caliptra_mbox_read_status returns Caliptra error codes, but does not return them to the caller.

libcaliptra/inc/caliptra_types.h
wmaroneAMD: Yeah I was thinking about that. Just pushed up a change that aligns them.

I'll do this on my pending branch too.

libcaliptra/inc/caliptra_api.h
FerralCoder: Details about the return codes used and what they indicate should be added to caliptra_api.h. Right now a caller has no way to know how to interpret the return value.

wmaroneAMD: https://github.com/chipsalliance/caliptra-sw/issues/678

jhand2: Sounds good, thanks

jhand2 (approved):

jhand2 (approved):

fdamato (approved):

PR #659, opened 2023-08-18
[feat] ROM mailbox infra for FIPS commands

mhatrevi:
This PR updates the ROM logic for accepting the following FIPS commands:
	1. SHUTDOWN
	2. VERSION
	3. SELF_TEST


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -122,39 +121,47 @@ impl FirmwareProcessor {
     /// handle_fatal_error is called. This prevents a race condition where the SoC
```
korran, line 121: Make it clear in this documentation that the returned transaction handle is for the FIRMWARE_LOAD command.

korran (approved):

PR #660, opened 2023-08-20
[fix] Update ECC384 and LMS verify function signatures

mhatrevi:
This PR changes return types on ECC384 and LMS verify to return verify_r and candidate key values respectively. The callers of these functions now compare the values with the expected to determine if the signature verification succeeded.


drivers/src/ecc384.rs
```diff
@@ -376,9 +378,42 @@ impl Ecc384 {
         digest: &Ecc384Scalar,
         signature: &Ecc384Signature,
     ) -> CaliptraResult<Ecc384Result> {
+        // Get the verify r result
```
korran, line 381: Can we #[inline(always)] this function?

drivers/src/ecc384.rs
```diff
@@ -376,9 +378,42 @@ impl Ecc384 {
...
+        let mut verify_r = self.verify_r(pub_key, digest, signature)?;
+
+        // compare the hardware generate `r` with one in signature
+        let result = if verify_r == signature.r {
```
korran, line 385: We need to make sure this comparison is unrolled.

drivers/src/ecc384.rs
korran: Hmm, maybe I'm misunderstand how you're planning on using verify_r

I originally thought verify() was going to be inlined, but I guess instead you're planning to call verify_r directly in a future PR?

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -191,13 +189,11 @@ impl FmcAliasLayer {
         env.key_vault.erase_key(auth_priv_key)?;
 
         // Verify the signature of the `To Be Signed` portion
-        let result = Crypto::ecdsa384_verify(env, auth_pub_key, tbs.tbs(), sig)?;
-        if cfi_launder(result) == Ecc384Result::Success {
-            cfi_assert!(result == Ecc384Result::Success);
-        } else {
-            cfi_assert!(result != Ecc384Result::Success);
+        let mut verify_r = Crypto::ecdsa384_verify(env, auth_pub_key, tbs.tbs(), sig)?;
+        if verify_r != sig.r {
```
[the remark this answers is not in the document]
mhatrevi: Added a (extra) cfi_assert_eq for checking the r value on success condition.

korran (approved):

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -245,13 +245,13 @@ impl InitDevIdLayer {
         let sig = okmutref(&mut sig)?;
 
         // Verify the signature of the `To Be Signed` portion
-        let result = Crypto::ecdsa384_verify(env, &key_pair.pub_key, tbs.tbs(), sig)?;
-        if cfi_launder(result) == Ecc384Result::Success {
-            cfi_assert!(result == Ecc384Result::Success);
-        } else {
-            cfi_assert!(result != Ecc384Result::Success);
+        let mut verify_r = Crypto::ecdsa384_verify(env, &key_pair.pub_key, tbs.tbs(), sig)?;
+        if cfi_launder(verify_r != sig.r) {
```
korran, line 249: Does this actually do what you intend? I think the compiler is still allowed to optimize the input to cfi_launder; it's the output that makes a difference.

Perhaps this should be cfi_launder(verify_r) != sig.r?

korran: Yeah, this is no good: For example:

```
use std::hint::black_box;

pub fn func() -> bool {
    let val = 42;
    black_box(35 == val)
}
ASM:

playground::func:
	movb	$0, -1(%rsp)
	leaq	-1(%rsp), %rax
	#APP
	#NO_APP
	movzbl	-1(%rsp), %eax
	retq
```

As you can see, the compiler optimized out the comparison. However,

```
use std::hint::black_box;

pub fn func() -> bool {
    let val = 42;
    35 == black_box(val)
}
playground::func:
	movl	$42, -4(%rsp)
	leaq	-4(%rsp), %rax
	#APP
	#NO_APP
	cmpl	$35, -4(%rsp)
	sete	%al
	retq
```

korran (changes requested):

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: I verified this check with and without launder; there were one and two memcmps respectively; that said, I think your suggestion makes; I'll change it if I still see the desired assembly.

korran: Take a look at func5 here:

https://godbolt.org/z/8vMqvsKMn

mhatrevi: cfi_launder(verify_r) != sig.r does not work; the compiler is stripping the cfi_assert! in this case. The original check is correct; I can use references though.

korran (approved):

PR #663, opened 2023-08-21
Implement tracking of minimum SVN in FMC.

bluegate010:
Also implements reporting of min-SVN from RT FW.


fmc/src/flow/rt_alias.rs
```diff
@@ -113,6 +113,10 @@ impl RtAliasLayer {
             .set_pcr_lock(caliptra_common::RT_FW_JOURNEY_PCR);
         cprintln!("[alias rt] Lock RT PCRs Done");
 
+        cprintln!("[alias rt] Populate DV");
```
rusty1968, line 116: Should we use bootstatus here?

runtime/src/lib.rs
```diff
@@ -255,6 +257,8 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         }
         #[cfg(feature = "test_only_commands")]
         CommandId::TEST_ONLY_HMAC384_VERIFY => HmacVerifyCmd::execute(drivers, cmd_bytes),
+        #[cfg(feature = "test_only_commands")]
+        CommandId::TEST_ONLY_INC_MIN_SVN => IncMinSvnCmd::execute(drivers),
```
jhand2, line 261: This command is a bit odd in that it should never succeed (even as a test-only command). I wonder if it makes more sense as part of one of the tests in `runtime/test-fw` (still as an RT test)."

Could be simple enough to add it as a boot tests (just assert that the RT min SVN slot is locked)

runtime/src/info.rs
```diff
@@ -7,16 +7,29 @@ use caliptra_drivers::CaliptraResult;
 pub struct FwInfoCmd;
 impl FwInfoCmd {
     pub(crate) fn execute(drivers: &Drivers) -> CaliptraResult<MailboxResp> {
-        let fuse_bank = drivers.soc_ifc.fuse_bank();
-
-        let runtime_svn = fuse_bank.runtime_fuse_svn();
-        let fmc_manifest_svn = fuse_bank.fmc_fuse_svn();
+        let runtime_svn = drivers.data_vault.rt_svn();
```
rusty1968, line 10: Need to use the handoff table instead of the driver api.

jhand2 (approved):

drivers/src/hand_off.rs
```diff
@@ -247,6 +248,9 @@ pub struct FirmwareHandoffTable {
     /// Index of RT SVN value in the Data Vault
     pub rt_svn_dv_hdl: HandOffDataHandle,
 
+    /// Index of RT Min SVN value in the Data Vault
+    pub rt_min_svn_dv_hdl: HandOffDataHandle,
```
FerralCoder, line 252: please keep documentation in sync by adding field to fmc/README.md

bluegate010: Done.

FerralCoder (dismissed):

jhand2 (dismissed):

jhand2 (approved):

korran (approved):

rusty1968 (approved):

PR #667, opened 2023-08-21
Reset non-fatal error register upon new command.

korran:


jhand2:
I also have a PR for this here: https://github.com/chipsalliance/caliptra-sw/pull/548 which includes an RT test. Just need to fix a test failure and add the same logic to ROM.

I'm fine with merging either PR.

korran:
> I also have a PR for this here: #548 which includes an RT test. Just need to fix a test failure and add the same logic to ROM.
> 
> I'm fine with merging either PR.

I stole the test-case from your PR and fixed the two failures.

jhand2 (approved):
Thanks!!

mhatrevi (dismissed):

mhatrevi (approved):

jhand2 (approved):

PR #668, opened 2023-08-22
Replace runtime key digest calculation with build time

wmaroneAMD:
This replaces OpenSSL usage with a build-time step that extracts the keys and generates a digest that is then plugged into an ELF section.

### DIFF
--- a/libcaliptra/examples/hwmodel/interface.c
+++ b/libcaliptra/examples/hwmodel/interface.c
@@ -9,7 +9,6 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
-#include <openssl/evp.h>
 
 #include <caliptra_top_reg.h>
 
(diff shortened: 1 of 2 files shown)


fdamato (approved):

libcaliptra/examples/hwmodel/interface.c
```diff
@@ -24,14 +23,14 @@
 
 struct caliptra_model_init_params init_params;
 struct caliptra_fuses fuses = {0};
+
+__attribute__((section("VPK_HASH"))) uint8_t vpk_hash[48];
+__attribute__((section("OPK_HASH"))) uint8_t opk_hash[48];
```
jhand2, line 28: It seems like it might be good to have the option to do this at runtime or build time.

If the SoC FW which uses libcaliptra is always bundled with a specific Caliptra image then this works. But if not, I think it is still desirable to extract they keys at runtime. wdyt?

wmaroneAMD: I think that's outside our scope? How that key digest is delivered to the firmware is going to be device and implementer specific (i.e. the SoC may do that digest with whatever library/hw it has,) and this way is simply a reflection on the "hwmodel" behavior.

We might need an "Developer's Guide" that explains everything.

wmaroneAMD: Also, IIRC the key digest is supposed to be burned in at manufacture, so most of the time you won't have the digest step at all.

jhand2: Ok, that makes sense. Agreed that if the SoC vendor needs to do something different they can handle that themselves.

jhand2 (approved):

PR #673, opened 2023-08-22
Get Profile draft changes

premalathak12:


hw-model/c-binding/examples/api/c_wrapper.c
```diff
@@ -0,0 +1,76 @@
...
+    uint32_t data_size = sizeof(CommandHdr);
+    const uint32_t OP_INVOKE_DPE_COMMAND = 0x44504543;
+    printf("%u\n", calculate_caliptra_checksum(OP_INVOKE_DPE_COMMAND, (uint8_t*)&cmdHdr, data_size));
+
+    INVOKE_DPE_COMMAND* invokeCmd = (INVOKE_DPE_COMMAND*)malloc(sizeof(INVOKE_DPE_COMMAND));
+    if (invokeCmd != NULL) {
+        printf("****INVOKE 3**********\n");
+        fflush(stdout);
+        invokeCmd->data_size = data_size;
+        invokeCmd->chksum = calculate_caliptra_checksum(OP_INVOKE_DPE_COMMAND, (uint8_t*)&cmdHdr, data_size);
```
jhand2, line 59: The data that is checksummed should include everything after the `chksum` member. I would recommend defining the command structure like this:

```c
typedef struct {
    uint32_t chksum;
    uint32_t data_size;
    uint8_t data[MAX_DATA_SIZE];
} INVOKE_DPE_COMMAND;
```

And then calculate the checksum like this:

```c
calculate_caliptra_checksum(OP_INVOKE_DPE_COMMAND, (uint8_t*)&cmd.data_size, sizeof(cmd->data_size) + cmd->data_size);
```

This will compute the checksum over all the bytes in the request, excluding the checksum field itself, and including the used bytes in `data`.

hw-model/c-binding/examples/api/caliptra_api.c
```diff
@@ -170,3 +194,33 @@ int caliptra_upload_fw(struct caliptra_model *model, struct caliptra_buffer *fw_
...
+
+
+int caliptra_get_profile(struct caliptra_model *model, struct caliptra_buffer *fw_buffer,uint32_t statusCheckRead, struct caliptra_output *test)
+{
+    const uint32_t OP_INVOKE_DPE_COMMAND = 0x44504543;
+    uint32_t *status;
+    int mStatus;
+    const uint32_t error_code = 0x3003000c;
+       struct caliptra_buffer out_buf = {
+        .data = (uint8_t *)test,
```
jhand2, line 206: `caliptra_output.data` is a variable-length array, so it has no backing memory. Any read/write to `data.data` is a buffer overflow.

jhand2:
Done in https://github.com/chipsalliance/caliptra-sw/pull/940

PR #679, opened 2023-08-23
Implement Crypto trait for Caliptra

sree-revoori1:


runtime/src/dpe_crypto.rs
```diff
@@ -78,21 +96,55 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+            AlgLen::Bit384 => {
+                let mut hmac_key = [0u8; AlgLen::Bit384.size()];
+                hmac_key.copy_from_slice(measurement.bytes());
+                hmac384_kdf(
+                    self.hmac384,
+                    Hmac384Key::Array4x12(&Array4x12::from(hmac_key)),
+                    b"derive_cdi",
+                    Some(info),
+                    &mut self.trng,
+                    KeyWriteArgs::new(KEY_ID_RT_CDI, KeyUsage::default().set_hmac_data_en()).into(),
```
jhand2, line 114: A few things here

* I think the input key should be `KEY_ID_RT_CDI`. We want to derive a CDI based on that
* The context argument can contain both the `info` and `measurement` parameters. The most convenient thing to do may be to hash them together and pass the output digest as the context.
* For the output CDI, we'll want to claim a keyid that isn't already being used by ROM/FMC.

runtime/src/dpe_crypto.rs
```diff
@@ -78,21 +96,55 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+            AlgLen::Bit256 => Err(CryptoError::Size),
+            AlgLen::Bit384 => {
+                hmac384_kdf(
+                    self.hmac384,
+                    KeyReadArgs::new(*cdi).into(),
+                    label,
+                    Some(info),
+                    &mut self.trng,
+                    KeyWriteArgs::new(
+                        KEY_ID_RT_PRIV_KEY,
```
jhand2, line 139: We'll want to pick another keyvault slot that's not currently in use. It looks like ROM and FMC use up to keyvault slot 5. So we can use 6 & 7 for the CDI and Private Key

runtime/src/dpe_crypto.rs
```diff
@@ -105,37 +157,75 @@ impl<'a> Crypto for DpeCrypto<'a> {
 
     fn ecdsa_sign_with_alias(
         &mut self,
-        _algs: AlgLen,
-        _digest: &Digest,
+        algs: AlgLen,
+        digest: &Digest,
     ) -> Result<EcdsaSig, CryptoError> {
-        Err(CryptoError::NotImplemented)
+        self.ecdsa_sign_with_derived(algs, digest, &KEY_ID_FMC_PRIV_KEY)
```
jhand2, line 163: Here we should sign with KEY_ID_RT_PRIV_KEY

jhand2 (approved):

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +101,171 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+                hmac384_kdf(
+                    self.hmac384,
+                    KeyReadArgs::new(*cdi).into(),
+                    label,
+                    Some(info),
+                    self.trng,
+                    KeyWriteArgs::new(
+                        KEY_ID_DPE_CDI,
+                        KeyUsage::default().set_ecc_key_gen_seed_en(),
+                    )
```
jhand2, line 154: This has a bit of a side-effect in that the input CDI is being modified. Maybe we should use another KV slot as scratch space (we can clear it before exiting the function)

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +101,171 @@ impl<'a> Crypto for DpeCrypto<'a> {
 
     fn derive_cdi(
```
jhand2, line 102: I think we may want to implement the `drop` trait for DpeCrypto and use it to clear the keyvault slots we used when the struct goes out of scope. drop is similar to a destructor.

```rust
impl Drop for DpeCrypto {
    fn drop(&mut self) {...}
}
```

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +101,171 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+                    self.hmac384,
+                    KeyReadArgs::new(KEY_ID_RT_CDI).into(),
+                    b"derive_cdi",
+                    Some(context.bytes()),
+                    self.trng,
+                    KeyWriteArgs::new(KEY_ID_DPE_CDI, KeyUsage::default().set_hmac_data_en())
+                        .into(),
+                )
+                .map_err(|_| CryptoError::CryptoLibError)?;
+                Ok(KEY_ID_RT_CDI)
```
jhand2, line 130: We should be returning KEY_ID_DPE_CDI here right?

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +112,193 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
-        _info: &[u8],
-    ) -> Result<Self::PrivKey, CryptoError> {
-        Err(CryptoError::NotImplemented)
+                hmac384_kdf(
+                    self.hmac384,
+                    KeyReadArgs::new(KEY_ID_RT_CDI).into(),
+                    b"derive_cdi",
+                    Some(context.bytes()),
+                    self.trng,
+                    KeyWriteArgs::new(KEY_ID_DPE_CDI, KeyUsage::default().set_hmac_data_en())
```
jhand2, line 137: The KeyUsage here should be:

```rust
KeyUsage::default()
    .set_hmac_key_en()
    .set_ecc_key_gen_seed_en()
```

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +112,193 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+                            KEY_ID_DPE_PRIV_KEY,
+                            KeyUsage::default().set_ecc_private_key_en(),
+                        )
+                        .into(),
+                    )
+                    .map_err(|_| CryptoError::CryptoLibError)?;
+                let pub_key = EcdsaPub {
+                    x: CryptoBuf::new(pub_key.x.as_bytes(), algs).map_err(|_| CryptoError::Size)?,
+                    y: CryptoBuf::new(pub_key.y.as_bytes(), algs).map_err(|_| CryptoError::Size)?,
+                };
```
jhand2, line 183: Using as_bytes doesn't work here. The reason is that the hardware consumes little endian words, while DPE is expecting to get the value as big-endian. Take the example X value `0xAABBCCDD_00112233`

When written to Caliptra's registers, this will have the byte representation `0xDDCCBBAA_33221100`. So we need to do endianness conversions between the two types. But as_bytes just taskes the raw byte representation.

Something like this should work (the `From` implementation in `drivers/src/array.rs` does the endianness conversion)

```rust
CryptoBuf::new(&<[u8; AlgLen::Bit384.size()]>::from(pub_key.x), algs).map_err(|_| CryptoError::Size)?
```

runtime/src/dpe_crypto.rs
```diff
@@ -78,64 +112,193 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
-        _algs: AlgLen,
-        _digest: &Digest,
+        algs: AlgLen,
+        digest: &Digest,
     ) -> Result<EcdsaSig, CryptoError> {
-        Err(CryptoError::NotImplemented)
+        let pub_key = EcdsaPub {
+            x: CryptoBuf::new(self.rt_pub_key.x.as_bytes(), algs).map_err(|_| CryptoError::Size)?,
+            y: CryptoBuf::new(self.rt_pub_key.y.as_bytes(), algs).map_err(|_| CryptoError::Size)?,
+        };
```
jhand2, line 197: Same here

jhand2 (approved):

PR #684, opened 2023-08-24
Execute FIPS Self Test in Runtime

rusty1968:
This commit enables Runtime to run FIPS Self Tests as follows : KATs, verification of firmware image in ICCM, and ROM integrity test. 

- Light refactor to decouple KAT crate from Rom dependency.
- Light refactor to decouple Image verification crate from ROM dependency.
- Ability for RT to defer execution of self test (basic job control). 

Runtime copies image from ICCM to mailbox SRAM , so it needs to fake a send a transaction and hold the mailbox lock while the image is being validated,. It defers the actual self test execution until the mailbox lock can be acquired. SoC must "poll" for self test completion.

- SoC must send a FIPS self test command one time to schedule the execution.
- Caliptra holds the mailbox lock while the self test is in progress.  
- Failure of FIPS self test is handled as fatal error.
- Upon successful self-test completion the mailbox lock is released and SoC will receive a positive acknowledgement when FIPS Self test command is re-issued.

### DIFF
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -64,7 +64,7 @@ pub const FMC_FAKE_WITH_UART: FwId = FwId {
 pub const APP_WITH_UART: FwId = FwId {
     crate_name: "caliptra-runtime",
     bin_name: "caliptra-runtime",
-    features: &["emu", "test_only_commands"],
+    features: &["emu", "test_only_commands", "fips_self_test"],
     workspace_dir: None,
 };
 
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -312,6 +312,7 @@ impl CaliptraError {
     pub const RUNTIME_GET_DEVID_CERT_FAILED: CaliptraError = CaliptraError::new_const(0x000E0013);
     pub const RUNTIME_CERT_CHAIN_CREATION_FAILED: CaliptraError =
         CaliptraError::new_const(0x000E0014);
+    pub const RUNTIME_SELF_TEST_IN_PROGREESS: CaliptraError = CaliptraError::new_const(0x000E0015);
 
     /// FMC Errors
     pub const FMC_GLOBAL_NMI: CaliptraError = CaliptraError::new_const(0x000F0001);
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -312,7 +312,7 @@ impl CaliptraError {
     pub const RUNTIME_GET_DEVID_CERT_FAILED: CaliptraError = CaliptraError::new_const(0x000E0013);
     pub const RUNTIME_CERT_CHAIN_CREATION_FAILED: CaliptraError =
         CaliptraError::new_const(0x000E0014);
-    pub const RUNTIME_SELF_TEST_IN_PROGREESS: CaliptraError = CaliptraError::new_const(0x000E0015);
+    pub const RUNTIME_SELF_TEST_IN_PROGRESS: CaliptraError = CaliptraError::new_const(0x000E0015);
 
     /// FMC Errors
     pub const FMC_GLOBAL_NMI: CaliptraError = CaliptraError::new_const(0x000F0001);
--- a/kat/src/lib.rs
+++ b/kat/src/lib.rs
@@ -16,6 +16,7 @@ Abstract:
 
 mod ecc384_kat;
 mod hmac384_kat;
+mod kats_env;
 mod lms_kat;
 mod sha1_kat;
 mod sha256_kat;
@@ -25,8 +26,45 @@ mod sha384acc_kat;
 pub use caliptra_drivers::{CaliptraError, CaliptraResult};
 pub use ecc384_kat::Ecc384Kat;
 pub use hmac384_kat::Hmac384Kat;
+pub use kats_env::KatsEnv;
 pub use lms_kat::LmsKat;
 pub use sha1_kat::Sha1Kat;
 pub use sha256_kat::Sha256Kat;
 pub use sha384_kat::Sha384Kat;
 pub use sha384acc_kat::Sha384AccKat;
+
+use caliptra_drivers::cprintln;
+
+/// Execute Known Answer Tests
+///
+/// # Arguments
+///
+/// * `env` - ROM Environment
+pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<()> {
+    cprintln!("[kat] ++");
+
+    cprintln!("[kat] sha1");
+    Sha1Kat::default().execute(env.sha1)?;
+
+    cprintln!("[kat] SHA2-256");
+    Sha256Kat::default().execute(env.sha256)?;
+
+    cprintln!("[kat] SHA2-384");
+    Sha384Kat::default().execute(env.sha384)?;
+
+    cprintln!("[kat] SHA2-384-ACC");
+    Sha384AccKat::default().execute(env.sha384_acc)?;
+
+    cprintln!("[kat] ECC-384");
+    Ecc384Kat::default().execute(env.ecc384, env.trng)?;
+
+    cprintln!("[kat] HMAC-384");
+    Hmac384Kat::default().execute(env.hmac384, env.trng)?;
+
+    cprintln!("[kat] LMS");
+    LmsKat::default().execute(env.sha256, env.lms)?;
+
+    cprintln!("[kat] --");
+
+    Ok(())
+}
--- a/runtime/src/fips.rs
+++ b/runtime/src/fips.rs
@@ -60,10 +60,74 @@ impl FipsVersionCmd {
 #[cfg(feature = "fips_self_test")]
 pub mod fips_self_test_cmd {
     use super::*;
+    use caliptra_common::{verifier::FirmwareImageVerificationEnv, FMC_ORG, RUNTIME_ORG};
+    use caliptra_drivers::ResetReason;
+    use caliptra_image_verify::ImageVerifier;
+    use zerocopy::AsBytes;
+
+    // Helper function to create a mutable slice from a memory region
+    unsafe fn create_slice(org: u32, size: usize) -> &'static mut [u8] {
+        let ptr = org as *mut u8;
+        core::slice::from_raw_parts_mut(ptr, size)
+    }
+
+    fn copy_and_verify_image(env: &mut Drivers) -> CaliptraResult<()> {
+        cprintln!("set dlen");
+        env.mbox.set_dlen(
+            env.persistent_data.get().manifest1.size
+                + env.persistent_data.get().manifest1.fmc.size
+                + env.persistent_data.get().manifest1.runtime.size,
+        );
+
+        cprintln!("copy manifest");
+        env.mbox
+            .copy_bytes_to_mbox(env.persistent_data.get().manifest1.as_bytes())?;
+
+        cprintln!("copy fmc");
+        let fmc = unsafe {
+            create_slice(
+                FMC_ORG,
+                env.persistent_data.get().manifest1.fmc.size as usize,
+            )
+        };
+        env.mbox.copy_bytes_to_mbox(fmc.as_bytes())?;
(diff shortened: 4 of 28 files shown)


runtime/src/fips.rs
```diff
@@ -80,14 +59,47 @@ impl FipsVersionCmd {
     }
 }
 
-pub struct FipsSelfTestCmd;
-impl FipsSelfTestCmd {
+pub mod fips_self_test_cmd {
```
FerralCoder, line 62: nit: this name is now inconsistent with the rest. Either change all or change none.

rusty1968: this is rust convention for modules.

builder/src/lib.rs
```diff
@@ -57,7 +57,7 @@ pub const FMC_WITH_UART: FwId = FwId {
 pub const APP_WITH_UART: FwId = FwId {
     crate_name: "caliptra-runtime",
     bin_name: "caliptra-runtime",
-    features: &["emu", "test_only_commands"],
+    features: &["emu", "test_only_commands", "fips_self_test"],
```
jhand2, line 60: Could you look at how big the `caliptra-runtime` image is with this feature enabled?

I usually just do `readelf -S path/to/caliptra-runtime` and then add the sizes of the `.text` and `.rodata` sections.

kat/src/lib.rs
```diff
@@ -25,8 +26,45 @@ mod sha384acc_kat;
...
+    Sha384AccKat::default().execute(env.sha384_acc)?;
+
+    cprintln!("[kat] ECC-384");
+    Ecc384Kat::default().execute(env.ecc384, env.trng)?;
+
+    cprintln!("[kat] HMAC-384");
+    Hmac384Kat::default().execute(env.hmac384, env.trng)?;
+
+    cprintln!("[kat] LMS");
+    LmsKat::default().execute(env.sha256, env.lms)?;
```
jhand2, line 65: Do we need to execute the LMS KAT? We don't use LMS except for to do the ImageVerification (which is also just to satisfy FIPS)

runtime/src/fips.rs
```diff
@@ -80,13 +59,99 @@ impl FipsVersionCmd {
...
+
+        cprintln!("copy manifest");
+        env.mbox.copy_bytes_to_mbox(env.manifest.as_bytes())?;
+
+        cprintln!("copy fmc");
+        let fmc = unsafe { crate::create_slice(FMC_ORG, env.manifest.fmc.size as usize) };
+        env.mbox.copy_bytes_to_mbox(fmc.as_bytes())?;
+        cprintln!("copy rt");
+        let rt = unsafe { crate::create_slice(RUNTIME_ORG, env.manifest.runtime.size as usize) };
+        env.mbox.copy_bytes_to_mbox(rt.as_bytes())?;
```
jhand2, line 92: We could maybe consider putting the rt and fmc regions in the MemoryRegions driver. This would allow rust to enforce accesses better.

runtime/src/lib.rs
```diff
@@ -205,7 +228,21 @@ impl<'a> Drivers<'a> {
...
+    // Run pending jobs before entering low power mode.
+    #[cfg(feature = "fips_self_test")]
+    if let SelfTestStatus::InProgress(execute) = drivers.self_test_status {
+        if drivers.mbox.lock() == false {
+            match execute(drivers) {
+                Ok(_) => drivers.self_test_status = SelfTestStatus::Done,
+                Err(e) => caliptra_drivers::report_fw_error_non_fatal(e.into()),
+            }
+        }
+    }
```
jhand2, line 244: Is there a specific reason SelfTest can't be blocking?

runtime/src/lib.rs
```diff
@@ -255,7 +292,18 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+        #[cfg(feature = "fips_self_test")]
+        CommandId::SELF_TEST => match drivers.self_test_status {
+            SelfTestStatus::Idle => {
+                drivers.self_test_status = SelfTestStatus::InProgress(fips_self_test_cmd::execute);
+                Ok(MailboxResp::default())
+            }
+            SelfTestStatus::Done => {
+                drivers.self_test_status = SelfTestStatus::Idle;
+                Ok(MailboxResp::default())
+            }
```
jhand2, line 304: I don't really understand this state machine. Can we really serve other commands while verification is running in the background? I would assume that would break our driver access model.

kat/src/lib.rs
rusty1968: I see your point. This is  a question for @JohnTraverAmd.

runtime/src/lib.rs
rusty1968: We need to  “reply with data” to conform with new FIPS API (response contains header + checksum), but the mailbox SRAM would already be mutated with the image.

runtime/src/lib.rs
rusty1968: We defer execution to the earliest moment in time when a lock can be acquired.

runtime/src/fips.rs
FerralCoder: agree, but then change them all instead of just changing this one

FerralCoder: sorry, ignore. The other one I was looking at is a struct, not mod.

runtime/src/fips.rs
rusty1968: I believe the memory regions will be replaced by the persistent data accessor soon. Would it be ok if we defer it to a later PR?

builder/src/lib.rs
rusty1968: According to the manifest, the size  of the RT binary image is 48276 bytes.

kat/src/lib.rs
korran: LMS is part of the firmware authentication, and fips requires us to reauthenticate. I assume since we do an LMS kat we're actually getting it certified, and thus it's required.

runtime/src/lib.rs
korran: The alternative approach is to make the image-verification digest generic, using the sha384 peripheral instead of sha384acc with the mailbox.

If we want to abandon calling back into ROM forever, perhaps just using the sha384 peripheral is better...

runtime/src/fips.rs
jhand2: Ya, that's ok with me

builder/src/lib.rs
jhand2: Thanks!

FerralCoder (approved):

error/src/lib.rs
```diff
@@ -303,6 +303,7 @@ impl CaliptraError {
     pub const RUNTIME_DISABLE_ATTESTATION_FAILED: CaliptraError =
         CaliptraError::new_const(0x000E0011);
     pub const RUNTIME_HANDOFF_INVALID_PARM: CaliptraError = CaliptraError::new_const(0x000E0012);
+    pub const RUNTIME_SELF_TEST_IN_PROGREESS: CaliptraError = CaliptraError::new_const(0x000E0013);
```
korran, line 306: typo: Should be RUNTIME_SELF_TEST_IN_PROGRESS

kat/src/lib.rs
jhand2: Ok, so we're saying that because we use LMS for firmware auth, we also need to run the LMS KAT. That's fine, just double checking.

runtime/src/lib.rs
korran: I think we should keep this run-self-test mailbox API as is. If a future version of FIPS ever needs us to test the sha384acc peripheral with something other than empty input data we'll need a mechanism to modify the mailbox.

test/tests/smoke_test.rs
```diff
@@ -362,3 +362,82 @@ fn smoke_test() {
...
+    // Confirm we can't re-start the FIPS self test.
+    hw.step_until_boot_status(
+        caliptra_runtime::RtBootStatus::RtFipSelfTestStarted.into(),
+        true,
+    );
+    let _resp = hw
+        .mailbox_execute(u32::from(CommandId::SELF_TEST), payload.as_bytes())
+        .unwrap_err();
+
+    hw.step_until_boot_status(
```
korran, line 425: I don't think it is realistic to expect callers to do this. Also, I'm not sure the boot status register is really intended to be used once we've fully booted.

Perhaps the mailbox response from the SELF_TEST command should explain to the user that the test is currently in progress.

Maybe it would be simpler if this were two separate commands `CommandId::SELF_TEST_START` and `CommandID:SELF_TEST_GET_RESULT`?

runtime/src/lib.rs
```diff
@@ -256,7 +291,18 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         #[cfg(feature = "test_only_commands")]
         CommandId::TEST_ONLY_HMAC384_VERIFY => HmacVerifyCmd::execute(drivers, cmd_bytes),
         CommandId::VERSION => FipsVersionCmd::execute(drivers),
-        CommandId::SELF_TEST => FipsSelfTestCmd::execute(drivers),
+        #[cfg(feature = "fips_self_test")]
+        CommandId::SELF_TEST => match drivers.self_test_status {
```
korran, line 295: I don't think you should overload the SELF_TEST command for the "start" and "get result" phases. The response in both cases is an empty success result, which isn't very informative. See my comments in smoke_test.rs.

runtime/src/lib.rs
```diff
@@ -256,7 +291,18 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+            SelfTestStatus::Idle => {
+                drivers.self_test_status = SelfTestStatus::InProgress(fips_self_test_cmd::execute);
```
jhand2, line 297: Does this need to take a function? Or could the handler code in goto_idle just call `fips_self_test_cmd::execute` directly? I feel like the current code is a bit hard to follow.

runtime/src/lib.rs
```diff
@@ -206,7 +229,19 @@ impl<'a> Drivers<'a> {
     }
 }
 
-fn wait_for_cmd(_mbox: &mut Mailbox) {
+/// Run pending jobs and enter low power mode.
+fn goto_idle(drivers: &mut Drivers) {
```
jhand2, line 233: nit: `go_idle`? I think the name `goto_idle` could be misinterpreted as performing a goto jump.

rusty1968: Done.

test/tests/smoke_test.rs
rusty1968: done

runtime/src/lib.rs
rusty1968: done

runtime/src/fips.rs
```diff
@@ -78,14 +57,120 @@ impl FipsVersionCmd {
...
+        cprintln!("[rt] Verify complete");
+        Ok(())
+    }
+
+    pub(crate) fn execute(env: &mut Drivers) -> CaliptraResult<()> {
+        caliptra_drivers::report_boot_status(RtFipSelfTestStarted.into());
         cprintln!("[rt] FIPS self test");
-        FipsModule::execute_kats(env)?;
+        copy_and_verify_image(env)?;
+        execute_kats(env)?;
```
mhatrevi, line 139: Shouldn't KATs be executed first before using the cryptographic primitives to verify the image?

mhatrevi (changes requested):

runtime/src/fips.rs
rusty1968: done.

rusty1968:
@mhatrevi , ajisaxena : please give feedback on changes in crates owned by msft.

mhatrevi (approved):

FerralCoder:
changes were included with PR 693

PR #685, opened 2023-08-24
[fix] Fail warm-reset flow if previous cold-reset was unsuccessful.

### LINKED ISSUE #628: If a warm reset occurs after image-copy to DCCM but before fmcalias derivation, FMC can access ldev key
Consider this scenario inside the [cold reset flow](https://github.com/chipsalliance/caliptra-sw/blob/27623ff53cc837d74fcb2a4b77d74780a163dfe1/rom/dev/src/flow/cold_reset/mod.rs#L64-L70):

* FirmwareProcessor::process() has finished, so ICCM contains the image
* warm reset comes in before FmcAliasLayer::derive() is complete
* ROM boots the warm reset flow, which does basically nothing but jump to FMC
* ldevid key is still usable in the keyvault by the FMC.

I think we need to make any warm-reset that occurs during the cold-reset flow a fatal error. 

Also, I wonder if the order of the code above would be a lot safer if we copied the from mailbox to ICCM AFTER we destroyed the ldevid key. No sense giving attackers a bunch of ROP gadgets to glitch into while the cpu still controls a super-powerful key.

mhatrevi:
This fix addresses: https://github.com/chipsalliance/caliptra-sw/issues/628

### DIFF
--- a/common/src/boot_status.rs
+++ b/common/src/boot_status.rs
@@ -64,7 +64,7 @@ pub enum RomBootStatus {
 
     // Cold Reset Statuses
     ColdResetStarted = COLD_RESET_BOOT_STATUS_BASE,
-    ColdResetComplete = COLD_RESET_BOOT_STATUS_BASE + 1,
+    ColdResetComplete = UPDATE_RESET_BOOT_STATUS_BASE - 1,
 
     // Update Reset Statuses
     UpdateResetStarted = UPDATE_RESET_BOOT_STATUS_BASE,
--- a/drivers/src/data_vault.rs
+++ b/drivers/src/data_vault.rs
@@ -70,7 +70,7 @@ impl From<ColdResetEntry48> for usize {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ColdResetEntry4 {
     FmcSvn = 0,
-    Reserved0 = 1,
+    RomColdBootStatus = 1,
     FmcEntryPoint = 2,
     EccVendorPubKeyIndex = 3,
     LmsVendorPubKeyIndex = 4,
@@ -388,6 +388,25 @@ impl DataVault {
     /// * `u32` - Vendor public key index
     pub fn lms_vendor_pk_index(&self) -> u32 {
         self.read_cold_reset_entry4(ColdResetEntry4::LmsVendorPubKeyIndex)
+    }
+
+    /// Set and lock the rom cold boot status.
+    ///
+    /// # Arguments
+    ///
+    /// * `status` - Rom Cold Boot Status
+    ///
+    pub fn set_rom_cold_boot_status(&mut self, status: u32) {
+        self.write_lock_cold_reset_entry4(ColdResetEntry4::RomColdBootStatus, status);
+    }
+
+    /// Get the rom cold boot status.
+    ///
+    /// # Returns
+    ///
+    /// * `u32` - Rom Cold Boot Status
+    pub fn rom_cold_boot_status(&self) -> u32 {
+        self.read_cold_reset_entry4(ColdResetEntry4::RomColdBootStatus)
     }
 
     /// Set the rt tcb component identifier.
--- a/rom/dev/src/flow/warm_reset.rs
+++ b/rom/dev/src/flow/warm_reset.rs
@@ -13,8 +13,10 @@ Abstract:
 --*/
 use crate::{cprintln, rom_env::RomEnv};
 use caliptra_cfi_derive::cfi_impl_fn;
+use caliptra_cfi_lib::{cfi_assert, cfi_assert_eq, cfi_launder};
 use caliptra_common::FirmwareHandoffTable;
-use caliptra_error::CaliptraResult;
+use caliptra_common::RomBootStatus::ColdResetComplete;
+use caliptra_error::{CaliptraError, CaliptraResult};
 
 /// Warm Reset Flow
 pub struct WarmResetFlow {}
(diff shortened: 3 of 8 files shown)


rom/dev/src/flow/warm_reset.rs
```diff
@@ -27,8 +28,15 @@ impl WarmResetFlow {
...
     #[inline(never)]
     #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
-    pub fn run(_env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
+    pub fn run(env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
         cprintln!("[warm-reset] ++");
+
+        // Check if previous Cold-Reset was successful.
+        if env.data_vault.rom_cold_boot_status() != ColdResetComplete.into() {
+            cprintln!("[warm-reset] Previous Cold-Reset was not successful.");
+            return Err(CaliptraError::ROM_WARM_RESET_UNSUCCESSFUL_PREVIOUS_COLD_RESET);
```
FerralCoder, line 37: wouldn't it be better to instead treat this case as another cold reset? Could we initiate a cold reset from here?

mhatrevi: Currently, that is not possible. All ColdReset datavault entries that were locked will remain locked on a warm-reset, and ROM won't be able to record cold-boot info on this path.

FerralCoder: What I meant here was to initiate a cold reset at the hardware level, not to try to run the cold boot flow while hardware is doing a warm reset.

korran: It's not safe to initiate a cold reset of Caliptra without also resetting the rest of the SoC; the measurements made early in boot would be lost and Caliptra would think that whatever SoC firmware is loading measurements into it is the SoC boot ROM, when in fact it could be attacker-controlled mutable firmware.

drivers/src/data_vault.rs
```diff
@@ -390,6 +390,25 @@ impl DataVault {
         self.read_cold_reset_entry4(ColdResetEntry4::LmsVendorPubKeyIndex)
     }
 
+    /// Set the rom cold boot status.
+    ///
+    /// # Arguments
+    ///
+    /// * `status` - Rom Cold Boot Status
+    ///
+    pub fn set_rom_cold_boot_status(&mut self, status: u32) {
```
korran, line 399: Maybe the name should make it clear is this also locking the entry? Or perhaps get rid of this function and call write_lock_cold_reset_entry4 directly?

rom/dev/src/flow/warm_reset.rs
```diff
@@ -27,8 +28,15 @@ impl WarmResetFlow {
     /// * `env` - ROM Environment
     #[inline(never)]
     #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
-    pub fn run(_env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
+    pub fn run(env: &mut RomEnv) -> CaliptraResult<Option<FirmwareHandoffTable>> {
         cprintln!("[warm-reset] ++");
+
+        // Check if previous Cold-Reset was successful.
+        if env.data_vault.rom_cold_boot_status() != ColdResetComplete.into() {
```
korran, line 35: Can we add some CFI defenses here? Read the register a second time and assert if it's different?

rom/dev/tests/test_fmcalias_derivation.rs
```diff
@@ -702,3 +702,63 @@ fn test_check_no_lms_info_in_datavault_on_lms_unavailable() {
...
+
+    let image_bundle = caliptra_builder::build_and_sign_image(
+        &TEST_FMC_WITH_UART,
+        &APP_WITH_UART,
+        ImageOptions::default(),
+    )
+    .unwrap();
+
+    assert!(hw
+        .upload_firmware(&image_bundle.to_bytes().unwrap())
```
korran, line 742: As a sanity check, maybe we could add another test where we pass bad firmware and make sure that ColdResetComplete is not true?

common/src/boot_status.rs
```diff
@@ -64,7 +64,7 @@ pub enum RomBootStatus {
 
     // Cold Reset Statuses
     ColdResetStarted = COLD_RESET_BOOT_STATUS_BASE,
-    ColdResetComplete = COLD_RESET_BOOT_STATUS_BASE + 1,
+    ColdResetComplete = UPDATE_RESET_BOOT_STATUS_BASE - 1,
```
korran, line 67: What is the purpose of this change? Does this collide with UpdateResetLoadManifestComplete ?

mhatrevi: This is being done to not have ColdResetStarted and ColdResetComplete differ by a single bit. With this change, ColdResetComplete will be 320 (UPDATE_RESET_BOOT_STATUS_BASE - 1) and UpdateResetStarted will be 321 (UPDATE_RESET_BOOT_STATUS_BASE). UpdateResetLoadManifestComplete is 322.

drivers/src/data_vault.rs
mhatrevi: This is inline with all other functions on the data vault driver. I have updated the comment on the function; if you feel strongly, we can change the naming for all these functions to indicate that the entry is also being locked, as a separate change.

rom/dev/src/flow/warm_reset.rs
mhatrevi: This would require a hardware change. We can discuss this as a separate issue.

rom/dev/src/flow/warm_reset.rs
mhatrevi: Added.

rom/dev/tests/test_fmcalias_derivation.rs
mhatrevi: Discussed offline; currently it is not possible to test the negative test case.

korran (approved):

korran (approved):

FerralCoder (approved):

PR #686, opened 2023-08-24
libcaliptra: idev, ldev, dpe commands

wmaroneAMD:
This adds support for the following mailbox commands:

OP_GET_IDEV_CSR
OP_GET_LDEV_CERT
OP_INVOKE_DPE_COMMAND

It also normalizes the new command API pattern, which will be necessary in the future to support asynchronous mailbox operations.

### DIFF
--- a/libcaliptra/inc/caliptra_enums.h
+++ b/libcaliptra/inc/caliptra_enums.h
@@ -31,3 +31,44 @@ enum toc_entry_id {
     MAX     = 0xFFFFFFFF,
 };
 
+// The below enums are placeholders to set up the baseline
+// required for communication of DPE commands to Caliptra
+// firmware.
+
+enum dpe_commands {
+    DPE_GET_PROFILE        = 0x1,
+    DPE_INITIALIZE_CONTEXT = 0x7,
+    DPE_DERIVE_CHILD       = 0x8,
+    DPE_CERTIFY_KEY        = 0x9,
+    DPE_SIGN               = 0xA,
+    DPE_ROTATE_CTX_HANDLE  = 0xE,
+    DPE_DESTROY_CTX        = 0xF,
+    DPE_GET_CERT_CHAIN     = 0x80,
+    DPE_EXTEND_TCI         = 0x81,
+    DPE_TAG_TCI            = 0x82,
+    DPE_GET_TAGGED_TCI     = 0x83,
+};
+
+enum dpe_error_codes {
+    DPE_NO_ERROR               = 0,
+    DPE_INTERNAL_ERROR         = 1,
+    DPE_INVALID_COMMAND        = 2,
+    DPE_INVALID_ARGUMENT       = 3,
+    DPE_ARGUMENT_NOT_SUPPORTED = 4,
+    DPE_INVALID_HANDLE         = 0x1000,
+    DPE_INVALID_LOCALITY       = 0x1001,
+    DPE_BADTAG                 = 0x1002,
+    DPE_MAXTCIS                = 0x1003,
+    DPE_PLATFORM_ERROR         = 0x1004,
+    DPE_CRYPTO_ERROR           = 0x1005,
+    DPE_HASH_ERROR             = 0x1006,
+    DPE_RAND_ERROR             = 0x1007,
+};
+
+#define DPE_PROFILE_256 1
+#define DPE_PROFILE_384 2
+
+enum dpe_profile {
+    P256Sha256 = DPE_PROFILE_256,
+    P384Sha384 = DPE_PROFILE_384,
+};
--- a/libcaliptra/inc/caliptra_types.h
+++ b/libcaliptra/inc/caliptra_types.h
@@ -64,6 +64,18 @@ struct caliptra_stash_measurement_resp {
     uint32_t                   dpe_result;
 };
 
+struct caliptra_get_idev_csr_resp {
+    struct caliptra_completion cpl;
+    uint32_t                   data_size;
+    uint8_t                    data[1024];
+};
+
+struct caliptra_get_ldev_cert_resp {
+    struct caliptra_completion cpl;
+    uint32_t                   data_size;
+    uint8_t                    data[1024];
+};
+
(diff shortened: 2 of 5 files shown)


libcaliptra/inc/caliptra_types.h
```diff
@@ -63,3 +63,117 @@ struct caliptra_stash_measurement_resp {
...
+    uint32_t magic;
+    uint32_t cmd_id;
+    uint32_t profile;
+};
+
+struct dpe_response {
+    uint32_t magic;
+    uint32_t status;
+    uint32_t profile;
+};
```
jhand2, line 96: nit: I would recommend naming these something like `dpe_cmd_hdr` and `dpe_resp_hdr`

libcaliptra/inc/caliptra_types.h
```diff
@@ -63,3 +63,117 @@ struct caliptra_stash_measurement_resp {
...
+struct caliptra_dpe_resp {
+    struct caliptra_completion cpl;
+    uint32_t                   data_size;
+    union {
+        struct dpe_new_handle_response            new_handle;
+        struct dpe_derive_child_response          derive_child;
+        struct dpe_certify_key_response           certify_key;
+        struct dpe_sign_response                  sign;
+        struct dpe_get_tagged_tci_response        get_tagged_tci;
+        struct dpe_get_certificate_chain_response get_certified_chain;
```
jhand2, line 176: nit: Can you also include a union member that is `uint8_t data[DPE_RESP_MAX]` ([max size defined here](https://github.com/chipsalliance/caliptra-sw/blob/c76b3ed2bfbd8d6ee9b7183ac011288491c6b173/common/src/mailbox_api.rs#L245C9-L245C9))

This is helpful in cases where the caller to DPE is just proxying arbitrary commands on behalf of some other component. We will use this for calling the Golang DPE acceptance tests.

libcaliptra/inc/caliptra_types.h
wmaroneAMD: Can do.

libcaliptra/inc/caliptra_types.h
wmaroneAMD: Can do. Just for my knowledge, is there an equivalent in the Rust codebase or does it just handle it on the fly?

jhand2: In caliptra-dpe, the response struct implements as_bytes(), so callers can always interpret it as a slice if needed: https://github.com/chipsalliance/caliptra-dpe/blob/a28b6b701f232b9982ad5f5e0efd5f87293f5c5d/dpe/src/response.rs#L31

And the the invoke dpe mailbox command does this when writing to the DPE response structure: https://github.com/chipsalliance/caliptra-sw/blob/c76b3ed2bfbd8d6ee9b7183ac011288491c6b173/runtime/src/invoke_dpe.rs#L22

jhand2 (approved):

jlmahowa-amd (dismissed):

libcaliptra/inc/caliptra_enums.h
```diff
@@ -31,3 +31,44 @@ enum toc_entry_id {
...
+// required for communication of DPE commands to Caliptra
+// firmware.
+
+enum dpe_commands {
+    DPE_GET_PROFILE        = 0x1,
+    DPE_INITIALIZE_CONTEXT = 0x7,
+    DPE_DERIVE_CHILD       = 0x8,
+    DPE_CERTIFY_KEY        = 0x9,
+    DPE_SIGN               = 0xA,
+    DPE_ROTATE_CTX_HANDLE  = 0xE,
```
fdamato, line 44: Not for this patch, can this eventually be converted from RUST and taken from the DPE Library ?

libcaliptra/src/caliptra_api.c
```diff
@@ -356,6 +356,37 @@ static int check_command_response(struct caliptra_completion *cpl, uint8_t *buff
...
+        .data = parcel->tx_buffer,
+        .len  = parcel->tx_bytes,
+    };
+
+    struct caliptra_buffer out_buf = {
+        .data = parcel->rx_buffer,
+        .len  = parcel->rx_bytes,
+    };
+
+    *((caliptra_checksum*)parcel->tx_buffer) = calculate_caliptra_checksum(parcel->command, parcel->tx_buffer, parcel->tx_bytes);
```
fdamato, line 376: shall we check for NULL first before dereferencing here ?

libcaliptra/inc/caliptra_enums.h
wmaroneAMD: I have attempted to run cbindgen on the DPE library and it consistently fails when attempting to process some of the impl statements for the structures. If there's a way to have it convert only structs and enums, that would be extremely helpful and I would replace everything with them.

jhand2: Let me know if there are any changes we can make to the struct/enum definitions to make this easier

wmaroneAMD: Unfortunately I wouldn't know what changes to make as I'm still ramping on Rust and most of my time thus far has me still working in the C world, and my time allocation has shifted away from Caliptra.

jhand2: Ack. I can take some time to look into cleaning this up in a follow-up CL. Could you open an issue for this?

wmaroneAMD: https://github.com/chipsalliance/caliptra-sw/issues/752

FerralCoder (approved):

nquarton (approved):

jhand2:
Can we get this PR merged? We would like to be able to add some DPE tests that leverage it.

cc @JohnTraverAmd @fdamato @jlmahowa-amd

jhand2 (approved):

JohnTraverAmd (approved):

fdamato (approved):

PR #689, opened 2023-08-25
Use a reference to persistent FHT in runtime firmware

jhand2:
Add the ability to get a reference to the FHT in persistent DCCM. This allows RT firmware to avoid copying the FHT onto the stack.

### DIFF
--- a/drivers/src/hand_off.rs
+++ b/drivers/src/hand_off.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 use bitfield::{bitfield_bitrange, bitfield_fields};
 use caliptra_error::CaliptraError;
-use zerocopy::{AsBytes, FromBytes};
+use zerocopy::{AsBytes, FromBytes, LayoutVerified};
 
 pub const FHT_MARKER: u32 = 0x54484643;
 pub const FHT_INVALID_ADDRESS: u32 = u32::MAX;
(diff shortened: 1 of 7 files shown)


drivers/src/hand_off.rs
```diff
@@ -427,6 +427,26 @@ impl FirmwareHandoffTable {
         None
     }
 
+    pub fn try_into_ref() -> Option<&'static FirmwareHandoffTable> {
+        let slice = unsafe {
```
korran, line 431: Assuming that FirmwareHandoffTable implements FromBytes and FHT_ORG is aligned properly, this is adequate:

```rs
let fht = unsafe { &*(FHT_ORG as *const FirmwareHandoffTable) };
```

We could make things a bit safer with something like this:

```rs
unsafe fn ref_from_addr<T: FromBytes>(addr: u32) -> &'static T {
  // LTO should be able to optimize this out to maintain panic_is_missing
  assert!(addr % core::mem::align_of<T>() == 0);
  assert!(core::mem::size_of::<u32>() == core::mem::size_of::<*const T>());
  &*(addr as *const T)
}
```

```rs
let fht = ref_from_addr::<FirmwareHandoffTable>(FHT_ORG);
```

drivers/src/hand_off.rs
```diff
@@ -427,6 +427,26 @@ impl FirmwareHandoffTable {
         None
     }
 
+    pub fn try_into_ref() -> Option<&'static FirmwareHandoffTable> {
```
korran, line 430: It is not sound for this function to be safe, as the aliasing rules for the returned reference will be violated as soon as somebody calls FirmwareHandoffTable::save().

This function needs to be in an actual driver that returns references with the same lifetime as the driver's &self, and the save() method needs to take the driver's (&mut self). That way, the type system will prevent aliasing violations.

korran: See #690 for what I'm thinking.

jhand2: Sounds good. I'm closing this PR in favor of #690

PR #690, opened 2023-08-25
Add safe abstraction for accessing persistent data.

korran:
There's too much unsafe code scattered around the code-base dereferencing global pointers. This PR puts all the persistent data into a single `PersistentData` struct, and provides a PersistentDataAcessor "driver" to mediate access to it, similar to the other drivers.

This PR just adds the new code; the behavior of the ROM is unchanged.

Related PRs:

* #702 
* #703
* #704


drivers/src/persistent.rs
```diff
@@ -0,0 +1,127 @@
...
+    reserved0: [u8; memory_layout::MAN1_SIZE as usize - size_of::<ImageManifest>()],
+
+    pub manifest2: ImageManifest,
+    reserved1: [u8; memory_layout::MAN2_SIZE as usize - size_of::<ImageManifest>()],
+
+    pub fht: FirmwareHandoffTable,
+    reserved2: [u8; memory_layout::FHT_SIZE as usize - size_of::<FirmwareHandoffTable>()],
+
+    // TODO: Do we want to hide these fields from the FMC/runtime and force them
+    // to go through the FHT addresses?
```
FerralCoder, line 23: I think that would be a good idea

korran: Perhaps, but I want to defer this to a future PR; there's not a super easy way to hide these from some crates and not others.

Also, I'm a bit nervous on relying on the FHT pointers, as if they're corrupted we could wander into undefined behavior land; I'd feel more comfortable if we just added a check that `(&persistent_data.fht.ldevid_tbs_addr) == (&persistent_data.ldevid_tbs as u32)`

jhand2 (approved):

FerralCoder (approved):

ajisaxena (approved):

PR #693, opened 2023-08-25
Implement FIPS commands for ROM and invoke SELF-TEST from Runtime.

jlmahowa-amd:
FIPS in ROM
- Enables VERSION, SELF_TEST, SHUTDOWN

This commit also enables Runtime to run FIPS Self Tests as follows : KATs, verification of firmware image in ICCM, and ROM integrity test.
- Light refactor to decouple KAT crate from Rom dependency.
- Light refactor to decouple Image verification crate from ROM dependency.
- Ability for RT to defer execution of self test (basic job control).
- Runtime copies image from ICCM to mailbox SRAM , so it needs to fake a send a transaction and hold the mailbox lock while the image is being validated,. It defers the actual self test execution until the mailbox lock can be acquired. SoC must "poll" for self test completion.
- ROM integrity check from RT using the RomInfo FHT field .
- SoC must send a FIPS self test command one time to schedule the execution.
- Caliptra holds the mailbox lock while the self test is in progress.
- Failure of FIPS self test is handled as fatal error.
- Upon successful self-test completion the mailbox lock is released and SoC will receive a positive acknowledgement when FIPS Self test Get Results command is re-issued.


nquarton (commented):
I don't have a better solution for the kats_env issue but maybe someone with more rust experience than me can think of something. Overall, it makes sense to me and the response handling looks good.

error/src/lib.rs
```diff
@@ -304,6 +304,7 @@ impl CaliptraError {
         CaliptraError::new_const(0x000E0011);
     pub const RUNTIME_HANDOFF_INVALID_PARM: CaliptraError = CaliptraError::new_const(0x000E0012);
     pub const RUNTIME_GET_DEVID_CERT_FAILED: CaliptraError = CaliptraError::new_const(0x000E0013);
+    pub const RUNTIME_SELF_TEST_IN_PROGREESS: CaliptraError = CaliptraError::new_const(0x000E0014);
```
mhatrevi, line 307: Spelling:  RUNTIME_SELF_TEST_IN_PROGRESS

kat/src/kats_env.rs
```diff
@@ -0,0 +1,29 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_drivers::Ecc384;
+use caliptra_drivers::{Hmac384, Lms, Sha1, Sha256, Sha384, Sha384Acc, Trng};
```
mhatrevi, line 4: Can combine these use statements.

kat/src/kats_env.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub struct KatsEnv<'a> {
+    // SHA1 Engine
+    pub sha1: &'a mut Sha1,
+
+    pub sha256: &'a mut Sha256,
```
mhatrevi, line 10: NIT: Consistent comments.

mhatrevi (changes requested):
Please see the review comments.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -126,18 +156,53 @@ impl FirmwareProcessor {
     fn process_mailbox_commands<'a>(
         soc_ifc: &mut SocIfc,
         mbox: &'a mut Mailbox,
+        env: &mut KatsEnv,
```
korran, line 159: Might be more future proof (for deferred measurements) to pass in the entire `&mut RomEnv` here, but we can defer that to a future PR.

runtime/src/fips.rs
```diff
@@ -37,55 +36,122 @@ impl FipsModule {
...
+    pub enum SelfTestStatus {
+        Idle,
+        InProgress(fn(&mut Drivers) -> CaliptraResult<()>),
+        Done,
+    }
 
-        cprintln!("[kat] Executing ECC-384 Engine KAT");
-        Ecc384Kat::default().execute(&mut env.ecc384, &mut env.trng)?;
+    fn copy_and_verify_image(env: &mut Drivers) -> CaliptraResult<()> {
+        cprintln!("write dummy cmd");
```
korran, line 62: Nit: these cprintln! statements aren't that informative. Not a blocker for merge.

runtime/src/fips.rs
```diff
@@ -37,55 +36,122 @@ impl FipsModule {
...
+pub mod fips_self_test_cmd {
+    use super::*;
+    use crate::RtBootStatus::{RtFipSelfTestComplete, RtFipSelfTestStarted};
+    use caliptra_common::{verifier::FirmwareImageVerificationEnv, FMC_ORG, RUNTIME_ORG};
+    use caliptra_drivers::ResetReason;
+    use caliptra_image_verify::ImageVerifier;
+    use zerocopy::AsBytes;
+
+    // Helper function to create a mutable slice from a memory region
+    unsafe fn create_slice(org: u32, size: usize) -> &'static mut [u8] {
```
korran, line 51: As other parts of the firmware take  &'static references to values in .rodata (which is inside the returned slice), returning &mut here is unsound. Please return &'static [u8] instead and use core::slice::from_raw_parts(ptr, size) below.

Since FMC_ORG and RUNTIME_ORG are in ICCM; an NMI would be raised if anybody were to write to these slices :)

runtime/src/fips.rs
```diff
@@ -37,55 +36,122 @@ impl FipsModule {
...
+
+        cprintln!("copy manifest");
+        env.mbox
+            .copy_bytes_to_mbox(env.persistent_data.get().manifest1.as_bytes())?;
+
+        cprintln!("copy fmc");
+        let fmc = unsafe {
+            create_slice(
+                FMC_ORG,
+                env.persistent_data.get().manifest1.fmc.size as usize,
```
korran, line 79: We're putting too much faith in fmc.size here, which may have been corrupted by a previously running "evil" firmware. This should have bounds checks to ensure this is within ICCM.

runtime/src/fips.rs
```diff
@@ -37,55 +36,122 @@ impl FipsModule {
...
+            )
+        };
+        env.mbox.copy_bytes_to_mbox(fmc.as_bytes())?;
+        cprintln!("copy rt");
+        let rt = unsafe {
+            create_slice(
+                RUNTIME_ORG,
+                env.persistent_data.get().manifest1.runtime.size as usize,
```
korran, line 87: We're putting too much faith in runtime.size here, which may have been corrupted by a previously running "evil" firmware. This should have bounds checks to ensure this is within ICCM.

runtime/src/lib.rs
```diff
@@ -329,6 +377,7 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
                 }
             }
             caliptra_common::wdt::stop_wdt(&mut drivers.soc_ifc);
+        } else {
```
korran, line 380: nitpick: empty else statement.

runtime/src/mailbox.rs
```diff
@@ -57,23 +81,43 @@ impl Mailbox {
...
+    }
+
+    pub fn copy_words_to_mbox(&mut self, buf: &[Unalign<u32>]) {
         let mbox = self.mbox.regs_mut();
         for word in buf {
             mbox.datain().write(|_| word.get());
         }
     }
 
+    pub fn copy_bytes_to_mbox(&mut self, buf: &[u8]) -> CaliptraResult<()> {
```
korran, line 99: To eliminate duplication, can write_response() call this function?

runtime/tests/integration_tests.rs
```diff
@@ -776,54 +756,3 @@ fn test_unimplemented_cmds() {
...
-    let resp = model
-        .mailbox_execute(u32::from(CommandId::GET_IDEV_INFO), payload.as_bytes())
-        .unwrap()
-        .unwrap();
-
-    GetIdevInfoResp::read_from(resp.as_slice()).unwrap();
-}
-
-#[test]
-fn test_idev_id_cert() {
```
korran, line 797: Why is this test being deleted?

runtime/tests/integration_tests.rs
```diff
@@ -776,54 +756,3 @@ fn test_unimplemented_cmds() {
     let resp = model.mailbox_execute(INVALID_CMD, payload.as_bytes());
     assert_eq!(resp, expected_err);
 }
-
-#[test]
-fn test_idev_id_info() {
```
korran, line 781: Why is this test being deleted?

jlmahowa-amd: Merge mistake. Restoring.

runtime/tests/integration_tests.rs
jlmahowa-amd: Merge mistake. Restoring.

runtime/src/mailbox.rs
jlmahowa-amd: Done.

runtime/src/lib.rs
jlmahowa-amd: Removed.

runtime/src/fips.rs
jlmahowa-amd: Done. Yes these didn't need to be mutable.

runtime/src/fips.rs
jlmahowa-amd: Removed.

runtime/src/fips.rs
jlmahowa-amd: Added checks for fmc_size and runtime_size. I'll let you confirm.

korran:
nquarton left a comment
> I don't have a better solution for the kats_env issue but maybe someone with more rust experience than me can think of something. Overall, it makes sense to me and the response handling looks good.

Yeah, having process_mailbox_commands() return the firmware-download transaction bound to the mailbox lifetime causes a lot of pain. What you've done with kat_env is probably as good as you can do without moving all the mailbox download logic inside process_mailbox_commands().

drivers/src/hand_off.rs
```diff
@@ -89,18 +89,29 @@ pub enum DataVaultRegister {
...
+                }),
             )),
             Vault::DataVault => match self.reg_type() {
                 1 => {
                     let entry = DataStore::DataVaultSticky4(
                         ColdResetEntry4::try_from(self.reg_num() as u8).unwrap_or_else(|_| {
-                            report_handoff_error_and_halt("Invalid ColdResetEntry4", 0xbadbad)
+                            report_handoff_error_and_halt(
+                                "Invalid ColdResetEntry4",
+                                CaliptraError::DRIVER_HANDOFF_INVALID_COLD_RESET_ENTRY4.into(),
```
korran, line 113: Why are we halting as opposed to returning Err(CaliptraError::DRIVER_HANDOFF_INVALID_COLD_RESET_ENTRY4)?

korran (approved):

drivers/src/hand_off.rs
korran: (doesn't need to be fixed in this PR, as this isn't new behavior, but I'm curious what is going on here)

jlmahowa-amd: Will need @rusty1968 to respond to this.

mhatrevi (approved):

jhand2 (approved):
LGTM for runtime

PR #696, opened 2023-08-25
Add more reserved bits in image signed data

bluegate010:
Just in case.

### DIFF
--- a/image/types/src/lib.rs
+++ b/image/types/src/lib.rs
@@ -278,7 +278,7 @@ pub struct VendorSignedData {
     /// Vendor End Date [ASN1 Time Format] For FMC alias certificate.
     pub vendor_not_after: [u8; 15],
 
-    reserved: [u8; 2],
+    reserved: [u8; 10],
 }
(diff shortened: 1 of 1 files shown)


image/types/src/lib.rs
```diff
@@ -290,7 +290,7 @@ pub struct OwnerSignedData {
     /// Owner End Date [ASN1 Time Format] For LDEV-Id certificate: Takes Preference over vendor end date
     pub owner_not_after: [u8; 15],
 
-    reserved: [u8; 2],
+    reserved: [u8; 10],
```
jhand2, line 293: Should vendor-signed data have the same? So far the approach as been "Allow the vendor to specify the field and optionally allow the owner to overwrite it"

bluegate010: Done.

jhand2 (approved):

FerralCoder (approved):

mhatrevi (approved):

PR #699, opened 2023-08-29
val-ROM canned certs and val-FMC feature added

nquarton:
- Canned LDEV and FMC Alias certs in val ROM
- val-FMC currently just jumps straight to runtime
- RT Alias cert support for val-FMC is a TODO


jlmahowa-amd (dismissed):

rom/dev/src/flow/val.rs
```diff
@@ -79,6 +213,53 @@ impl ValRomFlow {
...
+pub fn copy_canned_ldev_cert(env: &mut RomEnv) -> CaliptraResult<()> {
+    // Store signature
+    env.data_vault.set_ldev_dice_signature(&VAL_LDEV_SIG);
+
+    // Store pub key
+    env.data_vault.set_ldev_dice_pub_key(&VAL_LDEV_PUB_KEY);
+
+    // Copy TBS to DCCM
+    let tbs = &VAL_LDEV_TBS;
+    env.fht_data_store.ldevid_tbs_size = tbs.len() as u16;
```
korran, line 225: nitpick: For a noisier symptom if there's an overflow, prefer u16::try_from(tbs.len()).unwrap() to "as". Since the size is known at compile time it will probably compile to the same machine code.

nquarton: No problem, I updated both cases to this method

jlmahowa-amd (approved):

korran (approved):

rusty1968 (approved):

PR #702, opened 2023-08-29
Use PersistentDataAccessor in Runtime firmware.

korran:
The PersistentDataAccessor is introduced in #690 (Add safe abstraction for accessing persistent data).

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -32,16 +32,12 @@ pub use verify::EcdsaVerifyCmd;
 pub mod packet;
 use packet::Packet;
 
+use caliptra_common::cprintln;
 use caliptra_common::mailbox_api::CommandId;
-use caliptra_common::memory_layout::{
-    FHT_ORG, FHT_SIZE, FMCALIAS_TBS_ORG, FMCALIAS_TBS_SIZE, FUSE_LOG_ORG, FUSE_LOG_SIZE,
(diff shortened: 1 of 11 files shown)


jhand2 (dismissed):
LGTM. I assume you'll just re-target this at main once the parent commit is merged.

korran:
> LGTM. I assume you'll just re-target this at main once the parent commit is merged.

That's the plan.

rusty1968 (dismissed):

runtime/src/lib.rs
```diff
@@ -103,10 +98,11 @@ pub struct Drivers<'a> {
...
     pub ecc384: Ecc384,
 
-    pub fht: &'a mut FirmwareHandoffTable,
+    pub persistent_data: PersistentDataAccessor,
 
-    /// ImageManifest for the currently running image
-    pub manifest: &'a ImageManifest,
+    /// A copy of the ImageHeader for the currently running image
+    // TODO: Use persistent_data instead
+    pub manifest: ImageManifest,
```
jhand2, line 105: I think this can just be done in this PR right? There are only a couple places that use the manifest and they all look straightforward for using `persistent_data` instead.

korran: Done. FWIW, for first pass I was just trying to get rid of all the unsafe code.

jhand2 (approved):

PR #704, opened 2023-08-29
Use PersistentDataAccessor in ROM Firmware

korran:
I recommend reviewing each commit independently.

This allows us to remove quite a bit of unsafe code. The PersistentDataAccessor is introduced in https://github.com/chipsalliance/caliptra-sw/pull/690 (Add safe abstraction for accessing persistent data).

Also, support byte-length requests in mailbox driver. This allows for less awkward interaction with zerocopy as_bytes_mut(), which is useful when dealing with the persistent data directly rather than through word slices.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor.rs
@@ -188,7 +188,7 @@ impl FirmwareProcessor {
             core::slice::from_raw_parts_mut(ptr, core::mem::size_of::<ImageManifest>() / 4)
         };
 
-        txn.copy_request(slice)?;
+        txn.copy_request(slice.as_bytes_mut())?;
 
         let opt = ImageManifest::read_from(slice.as_bytes());
         let result = opt.is_some();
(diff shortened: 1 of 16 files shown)


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -181,24 +182,14 @@ impl FirmwareProcessor {
...
-        let slice = unsafe {
-            let ptr = MAN1_ORG as *mut u32;
-            core::slice::from_raw_parts_mut(ptr, core::mem::size_of::<ImageManifest>() / 4)
-        };
-
-        txn.copy_request(slice.as_bytes_mut())?;
-
-        let opt = ImageManifest::read_from(slice.as_bytes());
-        let result = opt.is_some();
-        if cfi_launder(result) {
```
korran, line 194: NOTE: This CFI check doesn't really make sense; opt will only be None when the slice is smaller than slice.len()*4, which is impossible.

Perhaps this code was expecting read_from() to some sort of validation?

rom/dev/src/pcr.rs
```diff
@@ -129,6 +130,7 @@ pub(crate) fn extend_pcrs(
 ///
 #[cfg_attr(not(feature = "no-cfi"), cfi_mod_fn)]
 pub fn log_pcr(
+    pcr_log: &mut [PcrLogEntry; 17],
```
mhatrevi, line 133: Suggest using a const instead of hardcoding.

mhatrevi (changes requested):

rom/dev/src/pcr.rs
korran: Replaced with a PcrLogArray typedef.

mhatrevi (approved):

PR #709, opened 2023-08-30
Reconstruct IDevId cert at runtime

### LINKED ISSUE #294: Consider reconstructing IDevID cert in Runtime
There are fuses for the vendor to program a signature over IDevID. It would be possible for Caliptra Runtime firmware to expose a mailbox command which takes some data (like the issuer name) and reconstructs the IDevID certificate.

A few things we would need to do for this to work:

* Develop strict requirements for vendor CAs on the contents of the IDevID cert they produce
* Pass some extra info in DataVault or DCCM from ROM to runtime
    * IDevID Public Key
    * IDevID Serial Number
    * IDevID UEID

cc @jameszhang-nvidia @mhatrevi @bluegate010

ArthurHeymans:
This adds 2 mailbox commands:
1) GET_IDEV_INFO which outputs the IDevId public key
2) GET_IDEV_CSR which appends the signature from the IdevId cert attr fuses

Closes #294

### DIFF
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -9,6 +9,7 @@ pub struct CommandId(pub u32);
 impl CommandId {
     pub const FIRMWARE_LOAD: Self = Self(0x46574C44); // "FWLD"
     pub const GET_IDEV_CSR: Self = Self(0x49444556); // "IDEV"
+    pub const GET_IDEV_CERT: Self = Self(0x49444543); // IDEC
     pub const GET_IDEV_INFO: Self = Self(0x49444549); // IDEI
     pub const GET_LDEV_CERT: Self = Self(0x4C444556); // "LDEV"
     pub const ECDSA384_VERIFY: Self = Self(0x53494756); // "SIGV"
@@ -51,6 +52,7 @@ impl From<CommandId> for u32 {
 #[allow(clippy::large_enum_variant)]
 pub enum MailboxResp {
     Header(MailboxRespHeader),
+    GetIdevCert(GetIdevCertResp),
     GetIdevCsr(GetIdevCsrResp),
     GetIdevInfo(GetIdevInfoResp),
     GetLdevCert(GetLdevCertResp),
@@ -65,6 +67,7 @@ impl MailboxResp {
     pub fn as_bytes(&self) -> &[u8] {
         match self {
             MailboxResp::Header(resp) => resp.as_bytes(),
+            MailboxResp::GetIdevCert(resp) => resp.as_bytes(),
             MailboxResp::GetIdevCsr(resp) => resp.as_bytes(),
             MailboxResp::GetIdevInfo(resp) => resp.as_bytes(),
             MailboxResp::GetLdevCert(resp) => resp.as_bytes(),
@@ -79,6 +82,7 @@ impl MailboxResp {
     pub fn as_bytes_mut(&mut self) -> &mut [u8] {
         match self {
             MailboxResp::Header(resp) => resp.as_bytes_mut(),
+            MailboxResp::GetIdevCert(resp) => resp.as_bytes_mut(),
             MailboxResp::GetIdevCsr(resp) => resp.as_bytes_mut(),
             MailboxResp::GetIdevInfo(resp) => resp.as_bytes_mut(),
             MailboxResp::GetLdevCert(resp) => resp.as_bytes_mut(),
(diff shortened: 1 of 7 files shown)


ArthurHeymans:
I'm not sure how to properly write tests for this other than some minimal validations like size. The fuses contain 0 in this setup.

common/src/mailbox_api.rs
```diff
@@ -152,6 +166,16 @@ impl GetIdevCsrResp {
     pub const DATA_MAX_SIZE: usize = 1024;
 }
 
+// GET_IDEV_INFO
+// No command-specific input args
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIdevInfoResp {
```
jhand2, line 173: Can you add this command to the runtime spec (runtime/README.md)?

common/src/mailbox_api.rs
```diff
@@ -140,7 +144,17 @@ impl Default for MailboxRespHeader {
...
 
 // GET_IDEV_CSR
-// No command-specific input args
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIdevCsrReq {
+    pub hdr: MailboxReqHeader,
+    pub data_size: u32,
+    pub data: [u8; GetIdevCsrReq::DATA_MAX_SIZE], // variable length
+}
```
jhand2, line 153: The idev CSR is a Certificate Signing Request generated by ROM that is self-signed. This is used during manufacturing so that the vendor's manufacturing CA can decide if they want to generate and sign a cert for the IDevID public key.

Instead of using the idev CSR command, we should define a new command (`GET_IDEV_CERT`) for for appending the signature.

common/src/mailbox_api.rs
```diff
@@ -140,7 +144,17 @@ impl Default for MailboxRespHeader {
 }
 
 // GET_IDEV_CSR
-// No command-specific input args
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIdevCsrReq {
+    pub hdr: MailboxReqHeader,
+    pub data_size: u32,
+    pub data: [u8; GetIdevCsrReq::DATA_MAX_SIZE], // variable length
```
jhand2, line 152: For clarity, maybe we should call these fields `tbs_size` and `tbs`

runtime/src/info.rs
```diff
@@ -25,3 +28,55 @@ impl FwInfoCmd {
...
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            };
+
+            if in_len > cmd.data.len() {
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            }
+            let mut resp_data = [0; GetIdevCsrResp::DATA_MAX_SIZE];
+            resp_data[..in_len].copy_from_slice(&cmd.data[..in_len]);
+
+            let cert = drivers.soc_ifc.get_idevid_cert_attr();
```
jhand2, line 61: This is a signature, not a cert. Consider changing name to something like `sig`

runtime/src/info.rs
```diff
@@ -25,3 +28,55 @@ impl FwInfoCmd {
...
+
+            let Ok(in_len) = usize::try_from(cmd.data_size) else {
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            };
+
+            if in_len > cmd.data.len() {
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            }
+            let mut resp_data = [0; GetIdevCsrResp::DATA_MAX_SIZE];
+            resp_data[..in_len].copy_from_slice(&cmd.data[..in_len]);
```
jhand2, line 59: X.509 encoding wraps the TBS and signature in an additional header and includes algorithm information along with the signature, so it's not sufficient to just concatenate them.

Instead, you can use [Ecdsa384CertBulder](https://github.com/chipsalliance/caliptra-sw/blob/8b23c0a6d1fb727c932d2f84b9ca4aa543214da8/x509/src/cert_bldr.rs#L187) and call `.build` which will do the proper X.509 encoding.

ArthurHeymans:
Changed the mailbox as requested to have the signature as input.

jhand2 (dismissed):

jhand2:
PR looks good, there's just a CI failure after rebasing. I'll merge once that is fixed.

jhand2 (approved):

PR #714, opened 2023-08-30
val-ROM image validation option

nquarton:


nquarton:
Replaced with https://github.com/chipsalliance/caliptra-sw/pull/729

PR #717, opened 2023-08-31
Implement Platform::get_certificate_chain

sree-revoori1:
The Platform implementation is complete with this commit.

We also add rtalias_tbs_size to the Firmware handoff table for use in the dice module, so we can piece the cert chain together.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -19,6 +19,7 @@ crypto.workspace = true
 platform.workspace = true
 ufmt.workspace = true
 zerocopy.workspace = true
+arrayvec.workspace = true
 
 [build-dependencies]
 caliptra_common = { workspace = true, default-features = false }
--- a/runtime/src/dice.rs
+++ b/runtime/src/dice.rs
@@ -14,6 +14,7 @@ use caliptra_x509::{Ecdsa384CertBuilder, Ecdsa384Signature};
 enum CertType {
     LDevId,
     FmcAlias,
+    RtAlias,
 }
 
 pub struct GetLdevCertCmd;
@@ -80,6 +81,18 @@ pub fn copy_fmc_alias_cert(
     cert_from_dccm(dv, persistent_data, cert, CertType::FmcAlias)
 }
 
+/// Copy RT Alias certificate produced by ROM to `cert` buffer
+///
+/// Returns the number of bytes written to `cert`
+#[inline(never)]
+pub fn copy_rt_alias_cert(
+    dv: &DataVault,
+    persistent_data: &PersistentData,
+    cert: &mut [u8],
+) -> CaliptraResult<usize> {
+    cert_from_dccm(dv, persistent_data, cert, CertType::RtAlias)
+}
+
 /// Copy a certificate from `dccm_offset`, append signature, and write the
 /// output to `cert`.
 fn cert_from_dccm(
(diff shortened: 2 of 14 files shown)


sree-revoori1:
Will add an integration test calling get_certificate_chain via the invoke DPE mbox cmd, but wanted to get thoughts on implementation first.

runtime/src/lib.rs
```diff
@@ -74,6 +74,15 @@ impl From<RtBootStatus> for u32 {
...
 pub const DPE_SUPPORT: Support = Support::all();
 
+pub struct CertChain {
+    pub cert: [u8; Self::MAX_CERT_CHAIN_SIZE],
+    pub cert_chain_size: usize,
+}
+
+impl CertChain {
+    const MAX_CERT_CHAIN_SIZE: usize = 4096;
+}
```
jhand2, line 84: Maybe ArrayVec would be better here. Then you can use `try_push` to push bytes into it.

sree-revoori1: Done. However it did result in extra stack allocation since I couldn't figure out a way to safely pass the ArrayVec to copy_fmc_alias_cert, copy_ldevid_cert, etc.

runtime/src/dice.rs
```diff
@@ -101,6 +116,10 @@ fn cert_from_dccm(
                 .get(..persistent_data.fht.fmcalias_tbs_size.into()),
             dv.fmc_dice_signature(),
         ),
+        CertType::RtAlias => (
+            persistent_data.rtalias_tbs.get(..RTALIAS_TBS_SIZE as usize),
```
jhand2, line 120: We should probably add rt alias size to the FHT in this PR. Otherwise we are generating the cert incorrectly, because RTALIAS_TBS_SIZE is the size of the whole DCCM region, not just the TBS portion of the cert.

runtime/Cargo.toml
```diff
@@ -19,6 +19,7 @@ crypto.workspace = true
 platform.workspace = true
 ufmt.workspace = true
 zerocopy.workspace = true
+arrayvec = { version = "0.7.4", default-features = false }
```
jhand2, line 22: We should add this as a workspace dependency in the root Cargo.toml

runtime/src/lib.rs
```diff
@@ -146,17 +150,54 @@ impl Drivers {
...
+        let mut cert_chain = ArrayVec::<u8, MAX_CERT_CHAIN_SIZE>::new();
+        for i in 0..cert_chain_size {
+            cert_chain
+                .try_push(
+                    *cert
+                        .get(i)
+                        .ok_or(CaliptraError::RUNTIME_CERT_CHAIN_CREATION_FAILED)?,
+                )
+                .map_err(|_| CaliptraError::RUNTIME_CERT_CHAIN_CREATION_FAILED)?;
+        }
```
jhand2, line 189: Can you move the cert chain building to a helper function? I think that will make the driver init code easier to understnand.

runtime/src/dice.rs
sree-revoori1: Done

runtime/Cargo.toml
sree-revoori1: Done

runtime/src/lib.rs
sree-revoori1: Agree, done

jhand2 (approved):

rom/dev/src/fht.rs
```diff
@@ -37,6 +37,9 @@ pub struct FhtDataStore {
     /// FmcAlias TBS size
     pub fmcalias_tbs_size: u16,
 
+    /// RtAlias TBS size
+    pub rtalias_tbs_size: u16,
```
jhand2, line 41: I think you can remove this since ROM won't ever populate this. Can just set rtalias_tbs_size to 0 when constructing the FHT.

sree-revoori1: Done.

jhand2 (approved):

bluegate010 (approved):

korran (approved):
Approval for drivers/

PR #720, opened 2023-09-01
Make sw-emulator and driver tests compatible with latest RTL

korran:
Various driver tests have been broken under verilator for awhile, and sw-emulator hadn't been updated with the new behavior:

* Mailbox FIFO should return 0 when empty
* Mailbox should enter error state when SoC clears exec bit unexpectedly.
* Hide hmac output from cpu with input from key-vault.

This PR updates the tests to expect what the hardware actually does, and updates sw-emulator to match.

See commit descriptions for more details.


sw-emulator/lib/periph/src/hmac_sha384.rs
```diff
@@ -432,7 +484,10 @@ impl HmacSha384 {
 
     fn op_complete(&mut self) {
         // Retrieve the tag
-        self.hmac.tag(self.tag.data_mut());
+        self.hmac.tag(self.tag.as_bytes_mut());
+        // Don't reveal the tag to the CPU if in the inputs came from the
```
mhatrevi, line 488: NIT: "if the inputs came from"

mhatrevi (dismissed):

sw-emulator/lib/periph/src/hmac_sha384.rs
korran: Whoops. Fixed.

mhatrevi (approved):

PR #722, opened 2023-09-05
FMC : Log PCR extensions

rusty1968:
This change logs a PCR entry to DCCM when a PCR is extended.

- Adds a 4-byte `pcr_log_index ` to the FHT to handoff.  
-  Adds tests to validate the log against the PCR values in hardware.

### DIFF
--- a/fmc/src/flow/pcr.rs
+++ b/fmc/src/flow/pcr.rs
@@ -23,9 +23,17 @@ Note:
 use crate::flow::tci::Tci;
 use crate::fmc_env::FmcEnv;
 use crate::HandOff;
-use caliptra_drivers::{okref, CaliptraResult, PcrId};
+use caliptra_drivers::{
+    cprintln,
+    memory_layout::{PCR_LOG_ORG, PCR_LOG_SIZE},
+    okref,
+    pcr_log::{PcrLogEntry, PcrLogEntryId},
+    CaliptraResult, PcrBank, PcrId,
+};
 
 use caliptra_common::{RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR};
+use caliptra_error::CaliptraError;
+use zerocopy::AsBytes;
 
 /// Extend current PCR
 ///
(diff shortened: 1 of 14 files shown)


fmc/src/flow/pcr.rs
```diff
@@ -61,11 +69,63 @@ fn extend_pcr_common(env: &mut FmcEnv, hand_off: &HandOff, pcr_id: PcrId) -> Cal
     let rt_tci: [u8; 48] = okref(&rt_tci)?.into();
     env.pcr_bank.extend_pcr(pcr_id, &mut env.sha384, &rt_tci)?;
 
+    log_pcr(
```
bluegate010, line 72: Optional but we can be a bit more efficient with the logs by refactoring to emit one log and indicating which PCRs it ends up in based on the reset reason. But I think that would be a pretty invasive change, and it doesn't look like we're logging too much here.

fmc/src/flow/pcr.rs
```diff
@@ -61,11 +69,63 @@ fn extend_pcr_common(env: &mut FmcEnv, hand_off: &HandOff, pcr_id: PcrId) -> Cal
...
+        PcrLogEntryId::FwImageManifest,
+        1 << pcr_id as u8,
+        &manifest_digest,
+    )
+}
+
+pub fn log_pcr(
+    pcr_bank: &mut PcrBank,
+    pcr_entry_id: PcrLogEntryId,
+    pcr_ids: u32,
```
bluegate010, line 96: If this function won't be used elsewhere in FMC, could make it non-public and change pcr_id to just accept a single value.

fmc/tests/test_rtalias.rs
```diff
@@ -45,7 +59,231 @@ fn test_boot_status_reporting() {
...
+
+        pcr.copy_from_slice(digest);
+    }
+
+    pcr
+}
 
-    let mut output = vec![];
-    hw.copy_output_until_exit_success(&mut output).unwrap();
+pub fn change_dword_endianess(data: &mut [u8]) {
```
bluegate010, line 284: Any need for this to be marked public?

fmc/src/flow/pcr.rs
```diff
@@ -23,9 +23,17 @@ Note:
...
+    okref,
+    pcr_log::{PcrLogEntry, PcrLogEntryId},
+    CaliptraResult, PcrBank, PcrId,
+};
 
 use caliptra_common::{RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR};
+use caliptra_error::CaliptraError;
+use zerocopy::AsBytes;
 
 /// Extend current PCR
```
bluegate010, line 38: This function erases the current PCR if the reset reason is an update reset. Can we have test coverage for that?

fmc/src/flow/pcr.rs
```diff
@@ -61,11 +69,63 @@ fn extend_pcr_common(env: &mut FmcEnv, hand_off: &HandOff, pcr_id: PcrId) -> Cal
     let rt_tci: [u8; 48] = okref(&rt_tci)?.into();
     env.pcr_bank.extend_pcr(pcr_id, &mut env.sha384, &rt_tci)?;
```
bluegate010, line 71: `extend_journey_pcr` is only invoked if the reset reason is cold-reset or update-reset. Do we have a way to trigger a reset whose reason is not either of those, so we can test that logic?

fmc/src/flow/pcr.rs
rusty1968: Let's defer it to a future PR then.

bluegate010:
Closing as this was completed in https://github.com/chipsalliance/caliptra-sw/pull/780

PR #727, opened 2023-09-05
Use SoC regs to configure `entropy_src` health test thresholds

rkr35:
See https://github.com/chipsalliance/caliptra-rtl/issues/190 for background.


drivers/test-fw/src/bin/csrng_fail_repcnt_soc_threshold_tests.rs
```diff
@@ -0,0 +1,51 @@
...
+use caliptra_test_harness::test_suite;
+
+fn test_boot_fail_repcnt_check() {
+    let csrng_reg = unsafe { CsrngReg::new() };
+    let entropy_src_reg = unsafe { EntropySrcReg::new() };
+    let mut soc_ifc_reg = unsafe { SocIfcReg::new() };
+    soc_ifc_reg
+        .regs_mut()
+        .cptra_i_trng_entropy_config_1()
+        .write(|w| w.repetition_count(20));
```
korran, line 34: In all these tests, do you mind setting these from the SoC side rather than the uC? If there's a problem in the hardware I'd like these tests to notice.

korran (approved):

mhatrevi (approved):

Issue #357, opened 2023-06-06
Runtime should use FHT table to locate cert signatures


rusty1968:
Runtime dice module is  retrieving the storage location of the FMC dice signature by invoking the data vault driver api directly.

```
        CertType::FmcAlias => (unsafe { &FMCALIAS_TBS_ORG[..] }, dv.fmc_dice_signature()),
```

It should be however,  using the following information fields of the handoff table:
```
    /// Index of FMC Certificate Signature R Component in the Data Vault.
    pub fmc_cert_sig_r_dv_hdl: HandOffDataHandle,

    /// Index of FMC Certificate Signature S Component in the Data Vault.
    pub fmc_cert_sig_s_dv_hdl: HandOffDataHandle,
```

PR #735, opened 2023-09-07
Add LdevId sig to FHT and use FHT to retrieve all signatures in RT

sree-revoori1:
We switch from using the datavault to retrieve cert signatures to using
the FHT in runtime. To do this, we add data vault handles to the
LDevId cert's signature's R and S coordinates.

Fixes https://github.com/chipsalliance/caliptra-sw/issues/357


drivers/src/hand_off.rs
```diff
@@ -275,10 +275,15 @@ pub struct FirmwareHandoffTable {
     /// Fuse log Address
     pub fuse_log_addr: u32,
 
+    // RtAlias public key.
```
jhand2, line 278: nit: rust style for doc comments (function, struct, member comments) is to start with `///`. If we every auto-generate documentation for these libraries, the docs will pick up all the doc comments.

runtime/src/dice.rs
```diff
@@ -78,54 +122,38 @@ pub fn copy_fmc_alias_cert(
...
 
-/// Copy a certificate from `dccm_offset`, append signature, and write the
-/// output to `cert`.
-fn cert_from_dccm(
-    dv: &DataVault,
-    persistent_data: &PersistentData,
+/// Create a certificate from a tbs and a signature and write the output to `cert`
+fn cert_from_tbs_and_sig(
+    tbs: Option<&[u8]>,
+    sig: Ecc384Signature,
```
jhand2, line 149: Pass by reference to avoid an additional copy

jhand2 (approved):

drivers/src/hand_off.rs
sree-revoori1: Done

jhand2 (approved):

drivers/src/hand_off.rs
```diff
@@ -275,21 +275,26 @@ pub struct FirmwareHandoffTable {
...
     pub fuse_log_addr: u32,
 
+    /// RtAlias public key.
     pub rt_dice_pub_key: Ecc384PubKey,
 
+    /// RtAlias certificate signature.
     pub rt_dice_sign: Ecc384Signature,
 
+    /// LdevId certificate signature.
+    pub ldevid_dice_sign: Ecc384Signature,
```
korran, line 285: Please update the docs at https://github.com/chipsalliance/caliptra-sw/blob/main/fmc/README.md with this change.

sree-revoori1: Done

jhand2 (dismissed):

fmc/README.md
```diff
@@ -139,10 +139,11 @@ fields may not be changed or removed). Table revisions with different Major Vers
...
 | rt_dice_sign          | 96           | FMC        | RT Alias DICE signature.                                                                                 |
+| ldevid_dice_sign      | 96           | FMC        | LDevID DICE signature.                                                                                 |
 | idev_dice_pub_key     | 96           | ROM        | Initial Device ID Public Key.                                                                            |
 | rom_info_addr         | 4            | ROM        | Address of ROMInfo struct describing the ROM digest and git commit.                                      |
 | rtalias_tbs_size      | 2            | FMC        | RT Alias TBS Size.                                                                                      |
-| reserved              | 128          |            | Reserved for future use.                                                                                 |
+| reserved              | 30           |            | Reserved for future use.                                                                                 |
```
FerralCoder, line 146: Should we consider increasing FHT size so that we have more buffer for future needs?

fmc/README.md
```diff
@@ -139,10 +139,11 @@ fields may not be changed or removed). Table revisions with different Major Vers
...
 
 *FHT is currently defined to be 512 bytes in length.*
```
FerralCoder, line 148: New fields also need to be described in the section directly below

sree-revoori1: Done.

fmc/README.md
sree-revoori1: Will defer to @jhand2 here.

FerralCoder (approved):

fmc/README.md
sree-revoori1: Spoke to Jordan, and we decided to increase the FHT size to 2048, since we already have 2k bytes reserved for the FHT in DCCM. I've updated the docs accordingly.

jhand2 (approved):

drivers/src/hand_off.rs
```diff
@@ -337,11 +342,12 @@ impl Default for FirmwareHandoffTable {
             ldevid_tbs_size: 0,
             fmcalias_tbs_size: 0,
             rtalias_tbs_size: 0,
-            reserved: [0u8; 126],
+            reserved: [0u8; 1566],
             ldevid_tbs_addr: 0,
             fmcalias_tbs_addr: 0,
             pcr_log_addr: 0,
             fuse_log_addr: 0,
+            ldevid_dice_sign: Ecc384Signature::default(),
```
FerralCoder, line 350: Probably for a separate PR, but we should re-order this Default to match the order of the actual structure. Better for clarity and maintenance

FerralCoder (dismissed):

drivers/src/hand_off.rs
FerralCoder: opened https://github.com/chipsalliance/caliptra-sw/issues/769

jhand2: Ack, I can follow up on this

mhatrevi (changes requested):
This change seems unnecessary. The reason to put the signature in the Datavault (andretrieve it from there) was that it provided locking protection.

jhand2 (dismissed):

jhand2:
@mhatrevi can you take another pass? It is now updated to just put the DV handles into the FHT (which is what we do for all the other DV handles too)

mhatrevi (dismissed):

bluegate010 (approved):

jhand2 (approved):

mhatrevi (approved):

PR #736, opened 2023-09-07
[feat] Add code coverage to sw-emulator

rusty1968:
This  change implements a simple code coverage metric using a bitmap.

### DIFF
--- a/hw-model/src/model_emulated.rs
+++ b/hw-model/src/model_emulated.rs
@@ -144,8 +144,7 @@ impl crate::HwModel for ModelEmulated {
 
     fn step(&mut self) {
         if self.cpu_enabled.get() {
-            self.cpu
-                .step(self.trace_fn.as_deref_mut(), Some(&mut self.code_coverage));
+            self.cpu.step(self.trace_fn.as_deref_mut());
         }
     }
 
--- a/hw-model/src/model_emulated.rs
+++ b/hw-model/src/model_emulated.rs
@@ -8,6 +8,7 @@ use std::path::Path;
 use std::rc::Rc;
 
 use caliptra_emu_bus::Clock;
+use caliptra_emu_cpu::cpu::CodeCoverage;
 use caliptra_emu_cpu::Cpu;
 use caliptra_emu_cpu::InstrTracer;
 use caliptra_emu_periph::ActionCb;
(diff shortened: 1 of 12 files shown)


rusty1968:
@mhatrevi , @ajisaxena , @korran

korran (approved):

mhatrevi (dismissed):

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -15,11 +15,36 @@ Abstract:
...
+}
+impl CodeCoverage {
+    pub fn new(capacity_in_bits: usize) -> Self {
+        Self {
+            bit_vec: BitVec::from_elem(capacity_in_bits, false),
+        }
+    }
+    pub fn log_execution(&mut self, pc: RvData) {
+        if (pc as usize) < self.bit_vec.len() {
+            self.bit_vec.set(pc as usize, true);
```
korran, line 36: Since instructions are 16-bit aligned, you could half the size of your bitvec and use (pc as usize / 2) as the index. Also when you get a 32-bit risc-v instruction, be sure to set all the bits associated with the instruciton.

korran (changes requested):

sw-emulator/lib/cpu/src/cpu.rs
rusty1968: Done. Considering size of instruction and marking all bytes corresponding to a instruction as true.

hw-model/src/model_emulated.rs
```diff
@@ -54,6 +55,7 @@ pub struct ModelEmulated {
     trace_fn: Option<Box<InstrTracer<'static>>>,
     ready_for_fw: Rc<Cell<bool>>,
     cpu_enabled: Rc<Cell<bool>>,
+    code_coverage: CodeCoverage,
```
korran, line 58: Remove this?

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -13,13 +13,53 @@ Abstract:
...
+            for i in 0..num_bytes {
+                let byte_index = (pc as usize) + i;
+                if byte_index < self.bit_vec.len() {
+                    self.bit_vec.set(byte_index, true);
+                }
+            }
+        }
+    }
+
+    pub fn count_executed(&self) -> usize {
```
korran, line 54: This functions aren't necessary; we'll compare to the ELF section sizes offline.

hw-model/src/model_emulated.rs
rusty1968: Done.

sw-emulator/lib/cpu/src/cpu.rs
rusty1968: Done.

korran (approved):

PR #743, opened 2023-09-08
Move DPE Instance to persistent data in DCCM

sree-revoori1:
This is necessary so that DPE measurements don't get cleared upon warm resets.
We will still need to add logic to RT to only initialize dpe upon
cold reset and validate the DpeInstance in DCCM, but these will
take longer and don't touch ROM so they can be done later.

Note: DpeInstance only uses ~ 3400 bytes currently, but I am reserving 4k bytes in DCCM for it.
This is in case we add new fields to the DpeInstance struct.

### DIFF
--- a/drivers/Cargo.toml
+++ b/drivers/Cargo.toml
@@ -16,12 +16,14 @@ caliptra-image-types.workspace = true
 caliptra-lms-types.workspace = true
 caliptra-registers.workspace = true
 cfg-if.workspace = true
+dpe = { workspace = true, optional = true }
 ufmt.workspace = true
 ureg.workspace = true
 zerocopy.workspace = true
 
 [features]
 emu = []
+runtime = ["dep:dpe"]
 fpga_realtime = ["caliptra-hw-model/fpga_realtime"]
 itrng = ["caliptra-hw-model/itrng"]
 verilator = ["caliptra-hw-model/verilator"]
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -3,6 +3,8 @@
 use core::{marker::PhantomData, mem::size_of, ptr::addr_of};
 
 use caliptra_image_types::ImageManifest;
+#[cfg(feature = "runtime")]
+use dpe::DpeInstance;
 use zerocopy::{AsBytes, FromBytes};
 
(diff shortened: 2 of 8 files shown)


runtime/src/lib.rs
```diff
@@ -176,6 +174,7 @@ impl Drivers {
         };
         let mut pcr_bank = PcrBank::new(PvReg::new());
         let dpe = Self::initialize_dpe(env, &mut pcr_bank, locality)?;
+        persistent_data.get_mut().dpe = dpe;
```
jhand2, line 177: We should only do this on cold reset

drivers/src/persistent.rs
```diff
@@ -47,6 +48,9 @@ pub struct PersistentData {
 
     pub fuse_log: FuseLogArray,
     reserved5: [u8; memory_layout::FUSE_LOG_SIZE as usize - size_of::<FuseLogArray>()],
+
+    pub dpe: DpeInstance,
```
jhand2, line 52: Hmm. I just realized depending on DPE here has the possibility of impacting ROM code (although maybe not since ROM doesn't use it).

@korran do you have any concerns with this?

Also @sree-revoori1 can you double check that the ROM binary is the same with and without this change?

sree-revoori1: It is not the same. I did cargo build with and without this change, and then did wc -c target/debug/caliptra-rom to find the number of bytes in the caliptra-rom binary. With this change, it is 7711896 bytes, without it is 7138048 bytes.

sree-revoori1: Done.

jhand2 (approved):

drivers/Cargo.toml
```diff
@@ -16,12 +16,14 @@ caliptra-image-types.workspace = true
 caliptra-lms-types.workspace = true
 caliptra-registers.workspace = true
 cfg-if.workspace = true
+dpe = { workspace = true, optional = true }
```
korran, line 19: Discussed offline: a future PR will move the DpeInstance struct to a new dpe-types crate, and caliptra-drivers will depend on that instead of all of dpe.

korran (approved):

PR #749, opened 2023-09-11
Move ROM's DATA_ORG to the start of DCCM

sree-revoori1:
This change reorganizes ROM's usage of DCCM by moving ROM's DATA_ORG to the start of DCCM and shrinking it, thereby making space for FMC and RT to use a large portion of DCCM that is guaranteed to be untouched by ROM.

We make the following changes:
- Reserve 1k bytes at the the beginning of DCCM
- Move ROM's DATA_ORG to the start of DCCM and make it occupy exactly 996 bytes
- Move the CFI_* regions and BOOT_STATUS_ORG to the remaining 28 bytes of the reserved region
- Move all following sections 1k bytes forward (except for STACK_ORG, ESTACK_ORG, and NSTACK_ORG)
- Reduce DATA_SIZE by 1k bytes to make space for the reserved 1k bytes at start of DCCM

### DIFF
--- a/drivers/src/memory_layout.rs
+++ b/drivers/src/memory_layout.rs
@@ -25,22 +25,23 @@ pub const ROM_ORG: u32 = 0x00000000;
 pub const MBOX_ORG: u32 = 0x30000000;
 pub const ICCM_ORG: u32 = 0x40000000;
 pub const DCCM_ORG: u32 = 0x50000000;
-pub const MAN1_ORG: u32 = 0x50000000;
-pub const MAN2_ORG: u32 = 0x50001800;
-pub const FHT_ORG: u32 = 0x50003000;
-pub const LDEVID_TBS_ORG: u32 = 0x50003800;
-pub const FMCALIAS_TBS_ORG: u32 = 0x50003C00;
-pub const RTALIAS_TBS_ORG: u32 = 0x50004000;
-pub const PCR_LOG_ORG: u32 = 0x50004400;
-pub const MEASUREMENT_LOG_ORG: u32 = 0x50004800;
-pub const FUSE_LOG_ORG: u32 = 0x50004C00;
-pub const BOOT_STATUS_ORG: u32 = 0x50005000;
-pub const CFI_VAL_ORG: u32 = 0x50005004;
-pub const CFI_MASK_ORG: u32 = 0x50005008;
-pub const CFI_XO_S0_ORG: u32 = 0x5000500C;
-pub const CFI_XO_S1_ORG: u32 = 0x50005010;
-pub const CFI_XO_S2_ORG: u32 = 0x50005014;
-pub const CFI_XO_S3_ORG: u32 = 0x50005018;
+// Region of ~ 1k bytes between DCCM_ORG and CFI_VAL_ORG is reserved for ROM's DATA_ORG
+pub const CFI_VAL_ORG: u32 = 0x500003E4;
+pub const CFI_MASK_ORG: u32 = 0x500003E8;
+pub const CFI_XO_S0_ORG: u32 = 0x500003EC;
+pub const CFI_XO_S1_ORG: u32 = 0x500003F0;
+pub const CFI_XO_S2_ORG: u32 = 0x500003F4;
+pub const CFI_XO_S3_ORG: u32 = 0x500003F8;
+pub const BOOT_STATUS_ORG: u32 = 0x500003FC;
+pub const MAN1_ORG: u32 = 0x50000400;
+pub const MAN2_ORG: u32 = 0x50001C00;
+pub const FHT_ORG: u32 = 0x50003400;
+pub const LDEVID_TBS_ORG: u32 = 0x50003C00;
+pub const FMCALIAS_TBS_ORG: u32 = 0x50004000;
+pub const RTALIAS_TBS_ORG: u32 = 0x50004400;
+pub const PCR_LOG_ORG: u32 = 0x50004800;
+pub const MEASUREMENT_LOG_ORG: u32 = 0x50004C00;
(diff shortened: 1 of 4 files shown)


drivers/src/memory_layout.rs
```diff
@@ -25,25 +25,26 @@ pub const ROM_ORG: u32 = 0x00000000;
...
+pub const CFI_XO_S3_ORG: u32 = 0x500003FC;
+pub const MAN1_ORG: u32 = 0x50000400;
+pub const MAN2_ORG: u32 = 0x50001C00;
+pub const FHT_ORG: u32 = 0x50003400;
+pub const LDEVID_TBS_ORG: u32 = 0x50003C00;
+pub const FMCALIAS_TBS_ORG: u32 = 0x50004000;
+pub const RTALIAS_TBS_ORG: u32 = 0x50004400;
+pub const PCR_LOG_ORG: u32 = 0x50004800;
+pub const FUSE_LOG_ORG: u32 = 0x50004C00;
+pub const BOOT_STATUS_ORG: u32 = 0x50004FE4;
```
korran, line 43: BOOT_STATUS_ORG is also modified by the ROM out-of-band and should be moved to the ROM section.

jhand2: If boot_status is set to non-zero, are we guaranteed to not continue booting to FMC? I want to make sure we don't have a use-case for reading/reporting this value in RT.

But also, iirc there are boot status registers, so I suppose we can use those if needed.

korran: I don't think it's a problem if any of these memory location are used by the runtime firmware. The important thing is that the ROM can make writes to this section without disturbing the persistent data.

drivers/src/persistent.rs
```diff
@@ -36,7 +36,7 @@ pub struct PersistentData {
 }
 impl PersistentData {
     pub fn assert_matches_layout() {
-        const P: *const PersistentData = memory_layout::DCCM_ORG as *const PersistentData;
+        const P: *const PersistentData = memory_layout::MAN1_ORG as *const PersistentData;
```
korran, line 39: You need to replace other references to DCCM_ORG in this file as well; that's why the tests are failing.

korran (changes requested):

drivers/src/memory_layout.rs
sree-revoori1: Done

drivers/src/persistent.rs
sree-revoori1: Done

drivers/src/memory_layout.rs
```diff
@@ -113,9 +114,10 @@ fn mem_layout_test_pcrlog() {
 }
 
 #[test]
+#[ignore]
```
sree-revoori1, line 117: Ignoring this test temporarily. Will update in https://github.com/chipsalliance/caliptra-sw/pull/743 when DPE is added to PerisistentData.

drivers/src/memory_layout.rs
```diff
@@ -25,25 +25,26 @@ pub const ROM_ORG: u32 = 0x00000000;
...
+pub const FHT_ORG: u32 = 0x50003400;
+pub const LDEVID_TBS_ORG: u32 = 0x50003C00;
+pub const FMCALIAS_TBS_ORG: u32 = 0x50004000;
+pub const RTALIAS_TBS_ORG: u32 = 0x50004400;
+pub const PCR_LOG_ORG: u32 = 0x50004800;
+pub const FUSE_LOG_ORG: u32 = 0x50004C00;
+pub const DATA_ORG: u32 = 0x50005000;
+pub const STACK_ORG: u32 = 0x5001A400;
+pub const ESTACK_ORG: u32 = 0x5001FC00;
+pub const NSTACK_ORG: u32 = 0x50020000;
```
korran, line 47: 0x50020000 is the end of DCCM; you can't move this.

sree-revoori1: Moved STACK_ORG, ESTACK_ORG, and NSTACK_ORG back to original addresses, and reduced size of DATA_ORG by 1k bytes to accommodate the new reserved section at the start of DCCM.

drivers/src/memory_layout.rs
```diff
@@ -61,7 +62,7 @@ pub const FMCALIAS_TBS_SIZE: u32 = 1024;
 pub const RTALIAS_TBS_SIZE: u32 = 1024;
 pub const PCR_LOG_SIZE: usize = 1024;
 pub const FUSE_LOG_SIZE: usize = 996;
```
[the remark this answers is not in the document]
sree-revoori1: Done

drivers/src/persistent.rs
```diff
@@ -65,7 +65,6 @@ impl PersistentData {
                 memory_layout::MEASUREMENT_LOG_ORG
             );
             assert_eq!(addr_of!((*P).fuse_log) as u32, memory_layout::FUSE_LOG_ORG);
-            assert_eq!(P.add(1) as u32, memory_layout::BOOT_STATUS_ORG);
```
korran, line 68: We still need to assert the size of the struct to make sure there's no overlap. How about:

```
assert_eq!(P.add(1) as u32, memory_layout::FUSE_LOG_ORG + memory_layout::FUSE_LOG_SIZE);
```

korran (changes requested):

drivers/src/persistent.rs
sree-revoori1: Done.

korran (approved):

mhatrevi (approved):

jhand2 (approved):

PR #756, opened 2023-09-12
PCR_QUOTE mailbox commands

sphinxc0re:
The initial implementation for the PCR_QUOTE mailbox command. The only thing not implemented yet are the monotonic reset counters.


korran:
Fix for the "weird behavior" in https://github.com/chipsalliance/caliptra-sw/pull/763

nquarton:
> need to add FIPS status response, see PR #734

Thanks for being on top of checking for this, Lou! It actually looks like this one is all good though. It includes and populates the MailboxRespHeader in the response struct.

FerralCoder:
> > need to add FIPS status response, see PR #734
> 
> Thanks for being on top of checking for this, Lou! It actually looks like this one is all good though. It includes and populates the MailboxRespHeader in the response struct.

Oh sorry, didn't realize this was in the header

FerralCoder (approved):

drivers/src/sha512.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct Sha512 {
+    sha512: Sha512Reg,
+}
+
+impl Sha512 {
+    pub fn new(sha512: Sha512Reg) -> Self {
+        Self { sha512 }
+    }
+
+    pub fn gen_pcr_hash(&mut self, nonce: Array4x8) -> CaliptraResult<Array4x12> {
```
jhand2, line 17: I think this should just be moved to the sha384 driver (which also uses the sha512 peripheral) since it's generating a sha384 hash.

drivers/src/sha512.rs
```diff
@@ -0,0 +1,40 @@
...
+        let status_reg = reg.gen_pcr_hash_status();
+
+        // Wait for the registers to be ready
+        while !status_reg.read().ready() {}
+
+        // Write the nonce into the register
+        reg.gen_pcr_hash_nonce().write(&nonce.into());
+
+        // Use the start command to start the digesting process
+        reg.gen_pcr_hash_ctrl().write(|ctrl| ctrl.start(true));
```
jhand2, line 29: Does mode need to be set first? https://github.com/chipsalliance/caliptra-sw/blob/9f1d077f7ba1e463c23c63b50b19809209c1e74c/drivers/src/sha384.rs#L266

runtime/src/lib.rs
```diff
@@ -316,7 +325,11 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         CommandId::GET_IDEV_CSR => Err(CaliptraError::RUNTIME_UNIMPLEMENTED_COMMAND),
         CommandId::GET_IDEV_INFO => IDevIdInfoCmd::execute(drivers),
         CommandId::GET_LDEV_CERT => Err(CaliptraError::RUNTIME_UNIMPLEMENTED_COMMAND),
+        CommandId::INCREMENT_PCR_RESET_COUNTER => {
+            IncrementPcrResetCounter::execute(drivers, cmd_bytes)
+        }
         CommandId::INVOKE_DPE => InvokeDpeCmd::execute(drivers, cmd_bytes),
+        CommandId::QUOTE_PCRS => get_pcr_quote(drivers, cmd_bytes),
```
jhand2, line 332: nit: Follow existing convention of <ReqStruct>::execute

runtime/src/pcr.rs
```diff
@@ -0,0 +1,107 @@
...
+    let DataStore::KeyVaultSlot(key_id) = priv_key_datastore else {
+        return Err(CaliptraError::DRIVER_BAD_DATASTORE_VAULT_TYPE);
+    };
+
+    let pub_key = drivers.data_vault.fmc_pub_key();
+
+    let priv_key_args = KeyReadArgs::new(key_id);
+    let priv_key = Ecc384PrivKeyIn::Key(priv_key_args);
+
+    let signature = drivers
```
jhand2, line 76: Need to use the PCR sign flow instead (probably needs to be added to the ecc driver). RT does not have direct access to the FMC alias key, it can only use it via the PCR sign hw support.

jhand2 (changes requested):

common/src/mailbox_api.rs
```diff
@@ -210,6 +215,28 @@ impl GetLdevCertResp {
...
+
+pub type PcrValue = [u8; 48];
+
+/// QUOTE_PCRS output
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct QuotePcrsResp {
+    pub hdr: MailboxRespHeader,
+    /// The PCR values
+    pub pcrs: [PcrValue; 32],
```
bluegate010, line 234: Might be useful to include the nonce in the response, just to allow the response to be cryptographically verifiable on its own.

runtime/src/pcr.rs
sphinxc0re: Okay, how would I go about implementing that? I don't see any mention of PCR sign hw support in the spec 🤔

runtime/src/lib.rs
sphinxc0re: I was wondering, what that was about. Since there isn't any data associated with those handlers, wouldn't it be better to just use modules? If that's not to your liking, maybe a trait makes sense in this case?

common/src/mailbox_api.rs
sphinxc0re: Since that's not how the spec defines it, should I just alter the spec with this PR or should the spec changes go into a different one?

sphinxc0re:
I'm going to need some feedback on the pcr_sign_flow implementation. Is there more documentation on the flow somewhere?

common/src/mailbox_api.rs
jhand2: You can update the spec in the same PR

common/src/mailbox_api.rs
```diff
@@ -270,6 +297,14 @@ pub struct StashMeasurementResp {
 // No command-specific input args
 // No command-specific output args
 
+// INCREMENT_PCR_RESET_COUNTER
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct IncrementPcrResetCounterReq {
```
jhand2, line 303: This is currently missing from the spec, but all Caliptra mailbox responses should start with `chksum` and `fips_status` fields. Could you update this PR with:

1. Add response structs for all commands
2. Update spec
3. Add these fields

drivers/src/ecc384.rs
```diff
@@ -280,6 +280,39 @@ impl Ecc384 {
...
+        data: &Ecc384Scalar,
+        trng: &mut Trng,
+    ) -> CaliptraResult<Ecc384Signature> {
+        let ecc = self.ecc.regs_mut();
+
+        // Wait for hardware ready
+        wait::until(|| ecc.status().read().ready());
+
+        // Copy digest
+        KvAccess::copy_from_arr(data, ecc.msg())?;
```
jhand2, line 294: I would expect that `pcr_sign` computes the message itself rather than taking as input. If you remove this, is the data that's output the same?

runtime/src/pcr.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn new() -> PcrResetCounter {
+        PcrResetCounter { counter: [0; 32] }
+    }
+
+    pub fn get(&self, id: PcrId) -> u32 {
+        self.counter[usize::from(id)]
+    }
+
+    pub fn increment(&mut self, id: PcrId) {
+        self.counter[usize::from(id)] += 1;
```
jhand2, line 33: This can overflow and panic. Can we add a check for this? Can just fail the command if an increment will overflow.

runtime/src/pcr.rs
```diff
@@ -0,0 +1,92 @@
...
+            });
+
+        let reset_ctrs = PcrBank::ALL_PCR_IDS
+            .into_iter()
+            .map(|pcr_id| drivers.pcr_reset.get(pcr_id))
+            .enumerate()
+            .fold([0; 32], |mut acc, (idx, reset_cnt)| {
+                acc[idx] = reset_cnt;
+                acc
+            });
```
jhand2, line 82: We already store these in a slice. Can we just copy the `counter` field directly?

runtime/src/pcr.rs
jhand2, line 1: Can you add tests for these mailbox commands in `runtime/tests/integration_tests.rs`?

ArthurHeymans:
> I've added an initial integration test for the pcr sign flow. The problem is that the runtime throws a `RUNTIME_GLOBAL_EXCEPTION` error when _reading_ from the pcr sign status register or any other pcr sign register for that matter. Can't really tell what the problem is here 🤔

IIRC The emulator does not implement this MMIO so it thinks unused memory is accessed which is that error.

jhand2:
> > I've added an initial integration test for the pcr sign flow. The problem is that the runtime throws a `RUNTIME_GLOBAL_EXCEPTION` error when _reading_ from the pcr sign status register or any other pcr sign register for that matter. Can't really tell what the problem is here 🤔
> 
> IIRC The emulator does not implement this MMIO so it thinks unused memory is accessed which is that error.

Ya, it looks 

> > I've added an initial integration test for the pcr sign flow. The problem is that the runtime throws a `RUNTIME_GLOBAL_EXCEPTION` error when _reading_ from the pcr sign status register or any other pcr sign register for that matter. Can't really tell what the problem is here 🤔
> 
> IIRC The emulator does not implement this MMIO so it thinks unused memory is accessed which is that error.

@sphinxc0re could you push what you have to your branch? I can pull the code and take a look, but it looks like there are currently some build failures.

sphinxc0re:
@jhand2 that should do the trick. I left in the debug `cprintln`s for convenience. I hope that's not a problem.

PR #757, opened 2023-09-12
Disable attestation when executing a mbox cmd in a warm reset

sree-revoori1:


runtime/src/lib.rs
```diff
@@ -360,6 +360,15 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
...
         enter_idle(drivers);
+        if drivers.mbox.cmd_busy() {
+            let reset_reason = drivers.soc_ifc.reset_reason();
+            if reset_reason == ResetReason::WarmReset {
+                let mut result = DisableAttestationCmd::execute(drivers);
+                if let Ok(mut resp) = result {
+                    Packet::copy_to_mbox(drivers, &mut resp);
+                }
+            }
+        }
```
jhand2, line 371: I think we want to do this outside of the loop so we only call disable attestation once. We should also print something to UART.

runtime/src/lib.rs
```diff
@@ -360,6 +360,15 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
     caliptra_drivers::report_boot_status(RtBootStatus::RtReadyForCommands.into());
     loop {
         enter_idle(drivers);
+        if drivers.mbox.cmd_busy() {
+            let reset_reason = drivers.soc_ifc.reset_reason();
+            if reset_reason == ResetReason::WarmReset {
+                let mut result = DisableAttestationCmd::execute(drivers);
```
jhand2, line 366: I think it would be good to add something to the [FwInfo command](https://github.com/chipsalliance/caliptra-sw/blob/b04acd1b50330379971ba084e5afea49f39a1de8/runtime/src/info.rs#L12) which says whether attestation is currently disabled. This will help callers troubleshoot what is wrong.

runtime/src/lib.rs
sree-revoori1: Done

runtime/src/lib.rs
sree-revoori1: Done

runtime/src/lib.rs
```diff
@@ -358,6 +361,18 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
     // Indicator to SOC that RT firmware is ready
     drivers.soc_ifc.assert_ready_for_runtime();
     caliptra_drivers::report_boot_status(RtBootStatus::RtReadyForCommands.into());
+    // Disable attestation if in the middle of executing an mbox cmd during warm reset
+    if drivers.mbox.cmd_busy() {
+        let reset_reason = drivers.soc_ifc.reset_reason();
+        if reset_reason == ResetReason::WarmReset {
+            let mut result = DisableAttestationCmd::execute(drivers);
+            if let Ok(mut resp) = result {
+                Packet::copy_to_mbox(drivers, &mut resp);
```
jhand2, line 370: I don't think we need to copy the response to the mailbox since we're just executing it ourselves.

But once we call disable attestation, we probably want to get the mailbox back into idle state so that other mailbox commands still work.

sree-revoori1: Done

runtime/src/lib.rs
```diff
@@ -358,6 +361,19 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> ! {
     // Indicator to SOC that RT firmware is ready
     drivers.soc_ifc.assert_ready_for_runtime();
     caliptra_drivers::report_boot_status(RtBootStatus::RtReadyForCommands.into());
+    // Disable attestation if in the middle of executing an mbox cmd during warm reset
+    if drivers.mbox.cmd_busy() {
+        let reset_reason = drivers.soc_ifc.reset_reason();
+        if reset_reason == ResetReason::WarmReset {
+            let mut result = DisableAttestationCmd::execute(drivers);
+            if let Ok(mut resp) = result {
```
jhand2, line 369: I think if this fails we want to fatal error

jhand2 (approved):

Issue #290, opened 2023-05-18
Consider including context when extending measurements at boot-time


bluegate010:
Came up in https://github.com/chipsalliance/caliptra-sw/pull/259.

Currently ROM measures raw uint8s and digests, where the only context for what each measurement indicates is the order the measurement was made. It would be moderately more robust/future-proof to also measure an identifier of some kind, to disambiguate what a given uint8 means.

PR #773, opened 2023-09-13
Measure the PCR entry ID into the PCR, to provide context hints for each entry

bluegate010:
Addresses https://github.com/chipsalliance/caliptra-sw/issues/290


bluegate010:
I'm okay if folks would rather have this not go in. Just a suggestion.

jhand2 (commented):
I think if data1 is relevant to the security of the measurement, it should just be included in the measurement. Otherwise I think it's confusing for the verifier since they can't do a direct lookup for the measurement of a piece of software. The measurement deposited into Caliptra would be different than what would be measured into a different RoT.

bluegate010:
After discussion, logging low-entropy values to PCRs without context hints is an existing practice so will leave this alone.

PR #778, opened 2023-09-14
UART driver: toggle a bit on every write.

korran:
This modification toggles generic-output-wire bit 8 every time a character is written, so the outside world can tell when we've written a new character without having to introspect internal signals.

As it is no longer necessary, remove the internal signal introspection from the verilator hardware model and have it look at the new toggle bit instead.


drivers/test-fw/src/bin/test_uart.rs
```diff
@@ -0,0 +1,24 @@
+// Licensed under the Apache-2.0 license
+
+//! A very simple ROM. Prints a serious of strings using the UART driver.
```
mhatrevi, line 3: NIT: series of strings

mhatrevi (approved):

drivers/test-fw/src/bin/test_uart.rs
korran: Will address in a future PR.

hw-model/src/lib.rs
```diff
@@ -975,7 +979,7 @@ mod tests {
             ..Default::default()
         })
         .unwrap();
-        model.step_until_output("hi").unwrap();
+        model.step_until_output("hii").unwrap();
```
fdamato, line 982: NP: why did we add an extra "i" ? :)

fdamato (approved):
approved for the hw model

hw-model/src/lib.rs
korran: Because sending the same character twice is now a special case (if the toggle bit doesn't change, the second character will be missed), and I wanted to just add a dummy check that it still works in this test.

PR #785, opened 2023-09-14
mailbox: add EXTEND_PCR command

leongross:
Initial implementation for the [`EXTEND_PCR`](https://github.com/chipsalliance/caliptra-sw/tree/main/runtime#extend_pcr) mailbox command.

### DIFF
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -305,13 +305,15 @@ pub struct ExtendPcrReq {
 pub type ExtendPcrReqErr = ();
 impl ExtendPcrReq {
     pub const DATA_MAX_SIZE: usize = 48;
+}
 
-    pub fn new(
-        hdr: MailboxReqHeader,
-        pcr_idx: u32,
-        data: [u8; 48],
-    ) -> Result<Self, ExtendPcrReqErr> {
-        Ok(ExtendPcrReq { hdr, pcr_idx, data })
+impl Default for ExtendPcrReq {
+    fn default() -> Self {
+        Self {
+            hdr: Default::default(),
+            pcr_idx: u32::default(),
+            data: [0u8; ExtendPcrReq::DATA_MAX_SIZE],
+        }
     }
 }
 // No command-specific output args
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -302,10 +302,26 @@ pub struct ExtendPcrReq {
     pub pcr_idx: u32,
     pub data_size: u32,
     pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
 }
 // No command-specific output args
 impl ExtendPcrReq {
-    pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE;
+    pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
+
+    pub fn new(
+        hdr: MailboxReqHeader,
+        pcr_idx: u32,
+        data_size: u32,
+        data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    ) -> Self {
+        ExtendPcrReq {
+            hdr,
+            pcr_idx,
+            data_size,
+            data,
+            _pad: [0u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
+        }
+    }
 }
 
 #[repr(C)]
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -301,19 +301,17 @@ pub struct ExtendPcrReq {
     pub hdr: MailboxReqHeader,
     pub pcr_idx: u32,
     pub data_size: u32,
-    pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
-    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4) % 4],
+    pub data: [u8; 48],
 }
 pub type ExtendPcrReqErr = ();
 impl ExtendPcrReq {
-    // pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
-    pub const DATA_MAX_SIZE: usize = 192;
+    pub const DATA_MAX_SIZE: usize = 48;
 
     pub fn new(
         hdr: MailboxReqHeader,
         pcr_idx: u32,
         data_size: u32,
-        data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+        data: [u8; 48],
     ) -> Result<Self, ExtendPcrReqErr> {
         let data_size = match data_size {
             0 | 255 => {
@@ -327,7 +325,6 @@ impl ExtendPcrReq {
             pcr_idx,
             data_size,
             data,
-            _pad: [0u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
         })
     }
 }
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -302,9 +302,9 @@ pub struct ExtendPcrReq {
     pub pcr_idx: u32,
     pub data_size: u32,
     pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
-    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4) % 4],
 }
-// No command-specific output args
+pub type ExtendPcrReqErr = ();
 impl ExtendPcrReq {
     pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
 
@@ -313,16 +313,24 @@ impl ExtendPcrReq {
         pcr_idx: u32,
         data_size: u32,
         data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
-    ) -> Self {
-        ExtendPcrReq {
+    ) -> Result<Self, ExtendPcrReqErr> {
+        let data_size = match data_size {
+            0 | 255 => {
+                return Err(());
+            }
+            ds => ds,
+        };
+
+        Ok(ExtendPcrReq {
             hdr,
             pcr_idx,
             data_size,
             data,
             _pad: [0u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
-        }
+        })
     }
 }
+// No command-specific output args
 
 #[repr(C)]
 #[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -1,6 +1,6 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_drivers::{CaliptraError, CaliptraResult};
+use caliptra_drivers::{sha384, CaliptraError, CaliptraResult};
 use core::mem::size_of;
 use zerocopy::{AsBytes, FromBytes, LayoutVerified};
 
@@ -305,7 +305,7 @@ pub struct ExtendPcrReq {
 }
 // No command-specific output args
 impl ExtendPcrReq {
-    pub const DATA_MAX_SIZE: usize = 80; // Req max size = (SHA384_BLOCK_BYTE_SIZE - SHA384_HASH_SIZE - 1)
+    pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE;
 }
 
 #[repr(C)]
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -54,7 +54,6 @@ impl From<CommandId> for u32 {
 #[cfg_attr(test, derive(PartialEq, Debug, Eq))]
 #[allow(clippy::large_enum_variant)]
 pub enum MailboxResp {
-    // ExtendPCR(ExtendPCRResp), // ExtendPCR does not return a response
     Header(MailboxRespHeader),
     GetIdevCert(GetIdevCertResp),
     GetIdevCsr(GetIdevCsrResp),
--- a/common/src/mailbox_api.rs
+++ b/common/src/mailbox_api.rs
@@ -17,6 +17,7 @@ impl CommandId {
     pub const INVOKE_DPE: Self = Self(0x44504543); // "DPEC"
     pub const DISABLE_ATTESTATION: Self = Self(0x4453424C); // "DSBL"
     pub const FW_INFO: Self = Self(0x494E464F); // "INFO"
+    pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
 
     // TODO: Remove this and merge with GET_LDEV_CERT once that is implemented
     pub const TEST_ONLY_GET_LDEV_CERT: Self = Self(0x4345524c); // "CERL"
@@ -53,6 +54,7 @@ impl From<CommandId> for u32 {
 #[cfg_attr(test, derive(PartialEq, Debug, Eq))]
 #[allow(clippy::large_enum_variant)]
 pub enum MailboxResp {
+    // ExtendPCR(ExtendPCRResp), // ExtendPCR does not return a response
     Header(MailboxRespHeader),
     GetIdevCert(GetIdevCertResp),
     GetIdevCsr(GetIdevCsrResp),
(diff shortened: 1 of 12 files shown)


common/src/mailbox_api.rs
```diff
@@ -276,6 +277,29 @@ impl Default for InvokeDpeReq {
     }
 }
 
+// EXTEND_PCR
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub value: [u8; ExtendPcrReq::DATA_MAX_SIZE], // variable length
```
bluegate010, line 286: It doesn't seem like this is variable-length? `ExtendPcrReq::read_from()` will fail if the caller doesn't pass in 512 bytes of data.

bluegate010 (commented):
Can we add test coverage for this?

common/src/mailbox_api.rs
jhand2: I think we may need a size field here to tell RT how much of `value` to interpret. Also same for StashMeasurementReq

leongross:
Also another topic to discuss: in the implementation of the [sha384](https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/sha384.rs#L142) there is a check that I quite don't understand:


```rust
pub fn pcr_extend(&mut self, id: PcrId, data: &[u8]) -> CaliptraResult<()> {
    let total_bytes = data.len() + SHA384_HASH_SIZE;
        if total_bytes > (SHA384_BLOCK_BYTE_SIZE - 1) {
            return Err(CaliptraError::DRIVER_SHA384_MAX_DATA_ERR);
        }
```

Why is it problematic, that `data.len() + 48 > 127`? Shouldn't this code check if it's `> 128` which is the maximum SHA384 block size `SHA384_BLOCK_BYTE_SIZE`?
This check implies, that the max size for `data` is `128 - 1 - 48 = 79`, is this intended, or am I maybe misunderstanding something?

bluegate010:
It's not clear from reading the [HW specification](https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/Caliptra_Hardware_Specification.pdf) whether multiple blocks would be supported, worth trying.

runtime/src/pcr.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
+        let cmd =
+            ExtendPcrReq::read_from(cmd_args).ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
+
+        let idx =
+            u8::try_from(cmd.pcr_idx).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        let pcr_index: PcrId =
+            PcrId::try_from(idx).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        drivers.pcr_bank.extend_pcr(
```
bluegate010, line 19: Will also want to append an entry to the PCR log.

runtime/src/pcr.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct ExtendPcrCmd;
+impl ExtendPcrCmd {
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
+        let cmd =
+            ExtendPcrReq::read_from(cmd_args).ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
+
+        let idx =
+            u8::try_from(cmd.pcr_idx).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        let pcr_index: PcrId =
+            PcrId::try_from(idx).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
```
bluegate010, line 17: May want a more descriptive error code. And tests for the error cases.

runtime/src/pcr.rs
bluegate010, line 19: There's not a _strong_ reason, but it seems like a good idea to reserve PCRs 0-3 for Caliptra, and not allow external callers to extend those. But I'm open to leaving it alone.

runtime/tests/integration_tests.rs
```diff
@@ -827,3 +827,31 @@ fn test_idev_id_cert() {
...
+    );
+
+    let cmd = ExtendPcrReq::new(
+        MailboxReqHeader { chksum: checksum },
+        cmd.pcr_idx,
+        cmd.data_size,
+        cmd.data,
+    );
+
+    let res = model.mailbox_execute(u32::from(CommandId::EXTEND_PCR), &cmd.as_bytes());
```
bluegate010, line 855: Can we also test other edge cases, like len=0 or 255, bad PCR index, etc.?

common/src/mailbox_api.rs
```diff
@@ -276,6 +277,36 @@ impl Default for InvokeDpeReq {
...
 
+// EXTEND_PCR
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data_size: u32,
+    pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
```
bluegate010, line 288: If DATA_MAX_SIZE % 4 == 0, _pad should be zero length. Wrap in an outer %?

common/src/mailbox_api.rs
```diff
@@ -276,6 +277,36 @@ impl Default for InvokeDpeReq {
...
+}
+// No command-specific output args
+impl ExtendPcrReq {
+    pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
```
bluegate010, line 292: It seems like this leaks HW details that would be nicer to hide. Is there a need to allow the data size to exceed 48 bytes? Alternatively, could we try rejiggering the PCR driver to handle multiple blocks? It's not clear to me that the current limitation is HW-based.

runtime/tests/integration_tests.rs
bluegate010: Should also test that the data we meant to extend got extended.

common/src/mailbox_api.rs
leongross: I think there is a need for the `data_size` attribute to exceed 48 bytes, since `data` is used as an input for the method `PcrBank::extend_pcr`, which can process more than 48 bytes at once in the internal hash digest. To me it looks like this function can take up to `128 - 1 - 48 = 79` bytes as payload (as we already discussed earlier) or am I confusing things here?

leongross: So you propose, that we make the `data_size` and `data` in `ExtendPcrReq` entirely variable (wihtout nay `DATA_MAX_SIZE`) and then handle multiple blocks in the `PcrBank::extend_pcr` method?

bluegate010: More or less. I think we still need some maximum size, but 79 bytes seems fairly arbitrary. Either 48 or 96 or 192 bytes seems a bit cleaner from an API perspective.

leongross: I see you point, I adjusted the command in such a way, that it should now be possible to pass a variable amount of data, currently limited to 192 bytes as you proposed.

leongross: Any thoughts on this? @bluegate010

common/src/mailbox_api.rs
```diff
@@ -293,6 +294,45 @@ impl Default for InvokeDpeReq {
...
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data_size: u32,
+    pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4) % 4],
+}
+pub type ExtendPcrReqErr = ();
+impl ExtendPcrReq {
+    // pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
```
bluegate010, line 309: Let's nix commented-out code.

common/src/mailbox_api.rs
bluegate010: Sorry for the delay! Will resolve this and add further comments inline.

common/src/mailbox_api.rs
```diff
@@ -293,6 +294,45 @@ impl Default for InvokeDpeReq {
...
 
+// EXTEND_PCR
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data_size: u32,
+    pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4) % 4],
```
ArthurHeymans, line 305: I think extending PCR is typically only done with a digest which has the size of the PCR reg itself?

drivers/src/pcr_bank.rs
```diff
@@ -241,8 +242,43 @@ impl PcrBank {
...
+    ) -> CaliptraResult<()> {
+        let data_block_size = (SHA384_BLOCK_BYTE_SIZE - SHA384_HASH_SIZE - 1) as u32;
+        let blocks: u32 = data.len() as u32 / data_block_size as u32;
+
+        for i in 0..blocks {
+            let d = data
+                .get((i * data_block_size) as usize..((i + 1) * data_block_size) as usize)
+                .ok_or(CaliptraError::DRIVER_PCR_BANK_EXTEND_INVALID_SIZE)?;
+
+            sha.pcr_extend(id, d)?;
```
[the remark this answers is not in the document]
leongross: Okay, let's do that.

runtime/src/pcr.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        drivers.pcr_bank.extend_pcr_variable(
+            pcr_index,
+            &mut drivers.sha384,
+            &cmd.data
+                .get(..cmd.data_size as usize)
+                .ok_or(CaliptraError::DRIVER_PCR_BANK_EXTEND_INVALID_SIZE)?,
+        )?;
+
+        // 2. Add log entry
```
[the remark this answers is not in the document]
leongross: So no logging at all or should we find another location to log this to?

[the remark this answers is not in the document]
bluegate010: I'm thinking no logging at all.

common/src/mailbox_api.rs
```diff
@@ -293,6 +294,44 @@ impl Default for InvokeDpeReq {
...
```
ArthurHeymans, line 305: only pub data: [u8; 48] is needed. No need for a size nor padding.

runtime/tests/integration_tests.rs
```diff
@@ -895,3 +895,82 @@ fn test_idev_id_cert() {
...
+            cmd.data,
+        )
+    }
+
+    let mut model = run_rt_test(None, None);
+    let payload_data = [0u8; ExtendPcrReq::DATA_MAX_SIZE];
+
+    let cmd = generate_mailbox_extend_pcr_req(4, payload_data.len(), payload_data).unwrap();
+    let res = model.mailbox_execute(u32::from(CommandId::EXTEND_PCR), &cmd.as_bytes());
+    assert!(res.is_ok());
```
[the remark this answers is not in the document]
leongross: As I see it this could only be done using the `ureg::Mmio`, which atm is not implemented in the emulator yet, right?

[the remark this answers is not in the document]
leongross: @bluegate010 any thoughts on this?

common/src/mailbox_api.rs
ArthurHeymans: Drop data_size too.

runtime/tests/integration_tests.rs
```diff
@@ -895,3 +898,108 @@ fn test_idev_id_cert() {
...
+        182, 215, 239, 9, 244, 123, 192, 213, 72, 247, 223, 97, 50, 203, 98, 85, 255, 185, 136,
+        117, 132, 45, 47, 144, 171, 30, 11, 237, 104, 239, 127, 74, 45, 22, 202, 113, 89, 231, 46,
+        208, 185, 237, 115, 123, 223, 231, 246, 254, 103, 112, 39, 33, 159, 23, 201, 17, 64, 144,
+        24, 179, 34, 251, 43, 5, 215, 169, 71, 118, 156, 103, 123, 101, 170, 5, 38, 190, 72, 75,
+        81, 223, 72, 131, 215, 100, 50, 150, 57, 34, 255, 172, 39, 211, 121, 252, 89, 24, 57, 15,
+        255, 175, 12, 201, 179, 23, 247, 117, 169, 254, 227, 206, 220, 210, 169, 217, 42, 172, 102,
+        146, 234, 16, 51, 40, 9, 37, 177, 251, 80, 134, 0, 113, 8, 244, 140, 23, 133, 119, 94, 216,
+        38, 247, 184, 174, 36, 0, 78, 151, 201, 134, 182, 124, 122, 105, 75,
+    ];
+
```
ArthurHeymans, line 951: I don't think this is the mailbox API is intended. You want to pass a digest of data, not the data itself. The mailbox itself is limited in size and PCRs often get extended with digests of code much larger than the mailbox itself.

runtime/tests/integration_tests.rs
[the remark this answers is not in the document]
jhand2: I think he just means we should populate `payload_data` with non-zero bytes.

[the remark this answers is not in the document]
leongross: @jhand2 my ping was regarding the implementation status of `ureg::MMio`

drivers/src/sha384.rs
```diff
@@ -20,10 +20,10 @@ use crate::{array::Array4x32, wait, Array4x12};
 use caliptra_error::{CaliptraError, CaliptraResult};
 use caliptra_registers::sha512::Sha512Reg;
 
-const SHA384_BLOCK_BYTE_SIZE: usize = 128;
+pub const SHA384_BLOCK_BYTE_SIZE: usize = 128;
```
jhand2, line 23: Don't think this needs to be public.

drivers/src/sha384.rs
```diff
@@ -20,10 +20,10 @@ use crate::{array::Array4x32, wait, Array4x12};
...
 const SHA384_BLOCK_LEN_OFFSET: usize = 112;
 const SHA384_MAX_DATA_SIZE: usize = 1024 * 1024;
-const SHA384_HASH_SIZE: usize = 48;
+pub const SHA384_HASH_SIZE: usize = 48;
```
jhand2, line 26: Same. This isn't being used outside this module.

drivers/src/lib.rs
```diff
@@ -42,7 +42,7 @@ mod persistent;
 pub mod printer;
 mod sha1;
 mod sha256;
-mod sha384;
+pub mod sha384;
```
jhand2, line 45: Don't think this is needed

drivers/test-fw/src/bin/pcrbank_tests.rs
```diff
@@ -84,9 +84,12 @@ fn test_write_protection_stickiness() {
     }
 }
 
+fn test_mailbox_pcrextend_cmd() {}
```
jhand2, line 87: Assume this will be filled out before merging?

common/src/mailbox_api.rs
```diff
@@ -293,6 +294,28 @@ impl Default for InvokeDpeReq {
...
+    pub const DATA_MAX_SIZE: usize = 48;
+
+    pub fn new(
+        hdr: MailboxReqHeader,
+        pcr_idx: u32,
+        data: [u8; 48],
+    ) -> Result<Self, ExtendPcrReqErr> {
+        Ok(ExtendPcrReq { hdr, pcr_idx, data })
+    }
+}
```
ArthurHeymans, line 316: Why have a function for this? It's just filling in a struct with 3 entries with 3 arguments, which is likely inefficient as arguments get copied in functions.

runtime/tests/integration_tests.rs
```diff
@@ -895,3 +899,129 @@ fn test_idev_id_cert() {
...
+    // How to read PCR regs from model?
+    // get pv = PvReg sha512pv.pcr_ctrl().at(id.into()).?
+
+    // prob. similar to drivers/test-fw/src/bin/sha384_tests.rs
+    // fn test_pcr_hash_extend_single_block_3() {
+    //     let mut sha384 = unsafe { Sha384::new(Sha512Reg::new()) };
+    //     let mut pcr_bank = unsafe { PcrBank::new(PvReg::new()) };
+
+    // #[peripheral(offset = 0x1001_8000, mask = 0x0000_7fff)]
+    // pub key_vault: KeyVault,
```
ArthurHeymans, line 992: Get rid of this commented out code?

runtime/tests/integration_tests.rs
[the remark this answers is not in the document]
jhand2: I don't think it's necessary to use ureg. Just make `extension_data` non-zero

common/src/mailbox_api.rs
leongross: This is an artifact from where the default constructor was not suitable, I will make it compact again.

runtime/tests/integration_tests.rs
```diff
@@ -895,3 +895,96 @@ fn test_idev_id_cert() {
...
+    // for i in 0..32 {
+    //     let e = entry.at(i as usize);
+    //     for ii in 0..12 {
+    //         //         // this fails, but why?
+    //         let ee = e.at(ii).read();
+    //     }
+    // }
+
+    let pcr_4 = pcr.pcr_ctrl().at(4);
+    let r = pcr_4.read();
```
ArthurHeymans, line 956: Are these used?

leongross:
Can we re-open this PR?

jhand2:
@leongross @ArthurHeymans told me he was taking over the PCR work and would open new PRs. Github isn't letting me reopen this PR ("The mailbox_extend_pcr was force-pushed or recreated"). Could you talk with Arthur and one of you open a new PR with these changes?

Sorry about the confusion.

Issue #786, opened 2023-09-14
Verilator test failure: test_fmcalias_derivation::test_pcr_log_across_update_reset


korran:
```
      6,338 <<< Executing mbox cmd 0x46574c44 (60612 bytes) from SoC
test test_pcr_log_across_update_reset has been running for over 60 seconds
    195,164 UART: 
    195,202 UART: Running Caliptra ROM ...
    196,178 UART: 
    205,444 UART: [state] CFI Enabled
    207,152 UART: [state] LifecycleState = Unprovisioned
    208,741 UART: [state] DebugLocked = No
    210,137 UART: [state] Watchdog Timer is not started because the device is not locked for debugging
    357,161 UART: ROM Digest: 11BC274A06B0086E7A15BE8553A83D8F7643D7BD1E82B02C05166D41C5E3A896
    367,411 UART: [kat] ++
    367,791 UART: [kat] sha1
    377,057 UART: [kat] SHA2-256
    379,713 UART: [kat] SHA2-384
    383,410 UART: [kat] SHA2-384-ACC
    385,449 UART: [kat] ECC-384
  3,954,255 UART: [kat] HMAC-384
  3,963,914 UART: [kat] LMS
  5,034,994 UART: [kat] --
  5,038,077 UART: [cold-reset] ++
  5,039,798 UART: [idev] ++
  5,040,218 UART: [idev] CDI.KEYID = 6
  5,041,094 UART: [idev] SUBJECT.KEYID = 7
  5,042,134 UART: [idev] UDS.KEYID = 0
  5,054,872 UART: [idev] Erasing UDS.KEYID = 0
  8,317,846 UART: [idev] Using Sha1 for KeyId Algorithm
  8,338,950 UART: [idev] --
  8,342,279 UART: [ldev] ++
  8,342,685 UART: [ldev] CDI.KEYID = 6
  8,343,577 UART: [ldev] SUBJECT.KEYID = 5
  8,344,619 UART: [ldev] AUTHORITY.KEYID = 7
  8,345,737 UART: [ldev] FE.KEYID = 1
  8,356,139 UART: [ldev] Erasing FE.KEYID = 1
 11,636,329 UART: [ldev] Signing Cert with AUTHORITY.KEYID = 7
 15,226,878 UART: [ldev] PUB.X = 504D38CA45D997901F48BA333A149A2FB2668B973AAC64D3B79ECC09A663F02ED2FABA133F5FA499677AEB7687C99B25
[...]

PR #792, opened 2023-09-14
Make test-fmc use FIRMWARE_LOAD to trigger the update reset.

korran:
This is necessary to work around the race condition in #786.


bluegate010:
Will probably also want to make the same change [here](https://github.com/chipsalliance/caliptra-sw/blob/b1f43f36f8a5e2d262b04c702b324baf8addd33c/fmc/tests/test_rtalias.rs#L179).

korran:
(Waiting for verilator tests to run to confirm that this fully fixes #786)

korran:
Confirmed that the tests passed under verilator.

fdamato (approved):
Approved for the HW Model Changes

PR #794, opened 2023-09-14
Add checksum verify to ROM command processing

nquarton:


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -570,6 +588,43 @@ impl FirmwareProcessor {
...
+    ///
+    /// # Arguments
+    /// * `txn` - Mailbox Receive Transaction
+    /// * `data` - Data buffer for the expected request
+    ///
+    /// # Returns
+    /// * `()` - Ok
+    ///    Error code on failure.
+    fn copy_req_verify_chksum(txn: &mut MailboxRecvTxn, data: &mut [u8]) -> CaliptraResult<()> {
+        // TODO: Some commands may have a variable size with a max instead of a fixed size
```
mhatrevi, line 601: Is there an additional change needed to address this TODO?

mhatrevi (changes requested):

rom/dev/src/flow/cold_reset/fw_processor.rs
nquarton: Good question. None of the commands we support in ROM currently have a variable request size, so it's not needed at the moment. I can change the wording on this so it's clearer this is sufficient for the current command set.

nquarton:
I don't know if I have the right failure behavior here. I am working on adding a test for the invalid checksum case. Should I be propagating it up from the process_mailbox_commands function like it is now? Or do we need to set a non-fatal error and continue? What about other processing failures?

nquarton:
Runtime calls a handler function from within the processing loop. In that method, failures are caught, reported as non-fatal errors, the mailbox status is set to failure, and the loop continues. Do we need to implement that? Or is this behavior of stopping execution what is wanted?

nquarton:
The last issue is with this current implementation is there is also a race condition where drop() for MailboxRecvTxn will complete the mailbox transaction (as a failure) before the error code is written by rom main.rs

mhatrevi (approved):

korran (approved):

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -559,6 +576,44 @@ impl FirmwareProcessor {
...
+    /// * `data` - Data buffer for the expected request
+    ///
+    /// # Returns
+    /// * `()` - Ok
+    ///    Error code on failure.
+    fn copy_req_verify_chksum(txn: &mut MailboxRecvTxn, data: &mut [u8]) -> CaliptraResult<()> {
+        // NOTE: Currently ROM only supports commands with a fixed request size
+        //       This check will need to be updated if any commands are added with a variable request size
+        if txn.dlen() as usize != data.len() {
+            return Err(CaliptraError::FW_PROC_MAILBOX_INVALID_REQUEST_LENGTH);
```
korran, line 592: Can we add a test case that checks for this error too?

nquarton: This is a good point, I can add that. Would it be alight to merge without it and I will add this week?

nquarton:
I will add the test suggested by Kor separately since I don't have time to address today: https://github.com/chipsalliance/caliptra-sw/issues/818

rom/dev/src/flow/cold_reset/fw_processor.rs
nquarton: Issue filed https://github.com/chipsalliance/caliptra-sw/issues/818

PR #803, opened 2023-09-15
Remove the need for PCR log verifier to swap endianness

bluegate010:


nquarton:
I agree with commenting out that check related to PCR in the fake rom boot test for now. You could make the argument it's not even needed there since the input to the certs doesn't match the collateral anyway. I'll either re-generate the certs so it passes or remove that check in the near future.

mhatrevi (approved):

test/src/derive.rs
```diff
@@ -328,15 +328,9 @@ impl Pcr0 {
...
-            &mut value,
-            swap_word_bytes(&input.vendor_pub_key_hash).as_bytes(),
-        );
-        extend(
-            &mut value,
-            swap_word_bytes(&input.owner_pub_key_hash).as_bytes(),
-        );
-        extend(&mut value, swap_word_bytes(&input.fmc_digest).as_bytes());
+        extend(&mut value, input.vendor_pub_key_hash.as_bytes());
+        extend(&mut value, input.owner_pub_key_hash.as_bytes());
```
korran, line 332: We don't want to do this: input.owner_pub_key_hash is the words (in the weird hardware format); by calling as_bytes() we're actually reversing it from the regular sha384 serialization.

This could me made clearer in this test by making input.owner_pub_key_hash a [u8; 48] with the regular serialization.

korran: Basically, we never want to convert from [u32; 12] to [u8; 48] with zerocopy .as_bytes()....

bluegate010: Ack, will close this PR

PR #805, opened 2023-09-15
Update bitflags 2.0.1 -> 2.4.0.

korran:
NOTE: This is actually a no-op, since the "newer" compatible version in dpe was overriding this old version. By using the same version as caliptra-dpe, we can avoid some confusion about which version of bitflags we're actually compiling against.

Also, unlike 2.0.1, 2.4.0 has been vetted by Google: https://github.com/google/rust-crate-audits/blob/main/audits.toml


FerralCoder (approved):

FerralCoder:
still relevant? Can we get this merged or closed?

korran:
> still relevant? Can we get this merged or closed?

This is still relevant, should be a no-op. Will merge as long as the ROM binary doesn't change.

PR #813, opened 2023-09-18
[fix] On returning generic error, store actual error in extended error info

### LINKED ISSUE #335: If ECC or LMS validation fails, the exact reason for the failure is lost
https://github.com/chipsalliance/caliptra-sw/blob/fe6cd119d29e6e1159a6e75a53a28b3b9c6f3f23/image/verify/src/verifier.rs#L316-L319

At the very least, the original failure reason should be saved somewhere for debugging purposes.

mhatrevi:
Addresses https://github.com/chipsalliance/caliptra-sw/issues/335


korran (commented):
We should add assertions for this extended error info to the existing test cases.

mhatrevi:
> We should add assertions for this extended error info to the existing test cases.

Discussed of offline; crypto function failures are not tested by integration tests.

korran (approved):

PR #819, opened 2023-09-19
Ensure that only PL0 is able to call certify key for X509s

sree-revoori1:


runtime/src/invoke_dpe.rs
```diff
@@ -23,21 +27,44 @@ impl InvokeDpeCmd {
             let hashed_rt_pub_key = crypto
                 .hash(AlgLen::Bit384, &rt_pub_key.to_der()[1..])
                 .map_err(|_| CaliptraError::RUNTIME_INITIALIZE_DPE_FAILED)?;
+            let image_header = pdata.manifest1.header;
```
jhand2, line 30: ImageHeader implements copy, so there's a chance this would be copied unnecessary. We should either

* Reference members as needed
* Make this local variable a borrow (`let image_header = &pdata.manifest.header`)

runtime/src/invoke_dpe.rs
```diff
@@ -23,21 +27,44 @@ impl InvokeDpeCmd {
...
+            let resp = match command {
+                Command::GetProfile => Ok(Response::GetProfile(
+                    dpe.get_profile(&mut env.platform)
+                        .map_err(|_| CaliptraError::RUNTIME_INVOKE_DPE_FAILED)?,
+                )),
+                Command::InitCtx(cmd) => cmd.execute(dpe, &mut env, locality),
+                Command::DeriveChild(cmd) => cmd.execute(dpe, &mut env, locality),
+                Command::CertifyKey(cmd) => {
+                    // PL1 cannot request X509
+                    if cmd.format == CertifyKeyCmd::FORMAT_X509 && image_header.flags & 1 == 0 {
```
jhand2, line 53: I think this misses checking whether locality == pl0_pauser. I also think checking flags directly might be a bit confusing. Maybe a helper like this would be good:

```rust
fn is_caller_pl0(drivers: &mut Drivers, caller_locality: u32) -> bool
```

sree-revoori1: Done

runtime/src/invoke_dpe.rs
sree-revoori1: Done.

runtime/src/invoke_dpe.rs
```diff
@@ -56,4 +83,8 @@ impl InvokeDpeCmd {
             Err(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)
         }
     }
+
+    fn is_caller_p1(pl0_pauser: u32, flags: u32, locality: u32) -> bool {
+        flags & 1 == 0 && locality != pl0_pauser
```
jhand2, line 88: nit: Can you defined a const for this? `const PL0_PAUSER_FLAG: u32 = 1`. Alternatively you could use bitflags, although that might be a little overkill since there's only one flag.

sree-revoori1: Done

jhand2 (approved):

PR #820, opened 2023-09-19
Validate DPE Instance in SRAM

sree-revoori1:
In RT-fw, if we are not in a cold reset, we are expected to use the DPE instance in SRAM. Before we use it, we need to validate it to ensure that the DPE context tree is well-formed (the array representing the contexts is actually a tree) and verify that the current TCI measurement of the root TCI node matches the latest runtime PCR value. Otherwise, we disable attestation.

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -6,13 +6,14 @@
 pub use crate::fips::{fips_self_test_cmd, fips_self_test_cmd::SelfTestStatus};
 
 use crate::{
-    dice, CptraDpeTypes, DpeCrypto, DpePlatform, Mailbox, DPE_SUPPORT, MAX_CERT_CHAIN_SIZE,
+    dice, CptraDpeTypes, DisableAttestationCmd, DpeCrypto, DpePlatform, Mailbox, DPE_SUPPORT,
+    MAX_CERT_CHAIN_SIZE,
 };
 
 use arrayvec::ArrayVec;
 use caliptra_drivers::{
-    CaliptraError, CaliptraResult, DataVault, Ecc384, KeyVault, Lms, PersistentDataAccessor,
-    ResetReason, Sha1, SocIfc,
+    cprintln, Array4x12, CaliptraError, CaliptraResult, DataVault, Ecc384, KeyVault, Lms,
+    PersistentDataAccessor, ResetReason, Sha1, SocIfc,
 };
 use caliptra_drivers::{Hmac384, PcrBank, PcrId, Sha256, Sha384, Sha384Acc, Trng};
 use caliptra_registers::mbox::enums::MboxStatusE;
@@ -21,6 +22,8 @@ use caliptra_registers::{
     mbox::MboxCsr, pv::PvReg, sha256::Sha256Reg, sha512::Sha512Reg, sha512_acc::Sha512AccCsr,
     soc_ifc::SocIfcReg, soc_ifc_trng::SocIfcTrngReg,
 };
+use dpe::context::{Context, ContextState};
+use dpe::tci::TciMeasurement;
 use dpe::{
     commands::{CommandExecution, DeriveChildCmd, DeriveChildFlags},
     context::ContextHandle,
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
bluegate010, line 1: It seems like `new_from_registers` has a good amount of code which is actually safe. Could the unsafe code be concentrated into a helper?

runtime/src/lib.rs
```diff
@@ -198,7 +199,56 @@ impl Drivers {
...
+
+        if reset_reason != ResetReason::ColdReset {
+            let dpe = &drivers.persistent_data.get().dpe;
+            let tci_type = u32::from_be_bytes(*b"RTJM");
+            let root_idx = dpe
+                .get_root_idx(tci_type, locality)
+                .map_err(|_| CaliptraError::RUNTIME_DPE_VALIDATION_FAILED)?;
+
+            // Ensure context/TCI tree is well-formed (all nodes reachable from root and no cycles)
+            if dpe.validate_context_tree(root_idx) {
```
bluegate010, line 212: There's a lot of code underneath this `if` block. Could the condition be inverted and we just bail early on failure?

runtime/src/lib.rs
sree-revoori1: Done.

runtime/src/lib.rs

runtime/src/drivers.rs
```diff
@@ -114,31 +105,110 @@ impl Drivers {
...
-                platform: DpePlatform::new(locality, hashed_rt_pub_key, &mut cert_chain),
+                platform: DpePlatform::new(locality, hashed_rt_pub_key, &mut drivers.cert_chain),
             };
-            let dpe = Self::initialize_dpe(env, &mut pcr_bank, locality)?;
-            persistent_data.get_mut().dpe = dpe;
+            let dpe = Self::initialize_dpe(env, &mut drivers.pcr_bank, locality)?;
+            drivers.persistent_data.get_mut().dpe = dpe;
         }
 
+        if reset_reason != ResetReason::ColdReset {
```
jhand2, line 114: nit: I think it might be easier to follow what's happening here with a bit of a refactor. Roughly

```rust
match reset_reason {
  ResetReason::ColdReset => {
    Self::initialize_dpe()?;
  },
  ResetReason::UpdateReset => {
    Self::check_dpe_structure()?;
    Self::update_dpe_rt_journey()?;
  },
  ResetReason::WarmReset => {
    Self::check_dpe_structure()?;
    Self::check_dpe_rt_journey_unchanged()?;
  }
}
```

runtime/src/drivers.rs
```diff
@@ -114,31 +105,110 @@ impl Drivers {
...
+            let dpe = &drivers.persistent_data.get().dpe;
+            // Find root node by finding the context with parent equal to ROOT_INDEX
+            let root_idx = dpe
```
jhand2, line 117: On UpdateReset I think we also need to update the TCI at the root to match the new PCR3 value.

runtime/src/drivers.rs
jhand2: Except I suppose for warm and update reset, instead of early return we want to call `disable_attestation` and keep booting.

runtime/src/drivers.rs
```diff
@@ -114,31 +105,110 @@ impl Drivers {
...
+            if root_idx >= dpe.contexts.len() {
+                return Err(CaliptraError::RUNTIME_DPE_VALIDATION_FAILED);
+            }
+
+            // Ensure context/TCI tree is well-formed (all nodes reachable from root and no cycles)
+            if !dpe.validate_context_tree(root_idx) {
+                // If SRAM Dpe Instance validation fails, disable attestation
+                let mut result = DisableAttestationCmd::execute(&mut drivers);
+                if result.is_ok() {
+                    cprintln!("Disabled attestation due to DPE validation failure");
```
jhand2, line 136: nit: In the print statements for these failures should we also print the error code?

runtime/src/drivers.rs
```diff
@@ -114,31 +105,110 @@ impl Drivers {
...
+                    drivers.mbox.set_status(MboxStatusE::DataReady);
```
jhand2, line 137: I think we want to avoid doing anything with the mailbox when calling disable attestation, since it wasn't invoked from the mailbox. This may require refactoring that command a little.

runtime/src/drivers.rs
```diff
@@ -78,34 +80,23 @@ impl Drivers {
...
-            EntropySrcReg::new(),
-            SocIfcTrngReg::new(),
-            &SocIfcReg::new(),
-        )?;
-
-        let mut sha384 = Sha384::new(Sha512Reg::new());
-        let mut ecc384 = Ecc384::new(EccReg::new());
-        let mut hmac384 = Hmac384::new(HmacReg::new());
-        let mut key_vault = KeyVault::new(KvReg::new());
+    pub fn new_from_registers() -> CaliptraResult<Self> {
```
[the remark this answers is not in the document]
sree-revoori1: Aren't these a bit orthogonal since unsafe doesn't necessarily imply that the function can't be called more than once? We already have the comment on new_from_registers telling us about safety.

runtime/src/drivers.rs
sree-revoori1: Done

runtime/src/drivers.rs
sree-revoori1: Done

runtime/src/drivers.rs
sree-revoori1: Done

runtime/src/drivers.rs
sree-revoori1: Done

runtime/src/drivers.rs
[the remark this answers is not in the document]
jhand2: They're technically orthogonal, it's just that calling an unsafe function requires you to write extra code to use it, so the idea is that people will be more careful with functions marked unsafe.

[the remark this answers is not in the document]
sree-revoori1: Done

jhand2 (approved):

PR #822, opened 2023-09-20
Add mailbox command parameter validity checking

### LINKED ISSUE #787: MB Command Parameter Checking for Run-Time
All Input Data must have parameter checking.  This mostly affects our MB implementation.

For example, if a size value is passed in with the MB command, we should verify that size value against max values and other bounds (like message ptr increments) to ensure that attackers can't attack via MB input.

sree-revoori1:
For invoke DPE, we check that data_size <= MAX_DATA_SIZE.

Fixes #787

### DIFF
--- a/runtime/src/invoke_dpe.rs
+++ b/runtime/src/invoke_dpe.rs
@@ -16,6 +16,11 @@ impl InvokeDpeCmd {
 
     pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
         if let Some(cmd) = InvokeDpeReq::read_from(cmd_args) {
+            // Validate data length
+            if cmd.data_size as usize > cmd.data.len() {
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            }
+
             let pdata = drivers.persistent_data.get();
             let rt_pub_key = pdata.fht.rt_dice_pub_key;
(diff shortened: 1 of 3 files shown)


runtime/src/invoke_dpe.rs
```diff
@@ -10,6 +10,14 @@ pub struct InvokeDpeCmd;
 impl InvokeDpeCmd {
     pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
         if let Some(cmd) = InvokeDpeReq::read_from(cmd_args) {
+            // Validate data length
+            let Ok(in_len) = usize::try_from(cmd.data_size) else {
```
jhand2, line 14: nit: Caliptra firmware won't ever run on a processor that is less than 32-bit, so `cmd.data_size as usize` is safe.

jhand2, line 14: It looks like we actually don't use `cmd.data_size`. I think we need to update the execute call to `.execute_serialized_command(&mut env, drivers.mbox.user(), &cmd.data[..cmd.data_size as usize])`

sree-revoori1: Done

sree-revoori1: Done.

jhand2 (dismissed):

jhand2 (approved):

PR #853, opened 2023-09-21
[fix] When in shutdown state all mailbox transactions are dropped and fatal…

rusty1968:
When in shutdown state all mailbox transactions are dropped and fatal/non-fatal error registers inform the SoC that Caliptra services have been shutdown.

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -1,0 +1,210 @@
+// Licensed under the Apache-2.0 license
+
+#![cfg_attr(not(feature = "fip-self-test"), allow(unused))]
+
+#[cfg(feature = "fips_self_test")]
+pub use crate::fips::{fips_self_test_cmd, fips_self_test_cmd::SelfTestStatus};
+
+use crate::{
+    dice, CptraDpeTypes, DpeCrypto, DpePlatform, Mailbox, DPE_SUPPORT, MAX_CERT_CHAIN_SIZE,
+};
+
+use arrayvec::ArrayVec;
+use caliptra_drivers::{
+    CaliptraError, CaliptraResult, DataVault, Ecc384, KeyVault, Lms, PersistentDataAccessor,
+    ResetReason, Sha1, SocIfc,
+};
+use caliptra_drivers::{Hmac384, PcrBank, PcrId, Sha256, Sha384, Sha384Acc, Trng};
+use caliptra_registers::mbox::enums::MboxStatusE;
+use caliptra_registers::{
+    csrng::CsrngReg, dv::DvReg, ecc::EccReg, entropy_src::EntropySrcReg, hmac::HmacReg, kv::KvReg,
+    mbox::MboxCsr, pv::PvReg, sha256::Sha256Reg, sha512::Sha512Reg, sha512_acc::Sha512AccCsr,
+    soc_ifc::SocIfcReg, soc_ifc_trng::SocIfcTrngReg,
+};
+use dpe::{
+    commands::{CommandExecution, DeriveChildCmd, DeriveChildFlags},
+    context::ContextHandle,
+    dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
+    support::Support,
+    DPE_PROFILE,
+};
+
+use crypto::{AlgLen, Crypto};
+
+pub struct Drivers {
+    pub mbox: Mailbox,
+    pub sha_acc: Sha512AccCsr,
+    pub data_vault: DataVault,
+    pub key_vault: KeyVault,
+    pub soc_ifc: SocIfc,
+    pub sha256: Sha256,
+
+    // SHA2-384 Engine
+    pub sha384: Sha384,
+
+    // SHA2-384 Accelerator
+    pub sha384_acc: Sha384Acc,
+
+    /// Hmac384 Engine
+    pub hmac384: Hmac384,
+
+    /// Cryptographically Secure Random Number Generator
+    pub trng: Trng,
+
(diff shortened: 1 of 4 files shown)


runtime/src/lib.rs
```diff
@@ -376,6 +189,18 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> CaliptraResult<()> {
...
+            cprintln!("[rt] is_shutdown");
+            caliptra_drivers::report_fw_error_non_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
+            caliptra_drivers::report_fw_error_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
+            loop {
+                // SoC firmware might be stuck waiting for Caliptra to finish
+                // executing this pending mailbox transaction. Notify them that
+                // we've failed.
+                unsafe { caliptra_drivers::Mailbox::abort_pending_soc_to_uc_transactions() };
+            }
+        }
```
jhand2, line 202: I think this needs to be inside the `is_cmd_ready` check. Otherwise we'll do this any time we wake up from any interrupt.

runtime/src/lib.rs
```diff
@@ -376,6 +189,18 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> CaliptraResult<()> {
...
+        if drivers.is_shutdown {
+            cprintln!("[rt] is_shutdown");
+            caliptra_drivers::report_fw_error_non_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
+            caliptra_drivers::report_fw_error_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
+            loop {
+                // SoC firmware might be stuck waiting for Caliptra to finish
+                // executing this pending mailbox transaction. Notify them that
+                // we've failed.
+                unsafe { caliptra_drivers::Mailbox::abort_pending_soc_to_uc_transactions() };
+            }
```
jhand2, line 201: Instead of doing this, we should be able to just fail this command and `continue`. On the next mailbox command we'll hit this if condition again and fail that command.

test/tests/smoke_test.rs
```diff
@@ -481,13 +481,10 @@ fn test_fips<T: HwModel>(hw: &mut T, fmc_version: u32, app_version: u32) {
...
 
-    // Check we are rejecting additional commands with the shutdown error code.
-    let expected_err = Err(ModelError::MailboxCmdFailed(0x000E0008));
-    let payload = MailboxReqHeader {
-        chksum: caliptra_common::checksum::calc_checksum(u32::from(CommandId::VERSION), &[]),
-    };
-    let resp = hw.mailbox_execute(u32::from(CommandId::VERSION), payload.as_bytes());
-    assert_eq!(resp, expected_err);
+    // Check we are rejecting additional commands.
+    let _ = hw
```
jhand2, line 485: might be good to check that the error code is the one we expect (and we don't happen to be failing for some other reason)

jhand2: I also think it might be worthwhile to test that several commands fail (like calling this in a loop 3 times).

I could see a regression which causes the next command to fail, but then Caliptra goes back into a normal state.

runtime/src/lib.rs
rusty1968: The semantics of receiving a command in shutdown should be handled the same way as ROM. In ROM we are handling as error.

jhand2: Right, I'm not suggesting changing that behavior. I just think the code could be written in a way that's more clear what's happening. Something like:

```rust
loop {
    enter_idle(drivers);
    if drivers.mbox.is_cmd_ready() {
        if drivers.is_shutdown {
            cprintln!("[rt] is_shutdown");
            caliptra_drivers::report_fw_error_non_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
            caliptra_drivers::report_fw_error_fatal(CaliptraError::RUNTIME_SHUTDOWN.into());
            unsafe { caliptra_drivers::Mailbox::abort_pending_soc_to_uc_transactions() };
            continue;
        }
        ...
    }
}
```

This will ensure every mailbox command is failed, but avoids having another infinite loop. I will also make managing the watchdog easier (especially when it is moved to `enter_idle`) since the normal watchdog behavior will still work.

runtime/src/drivers.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub pcr_bank: PcrBank,
+
+    pub cert_chain: ArrayVec<u8, MAX_CERT_CHAIN_SIZE>,
+
+    pub attestation_disabled: bool,
+
+    #[cfg(feature = "fips_self_test")]
+    pub self_test_status: SelfTestStatus,
+
+    pub is_shutdown: bool,
```
jhand2, line 72: This will be cleared on warm/update reset. Is that expected? Otherwise we should maybe put this in the FHT.

runtime/src/lib.rs
rusty1968: We can just leverage the fatal error flow  - propagate it as an error up to main.

runtime/src/drivers.rs
rusty1968: We don't make the shutdown condition persistent.

test/tests/smoke_test.rs
rusty1968: Done. I am sending different commands now and checking for shutdown error. Note this sequence exercised by rom and rt to make sure we are aligned in behavior.

jhand2 (approved):

Issue #807, opened 2023-09-15
Potential DoS waiting for sha384_acc to be locked


korran:
Consider this loop, used during image validation on both cold-reset and hitless-update:

https://github.com/chipsalliance/caliptra-sw/blob/9e4f4d673b22a5c23f9d9078872a09c3d02853cb/common/src/verifier.rs#L36-L42

If the SoC sends a hitless-update mailbox command, Caliptra will boot back to the ROM, and the SoC could then lock the sha384_acc peripheral, making the Caliptra ROM freeze up in the above loop, and causing the watchdog to eventually fire, bricking caliptra until the next power cycle.

korran:
We might be able to mitigate that one in the runtime firmware by refusing to do the hitless reboot unless we've already locked sha384acc. Need to write a test to ensure the ROM can handle sha384acc already being locked by the runtime.

korran:
Note that with the workaround to #841 added in #858, it is no longer possible to fix this issue without a ROM change.

* If we #858 as is, the runtime firmware can't lock the mailbox before resetting to fix https://github.com/chipsalliance/caliptra-sw/issues/807, as that would cause the AssumedLockState::NotAcquried attempt here to timeout in the watchdog.

* If we tweak the #858 to assume ShaAccLockState::Acquired for UpdateReset, we are trusting the runtime firmware too much, as an attacker with some control over the runtime firmware could do an update reset without the lock, and have the SoC grab the lock while the ROM assumes it already has the lock.

mhatrevi:
Closing this issue as it is no longer applicable due to https://github.com/chipsalliance/caliptra-sw/pull/862

PR #858, opened 2023-09-21
[fix] TOCTOU Condition in SHA-512 Accelerator Lock Acquisition

### LINKED ISSUE #841: TOCTOU Condition in SHA-512 Accelerator Lock Acquisition
The SHA-512 Accelerator was a shared component which could be accessed by both the
Caliptra firmware and the SoC, and could be used to compute SHA-384 (and SHA-512)
digests of shared memory regions, in particular the SRAM used to back the Mailbox.
Simultaneous use was restricted by means of a lock.

By exploiting a race condition within the ROM code used to acquire this lock, the SoC could
claim ownership of the peripheral while the ROM believed it had exclusive access. The SoC
could then cause digests of benign signed code to be computed in place of malicious
unsigned code that had been supplied to the ROM, which would result in the ROM accepting
the malicious code to be validly signed and proceed to execute it.

mhatrevi:
This fix addresses https://github.com/chipsalliance/caliptra-sw/issues/841 On a Cold/Warm Reset, the SHA accelerator is guaranteed to be locked for Caliptra. On a hitless update (aka Update Reset), it is expected that the SOC keeps the SHA Acc. unlocked. Given these guarantees, this change addresses the issue by passing the lock state information to the SHA Acc driver, which then acts on it by either attempting to acquire the SHA acc. lock or assuming it is already locked.

### DIFF
--- a/drivers/src/sha384acc.rs
+++ b/drivers/src/sha384acc.rs
@@ -24,6 +24,13 @@ const MAX_MAILBOX_CAPACITY_BYTES: u32 = 128 << 10;
 
 pub type Sha384Digest<'a> = &'a mut Array4x12;
 
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ShaAccLockState {
+    AssumedLocked = 0xAAAA_AAA5,
+    NotAcquired = 0x5555_555A,
+}
+
 pub struct Sha384Acc {
     sha512_acc: Sha512AccCsr,
 }
@@ -36,22 +43,46 @@ impl Sha384Acc {
     ///
     /// # Arguments
     ///
-    /// * None
+    /// * assumed_lock_state - The assumed lock state of the SHA384 Accelerator.
+    /// Note: Callers should pass assumed_lock_state=ShaAccLockState::NotAcquired
+    ///  unless they are the first caller to the peripheral after a cold/warm boot.
     ///
     /// # Returns
     ///
-    /// * `Sha384AccOp` - On, success, an object representing the SHA384 accelerator operation.
-    /// * 'None' - On failure to acquire the SHA384 Accelerator lock.
-    pub fn try_start_operation(&mut self) -> Option<Sha384AccOp> {
+    /// * On success, either an object representing the SHA384 accelerator operation or
+    /// 'None' if unable to acquire the SHA384 Accelerator lock.
+    /// On failure, an error code.
+    ///
+    pub fn try_start_operation(
+        &mut self,
+        assumed_lock_state: ShaAccLockState,
+    ) -> CaliptraResult<Option<Sha384AccOp>> {
         let sha_acc = self.sha512_acc.regs();
 
-        if sha_acc.lock().read().lock() && sha_acc.status().read().soc_has_lock() {
-            None
-        } else {
-            // We acquired the lock, or we already have the lock (such as at startup)
-            Some(Sha384AccOp {
-                sha512_acc: &mut self.sha512_acc,
-            })
+        match assumed_lock_state {
+            ShaAccLockState::NotAcquired => {
+                if sha_acc.lock().read().lock() {
+                    // Either SOC has the lock (correct state),
+                    // or the uC has the lock but the caller doesn't realize it (bug).
+                    Ok(None)
+                } else {
+                    // The uC acquired the lock just now.
+                    Ok(Some(Sha384AccOp {
+                        sha512_acc: &mut self.sha512_acc,
+                    }))
+                }
+            }
+            ShaAccLockState::AssumedLocked => {
+                if sha_acc.lock().read().lock() {
+                    // SHA Acc is locked and the caller is assuming that the uC has it.
+                    Ok(Some(Sha384AccOp {
+                        sha512_acc: &mut self.sha512_acc,
+                    }))
+                } else {
+                    // Caller expected uC to already have the lock, but uC actually didn't (bug)
+                    Err(CaliptraError::DRIVER_SHA384ACC_UNEXPECTED_ACQUIRED_LOCK_STATE)
+                }
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


drivers/src/sha384acc.rs
```diff
@@ -42,13 +49,13 @@ impl Sha384Acc {
...
-    pub fn try_start_operation(&mut self) -> Option<Sha384AccOp> {
+    pub fn try_start_operation(&mut self, lock_state: ShaAccLockState) -> Option<Sha384AccOp> {
         let sha_acc = self.sha512_acc.regs();
 
-        if sha_acc.lock().read().lock() && sha_acc.status().read().soc_has_lock() {
-            None
+        if lock_state == ShaAccLockState::NotAcquried && sha_acc.lock().read().lock() {
+            None // SOC has the lock.
         } else {
-            // We acquired the lock, or we already have the lock (such as at startup)
```
korran, line 51: Didn't you mention in the meeting that in the ShaAccLockState::Acquired case you wanted to read sha_acc.lock() and ensure that it was 1, fatal error otherwise?

korran: My suggestion, to make all the possibilities clearer:

```rs
match lock_state {
    ShaAccLockState::Acquired => {
        if sha_acc.lock().read().lock() {
            // The uC didn't just acquire the lock, and the caller is
            // assuming the uC already has it.
            Ok(Some(Sha384AccOp {
                sha512_acc: &mut self.sha512_acc,
            }))
        } else {
            // We expected to already have the lock, but clearly we didn't (bug)
            Err(CaliptraError::DRIVER_SHA384ACC_UNEXPECTED_ACQUIRED_LOCK_STATE)
        }
    }
    ShaAccLockState::NotAcquried => {
        if sha_acc.lock().read().lock() {
            // SoC has the lock, or the uC has the lock but the caller
            // doesn't realize it (bug)
            Ok(None)
        } else {
            Ok(Some(Sha384AccOp {
                sha512_acc: &mut self.sha512_acc,
            }))
        }
    }
}
```

korran: I don't love how much we have to trust the caller of this function.

drivers/src/sha384acc.rs
```diff
@@ -24,6 +24,13 @@ const MAX_MAILBOX_CAPACITY_BYTES: u32 = 128 << 10;
 
 pub type Sha384Digest<'a> = &'a mut Array4x12;
 
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ShaAccLockState {
```
korran, line 29: Since this is in the sha384acc module, it would be ok to call this AssumedLockState, which I think makes it clearer what is actually happening here.

drivers/src/sha384acc.rs
```diff
@@ -42,13 +49,13 @@ impl Sha384Acc {
     ///
     /// * `Sha384AccOp` - On, success, an object representing the SHA384 accelerator operation.
     /// * 'None' - On failure to acquire the SHA384 Accelerator lock.
-    pub fn try_start_operation(&mut self) -> Option<Sha384AccOp> {
```
korran, line 45: Add a comment that callers to set `assumed_lock_state` unless they are the first caller to the peripheral after boot.

korran:
I don't think the emulator sha384acc peripheral does the correct thing on warm_reset() and update_reset().

drivers/src/sha384acc.rs
mhatrevi: Updated

jhand2 (dismissed):

rom/dev/src/main.rs
```diff
@@ -126,15 +128,22 @@ pub extern "C" fn rom_entry() -> ! {
 
             /// Ecc384 Engine
             ecc384: &mut env.ecc384,
+
+            /// SHA Acc lock state.
+            /// SHA Acc is locked on Cold and Warm Resets;
+            /// it is unlocked on Update Reset.
```
korran, line 134: I don't think this is true. I ran some tests against the RTL, and as best I can tell the update reset does not affect the sha384acc lock state. If the runtime firmware locks the mailbox, it stays locked across the reset. It does appear that if the SoC has the lock, it loses it across the reset.

We don't have any great options to fix #807:

- If we leave this code as is, the runtime firmware can't lock the mailbox before resetting to fix #807, as that would cause the AssumedLockState::NotAcquried attempt here to timeout in the watchdog.

- If we change this code to ShaAccLockState::Acquired for UpdateReset, we are trusting the runtime firmware too much, as an attacker with some control over the runtime firmware could do an update reset without the lock, and have the SoC grab the lock while the ROM assumes it already has the lock. This could be mitigated by doing that verify-random-data-can-be-written-to-register hack I've mentioned before in the assume-acquired case.

drivers/src/sha384acc.rs
korran: I see you updated the variable name. Did you intend to update the type name too? If you prefer it this way, that's fine too...

drivers/src/sha384acc.rs
```diff
@@ -36,22 +43,46 @@ impl Sha384Acc {
...
+                } else {
+                    // The uC acquired the lock just now.
+                    Ok(Some(Sha384AccOp {
+                        sha512_acc: &mut self.sha512_acc,
+                    }))
+                }
+            }
+            ShaAccLockState::AssumedLockState => {
+                if sha_acc.lock().read().lock() {
+                    // The uC had the lock, and the caller is assuming (correctly) that the uC already has it.
```
korran, line 77: > the caller is assuming (correctly) that the uC already has it.

From the context of this code, we don't know that, as we're not verifying it here. If the SoC had the lock, we would still go into this branch.

korran: Maybe this could read:

"Somebody has the lock, and the caller is assuming that it's the uC".

rom/dev/src/main.rs
mhatrevi: I updated the comment stating that ROM will expect it to be unlocked on UpdateReset.

korran (approved):

jhand2 (approved):

PR #862, opened 2023-09-22
ROM/RT: Use sha384 instead of sha384acc for fw validation.

korran:
From the smoke tests:

Before this change:
```
 15,393,069 <<< Executing mbox cmd 0x46574c44 (64456 bytes) from SoC                                                                       
 15,393,125 UART: [fwproc] Waiting for Commands...                                                                                         
 15,426,175 UART: [fwproc] Received command 0x46574c44                                                                                     
 15,428,606 UART: [fwproc] Received Image of size 64456 bytes                                                                              
 22,617,849 UART: [fwproc] Image verified using Vendor ECC Key Index 0                                                                     
 22,699,547 UART: [fwproc] Loading FMC at address 0x40000000 len 12360                                                                     
 22,746,278 UART: [fwproc] Loading Runtime at address 0x40004000 len 46212                                                                 
 22,912,705 >>> mbox cmd response: success               
 ```

After this change: 
 ```
15,434,886 <<< Executing mbox cmd 0x46574c44 (64540 bytes) from SoC                                                                       
 15,434,956 UART: [fwproc] Waiting for Commands...                                                                                         
 15,468,968 UART: [fwproc] Received command 0x46574c44                                                                                     
 15,471,165 UART: [fwproc] Received Image of size 64540 bytes                                                                              
 23,697,753 UART: [fwproc] Image verified using Vendor ECC Key Index 0                                                                     
 23,782,916 UART: [fwproc] Loading FMC at address 0x40000000 len 12360                                                                     
 23,829,274 UART: [fwproc] Loading Runtime at address 0x40004000 len 46296                                                                 
 23,995,949 >>> mbox cmd response: success
```

When measuring a 58k ICCM image bundle, this adds approximately 1.1 million cycles (2.4ms at 450MHz).

The sha384 register access code is not currently tuned for high-bandwidth performance; I suspect I can probably speed it up 2x to 3x with some register-access optimizations in an optional followup PR.


jhand2 (approved):

mschilder123 (commented):
There is some write locking on mailbox, right? As in its content cannot be mutated during this checking and subsequent copying. Orthogonal to sha engine switch or this pr, really.

Do we FIPS KAT test this internal sha engine in time to be used during this image verification stage (in case we delay FIPS self-tests at a granular level)?

korran:
> There is some write locking on mailbox, right? As in its content cannot be mutated during this checking and subsequent copying. Orthogonal to sha engine switch or this pr, really.

Yes, the mailbox cannot be modified while the SoC has the lock (which it does at this time).
This is probably the best argument against this PR: We have to trust the hardware to lock the mailbox correctly, so may as well trust it to lock the sha384 peripheral too.

If we want to be extra paranoid, you could make an argument for using the sha384_acc peripheral to quickly hash the mailbox contents before copying to ICCM, and then the sha384 peripheral to hash the contents once they've been copied to ICCM as an extra defense against TOCTOU. (ICCM is always executable, so we MUST check before copy).

> Do we FIPS KAT test this internal sha engine in time to be used during this image verification stage (in case we delay FIPS self-tests at a granular level)?

All the KATS run before anything else happens.

mschilder123 (approved):

JohnTraverAmd (approved):

rusty1968 (approved):

PR #873, opened 2023-09-26
Software emulator glitching

sphinxc0re:


sw-emulator/app/src/cli.rs
```diff
@@ -0,0 +1,95 @@
+// Licensed under the Apache-2.0 license
```
korran, line 1: This seems like a major refactoring unrelated to glitching...

PR #879, opened 2023-09-27
.github/workflows: Move fuzzing build tests to separate job

benjamindoron:
As these use a different toolchain, parallelise and don't use time in the main job. Also avoids potential collisions in sccache.


benjamindoron:
Should also mention, this re-enables Clippy where we can

Issue #851, opened 2023-09-20
Verilator test failures: caliptra-runtime::integration_tests


korran:
```
50364:2023-09-20T01:45:54.8524368Z         FAIL [26955.341s] caliptra-runtime::integration_tests test_disable_attestation_cmd
52108:2023-09-20T02:35:19.0550460Z         FAIL [27705.854s] caliptra-runtime::integration_tests test_fips_cmd_api
53341:2023-09-20T03:14:50.3249292Z         FAIL [27293.130s] caliptra-runtime::integration_tests test_invoke_dpe_sign_and_certify_key_cmds
54099:2023-09-20T03:42:33.9457981Z         FAIL [27095.727s] caliptra-runtime::integration_tests test_keyvault
56727:2023-09-20T05:58:25.2853509Z         FAIL [38798.178s] caliptra-runtime::integration_tests test_fw_info
57214:2023-09-20T06:15:38.4889439Z         FAIL [19932.771s] caliptra-runtime::integration_tests test_stash_measurement
```
```
2023-09-20T06:15:38.4889439Z         FAIL [19932.771s] caliptra-runtime::integration_tests test_stash_measurement
2023-09-20T06:15:38.4889986Z 
2023-09-20T06:15:38.4890783Z --- STDOUT:              caliptra-runtime::integration_tests test_stash_measurement ---
2023-09-20T06:15:38.4891261Z 
2023-09-20T06:15:38.4891437Z running 1 test
2023-09-20T06:15:38.4895722Z       6,155 ready_for_fuses is high
2023-09-20T06:15:38.4896273Z       6,333 writing to cptra_bootfsm_go
2023-09-20T06:15:38.4896932Z test test_stash_measurement has been running for over 60 seconds
2023-09-20T06:15:38.4897533Z     195,201 UART: 
2023-09-20T06:15:38.4902426Z     195,256 UART: Running Caliptra ROM ...
2023-09-20T06:15:38.4902789Z     196,662 UART: 
2023-09-20T06:15:38.4903077Z     202,819 UART: [state] CFI Enabled
2023-09-20T06:15:38.4903441Z     205,066 UART: [state] LifecycleState = Unprovisioned
2023-09-20T06:15:38.4903787Z     207,326 UART: [state] DebugLocked = No
2023-09-20T06:15:38.4904235Z     209,296 UART: [state] Watchdog Timer is not started because the device is not locked for debugging
[...]

PR #893, opened 2023-09-29
Support for VALID_PAUSER and pauser of 0x1 for verilator and fpga

nquarton:
- Adds configurable soc_apb_pauser to HW model InitParams
- Adds setup of configurable VALID_PAUSER, configurable in HW model BootParams
- All PAUSER values default to 0x1

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -145,6 +145,9 @@ pub struct InitParams<'a> {
     pub trng_mode: Option<TrngMode>,
 
     pub wdt_timeout_cycles: u64,
+
+    // PAUSER value used for APB read/write cycles from SoC->caliptra
+    pub soc_apb_pauser: u32,
 }
 
 impl<'a> Default for InitParams<'a> {
@@ -174,11 +177,11 @@ impl<'a> Default for InitParams<'a> {
             etrng_responses,
             trng_mode: Default::default(),
             wdt_timeout_cycles: EXPECTED_CALIPTRA_BOOT_TIME_IN_CYCLES,
+            soc_apb_pauser: 0x1,
         }
     }
 }
 
-#[derive(Default)]
 pub struct BootParams<'a> {
     pub init_params: InitParams<'a>,
     pub fuses: Fuses,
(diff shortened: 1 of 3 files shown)


nquarton:
The failure is because I need to either add support for MBOX_VALID_PAUSER in the unit tests or use PAUSER of 0xFFFF_FFFF there. Currently the hw-model is what sets up MBOX_VALID_PAUSER

nquarton:
This addresses a failure documented here: https://github.com/chipsalliance/caliptra-sw/issues/851

hw-model/src/lib.rs
```diff
@@ -174,18 +177,33 @@ impl<'a> Default for InitParams<'a> {
...
+impl<'a> Default for BootParams<'a> {
+    fn default() -> Self {
+        Self {
+            init_params: Default::default(),
+            fuses: Default::default(),
+            fw_image: Default::default(),
+            initial_dbg_manuf_service_reg: Default::default(),
+            initial_repcnt_thresh_reg: Default::default(),
+            initial_adaptp_thresh_reg: Default::default(),
+            valid_pauser: 0x1,
```
rusty1968, line 204: Why does it need to be in BootParams and InitParams?

nquarton: One is for configuring the VALID_PAUSER register, the other is for which PAUSER will be used for bus accesses from "SoC". I'm not completely against having them be one thing, but I figured the control to individually control them would be wanted.

Additionally I'd like to add a test case to confirm accesses are dropped when there is a mismatch (which would require the individual control), but that's not supported in the emulator and I don't know if I can classify a test as invalid in the emulator environment.

hw-model/src/lib.rs
```diff
@@ -919,6 +956,18 @@ mod tests {
         model.soc_ifc().cptra_fuse_wr_done().write(|w| w.done(true));
         model.soc_ifc().cptra_bootfsm_go().write(|w| w.go(true));
 
+        // Set up the PAUSER as valid for the mailbox (using index 0)
+        model
+            .soc_ifc()
+            .cptra_mbox_valid_pauser()
+            .at(0)
+            .write(|_| 0x1);
```
jhand2, line 964: Should this be using the `valid_pauser` value from BootParams?

hw-model/src/lib.rs
```diff
@@ -951,6 +1000,18 @@ mod tests {
...
```
jhand2, line 1008: Same comment as above

hw-model/src/lib.rs
```diff
@@ -970,6 +1031,18 @@ mod tests {
...
```
jhand2, line 1039: Same

hw-model/src/lib.rs
nquarton: Good question, so these tests use new_unbooted and because of that, they don't have BootParams. Based on other setup I thought "new" was the right place for this. But I'm open to input on that.

jlmahowa-amd (approved):

jhand2 (approved):

PR #895, opened 2023-10-02
Expand FPGA memory map to include SHA acc

jlmahowa-amd:


hw-model/src/model_fpga_realtime.rs
```diff
@@ -19,21 +19,20 @@ use crate::Output;
 const SOC_PAUSER: u32 = 0xffff_ffff;
 
 // UIO mapping indices
-const GPIO_MAPPING: usize = 0;
-const MBOX_MAPPING: usize = 1;
-const SOC_IFC_MAPPING: usize = 2;
+const SOC_MAPPING: usize = 0;
```
jlmahowa-amd, line 22: "SOC" is too confusing. I should change this (and other re-names) to something like "FPGA_WRAPPER" instead.

korran: CALIPTRA_APB_MAPPING?

korran (approved):

PR #901, opened 2023-10-03
Fix pathLenConstraint for cert chain

### LINKED ISSUE #897: pathLenConstraint incorrect for LDevID and FMCAlias certificates
https://github.com/chipsalliance/caliptra-sw/blob/main/x509/build/build.rs#L66 shows pathLenConstraint of 0 for LDevID and FMCAlias. This is incorrect as LDevID and FMCAlias are issuers for CAs.

Proposal from @jhand2 :
IDevID (pathLen = 5) -> LDevID (pathLen = 4) -> FMC Alias (pathLen = 3) -> RT Alias (pathLen = 2) -> DPE IsCA = True (pathLen = 1) -> SoC Manager leaf (IsCA = False)
This would mean the SoC manager can make a CA key held by DPE, but they can only use it to sign one additional certificate layer.

jhand2:
Previously, all pathLenConstraints were set to 0, which will always fail verification. Update pathLenConstraint to
`IDevID (5) -> LDevID (4) -> FMC ALias (3) -> RT Alias (2)`.

Additionally, add AuthorityKeyIdentifier to the FMC Alias Key.

Add tests to ensure openssl can verify Caliptra's cert chain.

Fixes #897


jhand2:
TODO: I need to figure out how to update the FakeROM collateral but I can't figure out an easy way to do it.

Ideally we should check in a script to update it. cc @nquarton are there instructions?

x509/build/build.rs
```diff
@@ -96,7 +96,7 @@ fn gen_rt_alias_cert(out_dir: &str) {
     usage.set_digital_signature(true);
     let bldr = cert::CertTemplateBuilder::<EcdsaSha384Algo>::new()
         // Basic Constraints : CA = true, PathLen = 1
```
jhand2, line 98:
```suggestion
        // Basic Constraints : CA = true, PathLen = 2
```

varuns-nvidia (approved):

nquarton:
> TODO: I need to update the FakeROM collateral but I can't figure out an easy way to do it.
> 
> Ideally we should check in a script to update it. cc @nquarton are there instructions?

I need to come up with a better way (like you said, preferably a script) to do this. I ran a real ROM boot, manually copied out the data from the log, and formatted it with macros in a text editor.

I incorrectly assumed the canned cert wouldn't be too affected by tweaks and changes down the road and figured that manual process was okay. I'm offline most of this morning but I'll try to look into this more this afternoon.

Depending on the urgency maybe I can repeat that manual process on this PR for you since any other solution is likely going to  take a little more time.

nquarton:
We could just separate the expected certs for fake_rom and the real rom for now like this: https://github.com/jhand2/caliptra-sw/pull/2

I don't even know if we care too much about it not matching exactly longer term. But I can check on that.

FerralCoder (dismissed):

jhand2:
> TODO: I need to update the FakeROM collateral but I can't figure out an easy way to do it.
> 
> Ideally we should check in a script to update it. cc @nquarton are there instructions?

Updated ROM's build.rs to autogenerate this collateral from the checked in certs for the smoke test.

nquarton:
> > TODO: I need to update the FakeROM collateral but I can't figure out an easy way to do it.
> > Ideally we should check in a script to update it. cc @nquarton are there instructions?
> 
> Updated ROM's build.rs to autogenerate this collateral from the checked in certs for the smoke test.

This is fantastic. Thank you so much for doing this!

varuns-nvidia (approved):

nquarton (approved):

korran (approved):

mhatrevi (approved):

bluegate010 (approved):

PR #902, opened 2023-10-03
reproducible/hermetic binary builds

leongross:
The aim of this PR is to provide a deterministic build environment, which results in bit-by-bit identical firmware binaries.

Not reproducible for far:
- [ ] `lib.rmeta`


leongross:
@rusty1968 @jhand2 may you have a look at this?

korran:
This doesn't appear to be building the firmware binaries. See `./ci.sh check_frozen_images`.

leongross:
@korran does this fit the needs? Is generating and freezing the checksums all we are interested in?

leongross:
@korran does this implement what you have in mind? Is building the binaries `caliptra-rom-with-log.bin` and `caliptra-rom-no-log.bin` and recording their sha384 sums enough? Or should this also package and release other binaries (such as in the github releases)?

.github/Containerfiles/ubuntu_2204/Containerfile.base
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:22.04@sha256:aabed3296a3d45cede1dc866a24476c4d7e093aa806263c27ddaadbdce3c1054
+
+RUN apt update && apt install -y \
```
ArthurHeymans, line 3: Is 'update' good enough for this? The image would be different depending on when it was run, e.g. if ubuntu changes.

korran: Generally the idea with hermetic builds is that the input data to the build is immutable.

Our use of non-vendored cargo deps is arguably not hermetic (since we're downloading crates from the crates.io), but at least there Cargo.lock is ensuring that the hashes on the deps don't change.

Unfortunately, we don't have as good of a story for the OS or toolchain...

leongross: Since this is the base image it will only be built once since we once have to get the toolchain installed. This image should then be re-used for all other builds, for which it then can be used in a deterministic way.

leongross: I understand your concerns, maybe we can discuss some different approaches:

1. One alternative might be using Bazel as a hermetic build system. We could hash-pin all dependencies and rebuild wanted images at any time. I'm not a Bazel expert but afaik to make it efficient we would have to keep already built artifacts in caches to not rebuild the entire dependency chain. This would definitely require some set-up and running overhead.

2. Another alternative could be using NixOS as an build environment. There we can also hash-pin our dependencies and would have a reproducible build environment. The upsite I see here is, that we could use it like a "normal" Linux build environment and run the existing build scripts with little to non adjustment to them. On the other hand we would have to construct a nix config and set up some infrastructure to realize this.

Both variants would introduce more work than going with a docker like approach, but also they would considerably remove opacity compared to the simple docker approach.

Also I think the [reproducible-builds](https://reproducible-builds.org/docs/) website gives a great overview on what to consider when using hermetic builds.

MDr164: As discussed in this weeks sync: We should keep the Dockerfile approach for usability for now. The core parts are already reproducible to a satisfying degree due to Cargo keeping a lock file and the container gives a quick setup with a known toolchain and OS environment.

PR #907, opened 2023-10-03
Implement pauser PL based derive child/init context limits

sree-revoori1:
DeriveChild and InitalizeCtx can create new contexts in DPE. An attacker could use these commands to exhaust space in DPE's contexts array. To prevent this, we implement limits per pauser privilege level on the number of active contexts belonging to each pauser PL. If a DPE command would cause the number of active contexts to breach these limits, we do not execute that command and return an error.


sree-revoori1:
CI will fail for this PR until https://github.com/chipsalliance/caliptra-dpe/pull/211 is in and we update the DPE submodule, but locally it passes.

runtime/src/invoke_dpe.rs
```diff
@@ -45,18 +49,45 @@ impl InvokeDpeCmd {
             let command = Command::deserialize(&cmd.data[..cmd.data_size as usize])
                 .map_err(|_| CaliptraError::RUNTIME_INVOKE_DPE_FAILED)?;
             let flags = pdata.manifest1.header.flags;
+            let active_pl0_dpe_context_count = pdata.active_pl0_dpe_context_count;
+            let active_pl1_dpe_context_count = pdata.active_pl1_dpe_context_count;
```
jhand2, line 53: Now that we have APIs in DpeInstance, can we use those instead of storing this in persistent data.

sree-revoori1: Done

runtime/tests/integration_tests.rs
```diff
@@ -425,6 +430,148 @@ fn test_invoke_dpe_get_certificate_chain_cmd() {
...
+            u32::from(CommandId::INVOKE_DPE),
+            rotate_ctx_mbox_cmd.as_bytes(),
+        )
+        .unwrap()
+        .expect("We should have received a response");
+
+    assert!(rotate_ctx_resp_buf.len() <= std::mem::size_of::<InvokeDpeResp>());
+    let mut rotate_ctx_resp_hdr = InvokeDpeResp::default();
+    rotate_ctx_resp_hdr.as_bytes_mut()[..rotate_ctx_resp_buf.len()]
+        .copy_from_slice(&rotate_ctx_resp_buf);
```
jhand2, line 481: nit: I think these lines can collapse to `let rotate_ctx_resp = InvokeDpeResp::read_from(rotate_ctx_resp_buf)`

runtime/tests/integration_tests.rs
```diff
@@ -425,6 +430,148 @@ fn test_invoke_dpe_get_certificate_chain_cmd() {
...
+                )
+                .unwrap_err();
+            if let ModelError::MailboxCmdFailed(code) = resp {
+                assert_eq!(
+                    code,
+                    caliptra_drivers::CaliptraError::RUNTIME_PL0_ACTIVE_DPE_CONTEXT_THRESHOLD_EXCEEDED.into()
+                );
+            }
+            break;
+        }
```
jhand2, line 546: nit: If you update the loop to be `for i in 0..InvokeDpeCmd::PL0_DPE_ACTIVE_CONTEXT_THRESHOLD`, you can add this check outside the loop and ditch the `if`.

runtime/src/invoke_dpe.rs
```diff
@@ -91,7 +109,32 @@ impl InvokeDpeCmd {
         }
     }
 
-    fn is_caller_p1(pl0_pauser: u32, flags: u32, locality: u32) -> bool {
+    fn pl_context_threshold_exceeded(
```
jhand2, line 112: Code looks good. Can you also add a section to the runtime spec (somewhere near the other DPE details) explaining that we're doing this and why?

runtime/tests/integration_tests.rs
sree-revoori1: I tried it and the read_from call fails. Not totally sure, but I think it is due to missing the MailboxRespHeader.

runtime/tests/integration_tests.rs
sree-revoori1: But then derive_child_mbox_cmd would be out of scope, right? I agree this isn't the cleanest way to do it, but it would be a lot of code repetition to move this to outside of the loop I think.

runtime/src/invoke_dpe.rs
sree-revoori1: Done.

runtime/tests/integration_tests.rs
jhand2: Does it work if you do `let rotate_ctx_resp = InvokeDpeResp::read_from_prefix(rotate_ctx_resp_buf))`? I'm guessing it fails because the slice it larger than `read_from` is expecting.

sree-revoori1: No, strangely it still fails but now it fails on the call to DeriveChild. I'm guessing it incorrectly parsed the rotated context handle so derive child couldn't find it.

PR #918, opened 2023-10-05
Use Zeroize Crate to scrub memory.

### LINKED ISSUE #843: slice::fill(0) Does Not Always Zero Memory
Memory that is not zeroed stays resident for longer than strictly necessary, potentially
allowing it to be exposed by a separate memory exfiltration vulnerability.

mhatrevi:
This PR addresses issue https://github.com/chipsalliance/caliptra-sw/issues/843


rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -36,21 +37,16 @@ impl Ecdsa384SignatureAdapter for Ecc384Signature {
 }
 
 /// DICE  Layer Key Pair
-#[derive(Debug)]
+#[derive(Debug, Zeroize)]
```
bluegate010, line 40: Should these be `ZeroizeOnDrop`?

mhatrevi: We are explicitly calling the zeroize trait method on the critical structs, so just having Zeroize is good.

korran: It's worth considering in a future PR; I suspect we end up forgetting to zeroize many of these.

korran (approved):

rom/dev/src/flow/cold_reset/crypto.rs
mhatrevi: Sounds good.

jhand2 (approved):

PR #927, opened 2023-10-06
Add pauser control during execution for verilator and FPGA

nquarton:


nquarton:
I'm not sure if I am introducing the FPGA failure or not yet. It appears to be for a related test case, but I don't think there should have been any behavior change in that test from this PR for it.

jlmahowa-amd:
This PR does cause the FPGA error, confirmed by running the new test manually. What I think is happening is that Caliptra is issuing a SLVERR on the APB bus in response to the transaction with a different PAUSER. This error gets passed up to the PS AXI master by the APB-AXI Bridge, the OS's handler gets called and kills the process.
Can we handle this in SW?

Here is the specific error message I got:
Caused by:
  process didn't exit successfully: `/home/ubuntu/caliptra-sw/target/debug/deps/caliptra_hw_model-6d9a466b74b68363 test_mbox_pauser` (signal: 7, SIGBUS: access to undefined memory)

jlmahowa-amd (approved):

rusty1968 (approved):

PR #929, opened 2023-10-09
Always reset watchdog timer after enabling it to guarantee counting starts from zero

rusty1968:
- Move wdt from common to drivers
- Test activation and stoppage of WDT
- Always reset watchdog timer after enabling it to guarantee counting starts from zero.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -639,6 +639,7 @@ version = "0.1.0"
 dependencies = [
  "caliptra-cpu",
  "caliptra-drivers",
+ "caliptra-registers",
  "cfg-if 1.0.0",
  "ufmt",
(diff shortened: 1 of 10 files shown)


mhatrevi (changes requested):
Why is the move from common to drivers needed? Let's keep the change minimal and only add reset to start_wdt (which you have already done)

rusty1968:
> Why is the move from common to drivers needed? Let's keep the change minimal and only add reset to start_wdt (which you have already done)

This was feedback from @korran  during a standup meeting : he recommended moving to drivers anything that changes ROM. Other components have already been moved from common, e.g. the handoff module, printer, etc).

PR #930, opened 2023-10-09
[fix] Harden update reset flow

### LINKED ISSUE #817: If a warm reset occurs while ICCM is being written during a hitless update, CPU can execute an unvalidated instruction sequence
After the warm reset, the CPU eventually jumps to runtime firmware that is partially written.

By controlling the amount of time before the warm reset, an attacker may be able to get the CPU to perform a dangerous operation.

mhatrevi:
This PR fixes issue# https://github.com/chipsalliance/caliptra-sw/issues/817

### DIFF
--- a/drivers/src/data_vault.rs
+++ b/drivers/src/data_vault.rs
@@ -136,6 +136,7 @@ pub enum WarmResetEntry4 {
     RtEntryPoint = 1,
     ManifestAddr = 2,
     RtMinSvn = 3,
+    RomUpdateResetStatus = 4,
 }
 
 impl From<WarmResetEntry4> for u8 {
@@ -390,16 +391,6 @@ impl DataVault {
         self.read_cold_reset_entry4(ColdResetEntry4::LmsVendorPubKeyIndex)
     }
 
-    /// Set and lock the rom cold boot status.
-    ///
-    /// # Arguments
-    ///
-    /// * `status` - Rom Cold Boot Status
-    ///
-    pub fn set_rom_cold_boot_status(&mut self, status: u32) {
-        self.write_lock_cold_reset_entry4(ColdResetEntry4::RomColdBootStatus, status);
-    }
-
     /// Get the rom cold boot status.
     ///
     /// # Returns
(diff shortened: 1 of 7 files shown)


drivers/src/data_vault.rs
```diff
@@ -390,14 +391,19 @@ impl DataVault {
...
+    /// Set the rom cold boot status.
     ///
     /// # Arguments
     ///
     /// * `status` - Rom Cold Boot Status
+    /// * `lock_entry` - Lock the entry
     ///
-    pub fn set_rom_cold_boot_status(&mut self, status: u32) {
-        self.write_lock_cold_reset_entry4(ColdResetEntry4::RomColdBootStatus, status);
+    pub fn set_rom_cold_boot_status(&mut self, status: u32, lock_entry: bool) {
```
korran, line 401: It's not clear to the reader of calling code what these booleans mean. Perhaps this should follow the same pattern as write_lock_cold_reset_entry4() and be two separate functions?

`set_rom_cold_boot_status()` and `set_lock_cold_boot_status()`?

korran: Although I'm honestly not sure why we're not just calling the write_*entry4 functions directly....

mhatrevi: Removed the additional helper methods.

korran (approved):

rom/dev/src/flow/warm_reset.rs
```diff
@@ -31,12 +31,23 @@ impl WarmResetFlow {
     pub fn run(env: &mut RomEnv) -> CaliptraResult<()> {
         cprintln!("[warm-reset] ++");
 
-        // Check if previous Cold-Reset was successful.
-        if cfi_launder(env.data_vault.rom_cold_boot_status()) != ColdResetComplete.into() {
-            cprintln!("[warm-reset] Previous Cold-Reset was not successful.");
-            return Err(CaliptraError::ROM_WARM_RESET_UNSUCCESSFUL_PREVIOUS_COLD_RESET);
+        // Check if previous [Cold | Update] reset was successful.
+        if cfi_launder(env.data_vault.rom_cold_boot_status()) != ColdResetComplete.into()
+            && cfi_launder(env.data_vault.rom_update_reset_status()) != UpdateResetComplete.into()
```
korran, line 36: I don't think this logic is right. If rom_cold_boot_status() == ColdResetComplete, then we'll never go into this block...

korran (changes requested):
See comment

rom/dev/src/flow/warm_reset.rs
korran: These values can have three states: Default, Started, and Complete.

I think we want `rom_cold_boot_status() != ColdResetComplete || rom_update_reset_status() == UpdateResetStarted`

korran: Or maybe the slightly less glitch prone:

```
rom_cold_boot_status() != ColdResetComplete || (rom_update_reset_status() != UpdateResetDefaultValue && rom_update_reset_status() != UpdateResetCompleted)
```

mhatrevi: Fixed.

korran (dismissed):

korran (approved):

PR #935, opened 2023-10-10
[fix] Return fake random numbers if RNG h/w block unavailable

### LINKED ISSUE #888: ROM/Fake ROM to Support Integration by Booting without Random Numbers
Request is:
* Have a bit in Debug/Manufacturing register where ROM decides to skip RNG health tests for any RNG data it requires to boot if not in the production security state.

Background:
* For integration, RNG may take longer to setup and develop.  To help enable parrallel development and integration, it would be ideal to have a way to skip RNG requirements so ROM gets to ready for FW with minimal effort for the integrator.
* Another point here is that on first boot, before fuses are programmed, ROM can't hang if the random data isn't good enough.  ROM needs to be able to boot without a proper RNG configuration when not in a production state.

mhatrevi:
This change addresses issue# https://github.com/chipsalliance/caliptra-sw/issues/888

### DIFF
--- a/drivers/src/trng.rs
+++ b/drivers/src/trng.rs
@@ -1,16 +1,19 @@
 // Licensed under the Apache-2.0 license
 
+use core::array;
+
 use caliptra_error::{CaliptraError, CaliptraResult};
 use caliptra_registers::{
     csrng::CsrngReg, entropy_src::EntropySrcReg, soc_ifc::SocIfcReg, soc_ifc_trng::SocIfcTrngReg,
 };
 
-use crate::{trng_ext::TrngExt, Array4x12, Csrng};
+use crate::{trng_ext::TrngExt, Array4x12, Csrng, MfgFlags};
 
 #[repr(u32)]
 pub enum Trng {
     Internal(Csrng) = 0xb714a2b1,
     External(TrngExt) = 0xf3702ce3,
+    MfgMode() = 0x0c702ce3,
 
     // Teach the compiler that "other" values are possible to encourage it not
     // to get too crazy with optimizations. Match statements should handle `_`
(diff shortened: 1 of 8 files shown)


drivers/src/trng.rs
```diff
@@ -48,6 +57,48 @@ impl Trng {
         match self {
             Self::Internal(csrng) => Ok(csrng.generate12()?.into()),
             Self::External(trng_ext) => trng_ext.generate(),
+            Self::MfgMode(rng) => Ok(rng.generate12()?.into()),
```
korran, line 60: I don't think this match statement provides enough glitch resistance. If somebody can glitch the cpu into this branch when we're initializing the CFI PRNG, we lose significant CFI protections.

korran: Maybe that's ok if we add a bunch of cfi_asserts to FakeRng::generate()

drivers/src/trng.rs
```diff
@@ -48,6 +57,48 @@ impl Trng {
...
+    ///
+    /// # Errors
+    ///
+    /// None
+    pub fn generate16(&mut self) -> CaliptraResult<[u32; 16]> {
+        self.generate()
+    }
+
+    fn generate<const N: usize>(&mut self) -> CaliptraResult<[u32; N]> {
+        Ok(array::from_fn(|_| 0xdeadbeef_u32))
```
korran, line 102: We need multiple CFI protections here check that the flag is set and that we're in debug mode.

drivers/src/trng.rs
korran: we probably need to look a lot closer at all this assembly.

mhatrevi: cfi_asserts are added to ROM main function to ensure the TRNG config correctness.

drivers/src/trng.rs
mhatrevi: cfi_asserts are added to ROM main function to ensure the TRNG config correctness.

drivers/src/trng.rs
```diff
@@ -66,3 +76,44 @@ impl Trng {
         }
     }
 }
+
+pub struct FakeRng {}
```
korran, line 80: I don't think this struct is necessary; just return the bogus data in the Trng::generate() match arm.

rom/dev/src/main.rs
```diff
@@ -378,3 +363,36 @@ fn panic_is_possible() {
     // The existence of this symbol is used to inform test_panic_missing
     // that panics are possible. Do not remove or rename this symbol.
 }
+
+#[inline(always)]
+fn validate_trng_config(env: &mut RomEnv) {
+    if !cfi_launder(env.soc_ifc.debug_locked())
```
korran, line 369: We don't want to have these sorts of single branches. Nick wanted the only branches to be branches to cfi_panic

korran: Addressed by my push.

korran (approved):

PR #938, opened 2023-10-10
- Reset watchdog timer after enabling it to guarantee counting starts from zero

rusty1968:


rom/dev/tests/test_wdt_activation_and_stoppage.rs
```diff
@@ -0,0 +1,83 @@
...
+            rom: &rom,
+            security_state,
+            ..Default::default()
+        },
+        ..Default::default()
+    })
+    .unwrap();
+
+    // Ensure we are starting to count from zero.
+    hw.step_until(|m: &mut caliptra_hw_model::ModelEmulated| {
```
korran, line 36: Don't ever reference the type ModelEmulated. Use the DefaultHwModel alias instead, which uses the correct type depending on the feature flags.

mhatrevi (approved):

rom/dev/tests/test_wdt_activation_and_stoppage.rs
rusty1968: Done

PR #949, opened 2023-10-12
Check if the owner and vendor cert validty dates are present in RT Alias cert

rusty1968:


runtime/tests/integration_tests.rs
```diff
@@ -101,6 +102,44 @@ fn test_locked_dv_slot() {
...
+        .copy_from_slice("20260101000000Z".as_bytes());
+
+    let mut own_config = opts.owner_config.unwrap();
+
+    own_config
+        .not_before
+        .copy_from_slice("20270101000000Z".as_bytes());
+    own_config
+        .not_after
+        .copy_from_slice("20280101000000Z".as_bytes());
```
jhand2, line 125: nit: Can we define these strings as variables/constants at the top of the function? That we don't have to edit in multiple places if they need to be changed.

runtime/tests/integration_tests.rs
```diff
@@ -101,6 +102,47 @@ fn test_locked_dv_slot() {
...
+
+    let mut model = run_rt_test(Some(&firmware::runtime_tests::CERT), Some(opts));
+
+    let rt_resp = model.mailbox_execute(0x3000_0000, &[]).unwrap().unwrap();
+    let rt: &[u8] = rt_resp.as_bytes();
+
+    let rt_cert: X509 = X509::from_der(rt).unwrap();
+
+    let not_before: Asn1Time = Asn1Time::from_str("20270101000000Z").unwrap();
+    let not_after: Asn1Time = Asn1Time::from_str("20280101000000Z").unwrap();
```
jhand2, line 140: nit: Should these be populated with the `VENDOR_CONFIG` constants?

rusty1968: own config. done.

jhand2 (approved):

PR #954, opened 2023-10-12
[fix] CFI check for TRNG source

### LINKED ISSUE #920: Triple-verify TRNG MMIO read at CFI initialization in `rom_entry`
From Nick Galloway (gallowayn@google.com)

Out of reset, the Caliptra ROM does some initialization depending on whether it's a warm/cold boot, and will then jump to `rom_entry`. There's more initialization in there which is inlined from `rom_env::RomEnv::new_from_registers()`, but the important bit is an MMIO register read of the TRNG configuration. This is stored in a single bit read from the MMIO register, so it lacks the typical large hamming distance mitigation that other conditionally checked variables have to harden against bit flips. To mitigate this there should be an inlined triple verification step in the compiled `rom_entry` to verify that the register was read correctly to try to mitigate glitches sending the CFI library off to a random source the current version of Caliptra isn't designed for. (Something simple to check if the `Trng` enum is correctly _Internal_ or _External_ to match the register should be sufficient)

### LINKED ISSUE #921: Triple call CFI Counter reset when seeding the prng
From Nick Galloway (gallowayn@google.com)

Affter `RomEnv` has been created, `cfi_counter::CfiCounter::reset()` is called to seed the prng for the `CfiCounter`. That function is marked with `#[inline(never)]`, but it should instead be always inlined and called multiple times (say, three) so that it's harder to glitch past the prng seed from the trng. To successfully exploit this would require the ability to successfully glitch past the current non-inlined prng seed and then also a second time to glitch past something else (say FMC signing checks). After the initial prng seed all of the delays would be deterministic because the initial state of the `Xoshiro128` prng is itself deterministic, so this should be straightforward for someone able to glitch twice on a single run.

### LINKED ISSUE #922: Mitigate glitching of calls within Trng generate
From Nick Galloway (gallowayn@google.com)

An additional edge case that could be glitched in the early boot sequence before the first random delay happens:

* `rom_entry` calls `caliptra_cfi_lib::cfi_counter::CfiCounter::reset`
  * `reset` calls `caliptra_drivers::trng::Trng::generate`
    * `generate` uses `ureg::read_volatile_slice` and `compiler_builtins::mem::memcpy::memcpy`

Either of which could be attacked with a glitch to trigger a buffer overflow, such as with an instruction skip or bit flip

mhatrevi:
This fix addresses the following issues:

1.   https://github.com/chipsalliance/caliptra-sw/issues/920
2. https://github.com/chipsalliance/caliptra-sw/issues/921

image


andreslagarcavilla:
Needs rebase, please.

Also: "The Caliptra ROM is frozen; changes that affect the binary
require approval from the TAC." -- yes approved. This seems noisy at this stage

rom/dev/src/main.rs
```diff
@@ -70,10 +70,26 @@ pub extern "C" fn rom_entry() -> ! {
...
         cprintln!("[state] CFI Enabled");
         CfiCounter::reset(&mut env.trng);
+        CfiCounter::reset(&mut env.trng);
+        CfiCounter::reset(&mut env.trng);
     } else {
         cprintln!("[state] CFI Disabled");
     }
 
+    // Check if TRNG is correctly sourced as per hw config.
+    match env.trng {
```
nickg-ca, line 80: Does this compile to a single branch to then assert three times that the configuration is correct? If so I think it should be split out into three match statements so it's less easy to glitch past the single branch to the incorrect match. (Or possibly a function that does the verification, where that function is called several times)

korran: What do you think of https://github.com/chipsalliance/caliptra-sw/pull/979/commits/414433043c1f4daaace90e64ba19a4daa01f882e ?

korran: Merged that change into this PR. Thoughts?

rom/dev/src/start.S
```diff
@@ -104,14 +104,21 @@ _start:
...
 
-    la a0, DCCM_ORG         // dest
-    la a1, DCCM_SIZE        // len 
-    call _zero_mem256
+    // Zero out post-CFI Entropy region
+    // CFI region is used as an additional entropy source for CFI RNG,
+    // hence it is not cleared.
+    la a0, DCCM_POST_CFI_ENTROPY_ORG     // dest
+    la a1, DCCM_POST_CFI_ENTROPY_SIZE    // len 
+    call _zero_mem32
```
korran, line 121: Can we reserve the first 8 words and use zero_mem256? The unrolled loop is WAY faster.

rom/dev/src/main.rs
nickg-ca: Yeah, I think that's likely to look good when compiled. I'll need to have a look at the compiled version of this when I get a chance.

korran (dismissed):

nickg-ca:
Does this address #922 ? I only noticed that when looking at the compiled assembly which calls `ureg:: read_volatile_slice` and `compiler_builtins::mem::memcpy::memcpy`, either of which could be glitched to trigger a buffer overflow. For what it's worth I'm somewhat less concerned about #922 than the other issues.

korran:
> Does this address #922 ? I only noticed that when looking at the compiled assembly which calls `ureg:: read_volatile_slice` and `compiler_builtins::mem::memcpy::memcpy`, either of which could be glitched to trigger a buffer overflow. For what it's worth I'm somewhat less concerned about #922 than the other issues.

I don't think it does. I propose https://github.com/chipsalliance/caliptra-sw/pull/989 to solve #922.

korran (approved):

rusty1968 (approved):

PR #955, opened 2023-10-13
Add additional information to RT trap_handler

sree-revoori1:
This will make it easier to diagnose unexpected NMI errors.

### DIFF
--- a/runtime/src/main.rs
+++ b/runtime/src/main.rs
@@ -20,6 +20,7 @@ use caliptra_drivers::{
     report_fw_error_fatal, report_fw_error_non_fatal, Ecc384, Hmac384, KeyVault, Mailbox, Sha256,
     Sha384, Sha384Acc, SocIfc,
 };
+use caliptra_registers::soc_ifc::SocIfcReg;
 use caliptra_runtime::Drivers;
 use core::hint::black_box;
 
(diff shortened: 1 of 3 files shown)


runtime/src/main.rs
```diff
@@ -76,13 +88,37 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
...
 
+    {
+        let soc_ifc = soc_ifc.regs_mut();
+        let ext_info = soc_ifc.cptra_fw_extended_error_info();
+        ext_info.at(0).write(|_| trap_record.mcause);
+        ext_info.at(1).write(|_| trap_record.mscause);
+        ext_info.at(2).write(|_| trap_record.mepc);
+        ext_info.at(3).write(|_| trap_record.ra);
+        ext_info.at(4).write(|_| err_interrupt_status);
+    }
```
jhand2, line 120: This is really helpful! @rusty1968 @FerralCoder we should do this in FMC as well.

We should also document somewhere what we output in these registers so SoC integrators can find it somewhere.

jhand2 (approved):

runtime/src/main.rs
jhand2: Opened https://github.com/chipsalliance/caliptra-sw/issues/958 to track

PR #960, opened 2023-10-13
[fix] Do not stop WDT upon ROM exit.

rusty1968:


FerralCoder (approved):

runtime/src/main.rs
```diff
@@ -119,7 +122,15 @@ extern "C" fn nmi_handler(trap_record: &TrapRecord) {
...
     }
 
-    handle_fatal_error(caliptra_drivers::CaliptraError::RUNTIME_GLOBAL_NMI.into());
+    let mut error = CaliptraError::RUNTIME_GLOBAL_NMI;
+
+    let wdt_status = soc_ifc.regs().cptra_wdt_status().read();
+    if wdt_status.t1_timeout() || wdt_status.t2_timeout() {
+        cprintln!("WDT Expired");
+        error = CaliptraError::RUNTIME_GLOBAL_WDT_EXPIRED;
+    }
```
jhand2, line 131: tiny nit, could do this to be a little more idiomatic

```suggestion
    let wdt_status = soc_ifc.regs().cptra_wdt_status().read();
    let error = if wdt_status.t1_timeout() || wdt_status.t2_timeout() {
        cprintln!("WDT Expired");
        CaliptraError::RUNTIME_GLOBAL_WDT_EXPIRED
    } else {
        CaliptraError::RUNTIME_GLOBAL_NMI
    }
```

.github/workflows/fpga.yml
```diff
@@ -366,7 +366,9 @@ jobs:
           # Tests that are known to pass on the FPGA
           # TODO(kor): Fix all the tests and run them all here
           sudo ${VARS} "${TEST_BIN}/smoke_test-"* \
-              --skip fips_cmd_test_rt
+              --skip fips_cmd_test_rt \
+              --skip test_rt_wdt_timeout \
+              --skip test_fmc_wdt_timeout
```
jhand2, line 371: Do we know why these don't pass on the FPGA?

common/src/wdt.rs
```diff
@@ -97,5 +97,8 @@ pub fn restart_wdt(soc_ifc: &mut SocIfc) {
 /// * `soc_ifc` - SOC Interface
 ///
 pub fn stop_wdt(soc_ifc: &mut SocIfc) {
+    if !soc_ifc.debug_locked() {
+        return;
+    }
```
jhand2, line 102: Does this mean if we aren't in debug state we can never stop the wdt? I'm not sure this works for RT since it will be in low power state most of the time.

.github/workflows/fpga.yml
rusty1968: the conditional compilation is not passing the right timeout value.

common/src/wdt.rs
rusty1968: if we are debugging we don't want to start so we don't need to stop. it is like the rom.

jhand2: Oh, I missed that it was never getting started. Thanks 😄

mhatrevi (dismissed):

jhand2 (approved):

mhatrevi (approved):

bluegate010 (approved):

PR #962, opened 2023-10-13
Count both active and retired contexts when checking pl context limit

sree-revoori1:
Previously we were only counting active contexts in pl0 or pl1. However, there is still an opportunity for pl1 to cause a denial of service for pl0 by filling DPE's context array with retired contexts, since only active contexts are counted towards the limit.

### DIFF
--- a/runtime/src/invoke_dpe.rs
+++ b/runtime/src/invoke_dpe.rs
@@ -8,6 +8,7 @@ use dpe::{
     commands::{
         CertifyKeyCmd, Command, CommandExecution, DeriveChildCmd, DeriveChildFlags, InitCtxCmd,
     },
+    context::{Context, ContextState},
     response::{Response, ResponseHdr},
     DpeInstance,
 };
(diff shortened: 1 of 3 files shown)


runtime/tests/integration_tests.rs
```diff
@@ -587,7 +590,7 @@ fn test_pauser_privilege_level_dpe_context_thresholds() {
         };
 
         // If we are on the last call to DeriveChild, expect that we get a PL0_ACTIVE_DPE_CONTEXT_THRESHOLD_EXCEEDED error.
-        if i == InvokeDpeCmd::PL0_DPE_ACTIVE_CONTEXT_THRESHOLD - 1 {
+        if i == num_iterations - 1 {
```
jhand2, line 593: Shouldn't this fail when `i == num_iterations` not `num_iterations - 1`?

I think this command is just succeeding because if resp is not an error, the if on line 600 never gets entered.

Would recommend adding an else there which panics

runtime/src/invoke_dpe.rs
```diff
@@ -121,19 +122,25 @@ impl InvokeDpeCmd {
...
+            |c: &Context| non_inactive_context_pred(c) && c.locality == pl0_pauser;
+        let non_inactive_pl0_dpe_context_count = dpe
+            .count_contexts(non_inactive_contexts_in_locality_pred)
             .map_err(|_| CaliptraError::RUNTIME_INTERNAL)?;
-        let active_pl1_dpe_context_count = dpe
-            .count_active_contexts()
+        // the number of non-inactive pl1 dpe contexts is the total number of non-inactive contexts
+        // minus the number of non-inactive pl0 contexts, since a context can only be activated
+        // from pl0 or from pl1.
+        let non_inactive_pl1_dpe_context_count = dpe
```
jhand2, line 134: nit: Maybe replace `non_inactive` with `used`. Can add a comment that `used` includes Active and Retired contexts

runtime/src/invoke_dpe.rs
```diff
@@ -2,12 +2,13 @@
 
 use crate::{CptraDpeTypes, DpeCrypto, DpeEnv, DpePlatform, Drivers};
 use caliptra_common::mailbox_api::{InvokeDpeReq, InvokeDpeResp, MailboxResp, MailboxRespHeader};
-use caliptra_drivers::{CaliptraError, CaliptraResult};
+use caliptra_drivers::{cprintln, CaliptraError, CaliptraResult};
```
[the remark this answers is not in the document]
sree-revoori1: Done

runtime/src/invoke_dpe.rs
```diff
@@ -121,19 +122,25 @@ impl InvokeDpeCmd {
         locality: u32,
         dpe: &DpeInstance,
     ) -> CaliptraResult<()> {
-        let active_pl0_dpe_context_count = dpe
-            .count_active_contexts_in_locality(pl0_pauser)
+        let non_inactive_context_pred = |c: &Context| c.state != ContextState::Inactive;
+        let non_inactive_contexts_in_locality_pred =
+            |c: &Context| non_inactive_context_pred(c) && c.locality == pl0_pauser;
+        let non_inactive_pl0_dpe_context_count = dpe
+            .count_contexts(non_inactive_contexts_in_locality_pred)
```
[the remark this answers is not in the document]
sree-revoori1: Done

runtime/tests/integration_tests.rs
sree-revoori1: No, because the last iteration is num_iterations - 1 (i do see how the naming could be confusing). 

But the reason this makes sense is because before any of the DeriveChild calls in this test even run, we already have 2 non-inactive contexts - context[0] is retired and created during auto initialization, and context[1] is active due to the derive child call in drivers.rs. So then, we iterate [0, 7), and iterations the six iterations 0, 1, 2, 3, 4, 5, create 6 active contexts. Then we have 8 total active contexts. Then when num_iterations - 1 = i = 6, the derive child call would fail since we would go above the threshold.

runtime/src/invoke_dpe.rs
sree-revoori1: Done

runtime/tests/integration_tests.rs
jhand2: Ok makes sense. I was confused about why the test was passing before. I still think it would be good to add an `else { panic!(...); }` after line 605.

sree-revoori1: I did add the else clause which panics though.

runtime/src/invoke_dpe.rs
jhand2: I don't see this updated. Did it get missed in a commit?

sree-revoori1: strange, could you refresh? its updated on my end

sree-revoori1: oh did you mean updating the variable names as well to say used instead of non_inactive?

sree-revoori1: Updated the variable names and error codes as well from non_inactive -> used.

jhand2 (approved):

PR #964, opened 2023-10-13
FMC : Dump RISCV trap record information to the SoC interface.

rusty1968:


runtime/src/main.rs
```diff
@@ -70,16 +66,6 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
...
 
-    {
-        let mut soc_ifc = unsafe { SocIfcReg::new() };
-        let soc_ifc = soc_ifc.regs_mut();
-        let ext_info = soc_ifc.cptra_fw_extended_error_info();
-        ext_info.at(0).write(|_| trap_record.mcause);
-        ext_info.at(1).write(|_| trap_record.mscause);
-        ext_info.at(2).write(|_| trap_record.mepc);
-        ext_info.at(3).write(|_| trap_record.ra);
-    }
```
jhand2, line 81: Should we also call `log_trap_record` here? (And same in FMC exception handler)

runtime/src/main.rs
```diff
@@ -109,15 +95,7 @@ extern "C" fn nmi_handler(trap_record: &TrapRecord) {
...
     );
 
-    {
-        let soc_ifc = soc_ifc.regs_mut();
-        let ext_info = soc_ifc.cptra_fw_extended_error_info();
-        ext_info.at(0).write(|_| trap_record.mcause);
-        ext_info.at(1).write(|_| trap_record.mscause);
-        ext_info.at(2).write(|_| trap_record.mepc);
-        ext_info.at(3).write(|_| trap_record.ra);
-        ext_info.at(4).write(|_| err_interrupt_status);
```
jhand2, line 119: Would be good to keep the interrupt status. Could pass an `Option<u32>` to `log_trap_record` since the exception handler doesn't populate this value

runtime/src/main.rs
rusty1968: done.

runtime/src/main.rs
rusty1968: done

jhand2 (approved):

FerralCoder (approved):

PR #971, opened 2023-10-17
Adding Fake ROM and FMC notes to readme files

nquarton:


fmc/README.md
```diff
@@ -429,6 +429,19 @@ regardless of which reset path caused it to be executed.
...
+Fake FMC is a variation of the FMC intended to be used in the verification/enabling stages of development. The purpose is to greatly reduce the boot time for pre-Si environments by eliminating certain steps from the boot flow.
+
+**Differences from normal FMC:**
+Currently, Fake FMC directly proceeds to runtime without generating the RT Alias Cert. In the future, there will be a static cert and a corresponding private key will be used by runtime to support the DICE challenge flow.
+
+**How to use:**
+- Fake FMC is provided in the release along with the normal collateral.
+- The image builder exposes the argument "fake" that can be used to generate the fake versions
+
+Fake FMC should be used with the Fake ROM. Details can be found in the ROM readme.
```
rusty1968, line 443: Is production ROM + fake FMC a valid configuration?

nquarton: I wouldn't consider it valid. The DICE cert chain would be broken (once we introduce a canned runtime cert). I think in the current state, you would technically still be able to boot. But I don't think we want to guarantee or support that combination unless there is a strong need and use-case.

mhatrevi (approved):

rusty1968 (approved):

PR #977, opened 2023-10-19
Adds caliptra_set_wdt_timeout to C API

nquarton:


wmaroneAMD (approved):

alon-co2:
Are the correct registers used ?

caliptra_set_wdt_timeout()
calls caliptra_wdt_cfg_write(), that writes the parameter into register offsets **GENERIC_AND_FUSE_REG_CPTRA_WDT_CFG_0/1**.

Instead, it seems that caliptra_set_wdt_timeout() should call a new function like caliptra_wdt_timer1_timeout_period_write(), that should write the parameter into register offsets **GENERIC_AND_FUSE_REG_CPTRA_WDT_TIMER1_TIMEOUT_PERIOD_0/1**.

PR #984, opened 2023-10-21
Move the contents of rom/dev/ to rom/

bluegate010:
No need for the dev/ subdirectory.


korran:
> No need for the dev/ subdirectory.

The intention here is to have a 1.0/ directory here someday in addition to dev.

bluegate010:
> > No need for the dev/ subdirectory.
> 
> The intention here is to have a 1.0/ directory here someday in addition to dev/.

Ah ack.

PR #987, opened 2023-10-23
[fix] Return dpe_result for Stash Measurement command

### LINKED ISSUE #978: ROM Stash Measurement does not return dpe_result
The ROM implementation of Stash Measurement does not include the dpe_status in the response as it does in runtime. ROM can not actually call DPE code, but it can just report DPE_STATUS_SUCCESS (0x0)

The runtime implementation is here: runtime\src\[stash_measurement.rs](http://stash_measurement.rs/)

The Stash Measurement command in fw_processor.rs can be updated to something like
```
let mut resp = MailboxResp::StashMeasurement(StashMeasurementResp {
                hdr: MailboxRespHeader::default(),
                dpe_result,
            })
```

Instead of

`let mut resp = MailboxResp::default();`

mhatrevi:
This fix addresses issue# https://github.com/chipsalliance/caliptra-sw/issues/978

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -7,7 +7,7 @@ use std::{
     io::{stdout, ErrorKind, Write},
 };
 
-use caliptra_common::mailbox_api;
+use caliptra_common::mailbox_api::{self, StashMeasurementResp};
 use caliptra_emu_bus::Bus;
 use caliptra_hw_model_types::{
     ErrorInjectionMode, EtrngResponse, RandomEtrngResponses, RandomNibbles, DEFAULT_CPTRA_OBF_KEY,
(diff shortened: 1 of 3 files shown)


nquarton (dismissed):

jhand2 (dismissed):

jhand2 (dismissed):

hw-model/src/lib.rs
```diff
@@ -889,19 +889,22 @@ pub trait HwModel {
         let response = response.ok_or(ModelError::UploadMeasurementResponseError)?;
 
         // Get response as a response header struct
-        let response = mailbox_api::MailboxRespHeader::read_from(response.as_slice())
-            .ok_or(ModelError::UploadMeasurementResponseError)?;
+
+        let response: &StashMeasurementResp =
```
korran, line 893: From https://ip-v8.github.io/rust-ipv8/zerocopy/struct.LayoutVerified.html#method.new

> new verifies that bytes.len() == size_of::<T>() and that bytes is aligned to align_of::<T>(), and constructs a new LayoutVerified. If either of these checks fail, it returns None.

AFAICT, there is no guarantee that Vec<u8> has 4-byte alignment. Probably best to do:

```
let response = StashMeasurementResp::read_from(response.as_slice())
            .ok_or(ModelError::UploadMeasurementResponseError)?;
```

korran (approved):

PR #990, opened 2023-10-23
Adds itrng config, valid pauser config, and generic read/write to C API

nquarton:


wmaroneAMD (dismissed):

jlmahowa-amd (dismissed):

libcaliptra/examples/generic/main.c
```diff
@@ -10,6 +10,10 @@
 #include "caliptra_image.h"
 
 static const uint64_t wdt_timeout = 0xA0000000;         // approximately 5s for 500MHz clock
+static const uint16_t itrng_entropy_low_threshold = 0x1;
+static const uint16_t itrng_entropy_high_threshold = 0xFFFF;
+static const uint16_t itrng_entropy_repetition_count = 0xFFFF;
+static const uint32_t apb_pauser = 0x1;
```
FerralCoder, line 16: I suggest adding comments to all of these constants indicating that these are arbitrarily chosen example values and that integrators must appropriately customize for their SoC. We don't want anyone looking at this and thinking this are acceptable values for all usages.

nquarton: Oh good call. Originally I was going to look for more realistic values then I realized that may be even more misleading. Explicitly calling that out is needed.

FerralCoder (approved):

wmaroneAMD (approved):

korran (approved):
approval for sw-emulator.

PR #992, opened 2023-10-24
Emulation code coverage :  persist and merge coverage bitmaps

rusty1968:


hw-model/src/model_emulated.rs
```diff
@@ -54,6 +54,14 @@ pub struct ModelEmulated {
     trace_fn: Option<Box<InstrTracer<'static>>>,
     ready_for_fw: Rc<Cell<bool>>,
     cpu_enabled: Rc<Cell<bool>>,
+
+    image_tag: u64,
+}
+impl Drop for ModelEmulated {
+    fn drop(&mut self) {
+        let bitmap = self.code_coverage_bitmap();
+        let _ = caliptra_coverage::dump_emu_coverage_to_file(self.image_tag, bitmap);
```
korran, line 63: This will make a mess of /tmp; perhaps we should only do this if an environment variable is set (CPTRA_COVERAGE_PATH=...)?

rusty1968: sure, like we do in tracing.

coverage/src/lib.rs
```diff
@@ -1,12 +1,112 @@
...
+    let mut filename = rng.generate_name();
+    filename.push_str(&'-'.to_string());
+    filename.push_str(&tag.to_string());
+    filename.push_str(".bitvec");
+
+    let path = std::path::Path::new("/tmp").join(filename);
+
+    let file = File::create(path)?;
+    let mut writer = BufWriter::new(file);
+    serde_json::to_writer(&mut writer, &bitmap)?;
```
korran, line 63: json is a bit heavyweight for this, no? Doesn't have to be in this PR, we should consider doing something like

```
use zerocopy::AsBytes;

std::fs::write(path, bitmap.as_raw_slice().as_bytes())?;
```

rusty1968: Sure. We can do raw binary dumps as a follow up.

korran: Reading that back might be a bit tricky; simplest thing would be to use BitVec<u8> instead of BitVec<usize> (the default). Then this becomes:

```
std::fs::write(path, bitmap.as_raw_slice())?;
```

```
let bitmap = BitVec::from_vec(std::fs::read(path)?);
```

coverage/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow.workspace = true
+serde = {workspace = true, features = ["derive"]}
+serde_json.workspace  = true
+random_name_generator = { version = "0.3.2" }
```
korran, line 14: This thing is adding way too many deps to Cargo.lock relative to it's value IMHO. (regex? anstyle?)

Can you use hex::encode(rand::random::<[u8; 16]>()) intsead? We already have deps on `hex` and `rand`

Cargo.lock
```diff
@@ -866,11 +941,45 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
 dependencies = [
  "bitflags 1.3.2",
- "clap_lex",
+ "clap_lex 0.2.4",
  "indexmap",
  "textwrap",
 ]
 
+[[package]]
+name = "clap"
```
korran, line 950: Where is this dep coming from?

Cargo.lock
```diff
@@ -13,6 +13,15 @@ dependencies = [
  "cpufeatures",
```
korran, line 13: This file seems to have more changes than I expect. Can you revert it back and then let Cargo make only the necessary changes?

Cargo.lock
rusty1968: I don't know. I have added a cli app but it has no argument parsing at all.

Cargo.lock
rusty1968: will do.

hw-model/src/model_emulated.rs
rusty1968: Done

coverage/src/lib.rs
rusty1968: I will add it to the follow-up PR.

coverage/Cargo.toml
rusty1968: Done for this PR.

Cargo.lock
rusty1968: There is no longer a dependency on this.

Cargo.lock
rusty1968: Done

coverage/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+pub struct CoverageMap {
+    pub map: HashMap<u64, BitVec>,
+}
+
+impl CoverageMap {
+    pub fn new(paths: Vec<PathBuf>) -> Self {
+        let mut map = HashMap::<u64, BitVec>::default();
+        for path in paths {
+            let new_entry = get_entry_from_path(&path);
+            if map.contains_key(&new_entry.0) {
```
korran, line 26: FYI: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry

coverage/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+        Self { map }
+    }
+}
+pub struct CoverageMapEntry(u64, BitVec);
+pub fn get_entry_from_path(path: &PathBuf) -> CoverageMapEntry {
+    let filename = path.file_name().unwrap().to_str().unwrap();
+    let tag = filename
+        .split('-')
+        .nth(1)
+        .unwrap()
```
korran, line 44: nitpick: panicking where this should probably be an error

hw-model/src/model_emulated.rs
```diff
@@ -55,6 +55,21 @@ pub struct ModelEmulated {
     ready_for_fw: Rc<Cell<bool>>,
     cpu_enabled: Rc<Cell<bool>>,
     trace_path: Option<PathBuf>,
+
+    image_tag: u64,
+}
+impl Drop for ModelEmulated {
+    fn drop(&mut self) {
+        let cov_path =
+            std::env::var(caliptra_coverage::CPTRA_COVERAGE_PATH).unwrap_or_else(|_| "".into());
```
korran, line 64: probably should be .unwrap_or_default()

hw-model/src/model_emulated.rs
```diff
@@ -55,6 +55,21 @@ pub struct ModelEmulated {
...
+        if cov_path.is_empty() {
+            return;
+        }
+
+        let bitmap = self.code_coverage_bitmap();
+        let _ =
```
korran, line 70: nitpick: Should probably print a message when this happens.

hw-model/src/model_emulated.rs
```diff
@@ -125,6 +143,10 @@ impl crate::HwModel for ModelEmulated {
         let soc_to_caliptra_bus = root_bus.soc_to_caliptra_bus();
         let cpu = Cpu::new(BusLogger::new(root_bus), clock);
 
+        let mut hasher = DefaultHasher::new();
```
korran, line 146: nitpick: I think I saw a function above somewhere that does this; maybe call it?

korran: nitpick: probably could just be DefaultHasher::new().write(&params.rom).finish()

korran (approved):

coverage/src/lib.rs
rusty1968: follow up pr.

hw-model/src/model_emulated.rs
rusty1968: follow-up pr.

hw-model/src/model_emulated.rs
rusty1968: follow-up pr.

hw-model/src/model_emulated.rs
rusty1968: follow up pr.

PR #996, opened 2023-10-24
Emu test cov build fix attempt

nquarton:


nquarton:
These changes have been included in the PR they were based on

PR #997, opened 2023-10-24
FPGA ITRNG implementation

jlmahowa-amd:


jlmahowa-amd:
I believe the root cause of the repetition count failures is corruption on the AXI bus. Here is an example where a write to the itrng FIFO happens in the middle of writing zeroes to a fuse. In this scenario I force the fpga realtime model rust code to never write zero data, and the itrng_fifo_din takes the zero data from the fuse writes.
image
Here is another example where the itrng_fifo_din takes increment data from writes to the field entropy registers.
image

My chief suspect is the soc_adapter not complying with the AXI Lite spec.

.github/workflows/fpga.yml
```diff
@@ -16,8 +16,15 @@ on:
       rom-logging:
         default: true
         type: boolean
+      fpga-itrng:
```
korran, line 19: You need to include this parameter in the cache key for the bitstream (line 55)

.github/workflows/nightly-release.yml
```diff
@@ -69,6 +69,7 @@ jobs:
       artifact-suffix: -fpga-realtime-etrng-log
       extra-features: slow_tests
       rom-logging: true
+      fpga-itrng: FALSE
```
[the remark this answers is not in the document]
jlmahowa-amd: My bad. I initially had this passed as a string and forgot to come back to this.

.github/workflows/fpga.yml
```diff
@@ -47,7 +58,7 @@ jobs:
           # root directory.
           echo "rtl_cache_key=$(git rev-parse HEAD:hw-latest/fpga/src)-$(git hash-object hw-latest/fpga/fpga_configuration.tcl)-$(cd hw-latest/caliptra-rtl && git rev-parse HEAD:)-${{ env.CACHE_BUSTER }}" >> $GITHUB_OUTPUT
 
-          echo "kmod_cache_key=fpga-kernel-modules-$(git rev-parse HEAD:hw-latest/fpga/io_module)-$(git rev-parse HEAD:hw-latest/fpga/rom_backdoor)-${{ env.CACHE_BUSTER }}" >> $GITHUB_OUTPUT
+          echo "kmod_cache_key=fpga-kernel-modules-$(git rev-parse HEAD:hw-latest/fpga/io_module)-$(git rev-parse HEAD:hw-latest/fpga/rom_backdoor)-${{ inputs.fpga-itrng }}-${{ env.CACHE_BUSTER }}" >> $GITHUB_OUTPUT
```
korran, line 61: This is the wrong line. kmod_cache_key is for the kernel modules. We want this in rtl_cache_key.

korran (approved):

PR #1015, opened 2023-11-01
[fix] Remove PUF entropy for Xoshiro S* variables

mhatrevi:
This change reverts the non-zeroing of the Xoshiro S* variables. Although it would have been useful to include a bit of PUF state in the PNRG, reading uninitialized memory can lead to a double-bit ECC NMI, preventing Caliptra from booting.


cfi/lib/src/cfi_counter.rs
```diff
@@ -84,7 +84,7 @@ impl CfiCounter {
     /// Reset counter
     #[inline(always)]
     pub fn reset(trng: &mut caliptra_drivers::Trng) {
-        prng().mix_entropy_from_trng(trng);
```
korran, line 87: We still need this part; we need to mix the entropy from all three calls to this function.

rom/dev/test-fw/asm_tests.rs
```diff
@@ -69,31 +68,17 @@ extern "C" {
     fn _zero_mem256(dest: *mut u32, len: usize);
     fn _copy_mem32(dest: *mut u32, src: *const u32, len: usize);
     fn _zero_mem32(dest: *mut u32, len: usize);
```
korran, line 70: Remove _zero_mem32?

korran (approved):

PR #1017, opened 2023-11-01
Add ROM stress test for generating idev CSR.

zhalvorsen:
Boots Calitpra 1000 times with different UDS values to confirm valid CSR generation and expose possible serialization bugs.

### DIFF
--- a/rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
+++ b/rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
@@ -2,16 +2,15 @@
 
 use caliptra_builder::ImageOptions;
 use caliptra_drivers::{IdevidCertAttr, MfgFlags, X509KeyIdAlgo};
-use caliptra_hw_model::{Fuses, HwModel};
+use caliptra_hw_model::{DefaultHwModel, Fuses, HwModel};
+use caliptra_image_types::ImageBundle;
+use openssl::{rand::rand_bytes, x509::X509Req};
 use std::io::Write;
 
 use crate::helpers;
 
-#[test]
-fn test_generate_csr() {
+fn generate_csr(hw: &mut DefaultHwModel, image_bundle: &ImageBundle) -> Csrs {
     let mut output = vec![];
-    let (mut hw, image_bundle) =
-        helpers::build_hw_model_and_image_bundle(Fuses::default(), ImageOptions::default());
 
     // Set gen_idev_id_csr to generate CSR.
     let flags = MfgFlags::GENERATE_IDEVID_CSR;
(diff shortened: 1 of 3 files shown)


rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
```diff
@@ -54,3 +67,48 @@ fn test_idev_subj_key_id_algo() {
...
+        let (mut hw, image_bundle) =
+            helpers::build_hw_model_and_image_bundle(fuses, ImageOptions::default());
+
+        let Csrs {
+            uploaded,
+            downloaded: _,
+        } = generate_csr(&mut hw, &image_bundle);
+
+        // Ensure CSR is valid X.509
+        X509Req::from_der(&uploaded).unwrap_or_else(|_| {
```
jhand2, line 102: Can we also check that the signature verifies with the public key? I think it should be something like 

```rust
let pub = req.public_key().unwrap();
assert!(req.verify(pub).unwrap());
```

rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
```diff
@@ -54,3 +67,48 @@ fn test_idev_subj_key_id_algo() {
...
+    }
+}
+
+#[test]
+#[cfg_attr(not(feature = "slow_tests"), ignore)]
+fn test_generate_csr_stress() {
+    const NUM_TESTS: usize = 1000;
+
+    for _ in 0..NUM_TESTS {
+        let fuses = fuses_with_random_uds();
```
jhand2, line 92: It would be nice to make sure the test works on a single iteration to avoid regressions in the test code. And then the slow test would just run that `NUM_TESTS` times.

Could even just conditionally choose NUM_TESTS based on the config value.

rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
zhalvorsen: Thanks! I added the check.

rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
zhalvorsen: Good idea. Updated.

rom/dev/tests/rom_integration_tests/test_idevid_derivation.rs
```diff
@@ -54,3 +67,56 @@ fn test_idev_subj_key_id_algo() {
...
+        ..Default::default()
+    }
+}
+
+#[test]
+fn test_generate_csr_stress() {
+    #[cfg(feature = "slow_tests")]
+    const NUM_TESTS: usize = 1000;
+    #[cfg(not(feature = "slow_tests"))]
+    const NUM_TESTS: usize = 1;
```
jhand2, line 91: nit: Should be able to do something like

```rust
let num_tests: usize = if cfg!(feature = "slow_tests") {
    1000
} else {
    1
}
```

zhalvorsen: That's a lot better

mhatrevi (approved):

PR #1019, opened 2023-11-01
Check extended error info in test_rom_wdt_timeout

sree-revoori1:


rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs
```diff
@@ -107,4 +107,15 @@ fn test_rom_wdt_timeout() {
     .unwrap();
 
     hw.step_until(|m| m.soc_ifc().cptra_fw_error_fatal().read() == WDT_EXPIRED);
+
+    // no mcause if wdt times out
+    assert_eq!(hw.soc_ifc().cptra_fw_extended_error_info().at(0).read(), 0);
+    // no mscause if wdt times out
+    assert_eq!(hw.soc_ifc().cptra_fw_extended_error_info().at(1).read(), 0);
+    // mepc is a memory address so won't be 0
+    assert_ne!(hw.soc_ifc().cptra_fw_extended_error_info().at(2).read(), 0);
```
korran, line 116: Can you print out the address somewhere so folks can manually verify it looks sane?

rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs
```diff
@@ -107,4 +107,15 @@ fn test_rom_wdt_timeout() {
...
+    // return address won't be 0
+    assert_ne!(hw.soc_ifc().cptra_fw_extended_error_info().at(3).read(), 0);
+    // no error_internal_intr_r if wdt times out
+    assert_eq!(hw.soc_ifc().cptra_fw_extended_error_info().at(4).read(), 0);
```
korran, line 120: That's weird, I would expect on of these bits to be set:

https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.intr_block_rf.error_internal_intr_r#error_wdt_timer2_timeout_sts

korran: Ah, and it is on the FPGA:

```
thread 'test_wdt_activation_and_stoppage::test_rom_wdt_timeout' panicked at 'assertion failed: `(left == right)`
  left: `64`,
 right: `0`', rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs:120:5
```

So this is a problem with the emulator.

sree-revoori1: Removed this last line for now to make CI pass. I looked around in the sw-emulator but not sure what's going wrong.

rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs
sree-revoori1: I added some code printing it out, but if the watchdog timer expires, won't the nmi handler print the error codes to UART already?

rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs
jhand2: Can you add that assert back? Before we merge this I want to make sure all the right bits are getting set (since until they do, there is still a test gap 😁 )

jhand2: I think to fix this, in [these blocks](https://github.com/chipsalliance/caliptra-sw/blob/3df8b615e95ee5372c8df8e1b742056e0c014394/sw-emulator/lib/periph/src/soc_reg.rs#L1027-L1064), we want to set the [error_wdt_timer*_timeout_sts](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.intr_block_rf.error_internal_intr_r#error_wdt_timer2_timeout_sts.desc) bits in the `error_internal_intr_r` register.

Today the whole interrupt register block is just an array of `u32`s: https://github.com/chipsalliance/caliptra-sw/blob/3df8b615e95ee5372c8df8e1b742056e0c014394/sw-emulator/lib/periph/src/soc_reg.rs#L556C15-L556C15.

So the cheap way to write these bits would be to find the right index and set the bit in the corresponding if.

jhand2: There is a nuance in implementing this register that the bits are "write-one-to-clear" (`woclr`). So if we implement it in the cheap way I suggested above, that wouldn't actually work.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -551,10 +564,14 @@ struct SocRegistersImpl {
...
 
-    /// INTERNAL_FW_UPDATE_RESET_WAIT_CYCLES Register
+    /// INTR_BLOCK_RF Register
     #[register_array(offset = 0x0800)]
     intr_block_rf: [u32; INTR_BLOCK_SIZE / 4],
 
+    /// ERROR_INTERNAL_INTR_R Register
+    #[register(offset = 0x0814)]
+    error_internal_intr_r: ReadWriteRegister<u32, ErrorIntrT::Register>,
+
```
jhand2, line 574: These registers overlap, so I'm not sure this will work as expected. I think instead, I think you could define something like

```rust
const INTR_BLOCK_ERROR_INTERNAL_INTR_R_OFFSET = 0x14;

// And then when you write to it, do something like
self.intr_block_rf[INTR_BLOCK_ERROR_INTERNAL_INTR_R_OFFSET] = <register_value>;
```

rom/dev/tests/rom_integration_tests/test_wdt_activation_and_stoppage.rs
sree-revoori1: Done

sw-emulator/lib/periph/src/soc_reg.rs
sree-revoori1: Discussed and took a different approach of adding each register individually and removing intr_block_rf.

jhand2 (approved):
Great job!

mhatrevi (approved):

PR #1021, opened 2023-11-01
Enable FPGA Clock Gating

jlmahowa-amd:


korran:
To get the policy check to pass you'll have to rebase or merge from main.

korran (approved):

korran (changes requested):
Luke tells me this is not ready to merge yet

jlmahowa-amd:
Last trouble is that the clock gate conversion only works when building using the GUI. Should be close.

jlmahowa-amd:
Vivado took me for a ride, and the patch was fine as is. I did make a small optimization to only run the clock gate conversion pass when we have it enabled.

korran (approved):

PR #1022, opened 2023-11-01
[feat] Add additional CFI checks to firmware image library

### LINKED ISSUE #982: Fault injection hardening for verifier code
3 asks:
1. Ensure no constant time window between externally-visible signal and critical branch.
2. Add/confirm init-to-fail logic for boolean checks
3. Use multi-bit booleans

Details are captured in https://github.com/chipsalliance/caliptra-sw/blob/main/rom/dev/doc/threat-model.md#mitigation-2

mhatrevi:
This change addresses item 1 from the issue# https://github.com/chipsalliance/caliptra-sw/issues/982

### DIFF
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -17,6 +17,12 @@ pub const ROM_WITH_UART: FwId = FwId {
     features: &["emu"],
 };
 
+pub const ROM_WITH_UART_NO_CFI: FwId = FwId {
+    crate_name: "caliptra-rom",
+    bin_name: "caliptra-rom",
+    features: &["emu", "no-cfi"],
+};
+
 pub const ROM_FAKE_WITH_UART: FwId = FwId {
     crate_name: "caliptra-rom",
     bin_name: "caliptra-rom",
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -17,12 +17,6 @@ pub const ROM_WITH_UART: FwId = FwId {
     features: &["emu"],
 };
 
-pub const ROM_WITH_UART_NO_CFI: FwId = FwId {
-    crate_name: "caliptra-rom",
-    bin_name: "caliptra-rom",
-    features: &["emu", "no-cfi"],
-};
-
 pub const ROM_FAKE_WITH_UART: FwId = FwId {
     crate_name: "caliptra-rom",
     bin_name: "caliptra-rom",
--- a/fmc/tests/fmc_integration_tests/test_hand_off.rs
+++ b/fmc/tests/fmc_integration_tests/test_hand_off.rs
@@ -4,7 +4,7 @@ use caliptra_hw_model::{BootParams, HwModel, InitParams};
 
 #[test]
 fn test_hand_off() {
-    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART).unwrap();
+    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART_NO_CFI).unwrap();
 
     let image = caliptra_builder::build_and_sign_image(
         &firmware::FMC_WITH_UART,
--- a/fmc/tests/fmc_integration_tests/test_hand_off.rs
+++ b/fmc/tests/fmc_integration_tests/test_hand_off.rs
@@ -4,7 +4,7 @@ use caliptra_hw_model::{BootParams, HwModel, InitParams};
 
 #[test]
 fn test_hand_off() {
-    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART_NO_CFI).unwrap();
+    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART).unwrap();
 
     let image = caliptra_builder::build_and_sign_image(
         &firmware::FMC_WITH_UART,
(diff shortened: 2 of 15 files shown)


image/verify/src/verifier.rs
```diff
@@ -216,24 +221,32 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
 
         match key_idx {
             0..=SECOND_LAST_KEY_IDX => {
+                cfi_assert_le(cfi_launder(key_idx), SECOND_LAST_KEY_IDX);
                 let key = VendorPubKeyRevocation::from_bits_truncate(0x01u32 << key_idx);
-                if revocation.contains(key) {
+                if cfi_launder(revocation.contains(key)) {
                     Err(CaliptraError::IMAGE_VERIFIER_ERR_VENDOR_ECC_PUB_KEY_REVOKED)?;
+                } else {
+                    cfi_assert!(!revocation.contains(key));
```
korran, line 229: Need to double check the assembly that this re-does the full contains operations instead of just recomparing the result (which is what I suspect).

mhatrevi: Yeah, revocation should be laundered. Hardened it even further by laundering the key as well.

image

fmc/tests/fmc_integration_tests/test_hand_off.rs
```diff
@@ -4,7 +4,7 @@ use caliptra_hw_model::{BootParams, HwModel, InitParams};
 
 #[test]
 fn test_hand_off() {
-    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART).unwrap();
+    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART_NO_CFI).unwrap();
```
korran, line 7: revert this change

test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -541,7 +541,7 @@ fn fips_cmd_test_rt() {
         .set_debug_locked(false)
         .set_device_lifecycle(DeviceLifecycle::Unprovisioned);
 
-    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART).unwrap();
+    let rom = caliptra_builder::build_firmware_rom(&firmware::ROM_WITH_UART_NO_CFI).unwrap();
```
korran, line 544: revert this?

builder/src/firmware.rs
```diff
@@ -17,6 +17,12 @@ pub const ROM_WITH_UART: FwId = FwId {
     features: &["emu"],
 };
 
+pub const ROM_WITH_UART_NO_CFI: FwId = FwId {
```
korran, line 20: Either remove this or add it to REGISTERED_FW

image/verify/src/verifier.rs
```diff
@@ -216,24 +226,32 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
-        if reason == ResetReason::UpdateReset {
+        if cfi_launder(reason) == ResetReason::UpdateReset {
             let expected = self.env.vendor_ecc_pub_key_idx_dv();
-            if expected != key_idx {
+            if cfi_launder(expected) != key_idx {
                 Err(
                     CaliptraError::IMAGE_VERIFIER_ERR_UPDATE_RESET_VENDOR_ECC_PUB_KEY_IDX_MISMATCH,
                 )?;
+            } else {
+                cfi_assert_eq(expected, key_idx);
```
korran, line 251: We should be going back to the source here:

```
cfi_assert_eq(self.env.vendor_ecc_pub_key_idx_dv(), key_idx);
```

image/verify/src/verifier.rs
```diff
@@ -253,23 +271,31 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
-        if reason == ResetReason::UpdateReset {
+        if cfi_launder(reason) == ResetReason::UpdateReset {
             let expected = self.env.vendor_lms_pub_key_idx_dv();
-            if expected != key_idx {
+            if cfi_launder(expected) != key_idx {
                 Err(
                     CaliptraError::IMAGE_VERIFIER_ERR_UPDATE_RESET_VENDOR_LMS_PUB_KEY_IDX_MISMATCH,
                 )?;
+            } else {
+                cfi_assert_eq(expected, key_idx);
```
korran, line 295: Probably worth going back to the source here:

```
cfi_assert_eq(self.env.vendor_lms_pub_key_idx_dv(), key_idx);
```

image/verify/src/verifier.rs
```diff
@@ -760,15 +835,18 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
             0_u32
         } else {
+            cfi_assert!(!anti_rollback_disable);
             fuse_svn
         }
     }
 }
 
 #[cfg(all(test, target_family = "unix"))]
+#[cfg(feature = "no-cfi")]
```
korran, line 849: This is dangerous. With this change, `cargo test` no longer runs these tests.

korran: I'm not sure we have a lot of options to fix this. The first one that comes to mind is to define CFI_STATE_ORG and cfi_panic_handler

korran (approved):

jhand2 (approved):

PR #1023, opened 2023-11-02
Use unrolled CFI assertions (cfi_assert_eq_N_words())

korran:
This function should be used for double-checking important values like signatures and hashes.


image/verify/src/verifier.rs
```diff
@@ -520,10 +522,11 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
                             self.env.set_fw_extended_error(err.into());
                             CaliptraError::IMAGE_VERIFIER_ERR_VENDOR_LMS_VERIFY_FAILURE
                         })?;
-                if cfi_launder(candidate_key) != HashValue::from(lms_pub_key.digest) {
+                let pub_key_digest = HashValue::from(lms_pub_key.digest);
+                if candidate_key != pub_key_digest {
```
mhatrevi, line 526: candidate_key should be laundered. Most likely the else branch is optimized out.

image/verify/src/verifier.rs
```diff
@@ -548,10 +551,11 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
                 CaliptraError::IMAGE_VERIFIER_ERR_OWNER_LMS_VERIFY_FAILURE
             })?;
 
-        if cfi_launder(candidate_key) != HashValue::from(lms_pub_key.digest) {
+        let pub_key_digest = HashValue::from(lms_pub_key.digest);
+        if candidate_key != pub_key_digest {
```
mhatrevi, line 555: See comment above.

image/verify/src/verifier.rs
```diff
@@ -358,18 +358,20 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
 
         let fuses_digest = self.env.owner_pub_key_digest_fuses();
 
-        if cfi_launder(fuses_digest) != ZERO_DIGEST && cfi_launder(fuses_digest) != actual {
+        if fuses_digest == ZERO_DIGEST {
```
mhatrevi, line 361: fuse_digest must be laundered.

mhatrevi (changes requested):

image/verify/src/verifier.rs
korran: It can't optimize out the else branch; it's calling assembly code which is opaque to the compiler.

cfi_launder shouldn't make a difference here; the compiler still needs to do this comparison (it wasn't done anywhere else in this function).

mhatrevi (approved):

PR #1026, opened 2023-11-02
FIPS Functional Test Suite structure and initial tests

nquarton:


test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -380,209 +378,6 @@ fn smoke_test() {
     // TODO: Validate the rest of the fmc_alias certificate fields
 }
 
-fn test_fips<T: HwModel>(hw: &mut T, fmc_version: u32, app_version: u32) {
```
nquarton, line 383: I've relocated these tests to the FIPS test suite. If there is a concern with excluding these from smoke_test, we can try making them public and importing them so they run here as well. That will mean they are running twice in most cases, but it's an option.

jlmahowa-amd (dismissed):

test/tests/caliptra_integration_tests/smoke_test.rs
jlmahowa-amd: The location in smoke_test.rs was just to be with the other integration tests. The moved location is better.

rusty1968 (dismissed):

JohnTraverAmd (approved):

PR #1028, opened 2023-11-02
[feat] Add CFI checks to the driver library

mhatrevi:


drivers/src/ecc384.rs
```diff
@@ -254,7 +255,7 @@ impl Ecc384 {
         };
 
         // Pairwise consistency check.
-        if perform_pct {
+        if cfi_launder(priv_key_usage.ecc_private_key()) {
```
korran, line 258: I think this branch is too dangerous. Can we just remove the branch and always validate? Why would anyone want to generate a private key that can't be used as a private key?

drivers/src/ecc384.rs
```diff
@@ -336,12 +339,13 @@ impl Ecc384 {
     ) -> CaliptraResult<Ecc384Signature> {
         let mut sig_result = self.sign_internal(priv_key, data, trng);
         let sig = okmutref(&mut sig_result)?;
-        match self.verify(pub_key, data, sig)? {
+        let result = self.verify(pub_key, data, sig)?;
```
korran, line 342: Add `// TODO(kor): unrolled CFI comparison of verify result`

korran (commented):
Should we be wrapping these functions with cfi_impl_fn?

drivers/src/ecc384.rs
FerralCoder: I agree, I can't think of any valid reason why this PCT should ever be skipped.

mhatrevi: Discussed offline; we'll block creating a key-pair if [ecc_pkey_dest_valid](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.ecc_reg.ecc_kv_wr_pkey_ctrl#ecc_pkey_dest_valid.desc) is not set

mhatrevi:
> Should we be wrapping these functions with cfi_impl_fn?

Added.

drivers/src/sha256.rs
```diff
@@ -33,6 +35,8 @@ pub trait Sha256Alg {
 
     fn digest_init(&mut self) -> CaliptraResult<Self::DigestOp<'_>>;
     fn digest(&mut self, buf: &[u8]) -> CaliptraResult<Array4x8>;
+    #[cfg(not(feature = "no-cfi"))]
+    fn __cfi_digest(&mut self, buf: &[u8]) -> CaliptraResult<Array4x8>;
```
korran, line 39: Hmm, this is pretty messy. I wonder if we could use nested functions instead.

korran: As is, we need to update other users of this trait (in the fuzz tests)

korran: Here's a fairly nice solution:

https://github.com/chipsalliance/caliptra-sw/pull/1060

Probably the sweet spot of complexity vs code size is the second commit in that chain. The third makes it 3x more complex just to save 36 bytes (making the entire PR increase the size by just 4 bytes).

mhatrevi: Discussed offline; we'll decorate the Crypto::sha256_digest function with CFI instead of the Sha256::digest function.

korran (dismissed):

korran (approved):

jhand2 (approved):

FerralCoder (approved):

PR #1046, opened 2023-11-06
Add test triggering cpu fault and check extended error info

### LINKED ISSUE #1035: ROM test: Add test for CPU fault, and that extended error info is populated correctly
Test case could modify the ROM binary to do something that will obviously crash (replace whatever is at the `rom_entry` symbol with instructions that write to an invalid address?), and verify that the error-info is populated correctly, including that the mepc was set to the correct address.

sree-revoori1:
To do this, we write an illegal instruction at the rom_entry symbol in the rom binary.

Fixes #1035

### DIFF
--- a/rom/dev/tests/rom_integration_tests/test_cpu_fault.rs
+++ b/rom/dev/tests/rom_integration_tests/test_cpu_fault.rs
@@ -1,0 +1,59 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_builder::firmware;
+use caliptra_hw_model::{BootParams, HwModel, InitParams};
+use elf::{endian::LittleEndian, ElfBytes};
+
+#[test]
+fn test_cpu_fault() {
+    const GLOBAL_EXCEPTION: u32 = 0x01050002;
+
+    let rom_fwid = firmware::rom_from_env();
+
+    let elf_bytes = caliptra_builder::build_firmware_elf(rom_fwid).unwrap();
+    let mut rom = caliptra_builder::elf2rom(&elf_bytes).unwrap();
+    let elf = ElfBytes::<LittleEndian>::minimal_parse(&elf_bytes).unwrap();
+    let symbol_table = elf.symbol_table().unwrap().unwrap().0;
+    let string_table = elf.symbol_table().unwrap().unwrap().1;
+    let rom_entry_offset = symbol_table
+        .iter()
+        .find(|symbol| string_table.get(symbol.st_name as usize).unwrap() == "rom_entry")
+        .unwrap()
+        .st_value as usize;
+    println!("rom_entry_offset is {}", rom_entry_offset);
+
+    // Write an instruction that causes a cpu fault to the rom_entry offset
+    let illegal_instruction = [0xFF, 0xFF, 0xFF, 0xFF];
+    rom[rom_entry_offset..rom_entry_offset + illegal_instruction.len()]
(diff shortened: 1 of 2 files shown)


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -286,6 +286,12 @@ impl FirmwareProcessor {
                         resp.populate_chksum()?;
                         txn.send_response(resp.as_bytes())?;
                     }
+                    #[cfg(feature = "test_only_commands")]
+                    CommandId::TEST_ONLY_TRIGGER_CPU_FAULT => {
+                        // the print is necessary to ensure the compiler doesn't optimize out the unsafe code below
+                        cprintln!("Unsafe write to null mutable pointer");
```
jhand2, line 292: I still don't really understand why this is getting optimized out. Not sure if @korran would have any insight here.

Inline asm should ensure it doesn't get optimized out though.

korran: As I mentioned offline, I think we should do something like what I proposed at https://github.com/chipsalliance/caliptra-sw/issues/1035

sree-revoori1: Done

rom/dev/tests/rom_integration_tests/test_cpu_fault.rs
```diff
@@ -0,0 +1,59 @@
...
+use caliptra_builder::firmware;
+use caliptra_hw_model::{BootParams, HwModel, InitParams};
+use elf::{endian::LittleEndian, ElfBytes};
+
+#[test]
+fn test_cpu_fault() {
+    const GLOBAL_EXCEPTION: u32 = 0x01050002;
+
+    let rom_fwid = firmware::rom_from_env();
+    let mut rom = caliptra_builder::build_firmware_rom(rom_fwid).unwrap();
```
jhand2, line 12: Instead of calling build_firmware rom, you can call `caliptra_builder::elf2rom(&elf)` after you call `build_firmware_elf`. It'll probably make the test run a little faster.

rom/dev/tests/rom_integration_tests/test_cpu_fault.rs
```diff
@@ -0,0 +1,59 @@
...
+        "ROM Global Exception mcause=0x{:08X} mscause=0x{:08X} mepc=0x{:08X} ra=0x{:08X}",
+        mcause, mscause, mepc, ra,
+    );
+
+    // mcause must be illegal instruction
+    assert_eq!(mcause, 0x2);
+    // no mscause
+    assert_eq!(mscause, 0);
+    // mepc is a memory address so won't be 0
+    assert_ne!(mepc, 0);
```
jhand2, line 56: I think `mepc` is the value of the program counter at the faulting instruction. So I would expect to be able to do `assert_eq!(mepc, rom_entry_offset)`

sree-revoori1: Done

rom/dev/tests/rom_integration_tests/test_cpu_fault.rs
sree-revoori1: Done

jhand2 (approved):

mhatrevi (approved):

PR #1050, opened 2023-11-07
Add POPULATE_IDEV_CERT mailbox command to RT

### LINKED ISSUE #1048: Add an RT mailbox command to populate the IDevID Certificate
The Caliptra IDevID cert is generated by an external CA and not Caliptra itself. On some systems, the IDevID may be persisted somewhere outside Caliptra, like in flash alongside SoC firmware.

We should add a mailbox command `POPULATE_IDEV_CERT` which allows the SoC to provide a DER-encoded IDevID certificate on ever boot. Once the certificate has been populated, the DPE `GetCertificateChain` command will also include the IDevID cert. If this command is not called, the chain returned by `GetCertificateChain` will start at LDevID (as it does today).

sree-revoori1:
This command will allow the SoC to add the idev id cert to the start of the certificate chain.

Fixes #1048

### DIFF
--- a/runtime/src/info.rs
+++ b/runtime/src/info.rs
@@ -1,8 +1,10 @@
 // Licensed under the Apache-2.0 license
 
-use crate::{handoff::RtHandoff, Drivers};
+use crate::{handoff::RtHandoff, Drivers, MAX_CERT_CHAIN_SIZE, PL0_PAUSER_FLAG};
+use arrayvec::ArrayVec;
 use caliptra_common::mailbox_api::{
     FwInfoResp, GetIdevCertReq, GetIdevCertResp, GetIdevInfoResp, MailboxResp, MailboxRespHeader,
+    PopulateIdevCertReq,
 };
 use caliptra_drivers::{CaliptraError, CaliptraResult};
 use caliptra_x509::{Ecdsa384CertBuilder, Ecdsa384Signature};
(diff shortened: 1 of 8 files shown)


runtime/src/info.rs
```diff
@@ -80,3 +81,42 @@ impl IDevIdCertCmd {
...
+
+            // Insert IDevId certificate at the start of the certificate chain.
+            for i in 0..cert_size {
+                if i > drivers.cert_chain.len() {
+                    return Err(CaliptraError::RUNTIME_IDEV_CERT_POPULATION_FAILED);
+                }
+
+                drivers
+                    .cert_chain
+                    .try_insert(
```
jhand2, line 108: I believe every `try_insert` call will result in copying every element in the ArrayVec.

Another option for implementing this would to do something like:

```rust
let tmp_chain: ArrayVec::<u8, MAX_CERT_CHAIN_SIZE>::new();
tmp_chain.try_extend_from_slice(&cmd.cert.get(..cmd.cert.len());
tmp_chain.try_extend_from_slice(drivers.cert_chain.as_slice());
drivers.cert_chain = tmp_chain;
```

Ends up being less code and also more efficient.

runtime/tests/runtime_integration_tests/integration_tests.rs
```diff
@@ -487,6 +487,147 @@ fn test_invoke_dpe_get_certificate_chain_cmd() {
...
+        .unwrap()
+        .expect("We should have received a response");
+
+    let mut cert_chain_with_idev_cert = [0u8; 4096];
+    let cert_chain_len_with_idev_cert =
+        get_full_cert_chain(&mut model, &mut cert_chain_with_idev_cert);
+
+    // read idev cert from prefix of cert chain and parse it as X509
+    let idev_len = cert_chain_len_with_idev_cert - cert_chain_len_without_idev_cert;
+    X509::from_der(&cert_chain_with_idev_cert[..idev_len]).unwrap();
```
jhand2, line 628: nit: Could we also test that all the other certs are not corrupted? (i.e. LDevID -> RT certs can still be parsed)

runtime/src/info.rs
sree-revoori1: Thanks, this is much nicer. Done.

runtime/tests/runtime_integration_tests/integration_tests.rs
sree-revoori1: It would be difficult to parse the rest of the cert chain since we don't know the exact lengths of each cert, so instead of parsing, I checked that cert_chain_len_with_idev_cert[idev_len..] == cert_chain_without_idev_cert.

If you think this doesn't suffice, I can add parsing of the remaining 3 certs, but I think this would require me to add a backtracking algorithm that runs X509::from_der on prefixes of the cert chain.

sree-revoori1: Actually I don't think backtracking would even be needed. I think X509::from_der will parse as much of the prefix as possible by default, and ignore the rest so it shouldn't be too difficult to parse the entire cert chain. I will add it.

sree-revoori1: Added a function that recursively parses the entire cert chain.

runtime/tests/runtime_integration_tests/integration_tests.rs
```diff
@@ -487,6 +487,169 @@ fn test_invoke_dpe_get_certificate_chain_cmd() {
...
+    let cert_chunk_2 =
+        GetCertificateChainResp::read_from(&resp_hdr.data[..resp_hdr.data_size as usize]).unwrap();
+    out[cert_chunk_1.certificate_size as usize
+        ..cert_chunk_1.certificate_size as usize + cert_chunk_2.certificate_size as usize]
+        .copy_from_slice(&cert_chunk_2.certificate_chain[..cert_chunk_2.certificate_size as usize]);
+
+    cert_chunk_1.certificate_size as usize + cert_chunk_2.certificate_size as usize
+}
+
+fn parse_cert_chain(cert_chain: &[u8], cert_chain_size: usize) {
```
jhand2, line 575: I don't think this really needs to be recursive.

```rust
let mut i = 0;
while i < cert_chain.len() {
    let curr_cert = X509::from_der(cert_chain).unwrap();
    i += curr_cert.to_der().unwrap().len();
}
```

jhand2, line 575: Can you add a comment explaining that this function will panic if any of the chucks is not a valid x.509 cert

sree-revoori1: Done

sree-revoori1: Done

jhand2 (approved):

PR #1060, opened 2023-11-10
CFI derive: Use closures for inner function

korran:
This makes it possible to use this derive when implementing traits.

The first commit is super simple, but unfortunately the compiler has a hard time optimizing it, leading to significant bloat. 

The second commit is arguably simpler than head, and only bloats us up by 40 bytes. 

The third commit renames self so it can be passed to the closure, and only bloats (vs main) 4 additional bytes. Unfortunately, it's quite a bit more complex.


korran:
I like this PR, but we don't need it right now (sha256 was the only algorithm that used a trait, and since it's only used by LMS in tight inner loops, the mitigation had too much performance overhead).

PR #1066, opened 2023-11-13
Document the FW_INFO mailbox command

jhand2:


runtime/README.md
```diff
@@ -118,6 +118,26 @@ Table: `CALIPTRA_FW_LOAD` output arguments
 | chksum      | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips_status | u32      | Indicates if the command is FIPS approved or an error
 
+### FW\_INFO
```
korran, line 121: Is there a function for returning more details firmware versions? The ROM can provide its git commit and image hash...

jhand2: This just documents what exists today. If we want to add more things to this response that's fine (in a follow-up PR)

FerralCoder (approved):

jhand2:
This was done in https://github.com/chipsalliance/caliptra-sw/commit/ef4cf1efb7ad3c3ea89e4a832a40ea2081aa318d#diff-5891ec7eadfb449ec343281ee04e99dded25a8121dcdd9aedae243e47a811bab

PR #1074, opened 2023-11-14
Feature/sw emulator glitching

ArthurHeymans:
Add software glitching, by randomly skipping certain instructions.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -1206,6 +1206,22 @@ mod tests {
     }
 
     #[test]
+    fn test_glitching() {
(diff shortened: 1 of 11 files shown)


hw-model/src/lib.rs
```diff
@@ -1015,6 +1015,22 @@ mod tests {
...
+                rom: &gen_image_hi(),
+                ..Default::default()
+            },
+            ..Default::default()
+        })
+        .unwrap();
+
+        model.enable_glitching();
+
+        model.step_until_output("hii").unwrap();
```
ArthurHeymans, line 1031: this may just fail because of the glitching...

PR #1078, opened 2023-11-15
Allow tests be be executed against no-uart (prod) ROMs

korran:
I recommend reviewing each commit independently.

### DIFF
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -5,6 +5,16 @@
 
 use crate::FwId;
 
+pub fn rom_from_env() -> &'static FwId<'static> {
+    match std::env::var("CPTRA_ROM_TYPE").as_ref().map(|s| s.as_str()) {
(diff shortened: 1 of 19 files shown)


builder/src/firmware.rs
```diff
@@ -5,6 +5,16 @@
 
 use crate::FwId;
 
+pub fn rom() -> &'static FwId<'static> {
```
jhand2, line 8: nit: maybe call this something like `rom_from_env`? Makes it more clear in the test code how the ROM image is being configured.

mhatrevi (dismissed):

FerralCoder (dismissed):

builder/src/firmware.rs
korran: Done.

mhatrevi (approved):

jhand2 (approved):

PR #1084, opened 2023-11-16
[fix] Skip nightly release if calitpra-rtl submodule is stale

korran:
Last night, the release workflow made a release, even though the caliptra-rtl submodule was not up-to-date:

https://github.com/chipsalliance/caliptra-sw/actions/runs/6875915531

The syntax !"$CHECK_RELEASE_SYNC" does not work as intended; we should use -z instead.


nquarton (approved):

FerralCoder (approved):

leongross (approved):
Lgtm, let's get this in.

PR #1095, opened 2023-11-17
Extend the measurement into PCR31 in RT stash_measurement

sree-revoori1:


runtime/src/stash_measurement.rs
```diff
@@ -4,20 +4,27 @@ use crate::{dpe_crypto::DpeCrypto, CptraDpeTypes, DpePlatform, Drivers};
...
 pub struct StashMeasurementCmd;
 impl StashMeasurementCmd {
     pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
         if let Some(cmd) = StashMeasurementReq::read_from(cmd_args) {
+            // Extend the measurement into PCR31
+            drivers.pcr_bank.extend_pcr(
+                PCR_ID_STASH_MEASUREMENT,
+                &mut drivers.sha384,
+                cmd.measurement.as_bytes(),
+            )?;
```
jhand2, line 26: Ideally we would wait to extend the PCR until we're sure the rest of the command will succeed. This is still a little annoying because if pcr_extend fails, we also probably want to rollback the DeriveChild call which might be difficult to do in practice. But the PCR extend is for sure irreversible, so we should do it last.

sree-revoori1: The issue then is that we get borrow checker issues since we "cannot borrow `drivers.sha384` as mutable more than once at a time". I will look into refactoring this logic a bit

sree-revoori1: Fixed by limiting the scope of the first borrow of drivers.sha384.

runtime/src/stash_measurement.rs
```diff
@@ -4,65 +4,77 @@ use crate::{dpe_crypto::DpeCrypto, CptraDpeTypes, DpePlatform, Drivers};
...
-            let derive_child_resp = DeriveChildCmd {
-                handle: ContextHandle::default(),
-                data: cmd.measurement,
-                flags: DeriveChildFlags::MAKE_DEFAULT
-                    | DeriveChildFlags::CHANGE_LOCALITY
-                    | DeriveChildFlags::INPUT_ALLOW_CA
-                    | DeriveChildFlags::INPUT_ALLOW_X509,
-                tci_type: u32::from_be_bytes(cmd.metadata),
-                target_locality: locality,
+            if let DpeErrorCode::NoError = dpe_result {
```
jhand2, line 66: Today I learned you can do this to check enum variants :)

runtime/src/stash_measurement.rs
```diff
@@ -4,65 +4,77 @@ use crate::{dpe_crypto::DpeCrypto, CptraDpeTypes, DpePlatform, Drivers};
...
-            );
-            let mut env = DpeEnv::<CptraDpeTypes> {
-                crypto,
-                platform: DpePlatform::new(
-                    pdata.manifest1.header.pl0_pauser,
-                    hashed_rt_pub_key,
-                    &mut drivers.cert_chain,
-                ),
+            let dpe_result = {
+                let hashed_rt_pub_key = drivers.compute_rt_alias_sn()?;
```
jhand2, line 22: Since we're using the `?` operator here, I would expect `dpe_result` to be a `CaliptraResult`, but it seems like we don't have to return `Ok(error_code)` at the end of this block. So i'm a little confused about what happens to any errors from this funciton.

sree-revoori1: I tested this out by making compute_rt_alias_sn return an error, and it is working as expected - the stash measurement mailbox command exits the compute_rt_alias_sn and returns the same error.

jhand2 (approved):

PR #1097, opened 2023-11-20
WIP: Use newer rust version and set clippy args in Cargo.toml

ArthurHeymans:
Fix all the issues clippy finds with the new toolchain.

What about DPE? It's currently set to use the stable toolchain which errors on newer clippy.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,7 +1,7 @@
 # Licensed under the Apache-2.0 license
 
 [toolchain]
-channel = "1.70"
+channel = "1.74"
 targets = ["riscv32imc-unknown-none-elf"]
 profile = "minimal"
 components = ["rustfmt", "clippy"]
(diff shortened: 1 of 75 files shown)


rust-toolchain.toml
```diff
@@ -1,7 +1,7 @@
 # Licensed under the Apache-2.0 license
 
 [toolchain]
-channel = "1.70"
+channel = "1.74"
```
korran, line 4: I don't think we can make this change for the v1 caliptra ROM; we've spent a fair bit of time validating the early-boot machine code generated by 1.70 for CFI.

To do this, we'd need to move the ROM into its own workspace that stays at 1.70.

ArthurHeymans: That makes sense. I mostly wanted to experiment with this newer rust feature to be able to deny warnings from cargo rather than RUSTFLAGS environment.

It looks like clippy gained quite a few new lint checks so I could just change this PR to only fix those?

PR #1098, opened 2023-11-21
Truncate zeroes from mailbox req/resp

### LINKED ISSUE #1092: The allow truncating request for invoke_dpe mailbox command
`INVOKE_DPE` as-defined is supposed to allow truncating requests/responses to avoid writing extraneous zero bytes to the mailbox. The command handler needs to be updated to allow that.

sree-revoori1:
1. Implement as_bytes_partial{_mut} for all mailbox resp structs containing variable length arrays so that we don't write extraneous zeroes to the mailbox.
2. For all mailbox req structs containing variable length arrays, parse them from bytes by creating a default struct and copying the bytes into the prefix, since read_from will fail if extraneous zeroes are omitted from the req struct.

Fixes #1092

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -9,9 +9,9 @@ pub struct CommandId(pub u32);
 impl CommandId {
     pub const FIRMWARE_LOAD: Self = Self(0x46574C44); // "FWLD"
     pub const GET_IDEV_CSR: Self = Self(0x49444556); // "IDEV"
-    pub const GET_IDEV_CERT: Self = Self(0x49444543); // IDEC
-    pub const GET_IDEV_INFO: Self = Self(0x49444549); // IDEI
-    pub const POPULATE_IDEV_CERT: Self = Self(0x49444550); // IDEP
+    pub const GET_IDEV_CERT: Self = Self(0x49444543); // "IDEC"
+    pub const GET_IDEV_INFO: Self = Self(0x49444549); // "IDEI"
+    pub const POPULATE_IDEV_CERT: Self = Self(0x49444550); // "IDEP"
     pub const GET_LDEV_CERT: Self = Self(0x4C444556); // "LDEV"
     pub const ECDSA384_VERIFY: Self = Self(0x53494756); // "SIGV"
     pub const STASH_MEASUREMENT: Self = Self(0x4D454153); // "MEAS"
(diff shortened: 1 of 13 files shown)


hw-model/src/lib.rs
```diff
@@ -792,15 +792,14 @@ pub trait HwModel {
...
-        let response = match R::Resp::read_from(response_bytes.as_slice()) {
-            Some(response) => response,
-            None => {
-                return Err(ModelError::MailboxUnexpectedResponseLen {
-                    expected: mem::size_of::<R::Resp>() as u32,
-                    actual: response_bytes.len() as u32,
-                })
-            }
-        };
+        if response_bytes.len() > mem::size_of::<R::Resp>() {
```
korran, line 795: I don't think it's wise to do this for ALL commands. I think it's better to use my solution from https://github.com/chipsalliance/caliptra-sw/pull/1076, where we only do this for certain commands like GetLdevCertResp and InvokeDpeResp

korran: See https://github.com/chipsalliance/caliptra-sw/pull/1076/commits/f84c6b701d01ece874795a356b77bcbaf8d6c2b0

sree-revoori1: Thanks, I have updated it with your code.

api/src/mailbox.rs
```diff
@@ -432,6 +528,16 @@ pub struct InvokeDpeReq {
 
 impl InvokeDpeReq {
     pub const DATA_MAX_SIZE: usize = 512;
+
+    pub fn as_bytes_partial(&self) -> &[u8] {
```
korran, line 532: You could eliminate a lot of code duplication with something like this:

```rs

#[repr(C)]
#[derive(Debug, AsBytes, Default, FromBytes, PartialEq, Eq)]
pub struct MailboxRespHeaderVarSize {
    pub hdr: MailboxRespHeader,
    pub data_len: u32,
}
trait ResponseVarSize: AsBytes + FromBytes {
    fn data(&self) -> CaliptraResult<&[u8]> {
        // Will panic if sizeof<Self>() is smaller than MailboxRespHeaderVarSize
        // or Self doesn't have compatible alignment with
        // MailboxRespHeaderVarSize (should be impossible if MailboxRespHeaderVarSize is the first field)
        let (hdr, data) =
            LayoutVerified::<_, MailboxRespHeaderVarSize>::new_from_prefix(self.as_bytes())
                .unwrap();
        data.get(..hdr.data_len as usize)
            .ok_or(CaliptraError::MAILBOX_API_RESPONSE_DATA_LEN_TOO_LARGE)
    }
    fn partial_len(&self) -> usize {
        let (hdr, _) =
            LayoutVerified::<_, MailboxRespHeaderVarSize>::new_from_prefix(self.as_bytes())
                .unwrap();
        size_of::<MailboxRespHeaderVarSize>() + hdr.data_len as usize
    }
    fn as_bytes_partial(&self) -> CaliptraResult<&[u8]> {
        self.as_bytes()
            .get(..self.partial_len())
            .ok_or(CaliptraError::MAILBOX_API_RESPONSE_DATA_LEN_TOO_LARGE)
    }
    fn as_bytes_partial_mut(&mut self) -> CaliptraResult<&mut [u8]> {
        let partial_len = self.partial_len();
        self.as_bytes_mut()
            .get_mut(..partial_len)
            .ok_or(CaliptraError::MAILBOX_API_RESPONSE_DATA_LEN_TOO_LARGE)
    }
}
impl<T: ResponseVarSize> Response for T {
    const MIN_SIZE: usize = size_of::<MailboxRespHeaderVarSize>() + size_of::<u32>();
}


#[repr(C)]
#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
pub struct InvokeDpeResp {
    pub hdr: MailboxRespHeaderVarSize,
    pub data: [u8; InvokeDpeResp::DATA_MAX_SIZE], // variable length
}

impl InvokeDpeResp {
    pub const DATA_MAX_SIZE: usize = 2200;
}
impl ResponseVarSize for InvokeDpeResp {}
```

sree-revoori1: Thanks for writing this out. Also added a RequestVarSize trait.

api/src/mailbox.rs
```diff
@@ -298,6 +329,28 @@ pub struct GetIdevCertReq {
 }
 impl GetIdevCertReq {
     pub const DATA_MAX_SIZE: usize = 916; // Req max size = Resp max size - MAX_ECDSA384_SIG_LEN
+
+    pub fn as_bytes_partial(&self) -> &[u8] {
+        let unused_byte_count = Self::DATA_MAX_SIZE.saturating_sub(self.tbs_size as usize);
```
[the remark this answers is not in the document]
sree-revoori1: Done

api/src/mailbox.rs
```diff
@@ -164,45 +211,51 @@ pub enum MailboxReq {
...
     FipsVersion(MailboxReqHeader),
     FwInfo(MailboxReqHeader),
+    PopulateIdevCert(PopulateIdevCertReq),
+    GetIdevCert(GetIdevCertReq),
 
-    #[cfg(feature = "test_only_commands")]
+    #[cfg(all(feature = "test_only_commands"))]
     TestHmacVerify(HmacVerifyReq),
-    #[cfg(feature = "test_only_commands")]
+    #[cfg(all(feature = "test_only_commands"))]
```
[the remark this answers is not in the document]
sree-revoori1: It's not, previously I had two features here, "test_only_commands" and "runtime" but I removed "runtime" and forgot to remove the all.

jhand2 (approved):

mhatrevi (approved):

PR #1103, opened 2023-11-22
sw-emulator: implement PCR Hash digest and Sign

ArthurHeymans:


sw-emulator/lib/periph/src/hash_sha512.rs
```diff
@@ -565,6 +654,29 @@ impl HashSha512 {
         );
     }
 
+    fn pcr_hash_op_complete(&mut self) {
```
jhand2, line 657: I wasn't able to find documentation on what order everything is hashed in. Did you find it somewhere? Or just dig through the RTL?

Assuming this is the right order, this looks good to me :)

jhand2 (approved):
LGTM, just want to check on how the PCR digest is generated and if there's documentation so we can make sure the implementation here matches what the hardware does.

MDr164 (approved):

korran (approved):

PR #1104, opened 2023-11-22
Add PCR mailbox commands to quote and increment a reset counter

ArthurHeymans:


ArthurHeymans:
This merged https://github.com/chipsalliance/caliptra-sw/pull/1103 (sw-emulator changes). I created a separate PR to make it more digestable.

MDr164:
Policy check failure in CI likely related to #1102 being needed.

MDr164 (approved):

runtime/src/drivers.rs
```diff
@@ -66,6 +67,8 @@ pub struct Drivers {
 
     pub pcr_bank: PcrBank,
 
+    pub pcr_reset: PcrResetCounter,
```
jhand2, line 70: These should persist across Caliptra warm/update resets. I think they should be added to the `PersistentData` driver (as runtime-only, see DPE state for an example)

runtime/src/pcr.rs
```diff
@@ -0,0 +1,110 @@
...
+use caliptra_drivers::{
+    cprintln, hand_off::DataStore, CaliptraError, CaliptraResult, Ecc384PrivKeyIn, KeyReadArgs,
+    PcrBank, PcrId,
+};
+use core::ops::{Index, IndexMut};
+use enum_map::{EnumArray, EnumMap};
+use zerocopy::{transmute, FromBytes};
+
+pub struct PcrResetCounter {
+    counter: enum_map::EnumMap<PcrId, u32>,
```
jhand2, line 16: This seems like it could just be a `[u32; ALL_PCR_IDS.len()]`. And then populating the quote response is trivial.

ArthurHeymans: A previous version did that, but @sree-revoori1, wrote the following in #1111 (which is based on this PR):
"I think you might need a guard here to prevent potential panic since usize::from(id) could be out of array bounds" when accessing elements, so I looked into an enum array. I can revert it back.

sree-revoori1: You can run cargo test test_panic_missing within the runtime directory to check if your change induces potential panic. Then for the guard, it can just be an if statement checking that usize::from(id) is within the bounds of the array, and if it isn't it returns some sort of error.

drivers/Cargo.toml
```diff
@@ -23,6 +23,7 @@ zerocopy.workspace = true
 zeroize.workspace = true
 caliptra-cfi-lib = { workspace = true, default-features = false, features = ["cfi", "cfi-counter" ] }
 caliptra-cfi-derive.workspace = true
+enum-map.workspace = true
```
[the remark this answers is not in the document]
ArthurHeymans: Done

runtime/src/pcr.rs
ArthurHeymans: No panics created. The array of counters always has the exact length of the number of PcrId and PcrId is now used as an index, so no panic is ever possible.

drivers/src/memory_layout.rs
```diff
@@ -38,7 +38,8 @@ pub const PCR_LOG_ORG: u32 = 0x50004800;
 pub const MEASUREMENT_LOG_ORG: u32 = 0x50004C00;
 pub const FUSE_LOG_ORG: u32 = 0x50005000;
 pub const DPE_ORG: u32 = 0x50005400;
-pub const DATA_ORG: u32 = 0x50006400;
+pub const PCR_RESET_COUNTER_ORG: u32 = 0x50006400;
+pub const DATA_ORG: u32 = 0x50006000;
```
sree-revoori1, line 42: I think DATA_ORG should be 0x50006800

runtime/src/pcr.rs
sree-revoori1: Sounds good, thanks for checking.

drivers/src/memory_layout.rs
ArthurHeymans: Done

jhand2 (approved):

korran (approved):
Approval for caliptra-drivers.

Jordan: please address https://github.com/chipsalliance/caliptra-sw/issues/1185

PR #1110, opened 2023-11-28
Fix DPE HMAC bug where HMAC is always a zero array

sree-revoori1:
The function hmac_sign_with_derived incorrectly computes an HMAC by using an hmac key in the key vault and attempting to obtain an hmac tag as an array. This is not allowed due to wanting to prevent side-channel attacks.

With this change, we now derive an EC public key from the input CDI and KDF its entropy to get an hmac key in array form. This can then be used to calculate an hmac value in array form.


runtime/src/dpe_crypto.rs
```diff
@@ -281,21 +281,31 @@ impl<'a> Crypto for DpeCrypto<'a> {
         match algs {
             AlgLen::Bit256 => Err(CryptoError::Size),
             AlgLen::Bit384 => {
+                // derive an EC key pair from the CDI
+                let (_, pub_key) = Self::derive_key_pair(self, algs, cdi, label, info)?;
```
jhand2, line 285: nit: instead of pubkey can we call this `hmac_seed`? And add a comment explaining that the output point must be kept secret and why it is safe to do this (the output point is derived from the private key, so as long as that output is kept secret and not released outside of Caliptra, it is safe to use it as key material.

I want it to be clear that it shouldn't be treated as public data.

runtime/src/dpe_crypto.rs
```diff
@@ -281,21 +281,31 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+
+                // use the entropy from the EC public key to derive an HMAC key
+                let mut hmac_key = Array4x12::default();
+                let pub_key_x: [u8; 48] = pub_key
+                    .x
+                    .bytes()
+                    .try_into()
+                    .map_err(|_| CryptoError::Size)?;
```
jhand2, line 293: I think I would prefer to compute the sha384 hash of the whole public key and then use that output as the hmac key. The reason is that the distribution of a public key isn't completely random, its selected from a discrete set of points. Hashing it should ensure they key has the pull 384 bits of entropy.

bluegate010: Can we use a [KDF](https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/hmac384_kdf.rs)? A bit more NIST-friendly if we do it that way.

runtime/src/dpe_crypto.rs
sree-revoori1: Done

runtime/src/dpe_crypto.rs
sree-revoori1: As discussed in the meeting, I changed it to hash the pub key and feed that to the hmac kdf as ikm. We can revisit kdf-ing this hash later.

mschilder123: sha384 of the entire generated public key and then using that output directly as hmac-sha384 key is simplest and cryptographically fine wrt to bit munging and uniformity of the output.

A KDF step is more NIST-friendly but we already lost them on the entropy->private-key->public-key munging in the prior steps imo. And labels and different output lengths are not relevant for this single fixed sized output either.

I prefer simpler here. More comments in the code about the reasons for this particular work-around are probably nice for posterity.

runtime/src/dpe_crypto.rs
```diff
@@ -281,21 +281,41 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+                    Hmac384Tag::Array4x12(&mut hmac_key),
                 )
                 .map_err(|e| CryptoError::CryptoLibError(u32::from(e)))?;
 
+                // sign digest with HMAC key
                 let mut tag = Array4x12::default();
                 self.hmac384
                     .hmac(
-                        &KeyReadArgs::new(KEY_ID_DPE_PRIV_KEY).into(),
+                        &Hmac384Key::Array4x12(&hmac_key),
```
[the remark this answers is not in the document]
sree-revoori1: Done, https://github.com/chipsalliance/caliptra-dpe/pull/270 will zeroize hmac_seed, hmac_key is already being zeroized since Zeroize is derived for Array4xN, and I am now calling zeroize on hmac_ikm.

jhand2 (approved):

PR #1111, opened 2023-11-28
mailbox: Add EXTEND_PCR command

leongross:
* Re-opening of https://github.com/chipsalliance/caliptra-sw/pull/785

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,7 @@ dpe = { path = "dpe/dpe", default-features = false, features = ["dpe_profile_p38
 crypto = { path = "dpe/crypto", default-features = false }
 platform = { path = "dpe/platform", default-features = false }
 elf = "0.7.2"
+enum-map = "2.7.3"
 gdbstub = "0.6.3"
 gdbstub_arch = "0.2.4"
 getrandom = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,10 @@ bitfield = "0.14.0"
 bitflags = "2.4.0"
 bit-vec = "0.6.3"
 caliptra-api = { path = "api" }
-caliptra-cfi-lib = { path = "cfi/lib", default-features = false, features = ["cfi", "cfi-counter" ] }
+caliptra-cfi-lib = { path = "cfi/lib", default-features = false, features = [
+  "cfi",
+  "cfi-counter",
+] }
 caliptra-cfi-derive = { path = "cfi/derive" }
 caliptra_common = { path = "common", default-features = false }
 caliptra-coverage = { path = "coverage" }
@@ -104,7 +107,6 @@ caliptra-emu-derive = { path = "sw-emulator/lib/derive" }
 caliptra-emu-periph = { path = "sw-emulator/lib/periph" }
 caliptra-emu-types = { path = "sw-emulator/lib/types" }
 caliptra-error = { path = "error", default-features = false }
-caliptra-fpga-boss = { path = "ci-tools/fpga-boss" }
 caliptra-gen-linker-scripts = { path = "cpu/gen" }
 caliptra-hw-model = { path = "hw-model" }
 caliptra-hw-model-types = { path = "hw-model/types" }
@@ -123,7 +125,7 @@ caliptra-systemrdl = { path = "systemrdl" }
 caliptra-test = { path = "test" }
 caliptra-test-harness = { path = "test-harness" }
 caliptra-test-harness-types = { path = "test-harness/types" }
-caliptra-verilated = { path = "hw-latest/verilated",  features = ["verilator"] }
+caliptra-verilated = { path = "hw-latest/verilated", features = ["verilator"] }
 caliptra-x509 = { path = "x509", default-features = false }
 cbc = "0.1.2"
 cbindgen = { version = "0.24.0", default-features = false }
@@ -131,18 +133,17 @@ cfg-if = "1.0.0"
 chrono = "0.4.24"
 clap = { version = "3.2.14", default-features = false, features = ["std"] }
 convert_case = "0.6.0"
-dpe = { path = "dpe/dpe", default-features = false, features = ["dpe_profile_p384_sha384"] }
+dpe = { path = "dpe/dpe", default-features = false, features = [
+  "dpe_profile_p384_sha384",
+] }
 crypto = { path = "dpe/crypto", default-features = false }
 platform = { path = "dpe/platform", default-features = false }
 elf = "0.7.2"
-enum-map = "2.7.3"
 gdbstub = "0.6.3"
 gdbstub_arch = "0.2.4"
 getrandom = "0.2"
 hex = "0.4.3"
 lazy_static = "1.4.0"
-libftdi1-sys = { version = "1.1.2", features = ["libusb1-sys"] }
-libusb1-sys = "0.6.4"
 memoffset = "0.8.0"
 once_cell = "1.13"
 openssl = { version = "0.10", features = ["vendored"] }
@@ -151,7 +152,6 @@ proc-macro2 = "1.0.66"
 quote = "1.0"
 rand = "0.8"
 rfc6979 = "0.3.0"
-rusb = "0.9.3"
 serde = "1.0"
 serde_derive = "1.0.136"
 serde_json = "1.0"
@@ -160,9 +160,11 @@ sha3 = "0.10.8"
 smlang = "0.6.0"
 syn = "1.0.107"
 tinytemplate = "1.1"
-tock-registers = { git = "https://github.com/tock/tock.git"}
+tock-registers = { git = "https://github.com/tock/tock.git" }
 toml = "0.7.0"
-ufmt = { git = "https://github.com/korran/ufmt.git", rev = "1d0743c1ffffc68bc05ca8eeb81c166192863f33", features = ["inline"] }
+ufmt = { git = "https://github.com/korran/ufmt.git", rev = "1d0743c1ffffc68bc05ca8eeb81c166192863f33", features = [
+  "inline",
+] }
 uio = { version = "0.2.0" }
 ureg = { path = "ureg" }
 ureg-codegen = { path = "ureg/lib/codegen" }
@@ -174,7 +176,9 @@ zerocopy = "0.6.6"
 serial_test = "2.0.0"
 nix = "0.26.2"
 libc = "0.2"
-zeroize = { version = "1.6.0", default-features = false, features = ["zeroize_derive"] }
+zeroize = { version = "1.6.0", default-features = false, features = [
+  "zeroize_derive",
+] }
 regex = "1.10.2"
 
 [profile.firmware]
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -23,6 +23,7 @@ impl CommandId {
     pub const DPE_GET_TAGGED_TCI: Self = Self(0x47544744); // "GTGD"
     pub const INCREMENT_PCR_RESET_COUNTER: Self = Self(0x50435252); // "PCRR"
     pub const QUOTE_PCRS: Self = Self(0x50435251); // "PCRQ"
+    pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
 
     pub const TEST_ONLY_HMAC384_VERIFY: Self = Self(0x484D4143); // "HMAC"
 
@@ -602,6 +603,45 @@ impl Request for InvokeDpeReq {
     const ID: CommandId = CommandId::INVOKE_DPE;
     type Resp = InvokeDpeResp;
 }
+
+// EXTEND_PCR
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data_size: u32,
+    pub data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    _pad: [u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4) % 4],
+}
+pub type ExtendPcrReqErr = ();
+impl ExtendPcrReq {
+    // pub const DATA_MAX_SIZE: usize = sha384::SHA384_BLOCK_BYTE_SIZE - sha384::SHA384_HASH_SIZE - 1;
+    pub const DATA_MAX_SIZE: usize = 192;
+
+    pub fn new(
+        hdr: MailboxReqHeader,
+        pcr_idx: u32,
+        data_size: u32,
+        data: [u8; ExtendPcrReq::DATA_MAX_SIZE],
+    ) -> Result<Self, ExtendPcrReqErr> {
+        let data_size = match data_size {
+            0 | 255 => {
+                return Err(());
+            }
+            ds => ds,
+        };
+
+        Ok(ExtendPcrReq {
+            hdr,
+            pcr_idx,
+            data_size,
+            data,
+            _pad: [0u8; 4 - (ExtendPcrReq::DATA_MAX_SIZE % 4)],
+        })
+    }
+}
+// No command-specific output args
 
 #[repr(C)]
 #[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -227,6 +227,7 @@ pub enum MailboxReq {
     GetRtAliasCert(GetRtAliasCertReq),
     IncrementPcrResetCounter(IncrementPcrResetCounterReq),
     QuotePcrs(QuotePcrsReq),
+    ExtendPcr(ExtendPcrReq),
 
     #[cfg(feature = "test_only_commands")]
     TestHmacVerify(HmacVerifyReq),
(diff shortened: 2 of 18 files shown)


api/src/mailbox.rs
```diff
@@ -422,6 +504,30 @@ impl Request for InvokeDpeReq {
...
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data: [u8; 48],
+}
+pub type ExtendPcrReqErr = ();
+impl ExtendPcrReq {
+    pub const DATA_MAX_SIZE: usize = 48;
```
ArthurHeymans, line 517: Given that data is always 48 bytes, no more no less, 'MAX' is somewhat meaningsless? Maybe drop this?

api/src/mailbox.rs
```diff
@@ -422,6 +504,30 @@ impl Request for InvokeDpeReq {
...
+
+impl Default for ExtendPcrReq {
+    fn default() -> Self {
+        Self {
+            hdr: Default::default(),
+            pcr_idx: u32::default(),
+            data: [0u8; ExtendPcrReq::DATA_MAX_SIZE],
+        }
+    }
+}
```
ArthurHeymans, line 528: Not needed?

runtime/src/pcr.rs
```diff
@@ -0,0 +1,137 @@
...
+            digest: pcr_hash.into(),
+            signature_r: signature.r.into(),
+            signature_s: signature.s.into(),
+        }))
+    }
+}
+pub struct ExtendPcrCmd;
+impl ExtendPcrCmd {
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
+        // 1. Extend PCR
```
ArthurHeymans, line 116: Drop comment?

runtime/tests/runtime_integration_tests/integration_tests.rs
```diff
@@ -1269,3 +1317,138 @@ fn test_idev_id_cert() {
...
+        let cmd = ExtendPcrReq {
+            hdr: MailboxReqHeader { chksum: 0 },
+            pcr_idx: pcr_idx,
+            data: pcr_extension_data,
+        };
+
+        let checksum = caliptra_common::checksum::calc_checksum(
+            u32::from(CommandId::EXTEND_PCR),
+            &cmd.as_bytes()[4..],
+        );
```
ArthurHeymans, line 1337: Use the newer method on cmd to update checksum

runtime/tests/runtime_integration_tests/integration_tests.rs
```diff
@@ -1269,3 +1317,138 @@ fn test_idev_id_cert() {
         u32::from(caliptra_drivers::CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)
     );
 }
+
+#[test]
+fn test_extend_pcr_cmd() {
+    // PCR[4]' = sha384(PCR[4] || extension_data) = sha384(0x0...0 || extension_data)
```
ArthurHeymans, line 1323: What is this?

api/src/mailbox.rs
leongross: Currently not used, will remove this.

leongross:
rebased on ArthurHeymans/caliptra-sw/tree/pcr-mailbox-commands

api/src/mailbox.rs
```diff
@@ -590,6 +604,28 @@ impl Request for InvokeDpeReq {
...
+impl Default for ExtendPcrReq {
+    fn default() -> Self {
+        Self {
+            hdr: Default::default(),
+            pcr_idx: u32::default(),
+            data: [0u8; 48],
+        }
+    }
+}
+// No command-specific output args
```
ArthurHeymans, line 627: Remove?

api/src/mailbox.rs
```diff
@@ -590,6 +604,28 @@ impl Request for InvokeDpeReq {
...
+// EXTEND_PCR
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct ExtendPcrReq {
+    pub hdr: MailboxReqHeader,
+    pub pcr_idx: u32,
+    pub data: [u8; 48],
+}
+pub type ExtendPcrReqErr = ();
+impl ExtendPcrReq {}
```
ArthurHeymans, line 616: Unused?
Also it might be a good idea to implement the Request trait like other mailbox commands?

error/src/lib.rs
```diff
@@ -121,6 +121,8 @@ impl CaliptraError {
 
     pub const DRIVER_PCR_BANK_ERASE_WRITE_LOCK_SET_FAILURE: CaliptraError =
         CaliptraError::new_const(0x00070001);
+    pub const DRIVER_PCR_BANK_EXTEND_INVALID_SIZE: CaliptraError =
+        CaliptraError::new_const(0x00070002);
```
ArthurHeymans, line 125: Unused

hw-model/src/lib.rs
```diff
@@ -754,6 +754,20 @@ pub trait HwModel {
...
+        // ) -> caliptra_registers::soc_ifc_trng::RegisterBlock<BusMmio<Self::TBus<'_>>> {
+    ) -> caliptra_registers::pv::RegisterBlock<ureg::RealMmioMut> {
+        unsafe {
+            caliptra_registers::pv::RegisterBlock::new_with_mmio(
+                0x1001a000 as *mut u32,
+                core::default::Default::default(),
+            )
+        }
+    }
+
```
ArthurHeymans, line 770: unused

ArthurHeymans (approved):

runtime/src/pcr.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.counter[usize::from(id)]
+    }
+
+    /// Increment the selected reset counter.
+    /// Returns `false` in case of a counter overflow
+    pub fn increment(&mut self, id: PcrId) -> bool {
+        let old_value = self.counter[usize::from(id)];
+
+        if let Some(new_value) = old_value.checked_add(1) {
+            self.counter[usize::from(id)] = new_value;
```
sree-revoori1, line 39: Same here - I think you might need a guard here to prevent potential panic since usize::from(id) could be out of array bounds

runtime/src/pcr.rs
```diff
@@ -0,0 +1,124 @@
...
+    QuotePcrsResp,
+};
+use caliptra_drivers::{
+    cprintln, hand_off::DataStore, CaliptraError, CaliptraResult, Ecc384PrivKeyIn, KeyReadArgs,
+    PcrBank, PcrId,
+};
+use zerocopy::{transmute, FromBytes};
+
+pub struct PcrResetCounter {
+    counter: [u32; 32],
```
sree-revoori1, line 15: Can you pull this const 32 value from drivers/src/pcr_bank.rs? That way, if the size of the pcr bank ever changes, they wouldn't have to update this file.

api/src/mailbox.rs
```diff
@@ -590,6 +604,22 @@ impl Request for InvokeDpeReq {
     type Resp = InvokeDpeResp;
 }
 
+// EXTEND_PCR
+#[repr(C)]
```
sree-revoori1, line 608: Can you add ExtendPcrReq to the MailboxReq enum?

runtime/src/pcr.rs
```diff
@@ -0,0 +1,124 @@
...
+            self.counter[usize::from(id)] = new_value;
+
+            true
+        } else {
+            false
+        }
+    }
+}
+
+pub struct IncrementPcrResetCounter;
```
sree-revoori1, line 48: nit: rename this IncrementPcrResetCounterCmd to maintain consistency with other mbox cmds?

runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -161,7 +161,7 @@ fn get_ldev_cert(model: &mut DefaultHwModel) -> GetLdevCertResp {
     ldev_resp
 }
 
-fn get_fmc_alias_cert(model: &mut DefaultHwModel) -> GetFmcAliasCertResp {
+pub fn get_fmc_alias_cert(model: &mut DefaultHwModel) -> GetFmcAliasCertResp {
```
sree-revoori1, line 164: Maybe just move this function to common.rs?

runtime/tests/runtime_integration_tests/test_pcr.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    let fmc_resp = get_fmc_alias_cert(&mut model);
+    let fmc_cert: X509 = X509::from_der(&fmc_resp.data[..fmc_resp.data_size as usize]).unwrap();
+    let pkey = fmc_cert.public_key().unwrap().ec_key().unwrap();
+
+    assert!(sig.verify(&resp.digest, &pkey).unwrap());
+}
+
+#[test]
+fn test_extend_pcr_cmd() {
```
sree-revoori1, line 80: Can you break this test into multiple tests? I think one test for 0, 1.0, 2.0, etc would be good.

runtime/tests/runtime_integration_tests/test_pcr.rs
```diff
@@ -0,0 +1,203 @@
+// Licensed under the Apache-2.0 license
+
```
sree-revoori1, line 2: Can you add all new tests to test-coverage.md? And also add any test gaps that may have come up due to this PR?

runtime/src/pcr.rs
```diff
@@ -0,0 +1,124 @@
...
+            .ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
+
+        let index =
+            u8::try_from(cmd.index).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        let pcr =
+            PcrId::try_from(index).map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        if !drivers.pcr_reset.increment(pcr) {
+            return Err(CaliptraError::RUNTIME_INTERNAL);
```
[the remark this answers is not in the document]
leongross: This PR should only implement the PCR_EXTEND feature.

But since this code relies on #1104, I rebased it on this branch because it has not been merged yet. I am not the author of the pcr reset code, so I think a review on this code fits better into into PR #1104

runtime/src/pcr.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    pub fn get(&self, id: PcrId) -> u32 {
+        self.counter[usize::from(id)]
+    }
+
+    /// Increment the selected reset counter.
+    /// Returns `false` in case of a counter overflow
+    pub fn increment(&mut self, id: PcrId) -> bool {
+        let old_value = self.counter[usize::from(id)];
```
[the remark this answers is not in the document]
leongross: See #1104

runtime/src/pcr.rs
leongross: See #1104

runtime/src/pcr.rs
leongross: #1004

runtime/tests/runtime_integration_tests/test_pcr.rs
leongross: What do you mean with 'test gaps'?

runtime/tests/runtime_integration_tests/test_certs.rs
leongross: #1104

runtime/tests/runtime_integration_tests/test_pcr.rs
jhand2: This doc covers tests in runtime firmware. If there's any features that get added here but are not yet tested, you can add an item to the "Test Gaps" section at the bottom. But if not that's fine (I still need to do a pass to check if I see anything).

jhand2:
I think this PR looks good, but it depends on some comments being resolved in https://github.com/chipsalliance/caliptra-sw/pull/1104 (since this is built on top of that). Let's try to get 1104 merged and then I'll do a final pass on this PR.

error/src/lib.rs
```diff
@@ -400,6 +400,10 @@ impl CaliptraError {
     pub const RUNTIME_INCREMENT_PCR_RESET_MAX_REACHED: CaliptraError =
         CaliptraError::new_const(0x000E003A);
 
+    /// PCR Runtime Errors
+    pub const RUNTIME_PCR_RESERVED: CaliptraError = CaliptraError::new_const(0x000E0015);
```
sree-revoori1, line 404: These errror codes are already used by other runtime errors. Can you make them 0x000E003B and 0x000E003C respectively?

runtime/doc/test-coverage.md
```diff
@@ -134,6 +134,32 @@ Check whether the digital signature returned by Sign command can be verified usi
 Check that the Sign command fails on simulated contexts as simulation context do not allow signing | **TestSignSimulation** | StatusInvalidArgument
 Calls and tests behavior of the DPE command Sign with the Symmetric flag set | **TestSignSymmetric** | N/A
 
+<br><br>
```
sree-revoori1, line 137: I think you might be using an old version of this file, since it no longer has a mailbox command tests or wycheproof tests section. Can you add these tests to the pcr tests section?

drivers/src/pcr_bank.rs
```diff
@@ -14,9 +14,10 @@ Abstract:
 
 use crate::{Array4x12, CaliptraError, CaliptraResult, Sha384};
 use caliptra_registers::pv::PvReg;
+use enum_map::Enum;
 
 /// PCR Identifier
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Enum)]
```
sree-revoori1, line 20: Why do we need to derive Enum here?

Cargo.toml
```diff
@@ -135,6 +135,7 @@ dpe = { path = "dpe/dpe", default-features = false, features = ["dpe_profile_p38
 crypto = { path = "dpe/crypto", default-features = false }
 platform = { path = "dpe/platform", default-features = false }
 elf = "0.7.2"
+enum-map = "2.7.3"
```
jhand2, line 138: I don't think enum-map is being used anymore. Should this be removed here and in other Cargo.toml files?

ArthurHeymans (commented):
Do you need to update cargo.lock and cargo.toml ?

runtime/tests/runtime_integration_tests/test_pcr.rs
```diff
@@ -72,3 +74,136 @@ fn test_pcr_quote() {
...
+    // 1.1 Checking for PCR values using PCR_QUOTE
+    let pcrs = get_model_pcrs(&mut model);
+    assert_eq!(
+        pcrs[4],
+        [
+            245, 123, 183, 237, 130, 198, 174, 74, 41, 230, 201, 135, 147, 56, 197, 146, 199, 212,
+            42, 57, 19, 85, 131, 232, 204, 190, 57, 64, 242, 52, 75, 14, 182, 235, 133, 3, 219, 15,
+            253, 106, 57, 221, 208, 12, 208, 125, 131, 23
+        ]
```
jhand2, line 133: Instead of hard-coding this (and other expected values) I think we should just compute the hashes on the test side with the `sha2` crate and compare them to the expected output.

leongross: Since the calculation of sha2 here is depending on the hardware model, I suggest not recalculating it using the same model, because then we could potentially be comparing two erroneous results that are the same but still wrong.

The values I used in the test were calculated with a python script that uses battle-tested software implementations of the hash functions.
For further clarification, I will add the python script to this PR.

ArthurHeymans: It does not depend on the hardware model. The Quote command does something to the SHA engine resulting in a wrong result on the FPGA. There is no need to have a script just to compute basic hash values. The runtime cost of computing it is little and it's much clearer what is going on than comparing a precomputed output. Maybe rust even computes the hash at compiletime as the data is static?

jhand2: I'm just recommending you use a 3rd-party rust software crate (unrelated to Caliptra) to compute the hash of zeros instead of computing it elsewhere and pasting it here.

Fwiw, I don't think doing this will fix the CI failure you're seeing. But I do think it would make the test more readable/maintainable.

I'll try to spend some time on the FPGA today to see if I can see why this might be happening.

leongross: Sounds good, I am working on that.

MDr164: @jhand2 CI fails due to weird behavior of the SHA engine as demonstrated here #1233

jhand2: Looks like the hardware folks were able to reproduce this bug in simulation: https://github.com/chipsalliance/caliptra-rtl/issues/375

So the proposed workaround is to re-initialize the SHA engine after doing `gen_pcr_hash` and before `pcr_extend`.

I'm thinking the easier option would be to do it at the [beginning of pcr_extend](https://github.com/chipsalliance/caliptra-sw/blob/0bbb29f66eac7d65bb606a8df766796125033300/drivers/src/sha384.rs#L180). Something like `self.sha512.regs_mut().ctrl().write(|w| w.init(true);`. I confirmed that this resolves the issue on my FPGA.

Let's do that workaround for now, and then we can decide later if we think it's worth taking the RTL fix in 1.0 or not.

jhand2: This PR should fix the problem: https://github.com/chipsalliance/caliptra-sw/pull/1249

leongross:
@jhand2 any thoughts on this?

jhand2:
> @jhand2 any thoughts on this?

I set https://github.com/chipsalliance/caliptra-sw/pull/1254 to merge, but it looks like there might be some things in this PR in addition to that one. If you want you can rebase this onto main and I can take a look at the diffs.

PR #1132, opened 2023-11-30
Implement execute_dpe_cmd to reduce duplication in RT tests

sree-revoori1:
This command takes in a DPE command and returns a DPE response. It reduces duplication in INVOKE_DPE calls in RT testing significantly.


jhand2 (commented):
LGTM, but I don't think the DPE submodule commit exists in the upstream (since the newly merged commit has a different hash). Can you update that?

sree-revoori1:
Updated the DPE commit.

jhand2 (approved):

PR #1139, opened 2023-12-04
Expand the smoke test with fmc-alias, rt-alias, and hitless update validation

korran:
Please review each commit separately.


test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -395,6 +397,114 @@ fn smoke_test() {
...
+
+    let rt_dice_tcb_info = DiceTcbInfo::find_single_in_cert(rt_alias_cert_der).unwrap();
+    assert_eq!(
+        rt_dice_tcb_info,
+        Some(DiceTcbInfo {
+            vendor: Some("Caliptra".into()),
+            model: Some("RT".into()),
+            svn: Some(0x100),
+            fwids: vec![DiceFwid {
+                // FMC
```
jhand2, line 426: nit: RT?

test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -510,6 +510,147 @@ fn smoke_test() {
         .cptra_hw_error_non_fatal()
         .read()
         .mbox_ecc_unc());
+
+    let image2 = caliptra_builder::build_and_sign_image(
```
jhand2, line 514: Maybe the ship has sailed, but the smoke test is a bit hard to follow since it does so many things. Would it make sense to factor the checks into helpers to make it a little easier to follow. Something like:

```
// Load firmware

verify_uart(...);
verify_cert_fields(...);

// Update firmware

verify_updated_cert_fields(...);
```

jhand2 (approved):

test/tests/caliptra_integration_tests/smoke_test.rs
korran: I do want to do a cleanup commit at some point, moving boilerplaty stuff like the golden-data checks to helper functions or macros. In general though, I think we want to keep important assertions like "the rt-alias must be signed by the fmc-alias" in the top-level test. It's more important that the assertions are clear and "obviously right at a glance". 

The next major change I want to make to the test is to parameterize the UDS and lifecycle/debug values.

jhand2: Ack. Seems fine clean this up in a follow up. I agree having all the asserts in one place is helpful.

test/tests/caliptra_integration_tests/smoke_test.rs
korran: Done

jhand2 (approved):

PR #1141, opened 2023-12-04
Enable CFI in FMC.

rusty1968:
- Enable CFI protection feature in the drivers crate.
- Use CFI protection in RtAlias key derivation.


FerralCoder (approved):

fmc/src/main.rs
```diff
@@ -44,6 +45,16 @@ pub extern "C" fn entry_point() -> ! {
         Err(e) => report_error(e.into()),
     };
 
+    if !cfg!(feature = "no-cfi") {
+        cprintln!("[state] CFI Enabled");
+        let mut entropy_gen = || env.trng.generate().map(|a| a.0);
+        CfiCounter::reset(&mut entropy_gen);
```
korran, line 51: We should only need to do this once; we can trust that there's already some entropy in there from the ROM so it's not a big deal if somebody glitches past this...

korran: @mhatrevi Is this even necessary? Does the ROM not leave the cfi counter in a healthy state?

rusty1968: ROM is corrupting cfi state before handing off.

korran: Ok, I'd like to get some more info on why we do that, but it doesn't need to block this PR.

FerralCoder (approved):

PR #1150, opened 2023-12-06
Add PAUSER Privilege level tests

sree-revoori1:


runtime/src/drivers.rs
```diff
@@ -416,4 +428,55 @@ impl Drivers {
...
+            if i < dpe.contexts.len()
+                && i < context_has_tag.len()
+                && i < context_tags.len()
+                && dpe.contexts[i].state != ContextState::Active
+            {
+                context_has_tag[i] = U8Bool::new(false);
+                context_tags[i] = 0;
+            }
+        });
+    }
```
jhand2, line 481: Similar comments for these, I think they might make sense as members for `Drivers`

jhand2 (commented):
The testing here is really thorough, great job!

runtime/src/drivers.rs
```diff
@@ -416,4 +428,55 @@ impl Drivers {
         drivers.cert_chain = cert_chain;
         Ok(())
     }
+
+    pub fn pl_context_threshold_exceeded(
```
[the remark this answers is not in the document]
sree-revoori1: First two bullets done. For the third bullet, discussed in chat but the problem with making them member functions is that we get borrow checker errors since we mutably borrow drivers in other places in the code and it would take a large refactor to fix this.

runtime/src/drivers.rs
sree-revoori1: See other comment.

jhand2 (approved):

PR #1154, opened 2023-12-06
Test mbox pausers and RT journey PCR are added to DPE upon RT startup

### LINKED ISSUE #1094: [RT Test] Add a test checking that measurements in the measurement log are added to DPE upon initializing drivers

### LINKED ISSUE #1116: [RT Test] Check that mailbox valid pausers are measured into DPE upon RT startup

sree-revoori1:
Fixes #1116 
Fixes #1094 

Also, add a test that checks that measurement log entries are added to DPE upon RT startup

### DIFF
--- a/runtime/test-fw/src/mbox_responder.rs
+++ b/runtime/test-fw/src/mbox_responder.rs
@@ -1,0 +1,78 @@
+// Licensed under the Apache-2.0 license
+
+#![no_main]
+#![no_std]
+
+use caliptra_common::mailbox_api::CommandId;
+use caliptra_drivers::{pcr_log::RT_FW_JOURNEY_PCR, Array4x12};
+use caliptra_registers::mbox::enums::MboxStatusE;
+use caliptra_runtime::{ContextState, Drivers};
+use caliptra_test_harness::{runtime_handlers, test_suite};
+use zerocopy::AsBytes;
+
+fn mbox_responder() {
+    let mut drivers = unsafe { Drivers::new_from_registers().unwrap() };
+    assert!(drivers.persistent_data.get().fht.is_valid());
+    let mut mbox = drivers.mbox;
+
+    loop {
+        while !mbox.is_cmd_ready() {
+            // Wait for a request from the SoC.
+        }
+        let cmd = mbox.cmd();
+
+        match cmd {
+            // Read RT_FW_JOURNEY_PCR
+            CommandId(0x1000_0000) => {
+                let rt_journey_pcr: [u8; 48] = drivers.pcr_bank.read_pcr(RT_FW_JOURNEY_PCR).into();
+                mbox.write_response(&rt_journey_pcr).unwrap();
+                mbox.set_status(MboxStatusE::DataReady);
+            }
+            // Reconstruct valid mbox pauser hash
+            CommandId(0x2000_0000) => {
+                const PAUSER_COUNT: usize = 5;
+                let mbox_valid_pauser: [u32; PAUSER_COUNT] = drivers.soc_ifc.mbox_valid_pauser();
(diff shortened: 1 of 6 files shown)


runtime/test-fw/src/mbox_responder.rs
```diff
@@ -0,0 +1,74 @@
...
+                    if mbox_pauser_lock[i] {
+                        valid_pausers[num_valid_pausers] = mbox_valid_pauser[i];
+                        num_valid_pausers += 1;
+                    }
+                }
+                let valid_pauser_hash: [u8; 48] = drivers
+                    .sha384
+                    .digest(valid_pausers[..num_valid_pausers].as_bytes())
+                    .unwrap()
+                    .into();
```
jhand2, line 48: nit: could do something like 
```suggestion
                const PAUSER_COUNT: usize = 5;
                let mbox_valid_pauser: [u32; PAUSER_COUNT] = drivers.soc_ifc.mbox_valid_pauser();
                let mbox_pauser_lock: [bool; PAUSER_COUNT] = drivers.soc_ifc.mbox_pauser_lock();
                let mut digest_op = sha384.digest_init().unwrap();
                for i in 0..PAUSER_COUNT {
                    if mbox_pauser_lock[i] {
                        digest.update(mbox_valid_pauser[i].as_bytes());
                    }
                }
                
                let mut valid_pauser_hash = Array4x12::default();
                digest_op.finalize(&mut valid_pauser_hash).unwrap();
```

runtime/tests/runtime_integration_tests/test_boot.rs
```diff
@@ -75,3 +77,30 @@ fn test_update() {
...
+    let valid_pauser_hash_resp = model.mailbox_execute(0x2000_0000, &[]).unwrap().unwrap();
+    let valid_pauser_hash: &[u8] = valid_pauser_hash_resp.as_bytes();
+
+    const MEASUREMENT_SIZE: usize = DPE_PROFILE.get_hash_size();
+    let mut boot_tci_data = [0u8; 3 * MEASUREMENT_SIZE];
+    boot_tci_data[MEASUREMENT_SIZE..2 * MEASUREMENT_SIZE].copy_from_slice(rt_journey_pcr);
+    boot_tci_data[2 * MEASUREMENT_SIZE..].copy_from_slice(valid_pauser_hash);
+    let hashed_boot_tci_data = model
+        .compute_sha512_acc_digest(&boot_tci_data, ShaAccMode::Sha384Stream)
+        .unwrap();
```
jhand2, line 97: I think this might be more clear if you use the https://docs.rs/sha2/latest/sha2/ crate and call update for each measurement, instead of using the accelerator.

runtime/test-fw/src/mbox_responder.rs
sree-revoori1: Done

runtime/tests/runtime_integration_tests/test_boot.rs
sree-revoori1: It looks like there's some differences in the caliptra-internal sha384 and the normal sha384. I tried hashing some simple things like [0u8; 48] and "hello world" and the sha2::Sha384 hash and the caliptra sha384 hash differed. I ended up creating a new mbox cmd in mbox_responder that reads data from the mailbox, hashes it, and writes it back to the mbox so that we can use the same hasher that caliptra uses in the tests.

jhand2: >  I tried hashing some simple things like [0u8; 48] and "hello world" and the sha2::Sha384 hash and the caliptra sha384 hash differed

Huh, that's unexpected. We should maybe look into why that is.

jhand2 (approved):

PR #1165, opened 2023-12-07
Code quality assurance : Extend code coverage to ICCM.

rusty1968:
This PR extends the existing ROM coverage infra-structure in the sw-emulator with a bitmap vector for ICCM and adds coverage reports for FMC and RT to the coverage app. Coverage analysis is done via the per-submit ci.sh script. 
Note : Additionally, removes unused code from cpu crate.


hw-model/src/model_emulated.rs
```diff
@@ -66,14 +66,14 @@ impl Drop for ModelEmulated {
...
-        let bitmap = self.code_coverage_bitmap();
+        let (bitmap, _) = self.code_coverage_bitmap();
         let _ =
             caliptra_coverage::dump_emu_coverage_to_file(cov_path.as_str(), self.image_tag, bitmap);
     }
 }
 
 impl ModelEmulated {
-    pub fn code_coverage_bitmap(&self) -> &bit_vec::BitVec {
+    pub fn code_coverage_bitmap(&self) -> (&bit_vec::BitVec, &bit_vec::BitVec) {
```
korran, line 76: For readability, can you replace this tuple with 

```
struct CoverageBitmaps<'a> {
  rom: &'a bit_vec::BitVec,
  iccm: &'a bit_vec::BitVec,
}
```

(basically the same thing, but with field names)

rusty1968: Will do.

rusty1968: Done

jlmahowa-amd (approved):

cpu/src/nmi.S
```diff
@@ -82,12 +82,3 @@ _nmi_handler:
 
     mret
 
-.section .init.text, "ax"
```
korran, line 85: Was this change intentional? The PR description doesn't mention it...

rusty1968: Yes. This is unused code. Made it a stand alone commit.

korran (approved):

hw-model/src/lib.rs
```diff
@@ -540,6 +540,7 @@ pub trait HwModel {
                 }
             }
             writeln!(hw.output().logger(), "ready_for_fw is high")?;
+            hw.cover_fw_mage(fw_image);
```
korran, line 543: @rusty1968 to file an issue to address ICCM upgrades. We need to make sure that coverage from the new firmware image doesn't get associated with the old one:

Suggestion:

* Have HwModel::upload_firmware() call cover_fw_image(); this will take care of most of the simple upgrade tests

* For every instruction, compare the instruction read out of iccm with the one read from the covered image, and error if they don't match.

PR #1167, opened 2023-12-08
Auto-initialize DPE with the RT Journey PCR

### LINKED ISSUE #1118: [RT Test] Check that measurements are stored in DPE when STASH_MEASUREMENT is called

### LINKED ISSUE #1169: Rethink tagging behavior
Few questions that need to be answered regarding tagging:

How should tagging work for retired contexts? I see 3 options:

1. Disallow TAG_TCI and GET_TAGGED_TCI on retired contexts
2. Allow TAG_TCI and GET_TAGGED_TCI on retired contexts
3. Disallow TAG_TCI but allow GET_TAGGED_TCI on retired contexts

We currently do 1, but I think 3 could be better. This would mean users would be able to tag a context, derive a child from it, and still have access to its TCIs directly. 

If we tag a default context, and then derive a child from that context with retains-parent = false, and make-default = true, we would get a new context with the same handle as the tagged context, except without the tag. Should the tag carry over to the new context?

sree-revoori1:
Currently, there is a bug in RT where Drivers::get_dpe_root_context_idx returns the incorrect DPE context index. This function searches for the context whose parent is 0xFF, which indicates that that context is the root. However, DPE is auto-initialized and then DeriveChild is called in order to measure the RT journey PCR into a child context. This means the root context does not contain the RT journey PCR which is incorrect.

With the change in https://github.com/chipsalliance/caliptra-dpe/pull/278, we are able to auto-initialize DPE with a measurement. Using that, we now auto-initialize DPE with the RT journey PCR so that we don't have an extraneous default measurement in DPE. 

This PR also updates tagging behavior to allow GET_TAGGED_TCI on retired contexts and disallow TAG_TCI on retired contexts. The test has been updated accordingly.

This PR also adds a test that checks that measurements are stored in DPE when STASH_MEASUREMENT is called.

Fixes #1118 
Fixes #1169

### DIFF
--- a/runtime/src/invoke_dpe.rs
+++ b/runtime/src/invoke_dpe.rs
@@ -10,7 +10,7 @@ use dpe::{
     },
     context::{Context, ContextState},
     response::{Response, ResponseHdr},
-    DpeInstance, MAX_HANDLES,
+    DpeInstance, U8Bool, MAX_HANDLES,
 };
 use zerocopy::{AsBytes, FromBytes};
 
@@ -78,10 +78,7 @@ impl InvokeDpeCmd {
                     {
                         return Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL);
                     }
-                    let derive_child_resp = cmd.execute(dpe, &mut env, locality);
-                    // clear tags for retired contexts
-                    Drivers::clear_tags_for_non_active_contexts(dpe, context_has_tag, context_tags);
-                    derive_child_resp
+                    cmd.execute(dpe, &mut env, locality)
                 }
                 Command::CertifyKey(cmd) => {
                     // PL1 cannot request X509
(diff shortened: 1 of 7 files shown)


runtime/src/invoke_dpe.rs
```diff
@@ -131,4 +128,21 @@ impl InvokeDpeCmd {
...
+    pub fn clear_tags_for_inactive_contexts(
+        dpe: &mut DpeInstance,
+        context_has_tag: &mut [U8Bool; MAX_HANDLES],
+        context_tags: &mut [u32; MAX_HANDLES],
+    ) {
+        (0..MAX_HANDLES).for_each(|i| {
+            if i < dpe.contexts.len()
+                && i < context_has_tag.len()
+                && i < context_tags.len()
+                && dpe.contexts[i].state != ContextState::Active
```
jhand2, line 141: Should this be for `dpe.contexts[i].state == ContextState::Inactive`?

sree-revoori1: Yes, thanks for catching. Done.

dpe
jhand2, line 1: I think this DPE commit doesn't exist in caliptra-dpe upstream

sree-revoori1: Fixed

jhand2 (approved):

PR #1178, opened 2023-12-08
Add Update Reset tests to RT

### LINKED ISSUE #1134: [RT Test] Test DPE structure validation upon update reset
We can trigger a hitless update with the FW_LOAD mbox cmd. Load bad data into SRAM so that the DPE structure becomes invalid and check that validation fails.

### LINKED ISSUE #1145: [RT Test] Check Context tag structure upon warm/update reset

### LINKED ISSUE #1161: [RT Test] Check that RT_FW_JOURNEY_PCR is updated correctly upon hitless update

sree-revoori1:
Fixes #1161 
Fixes #1145 
Fixes #1134

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_update_reset.rs
+++ b/runtime/tests/runtime_integration_tests/test_update_reset.rs
@@ -1,0 +1,212 @@
+// Licensed under the Apache-2.0 license
+
+use std::mem::size_of;
+
+use caliptra_builder::{
+    firmware::{self, APP_WITH_UART, FMC_WITH_UART},
+    ImageOptions,
+};
+use caliptra_common::mailbox_api::{
+    CommandId, FwInfoResp, MailboxReq, MailboxReqHeader, TagTciReq,
+};
+use caliptra_error::CaliptraError;
+use caliptra_hw_model::HwModel;
+use caliptra_runtime::RtBootStatus;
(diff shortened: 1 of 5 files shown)


runtime/tests/runtime_integration_tests/test_update_reset.rs
```diff
@@ -0,0 +1,212 @@
...
+        ImageOptions::default(),
+    )
+    .unwrap()
+    .to_bytes()
+    .unwrap();
+    model
+        .mailbox_execute(u32::from(CommandId::FIRMWARE_LOAD), &updated_fw_image)
+        .unwrap();
+
+    let rt_journey_pcr_resp = model.mailbox_execute(0x1000_0000, &[]).unwrap().unwrap();
```
jhand2, line 41: nit: for all the mbox_responder command IDs, could you define consts or an enum somewhere in the test code? I think it will make it easier to well what the tests are doing.

sree-revoori1: I tried to do this earlier, but its a bit hard since mbox_responder lives in a separate crate than the tests, so we would have to define these codes in real RT firmware which isn't preferable. The reason we can do this for FW_LOAD is that it's already a mbox command in RT. 

I agree though and I'll try to refactor the test code in a future PR so that this is possible.

jhand2 (approved):

PR #1182, opened 2023-12-09
Diversify Runtime error codes

sree-revoori1:
Also, in places where we disable attestation, signal a non fatal error to SoC so there is some way to find out what went wrong, other than looking at UART.

After this, we still need to find a way to pass DPE error code info up to runtime, instead of throwing it away by using map_err.

### DIFF
--- a/runtime/src/dice.rs
+++ b/runtime/src/dice.rs
@@ -155,6 +155,7 @@ pub fn copy_ldevid_cert(
         .get(..persistent_data.fht.ldevid_tbs_size.into());
     let sig = ldevid_dice_sign(persistent_data, dv)?;
     cert_from_tbs_and_sig(tbs, &sig, cert)
+        .map_err(|_| CaliptraError::RUNTIME_GET_LDEVID_CERT_FAILED)
 }
 
 // Retrieve the r portion of the FMC cert signature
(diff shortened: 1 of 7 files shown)


runtime/src/dice.rs
```diff
@@ -233,17 +243,23 @@ pub fn copy_rt_alias_cert(
...
+        CaliptraError::RUNTIME_GET_RT_ALIAS_CERT_FAILED,
+    )
 }
 
 /// Create a certificate from a tbs and a signature and write the output to `cert`
 fn cert_from_tbs_and_sig(
     tbs: Option<&[u8]>,
     sig: &Ecc384Signature,
     cert: &mut [u8],
+    err: CaliptraError,
```
jhand2, line 259: nit: instead of changing the function signature, could use `map_err` at the callsites. Might be a little more clear what is happening.

jhand2 (approved):

runtime/src/dice.rs
jhand2: Since this is the only comment, I'm happy to just merge this anyway. Up to you.

sree-revoori1: Since I followed this pattern in two places, decided to just make the change. Reads a lot simpler too.

jhand2 (approved):

PR #1184, opened 2023-12-11
Nightly: Publish test matrix HTML to chipsalliance.github.io/caliptra-sw

korran:
Tested in a different repo:

Release Workflow:
https://github.com/korran-testorg/caliptra-sw/actions/runs/7172400508

Release:
https://github.com/korran-testorg/caliptra-sw/releases/tag/release_v20231211_2

Published HTML:
https://korran-testorg.github.io/caliptra-sw


ci-tools/test-matrix/src/html.rs
```diff
@@ -0,0 +1,95 @@
...
+struct TemplateContext<'a> {
+    pub matrix: &'a TestMatrix,
+    pub javascript: &'static str,
+    pub style: &'static str,
+    pub run: &'a Run,
+    pub run_infos: Vec<MaybeSelected<&'a RunInfo>>,
+}
+
+#[derive(Serialize)]
+pub struct MaybeSelected<T: Serialize> {
```
jhand2, line 22: nit: Can you add a comment explaining what it means or an item to be selected? Is this a filtering mechanism?

ci-tools/test-matrix/src/html.rs
```diff
@@ -0,0 +1,95 @@
...
+impl<'a, T: Serialize + Debug> Debug for MaybeSelected<&'a T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MaybeSelected")
+            .field("selected", &self.selected)
+            .field("value", &self.value)
+            .finish()
+    }
+}
+
+pub fn format(run: &Run, toc: &[RunInfo], matrix: &TestMatrix) -> String {
```
jhand2, line 49: Can you add a comment with input/output contract? Is the output here HTML?

ci-tools/test-matrix/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+        RunInfo {
+            id: run.created_at.format("%F-%H%M%S").to_string(),
+            display_name: run.created_at.format("%F %H:%M:%S").to_string(),
+        }
+    }
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let www_out = std::env::var("CPTRA_WWW_OUT").unwrap();
```
jhand2, line 191: If these aren't present, we should print a message what what the caller should do to populate them rather than panicking.

ci-tools/test-matrix/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+    const ORG: &str = "chipsalliance";
+    const REPO: &str = "caliptra-sw";
+
+    let octocrab = Octocrab::builder().personal_token(token).build()?;
+    let release_runs = octocrab
+        .workflows(ORG, REPO)
+        .list_runs("nightly-release.yml")
+        .branch("main")
+        .send()
+        .await?;
```
jhand2, line 202: From what I can tell, the token is only needed for publishing the results. Would it be simple enough to support running locally as the default to support contributing to this tool? And then an option can be passed to publish, which would require a token.

In terms of reviewing this HTML generation parts of this PR, I found it a bit hard to review without being able to run the tool and see what HTML it generates.

jhand2: Oh, I see you posted the output. In that case this comment isn't necessarily blocking.

ci-tools/test-matrix/src/html.rs
jhand2: I see this is for the modal with the test output. Still would be good to have a comment explaining this.

ci-tools/test-matrix/src/html/index.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Caliptra test matrix</title>
```
jhand2, line 4: nit: Would be good to also have a heading at the top with this same title

jhand2 (approved):
Overall LGTM. My comments are mainly cosmetic and can be handled as follow-ups if we want.

ci-tools/test-matrix/src/main.rs
korran: GitHub doesn't provide workflow artifacts to unauthenticated users. A future version of this tool may look at the public release artifacts instead (they don't exist yet).

ci-tools/test-matrix/src/html.rs
korran: Done

ci-tools/test-matrix/src/main.rs
korran: Added expectation message. FWIW, this isn't really intended to be used by humans.

ci-tools/test-matrix/src/html/index.html
korran: Done. See https://korran-testorg.github.io/caliptra-sw

PR #1190, opened 2023-12-12
Add persistent data tests in RT

### LINKED ISSUE #1185: Call PersistentData::assert_matches_layout() from a calitptra-runtime test case
The PersistentData struct has much [conditional compilation](https://github.com/chipsalliance/caliptra-sw/blob/b2f09dfdf0b03933aee5df9c42db0498f8383c70/drivers/src/persistent.rs#L54) based on the "runtime feature"

The [driver test](https://github.com/chipsalliance/caliptra-sw/blob/b2f09dfdf0b03933aee5df9c42db0498f8383c70/drivers/test-fw/src/bin/persistent_tests.rs#L10) calls PersistentData::assert_matches_layout() without the "runtime" feature. We need to make sure the layout is also correct when the "runtime" feature is enabled. I recommend adding a test case to the runtime crate that calls this with the runtime feature enabled.

sree-revoori1:
The PersistentData struct uses the "runtime" cfg flag to add new persistent data without changing the ROM binary. The driver test for persistent data layout calls
PersistentData::assert_matches_layout() without the "runtime" feature. This PR adds a runtime test that calls assert_matches_layout() so that we can test the fields in PersistentData behind the "runtime" flag.

fixes #1185

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_boot.rs
+++ b/runtime/tests/runtime_integration_tests/test_boot.rs
@@ -34,6 +34,15 @@ fn test_boot() {
 }
 
 #[test]
+/// This test differs from the drivers' test_persistent() in that it is ran with the "runtime" flag so
+/// it allows us to test conditionally compiled runtime-only persistent data that ROM/FMC may have corrupted.
+fn test_persistent_data() {
+    let mut model = run_rt_test(Some(&firmware::runtime_tests::PERSISTENT_RT), None, None);
+
+    model.step_until_exit_success().unwrap();
+}
+
+#[test]
 fn test_fw_version() {
     let mut model = run_rt_test(None, None, None);
     model.step_until(|m| {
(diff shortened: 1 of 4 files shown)


runtime/tests/runtime_integration_tests/test_boot.rs
```diff
@@ -33,6 +33,13 @@ fn test_boot() {
     model.step_until_exit_success().unwrap();
 }
 
+#[test]
+fn test_persistent_data() {
```
jhand2, line 37: Would be good to comment on why this test is different from https://github.com/chipsalliance/caliptra-sw/blob/f845e85c1c3192823eae16e37ff31b0362d18efe/drivers/test-fw/src/bin/persistent_tests.rs#L9C1-L9C1. I think basically the idea is to make sure ROM/FMC didn't corrupt something in persistent data. Which I agree is important.

jhand2 (approved):

runtime/tests/runtime_integration_tests/test_boot.rs
sree-revoori1: Done

jhand2 (approved):

PR #1193, opened 2023-12-13
Turn the mailbox loop into WFI

ArthurHeymans:
- Set up interrupt for mailbox CMD:
  * MSTATUS
  * MEIE
  * Disable MISC clock gating
  * PIC priority and enabled for SOC interrupts
  * Enable SOC MBOX notif interrupts

- Enter halted state with interrupts enabled after handling mailbox


runtime/src/lib.rs
```diff
@@ -102,13 +103,15 @@ fn enter_idle(drivers: &mut Drivers) {
...
+        return;
     }
 
-    // TODO: Enable interrupts?
-    //#[cfg(feature = "riscv")]
-    //unsafe {
-    //core::arch::asm!("wfi");
-    //}
+    #[cfg(feature = "riscv")]
+    if cfg!(feature = "fpga_realtime") {
```
jlmahowa-amd, line 111: This cfg statement wouldn't run on real silicon.

cpu/src/csr.rs
```diff
@@ -0,0 +1,25 @@
+// Licensed under the Apache-2.0 license
+
+#[cfg(feature = "riscv")]
```
jhand2, line 3: Could you link to the relevant VeeR manual sections in comments in this file to make it easier for folks reading this code?

cpu/src/csr.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn mie_enable_external_interrupts() {
+    const MEIE: usize = 1 << 11;
+    unsafe {
+        core::arch::asm!("csrrs zero, mie, {r}", r = in(reg) MEIE);
+    }
+}
+
+#[cfg(feature = "riscv")]
```
jhand2, line 11: nit: IIUC, these mechanisms are veer-specific right? Should we gate on a `veer` feature flag rather than a generic `riscv` feature flag?

runtime/src/lib.rs
```diff
@@ -186,6 +189,23 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+
+        // Misc clock gating breaks the state of MEPC so disable it
+        caliptra_cpu::csr::mcgc_disable_misc_cg();
+
+        // Set highest priority so that Int can wake CPU
+        drivers.pic.int_set_max_priority(IntSource::SocIfcNotif);
+        drivers.pic.int_enable(IntSource::SocIfcNotif);
+
+        drivers.soc_ifc.enable_mbox_notif_interrupts();
+    }
```
jhand2, line 207: nit: I would prefer making this a top level helper rather than defining a function inside this funciton

runtime/src/lib.rs
```diff
@@ -102,13 +103,15 @@ fn enter_idle(drivers: &mut Drivers) {
                 Err(e) => caliptra_drivers::report_fw_error_non_fatal(e.into()),
             }
         }
+        // Don't enter low power mode when in progress
+        return;
```
jhand2, line 107: If we enter the `Ok` case on line 102, I think we do want to enter a sleep state (because our in-progress operation was completed). In that case we don't need to do anything else until the caller calls `GET_RESULTS`.

So I think we just want this return if the case where `drivers.mbox.lock() == true` (the else case).

runtime/src/main.rs
```diff
@@ -78,17 +78,27 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
...
+
+    let soc_ifc_regs = soc_ifc.regs_mut();
+    let intr_block = soc_ifc_regs.intr_block_rf();
+
+    // TODO implement MMIO reg for emulator
+    if cfg!(feature = "fpga_realtime") {
+        let notif_interrupt_status = intr_block.notif_internal_intr_r();
+
+        // On wake from mailbox command, clear the interrupt status and mret to continue execution
+        if notif_interrupt_status.read().notif_cmd_avail_sts() && trap_record.mcause == 0xF000_1002
```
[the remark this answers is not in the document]
ArthurHeymans: Done

runtime/src/lib.rs
```diff
@@ -186,6 +189,23 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
 pub fn handle_mailbox_commands(drivers: &mut Drivers) -> CaliptraResult<()> {
+    #[cfg(feature = "riscv")]
+    // TODO implement in emulator
+    fn setup_mailbox_wfi(drivers: &mut Drivers) {
+        use caliptra_drivers::IntSource;
+
+        caliptra_cpu::csr::mie_enable_external_interrupts();
+
+        // Misc clock gating breaks the state of MEPC so disable it
+        caliptra_cpu::csr::mcgc_disable_misc_cg();
```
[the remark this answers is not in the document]
ArthurHeymans: Removed

runtime/src/lib.rs
ArthurHeymans: Done

runtime/src/lib.rs
ArthurHeymans: Done

jhand2 (dismissed):
LGTM. Remaining comments are minor.

jhand2:
@ArthurHeymans it looks like the register code needs to be regenerated:

```
"registers/src/el2_pic_ctrl.rs" does not match the generator output. If this is unexpected, ensure that the caliptra-rtl submodule is pointing to the correct commit and/or run "git submodule update". Otherwise, run registers/update.sh to update this file.
```

ArthurHeymans:
> @ArthurHeymans it looks like the register code needs to be regenerated:
> 
> ```
> "registers/src/el2_pic_ctrl.rs" does not match the generator output. If this is unexpected, ensure that the caliptra-rtl submodule is pointing to the correct commit and/or run "git submodule update". Otherwise, run registers/update.sh to update this file.
> ```

No CI tests with the 'main' submodule afaict. So a rebase is needed.

jhand2 (approved):

jhand2 (approved):

mhatrevi (approved):

korran:
I fear this PR has made the FPGA tests flaky. Seeing lots of different tests freeze up before processing a command:

https://github.com/chipsalliance/caliptra-sw/actions/runs/7791077742/job/21246531267?pr=1315

jhand2:
> I fear this PR has made the FPGA tests flaky. Seeing lots of different tests freeze up before processing a command:
> 
> https://github.com/chipsalliance/caliptra-sw/actions/runs/7791077742/job/21246531267?pr=1315

Looking. This looks like it just impacts `test_fips_shutdown`, right? Which has special handling here, so it may be a fairly straightforward fix.

korran:
> Looking. This looks like it just impacts test_fips_shutdown, right? 

No, here's another example:

https://github.com/chipsalliance/caliptra-sw/actions/runs/7791077742/job/21247171576

I'm not sure whether this PR or #1297 is to blame. Need more data...

korran:
> I'm not sure whether this PR or https://github.com/chipsalliance/caliptra-sw/pull/1297 is to blame. Need more data...

Smells like #1297. It has failed 2 out of 3 runs, this PR has passed 3 out of 3.

korran:
Update:

I think it's actually the combination of #1297 and #1193 that's to blame for these test failures:

https://github.com/chipsalliance/caliptra-sw/blob/d0645c4d5ae7abd52edc5e739e2303172485f1f3/runtime/src/lib.rs#L268-L272

For proper glitch protection, we probably want CfiCounter::delay() to be after enter_idle(). By randomizing the amount of delay before we sleep, we're probably exposing a race condition in the wfi code (my guess is that the interrupt is firing before we go to sleep). #1316 does this.

However, we still have the race condition to worry about. I believe the correct fix is to not set mie directly but instead only enable interrupts when going to sleep by setting the haltie bit in the mpmc register when going to sleep.

With regards to why we're getting the NMI, our veer instance has been configured with RV_FAST_INTERRUPT_REDIRECT:

https://github.com/chipsalliance/caliptra-rtl/blob/bbc9a329abb109abee5befc39419a5336672d036/src/riscv_core/veer_el2/rtl/common_defines.sv#L212

This means that the regular risc-v external interrupt dispatch is ignored, and instead the CPU is using a vector table from the `meivt`. Because we don't set this, it's likely that it's using address 0, and jumping to crazy-land, causing the unexpected NMI. 

See section 6.6.1 "Fast Interrupt Redirect" in the veer docs for more info.

PR #1195, opened 2023-12-13
Test that attestation fails after calling DisableAttestation

### LINKED ISSUE #1115: [RT Test] Check that the RT alias key is different from the key signing DPE certs

### LINKED ISSUE #1117: [RT Test] Verify that DPE attestation flow fails after DISABLE_ATTESTATION is called

sree-revoori1:
There are a couple of bugs related to disable_attestation which this PR fixes:

1. The rt_pub_key does not match KEY_ID_RT_PRIV_KEY if disable attestation is called. This is due to the rt_pub_key being taken from the FHT, meaning the pub key matches the original RT alias key derived in FMC. In DisableAttestation, we re-generate the private key meaning the pub key would no longer match the new private key. To fix this, we update the rt_pub_key in the FHT when disable attestation is called.
2. The attestation_disabled flag gets cleared upon update/warm resets. In order to re-enable attestation, Caliptra needs to undergo cold reset. But since the attestation_disabled flag is not in persistent memory, it gets cleared even on warm/update reset. To fix this, we move attestation_disabled to persistent memory. 

fixes #1117 
fixes #1115


runtime/tests/runtime_integration_tests/test_disable.rs
```diff
@@ -30,4 +36,31 @@ fn test_disable_attestation_cmd() {
...
+    };
+
+    let certify_key_cmd = CertifyKeyCmd {
+        handle: ContextHandle::default(),
+        label: TEST_LABEL,
+        flags: CertifyKeyFlags::empty(),
+        format: CertifyKeyCmd::FORMAT_X509,
+    };
+    // the signature should be unable to be verified so expect a CFI error triggered
+    // by the ecc384 driver
```
jhand2, line 59: Hmm, that's unexpected. I would expect the driver to be able to verify the signature. The CDI used to derive the key won't be the same, so the RT pubkey won't match the key in the RT cert. But the key itself should be usable to sign stuff, right?

sree-revoori1: It's because the RT alias public key which we get from the FHT no longer matches the zeroed private key, so verification fails in the driver itself.

sree-revoori1: I am now updating the rt pub key whenever disable attestation is called so that the pub key and the priv key match.

drivers/src/persistent.rs
```diff
@@ -74,6 +76,17 @@ pub struct PersistentData {
...
+    #[cfg(feature = "runtime")]
+    pub rt_pub_key: Ecc384PubKey,
+    #[cfg(feature = "runtime")]
+    pub attestation_disabled: U8Bool,
+    #[cfg(feature = "runtime")]
+    reserved8: [u8; memory_layout::DISABLE_ATTESTATION_SIZE as usize
+        - size_of::<Ecc384PubKey>()
+        - size_of::<U8Bool>()],
+    #[cfg(not(feature = "runtime"))]
+    rt_pub_key: [u8; memory_layout::DISABLE_ATTESTATION_SIZE as usize],
```
jhand2, line 89: If we do the above, we can probably just modify the FHT in-place (or store in the drivers struct). Since we'll be updating it on every boot anyway.

runtime/tests/runtime_integration_tests/test_disable.rs
```diff
@@ -1,33 +1,137 @@
...
+        &BigNum::from_slice(&certify_key_resp.derived_pubkey_x).unwrap(),
+        &BigNum::from_slice(&certify_key_resp.derived_pubkey_y).unwrap(),
+    )
+    .unwrap();
+    // check that signature is unable to be verified by the pub key
+    assert!(!sig.verify(&TEST_DIGEST, &ecc_pub_key).unwrap());
+}
+
+#[test]
+fn test_attestation_disabled_flag_after_update_reset() {
```
jhand2, line 91: In this test we should also make sure that attestation still fails.

drivers/src/persistent.rs
```diff
@@ -74,6 +76,17 @@ pub struct PersistentData {
 
     #[cfg(not(feature = "runtime"))]
     pcr_reset: [u8; memory_layout::PCR_RESET_COUNTER_SIZE as usize],
+
+    #[cfg(feature = "runtime")]
+    pub rt_pub_key: Ecc384PubKey,
+    #[cfg(feature = "runtime")]
+    pub attestation_disabled: U8Bool,
```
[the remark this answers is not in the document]
sree-revoori1: Good catch, done.

drivers/src/persistent.rs
sree-revoori1: Done

runtime/tests/runtime_integration_tests/test_disable.rs
sree-revoori1: Done

jhand2 (approved):

korran (approved):
Approval for drivers/

PR #1197, opened 2023-12-14
Add test triggering CPU fault in RT

### LINKED ISSUE #1176: [RT Test] Add test for RT global exception

sree-revoori1:
This test differs from the ROM cpu fault test in that it uses a test only mailbox command to make an unsafe write to a null pointer, thereby triggering a cpu fault and the
RUNTIME_GLOBAL_EXCEPTION.

We can't copy the ROM cpu fault test due to image verification - any changes to the RT firmware image will be caught upon cold reset and an error will be thrown before triggering a cpu fault.

fixes #1176

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -25,6 +25,7 @@ impl CommandId {
     pub const QUOTE_PCRS: Self = Self(0x50435251); // "PCRQ"
 
     pub const TEST_ONLY_HMAC384_VERIFY: Self = Self(0x484D4143); // "HMAC"
+    pub const TEST_ONLY_TRIGGER_CPU_FAULT: Self = Self(0x54504355); // "TCPU"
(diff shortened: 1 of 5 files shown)


jhand2:
> We can't copy the ROM cpu fault test due to image verification - any changes to the RT firmware image will be caught upon cold reset and an error will be thrown before triggering a cpu fault.

I think after you've modified the ELF, you can regenerate the image header like this: https://github.com/chipsalliance/caliptra-sw/blob/3f5b5ffda182b53fbb4da6b85e32b56cf4371954/rom/dev/tests/rom_integration_tests/test_image_validation.rs#L2077

This will resign the image so it should pass verification. Let me know if this isn't too involved. I'd prefer not to have too many test-only commands if we can avoid it. Especially ones that do dangerous things.

PR #1198, opened 2023-12-14
Mark min-svn fields as deprecated.

bluegate010:


image/verify/src/verifier.rs
```diff
@@ -714,7 +714,8 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
                 Err(CaliptraError::IMAGE_VERIFIER_ERR_FMC_SVN_GREATER_THAN_MAX_SUPPORTED)?;
             }
 
-            if verify_info.svn < verify_info.min_svn {
+            // TODO: remove this check.
+            if verify_info.svn < verify_info.deprecated_min_svn {
```
korran, line 718: Aha, so we are looking at this field...

bluegate010: Yes but it's just a sanity check, comparing two numbers in the signed header against each other. No other logic being performed on the field. Would be a safe two-liner change to just remove this field.

FerralCoder (approved):

builder/src/lib.rs
```diff
@@ -408,10 +408,10 @@ pub fn elf_size(elf_bytes: &[u8]) -> io::Result<u64> {
 #[derive(Clone)]
 pub struct ImageOptions {
     pub fmc_version: u32,
-    pub fmc_min_svn: u32,
+    pub deprecated_fmc_min_svn: u32,
```
mhatrevi, line 411: Can we name this 'reserved'?

drivers/src/fuse_log.rs
```diff
@@ -19,16 +19,16 @@ use zeroize::Zeroize;
...
-    FuseFmcSvn = 5,                 // 4 bytes
-    ManifestRtSvn = 6,              // 4 bytes
-    ManifestRtMinSvn = 7,           // 4 bytes
-    FuseRtSvn = 8,                  // 4 bytes
-    VendorLmsPubKeyIndex = 9,       // 4 bytes  (From Manifest)
-    VendorLmsPubKeyRevocation = 10, // 4 bytes  (From Fuse)
+    VendorEccPubKeyIndex = 1,        // 4 bytes  (From Manifest)
+    VendorEccPubKeyRevocation = 2,   // 4 bytes  (From Fuse)
+    ManifestFmcSvn = 3,              // 4 bytes
+    ManifestDeprecatedFmcMinSvn = 4, // 4 bytes
```
mhatrevi, line 25: Can we name this to 'reserved'?

image/elf/src/lib.rs
```diff
@@ -24,7 +24,7 @@ use std::path::PathBuf;
 pub struct ElfExecutable {
     version: u32,
     svn: u32,
-    min_svn: u32,
+    deprecated_min_svn: u32,
```
mhatrevi, line 27: This can be removed, right?

image/elf/src/lib.rs
```diff
@@ -140,9 +140,9 @@ impl ImageGenratorExecutable for ElfExecutable {
         self.svn
     }
 
-    /// Executable Minimum Security Version Number
-    fn min_svn(&self) -> u32 {
-        self.min_svn
+    /// Executable deprecated minimum SVN
+    fn deprecated_min_svn(&self) -> u32 {
```
mhatrevi, line 144: We can remove this.

image/types/src/lib.rs
```diff
@@ -372,8 +372,8 @@ pub struct ImageTocEntry {
     /// Security Version Number
     pub svn: u32,
 
-    /// Minimum Security Version Number
-    pub min_svn: u32,
+    /// Deprecated field, must be less than or equal to SVN.
+    pub deprecated_min_svn: u32,
```
mhatrevi, line 376: Can we name this to 'reserved'?

image/verify/src/verifier.rs
mhatrevi: Can we remove this?

image/verify/src/verifier.rs
```diff
@@ -749,7 +750,7 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
 
         let log_info: ImageSvnLogInfo = ImageSvnLogInfo {
             manifest_svn: verify_info.svn,
-            manifest_min_svn: verify_info.min_svn,
+            manifest_deprecated_min_svn: verify_info.deprecated_min_svn,
```
mhatrevi, line 753: We should just rename the field to 'reserved' in the image and remove it from verifier function, test cases etc.

mhatrevi (changes requested):
IMO, we should rename the field to 'reserved' in the image and remove it from other places (verifier, tests etc)

image/verify/src/verifier.rs
bluegate010: Will do in a follow-up.

builder/src/lib.rs
bluegate010: Have removed.

bluegate010:
> IMO, we should rename the field to 'reserved' in the image and remove it from other places (verifier, tests etc)

Have removed test cases and builder code related to this, marked the field as reserved in the firmware header and fuse log record, and will issue a follow-up PR to remove the remaining logic in ROM that operates on this field.

image/verify/src/verifier.rs
```diff
@@ -749,7 +750,7 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
 
         let log_info: ImageSvnLogInfo = ImageSvnLogInfo {
             manifest_svn: verify_info.svn,
-            manifest_min_svn: verify_info.min_svn,
+            reserved: 0,
```
korran, line 753: This is probably causing your ROM diff...

image/verify/src/verifier.rs
```diff
@@ -827,7 +829,7 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
```
korran, line 832: and here

image/verify/src/verifier.rs
bluegate010: Ah, fixed.

mhatrevi (approved):

rusty1968 (approved):

korran (approved):

PR #1200, opened 2023-12-14
Use sha2::Sha384 to hash in RT tests instead of mbox responder

### LINKED ISSUE #1125: [RT Test] Check that PCR31 is updated in STASH_MEASUREMENT

### LINKED ISSUE #1166: Investigate Sha2::Sha384 hashes not matching Sha384 driver hashes

sree-revoori1:
Previously, we were using the sha384 driver via mbox responder for hashing in RT tests since the sha2::Sha384 hash wasn't matching what we were getting from the driver. I discovered that this is due to the endianness being different.

Now, we can use sha2::Sha384 instead of mbox responder if we just flip the endianness of the sha2::Sha384 hash.

Also, add a test to check that PCR31 is extended in stash_measurement. This was previously attempted, but I couldn't get the test to pass, again due to endianness being flipped. 

fixes #1166 
fixes #1125

### DIFF
--- a/runtime/test-fw/src/mbox_responder.rs
+++ b/runtime/test-fw/src/mbox_responder.rs
@@ -6,7 +6,10 @@
 use core::mem::size_of;
 
 use caliptra_common::mailbox_api::CommandId;
-use caliptra_drivers::{pcr_log::RT_FW_JOURNEY_PCR, Array4x12};
+use caliptra_drivers::{
+    pcr_log::{PCR_ID_STASH_MEASUREMENT, RT_FW_JOURNEY_PCR},
+    Array4x12,
+};
 use caliptra_registers::{mbox::enums::MboxStatusE, soc_ifc::SocIfcReg};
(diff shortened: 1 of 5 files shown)


runtime/test-fw/src/mbox_responder.rs
```diff
@@ -62,12 +65,11 @@ fn mbox_responder() {
                 mbox.write_response(digest.as_bytes()).unwrap();
```
jhand2, line 65: Instead of using `flip_endianness`, I think you should use `let out: [u8; 48] = digest.into();`. This will do the endianness conversion as from of the `From` trait implementation: https://github.com/chipsalliance/caliptra-sw/blob/23cc74f49ca36e82b3963f2888208244012528ec/drivers/src/array.rs#L110

sree-revoori1: Done

jhand2 (approved):

PR #1207, opened 2023-12-18
DPE: Use DPE as a dependency crate

ArthurHeymans:
Rather use DPE as a crate dependency than a git submodule, as it makes it easier to handle.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,10 +35,6 @@ jobs:
         run: |
           echo "Build-Test: release_ref=$(git rev-parse HEAD)"
 
-      - name: Pull dpe submodule
-        run: |
-          git submodule update --init dpe
-
       - name: Install required packages
         run: |
           sudo apt-get install libftdi1-dev libusb-1.0-0-dev golang-1.20-go
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -327,10 +327,6 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
 
-      - name: Pull dpe submodule
-        run: |
-          git submodule update --init dpe
-
       - name: 'Download FPGA Bitstream Artifact'
         uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/fw-test-emu.yml
+++ b/.github/workflows/fw-test-emu.yml
@@ -45,10 +45,6 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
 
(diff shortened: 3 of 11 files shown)


ArthurHeymans:
DPE contains both go and rust code, which CI both use. Maybe it's possible to parse the cargo lock file and update the go deps for CI, but then a submodule might be easier after all.

ArthurHeymans:
> DPE contains both go and rust code, which CI both use. Maybe it's possible to parse the cargo lock file and update the go deps for CI, but then a submodule might be easier after all.

Cargo can print out the used DPE version which in turn can be used for the golang dep.

jhand2 (changes requested):
@ArthurHeymans The issue is that there's still a fair bit of churn, so there are likely to be changes in that repo that will break the build here. Using a submodule allows us target a specific commit so we can wait to update the submodule in a PR that also fixes up any of the API calls from caliptra-sw

ArthurHeymans:
> @ArthurHeymans The issue is that there's still a fair bit of churn, so there are likely to be changes in that repo that will break the build here. Using a submodule allows us target a specific commit so we can wait to update the submodule in a PR that also fixes up any of the API calls from caliptra-sw

Cargo also features this "cargo update -p dpe --precise <commit hash>" and the Cargo.lock file should use only a specific commit.

jhand2:
> > @ArthurHeymans The issue is that there's still a fair bit of churn, so there are likely to be changes in that repo that will break the build here. Using a submodule allows us target a specific commit so we can wait to update the submodule in a PR that also fixes up any of the API calls from caliptra-sw
> 
> Cargo also features this "cargo update -p dpe --precise " and the Cargo.lock file should use only a specific commit.

We also need to keep the golang dependency in sync with the rust dependency. This seems more complicated than just using a submodule.

Is there a specific thing you're trying to do that is difficult with a submodule?

ArthurHeymans:
> > > @ArthurHeymans The issue is that there's still a fair bit of churn, so there are likely to be changes in that repo that will break the build here. Using a submodule allows us target a specific commit so we can wait to update the submodule in a PR that also fixes up any of the API calls from caliptra-sw
> > 
> > 
> > Cargo also features this "cargo update -p dpe --precise " and the Cargo.lock file should use only a specific commit.
> 
> We also need to keep the golang dependency in sync with the rust dependency. This seems more complicated than just using a submodule.
> 
> Is there a specific thing you're trying to do that is difficult with a submodule?

Not really but I just often forget to do update submodule.

ArthurHeymans:
Probably not worth it as both rust and go programs depend on the submodule

PR #1210, opened 2023-12-19
libcaliptra: adding tests to exercise all commands

nquarton:


fdamato (approved):
Maybe this in future could be modularized a little bit more. 
For example we can have a table {.test. .params. support_flags} // support_flags will indicate for now whether the test is supported by ROM or FW (or both).  then you can have a single function test_all_commands(flags).... that can be called twice (once for ROM and once for RTFW).
However I don`t think it should gate this commit

PR #1218, opened 2023-12-26
Add initial support for mailbox notify interrupts in the sw-emulator

ArthurHeymans:


ArthurHeymans:
Updated this to implement a halted state.

jlmahowa-amd (approved):
Are the TODO's still relavent?

ArthurHeymans:
> Are the TODO's still relavent?

Those indeed need to be updated. Thanks

jhand2 (commented):
Just a couple small comments. But @korran should also look since I'm not too familiar with how the hardware works here.

sw-emulator/lib/cpu/src/pic.rs
```diff
@@ -0,0 +1,559 @@
...
+    }
+    fn set(&self, idx: u8, val: bool) {
+        let mask = 1 << idx;
+        if val {
+            self.bits.set(self.bits.get() | mask);
+        } else {
+            self.bits.set(self.bits.get() & !mask);
+        }
+    }
+}
```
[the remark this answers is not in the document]
korran: I don't think that's useful here.

1. bitfield is meant to implement types that give strongly typed access to explicit bits (or bit ranges) inside an integer. What we need here is indexed access to arbitrary bits.

2. Bits32 uses interior mutability (Cell<u32>) for performance reasons, allowing it to be easily modified through multiple &Bit32 references to the same value.

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -408,27 +412,41 @@ impl<TBus: Bus> Cpu<TBus> {
...
                 TimerAction::SetGlobalIntEn { en } => self.global_int_en = *en,
                 TimerAction::SetExtIntEn { en } => self.ext_int_en = *en,
+                TimerAction::Halt => self.halted = true,
                 _ => {}
             }
         }
 
+        // We are in a halted state. Don't continue executing but poll the bus for interrupts
+        if self.halted {
+            self.clock.timer().schedule_poll_in(1);
```
korran, line 445: For efficiency reasons, this isn't how this was intended to work. Peripherals should schedule their own polls for actions that will fire interrupts.

For example, a timer peripheral might schedule a poll when the timer expires, and then when polled, fire the interrupt.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -328,6 +405,27 @@ impl Bus for SocRegistersInternal {
 
     fn poll(&mut self) {
         self.regs.borrow_mut().poll();
+
+        let mut regs = self.regs.borrow_mut();
+        if regs.mailbox.get_notif_irq() {
```
korran, line 412: One change I might consider making in the future is to put this filtering state inside some sort of Rc<SocIrqFilter>, and just have peripherals call a .set_irq(true) on it to dispatch interrupts.

sw-emulator/lib/cpu/src/cpu.rs
korran: I think this might still work if you delete this line.

korran:
Looks great!

sw-emulator/lib/cpu/src/csr_file.rs
```diff
@@ -165,10 +181,45 @@ impl CsrFile {
...
+                }
+                if addr == Csr::MSTATUS as usize {
+                    self.timer.schedule_action_in(
+                        0,
+                        TimerAction::SetGlobalIntEn {
+                            en: val & MSTATUS_MIE != 0,
+                        },
+                    );
+                    // Let's see if the soc wants to interrupt
+                    self.timer.schedule_poll_in(2);
```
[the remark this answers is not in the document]
korran: Yes, this is the best way to do this. The fact that we can delay it a few cycles is nice because it gives us the option to make things behave more like the real hardware.

sw-emulator/lib/cpu/src/cpu.rs
ArthurHeymans: Done. Previous iterrations didn't schedule a poll inside the mailbox, but now that's the case.

sw-emulator/lib/cpu/src/pic.rs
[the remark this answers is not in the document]
jhand2: Ok sounds good. Happy to leave this as-is

jhand2 (approved):

korran (approved):

PR #1230, opened 2024-01-06
Implement hash chain derivation in FMC.

bluegate010:
This commit sets the max SVN value to 64. A hash chain of that length costs ~160k additional cycles to derive.

This commit also allocates reserved owner-signed bits for an epoch field, which is used when deriving the hash chain root.


bluegate010:
Will be mailing this as sub-PRs.

PR #1239, opened 2024-01-08
Add runtime crypto routines for hashing and HMACing data from KV.

bluegate010:
Due to hardware limitations these use HMAC and ECDSA keygen, respectively.

### DIFF
--- a/runtime/src/hmac.rs
+++ b/runtime/src/hmac.rs
@@ -1,0 +1,119 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_common::{crypto::Ecc384KeyPair, keyids::KEY_ID_TMP};
+use caliptra_drivers::{
+    hmac384_kdf, Array4x12, Ecc384PrivKeyOut, Ecc384PubKey, Hmac384Data, Hmac384Key, Hmac384Tag,
+    KeyId, KeyReadArgs, KeyUsage, KeyWriteArgs,
+};
+use caliptra_error::CaliptraResult;
+use zerocopy::AsBytes;
+use zeroize::Zeroize;
+
+use crate::Drivers;
+
+// Generate an ECC keypair
+fn ecc384_key_gen(
+    drivers: &mut Drivers,
+    input: KeyId,
+    label: &[u8],
+    priv_key: KeyId,
+) -> CaliptraResult<Ecc384KeyPair> {
+    hmac384_kdf(
+        &mut drivers.hmac384,
+        KeyReadArgs::new(input).into(),
+        label,
+        None,
+        &mut drivers.trng,
+        KeyWriteArgs::new(
+            KEY_ID_TMP,
+            KeyUsage::default()
+                .set_hmac_key_en()
+                .set_ecc_key_gen_seed_en(),
+        )
+        .into(),
+    )?;
+
+    let pub_key = drivers.ecc384.key_pair(
+        &KeyReadArgs::new(KEY_ID_TMP).into(),
+        &Array4x12::default(),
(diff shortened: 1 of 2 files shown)


runtime/src/crypto.rs
jhand2, line 1: nit: I'd prefer to make this specifically an `hmac` interface rather than a generic crypto driver.

runtime/src/crypto.rs
```diff
@@ -0,0 +1,128 @@
...
+    hmac384_kdf, Array4x12, Ecc384PrivKeyOut, Ecc384PubKey, Hmac384Data, Hmac384Key, Hmac384Tag,
+    KeyId, KeyReadArgs, KeyUsage, KeyWriteArgs,
+};
+use caliptra_error::CaliptraResult;
+use zerocopy::AsBytes;
+use zeroize::Zeroize;
+
+use crate::Drivers;
+
+pub enum Crypto {}
```
jhand2, line 14: why enum instead of struct?

runtime/src/crypto.rs
```diff
@@ -0,0 +1,128 @@
...
+                    .set_ecc_key_gen_seed_en(),
+            )
+            .into(),
+        )
+    }
+
+    // Perform an "HMAC" with a key from KV by first using it to derive an
+    // ECC keypair, then hashing the public key coordinates into an HMAC key.
+    // This roundabout mechanism is necessary because the hardware does not
+    // directly support exposing an HMAC computed with key material from KV.
```
jhand2, line 93: probably worth a note that this is only secure because the generated public key is kept secret. Just so future contributors know the public key shouldn't be exposed.

runtime/src/crypto.rs
```diff
@@ -0,0 +1,128 @@
...
+            &Hmac384Data::Slice(data),
+            &mut drivers.trng,
+            Hmac384Tag::Array4x12(&mut hmac_output),
+        )?;
+
+        // Clean up state.
+        unsafe { caliptra_drivers::Sha384::zeroize() }
+        pubkey_digest.zeroize();
+        keypair.pub_key.zeroize();
+        drivers.key_vault.erase_key(keypair.priv_key)?;
```
jhand2, line 124: if `drivers.hmac384.hmac` fails, these things won't be zeroized

jhand2 (changes requested):

runtime/src/crypto.rs
bluegate010: It was the pattern I see on other crypto implementations in [FMC](https://github.com/chipsalliance/caliptra-sw/blob/e61eb66bfcd42e56438c1548272c7b568ce9f8d0/fmc/src/flow/crypto.rs#L16) and [ROM](https://github.com/chipsalliance/caliptra-sw/blob/e61eb66bfcd42e56438c1548272c7b568ce9f8d0/rom/dev/src/flow/cold_reset/crypto.rs#L50). Wasn't sure what was more idiomatic.

runtime/src/crypto.rs
bluegate010: Fixed.

runtime/src/crypto.rs
bluegate010: Added a note.

runtime/src/crypto.rs
bluegate010: After reading up it looks like this pattern prevents someone from making an instance of a `Crypto`.

runtime/src/crypto.rs
bluegate010: Done

jhand2 (approved):

PR #1245, opened 2024-01-09
Improve Mbox responder

### LINKED ISSUE #1174: Improve RT Mbox responder
- Add trap handler and NMI handler
- Handle errors better and write to error registers
- Define command IDs somewhere

sree-revoori1:
This PR adds better error handling and cleaner code flow to the mbox responder.

fixes #1174


runtime/test-fw/src/mbox_responder.rs
```diff
@@ -1,26 +1,91 @@
...
 use zerocopy::{AsBytes, FromBytes};
 
 const FW_LOAD_CMD_OPCODE: u32 = CommandId::FIRMWARE_LOAD.0;
 
-fn mbox_responder() {
-    let mut drivers = unsafe { Drivers::new_from_registers().unwrap() };
-    assert!(drivers.persistent_data.get().fht.is_valid());
-    let mut mbox = drivers.mbox;
+#[cfg(feature = "std")]
+pub fn main() {}
```
jhand2, line 22: In what circumstances would this be build with `std` enabled? Especially since if it is, I think it wouldn't do anything since the main implementation is empty.

sree-revoori1: Removed.

jhand2 (approved):

PR #1249, opened 2024-01-10
Initialize SHA hardware before PCR extensions in SHA384 driver

jhand2:
To workaround https://github.com/chipsalliance/caliptra-rtl/issues/375, the SHA512 peripheral must be reinitialized between gen_pcr_hash and pcr_extend. To simplify when the registers get cleared, always reinitialize the hardware before calling pcr_extend. This isn't done at the end of gen_pcr_hash since other code in firmware still expects to be able to read the PCR hash (such as for signing) after gen_pcr_hash is called.

### DIFF
--- a/drivers/src/sha384.rs
+++ b/drivers/src/sha384.rs
@@ -165,6 +165,10 @@ impl Sha384 {
         // Wait for the registers to be ready
         wait::until(|| status_reg.read().ready());
 
+        // Initialize SHA hardware to clear write lock
+        reg.ctrl().write(|w| w.init(true));
+        wait::until(|| status_reg.read().ready());
+
         if status_reg.read().valid() {
             Ok(reg.gen_pcr_hash_digest().read().into())
         } else {
--- a/drivers/test-fw/src/bin/sha384_tests.rs
+++ b/drivers/test-fw/src/bin/sha384_tests.rs
@@ -264,6 +264,11 @@ fn test_pcr_hash_extend_single_block() {
     ];
     pcr_bank.erase_all_pcrs();
 
+    // Call gen_pcr_hash first to ensure that software workaround for
+    // https://github.com/chipsalliance/caliptra-rtl/issues/375 in SHA
+    // driver works as expected.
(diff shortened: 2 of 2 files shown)


ArthurHeymans:
Are there other use cases of the SHA512 engine where PCR digest generation could cause problems? If so would it make sense to simply add this as a cleanup in the PCR quote runtime command, more precisely after the PCR signature is generated?

ArthurHeymans:
> Are there other use cases of the SHA512 engine where PCR digest generation could cause problems? If so would it make sense to simply add this as a cleanup in the PCR quote runtime command, more precisely after the PCR signature is generated?

I updated https://github.com/chipsalliance/caliptra-sw/pull/1233 . It looks like all SHA512 engine ops are impacted by the PCR digest op.

I created https://github.com/chipsalliance/caliptra-sw/pull/1251 as an alternative . It has the advantage that the frozen images are not changed.

jhand2:
@ArthurHeymans Based on the latest discussion in https://github.com/chipsalliance/caliptra-rtl/issues/375, I updated this PR to call init directly after gen_pcr_hash. Can you confirm that this works with your PCR_EXTEND PR?

Great news if this works since it doesn't require a ROM change (gen_pcr_hash isn't called in ROM).

mhatrevi (approved):

ArthurHeymans:
> @ArthurHeymans Based on the latest discussion in [chipsalliance/caliptra-rtl#375](https://github.com/chipsalliance/caliptra-rtl/issues/375), I updated this PR to call init directly after gen_pcr_hash. Can you confirm that this works with your PCR_EXTEND PR?
> 
> Great news if this works since it doesn't require a ROM change (gen_pcr_hash isn't called in ROM).

This indeed seems to work. https://github.com/ArthurHeymans/caliptra-sw/tree/HW_Workaround_PCR_hash does roughly the same but with a cargo feature to guard that line. If that's not needed this is the better ptach.

ArthurHeymans (approved):

PR #1252, opened 2024-01-10
Add warm reset tests to RT

### LINKED ISSUE #1126: [RT Test] Check that attestation is disabled if mbox_busy during a warm reset

### LINKED ISSUE #1133: [RT Test] Trigger warm reset and check that DPE structure is valid upon RT initialization
Also check persistence of the DPEInstance upon warm reset. We can't really check this as easily for update resets since we update the root context measurement, so the DPE instance before and after update reset would be different, which isn't the case for warm reset.

### LINKED ISSUE #1170: [RT Test] Check that RT Journey PCR unchanged upon warm reset

sree-revoori1:
fixes #1170 
fixes #1133 
fixes #1126


runtime/tests/runtime_integration_tests/test_warm_reset.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    // Wait for boot
+    while !model
+        .soc_ifc()
+        .cptra_flow_status()
+        .read()
+        .ready_for_runtime()
+    {
+        model.step();
+    }
```
jhand2, line 74: nit:
```suggestion
    model.step_until(|m| {
        m.soc_ifc().cptra_flow_status().read().ready_for_runtime()
    });
```

runtime/tests/runtime_integration_tests/test_warm_reset.rs
```diff
@@ -0,0 +1,180 @@
...
+            fmc_key_manifest_svn: 0b1111111,
+            ..Default::default()
+        },
+        fw_image: Some(&image.to_bytes().unwrap()),
+        ..Default::default()
+    })
+    .unwrap();
+
+    // Wait for boot
+    while !model
```
jhand2, line 144: same as above

runtime/tests/runtime_integration_tests/test_warm_reset.rs
sree-revoori1: Done

runtime/tests/runtime_integration_tests/test_warm_reset.rs
sree-revoori1: Done

jhand2 (approved):

PR #1257, opened 2024-01-12
Adding revision and digest for FW, FMC, and ROM to FW_INFO cmd

nquarton:


runtime/src/info.rs
```diff
@@ -17,6 +18,7 @@ impl FwInfoCmd {
         let runtime_svn = handoff.rt_svn()?;
```
rusty1968, line 18: Would failure to report FwInfo be classified as fatal?

bluegate010: I don't think it should be fatal. If the caller considers it fatal they can issue the FIPS shutdown command.

nquarton: I agree I don't think we'd want it to be fatal. Currently it would bubble that result back to the mailbox command handler and report a non-fatal error here:
image

rusty1968: Will all integrators enable FIPS mode without exception ?

nquarton: We may have to find someone else to help answer that. I wasn't aware there was a mode for FIPS. What does this mode affect?

jhand2: Also if this fails they can call DisableAttestation, fail to boot, or fail to forward requests to Caliptra to mitigate any possible security issues.

rusty1968 (approved):

jhand2 (approved):

fdamato (approved):

PR #1265, opened 2024-01-12
Allocate fields for communicating hash chain slot to RT FW.

bluegate010:
As the last pre-existing field was not word-aligned, this also allocates a 16-bit field for communicating the max SVN used by FMC when deriving the hash chain, as a purely informative value.

A follow-up commit will populate these fields from FMC.

### DIFF
--- a/drivers/src/hand_off.rs
+++ b/drivers/src/hand_off.rs
@@ -289,7 +289,19 @@ pub struct FirmwareHandoffTable {
     /// RtAlias TBS Size.
     pub rtalias_tbs_size: u16,
 
+    /// Maximum value RT FW SVN can take.
+    #[cfg(any(feature = "fmc", feature = "runtime"))]
+    pub rt_hash_chain_max_svn: u16,
+
+    /// Index of RT hash chain value in the Key Vault.
+    #[cfg(any(feature = "fmc", feature = "runtime"))]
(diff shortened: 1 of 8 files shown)


drivers/src/hand_off.rs
```diff
@@ -289,7 +289,19 @@ pub struct FirmwareHandoffTable {
...
+
+    /// Index of RT hash chain value in the Key Vault.
+    #[cfg(any(feature = "fmc", feature = "runtime"))]
+    pub rt_hash_chain_kv_hdl: HandOffDataHandle,
+
     /// Reserved for future use.
+    #[cfg(any(feature = "fmc", feature = "runtime"))]
+    pub reserved: [u8; 1636],
+
+    #[cfg(not(any(feature = "fmc", feature = "runtime")))]
```
FerralCoder, line 304: I think it would be better to have these fields defined always, not based on configuration. I feel like it will get tricky to manage if the structure starts looking different depending on how the FW was built.

bluegate010: I'd like to always define these as well, it's just that without these cfg-guards the ROM binary changes.

FerralCoder: ahh, got it.

FerralCoder (approved):

jhand2 (approved):

korran (approved):
Approval for drivers.

PR #1283, opened 2024-01-19
Adding RTL hash tool for release

nquarton:
- Adds tool to generate a hash from the Caliptra RTL
- Adds step in release process to generate hash and add tool to packaged files


nquarton:
The result of these changes is the release contains the rtl_hash.sh script and rtl_hash_file_list.txt in the top-level directory and the release notes have been updated to include the RTL hash calculated during the release.

image

I'm open to any other opinions on where those files should go (I considered the caliptra-rtl directory within the release too) or formatting of the release notes.

jlmahowa-amd (approved):

ci-tools/release/tools/rtl_hash.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+# Get args
+rtl_path=$1
+expected_file_list=$2
+
+# Read expected file list
+expected_file_list=$(cat $expected_file_list)
+# Prepend rtl path
+expected_file_list=$(echo "$expected_file_list" | sed "s@^@"$rtl_path"/@")
```
korran, line 38: Why not replace the above 7 lines with:

```
expected_file_list=$(cat "$1" | sed "s@^@"$rtl_path"/@")
```

ci-tools/release/tools/rtl_hash.sh
```diff
@@ -0,0 +1,62 @@
...
+			echo "Missing expected files: "
+		fi
+		missing_files=$(($missing_files + 1))
+		echo "  $line"
+	fi
+done < <(echo "$expected_file_list")
+
+# Calculate the hash (only if no files were missing)
+if [ "$missing_files" -eq 0 ]; then
+	hash=$(cat $expected_file_list | sha1sum | tr -d "\n *-")
```
korran, line 56: sha1 is compromised. sha256sum or sha384sum?

nquarton: Honestly the main reason I used this was because it was shorter and matched what git uses for the hashes listed above it in the notes. My understanding was sha1 was okay for non-security based uses like this where it's just necessary to identify something uniquely. But I can see the argument that this tool may have some security implications if we don't trust everyone with access to the RTL during integration.

ci-tools/release/tools/rtl_hash.sh
nquarton: Good point I can make that more concise.

ci-tools/release/tools/rtl_hash.sh
FerralCoder: I would not consider this a security feature. It is already going to have to be taken on trust and faith that an integrator is truly using the RTL they claim to be using.

nquarton: That was my original thinking. I'm open to either way on this though. I like that it's more concise and consistent with git with sha1. But I can switch it to sha256 if you feel more comfortable with it, Kor.

korran: I don't have a strong opinion, other than I know sometimes having sha1 deps can cause pain with NIST.

korran: (As soon as github supports sha256 I would love to migrate the repo)

nquarton:
Thanks for Kor for running a release in his test repo for me. No discrepancies from my local testing and I confirmed I can generate the same RTL hash on the released files on my system with the included copy of the tool.

https://github.com/korran-testorg/caliptra-sw/releases/tag/release_v20240125_0

jlmahowa-amd (approved):

korran (dismissed):

ci-tools/release/tools/rtl_hash.sh
nquarton: This is updated to sha384 now

JohnTraverAmd (approved):

PR #1291, opened 2024-01-25
ROM: Initialize mbox sram in startup assembly.

korran:
This is a workaround for chipsalliance/caliptra-rtl#399.


drivers/src/soc_ifc.rs
```diff
@@ -285,7 +285,7 @@ impl SocIfc {
         self.soc_ifc
             .regs_mut()
             .cptra_flow_status()
-            .write(|w| w.ready_for_runtime(true));
+            .modify(|w| w.ready_for_runtime(true));
```
jhand2, line 288: Just for my education, what's the difference?

rom/dev/src/start.S
```diff
@@ -108,6 +108,33 @@ _start:
...
+    li t1, 1
+    sw t1, 0(t0)
+
+    // Tell the SoC it can use the mailbox, as long as we're not doing CSR
+    // generation
+    li t0, 0x300300bc  // CPTRA_DBG_MANUF_SERVICE_REG
+    lw t1, 0(t0)
+
+    andi t1, t1, 0x1   // MFG_FLAG_GEN_IDEV_CSR
+    bne t1, x0, 1f
```
jhand2, line 131: Is there a specific reason to move this here instead of later once we've jumped to Rust code?

jhand2 (approved):

drivers/src/soc_ifc.rs
korran: .modify() is a ready-modify-write that only sets the ready_for_runtime field. .write() does a write that sets all the other fields to the reset value.

rom/dev/src/start.S
korran: Two reasons:

* We want to clear the mailbox before we clear the ICCM and DCCM so the SoC can stash their measurement as soon as we're done clearing the mailbox. (and we can't clear DCCM from Rust code, as DCCM contains the stack)

* We want to use the _zero_mem256 function, which is much faster than memset.

drivers/src/soc_ifc.rs
korran: See docs at https://github.com/chipsalliance/caliptra-sw/blob/955115deef9fd788b78127a5cc06dc5e8eb69aa4/ureg/src/lib.rs#L506

and

https://github.com/chipsalliance/caliptra-sw/blob/955115deef9fd788b78127a5cc06dc5e8eb69aa4/ureg/src/lib.rs#L588

jhand2: Makes sense, thanks :)

rom/dev/src/start.S
jhand2: I just meant the ready_for_fw part. It seems like the reason is just to allow the SoC to start doing mailbox writes sooner.

korran: Yes, every nanosecond counts when you're trying to meet PCIe reset requirements. Before this PR, they could write to the mailbox immediately after writing fuses. Now they HAVE to wait for ready_for_fw, so let's signal that as soon as we can, rather than forcing them to wait for us to clear iccm and DCCM.

Neal-liu:
If HW/RTL fixes the bug, is this patch still necessary?
Should this patch be merged into all Caliptra 1.x versions since mailbox SRAM uncorrectable ECC errors might occur?
Thanks.

PR #1294, opened 2024-01-29
Update DPE version

sree-revoori1:
1. Rename DeriveChild -> DeriveContext
2. Remove traces of ExtendTci
3. Check DPE context threshold limits in DeriveContext only if the recursive flag is not set
4. Add cert_validity to platform
5. Add subject key identifier to platform


dpe
jhand2, line 1: I assume this needs to be updated with the latest caliptra-dpe commit for the verification tests

runtime/src/dpe_platform.rs
```diff
@@ -96,26 +101,29 @@ impl Platform for DpePlatform<'_> {
...
 
-    fn get_issuer_sn(&mut self, out: &mut [u8; MAX_SN_SIZE]) -> Result<usize, PlatformError> {
-        let sn = self.hashed_rt_pub_key.bytes();
-        let size = min(MAX_SN_SIZE, sn.len());
-        // Prevents potential panic
-        if size > out.len() || size > sn.len() {
-            return Err(PlatformError::IssuerNameError(0));
+    fn get_signer_identifier(&mut self) -> Result<SignerIdentifier, PlatformError> {
+        let mut ski = [0u8; MAX_SKI_SIZE];
+        let hashed_rt_pub_key = self.hashed_rt_pub_key.bytes();
```
jhand2, line 106: nit: Can you link to the code in `x509/` that generates the RT SKI?

runtime/src/invoke_dpe.rs
```diff
@@ -71,11 +71,14 @@ impl InvokeDpeCmd {
                     }
                     cmd.execute(dpe, &mut env, locality)
                 }
-                Command::DeriveChild(cmd) => {
-                    Drivers::is_dpe_context_threshold_exceeded(
-                        pl0_pauser, flags, locality, dpe, false,
-                    )?;
-                    if DeriveChildCmd::changes_locality(&cmd)
+                Command::DeriveContext(cmd) => {
+                    // DeriveContext can only create new contexts if the recursive flag is not set.
```
jhand2, line 75: nit: I think this comment is a bit confusing. Maybe something like:
```suggestion
                    // If the recursive flag is not set, DeriveContext will generate a new context.
                    // If recursive _is_ set, it will extend the existing one, which will not count
                    // against the context threshold.
```

test/dpe_verification/dpe_test.go
```diff
@@ -3,13 +3,15 @@
 package dpe
 
 import (
-	"github.com/chipsalliance/caliptra-dpe/verification"
 	"os"
 	"testing"
+
+	"github.com/chipsalliance/caliptra-dpe/verification/client"
+	DpeTesting "github.com/chipsalliance/caliptra-dpe/verification/testing"
```
jhand2, line 10: nit: Generally golang packages are all lower case. `dpetesting` would probably be the idiomatic option

runtime/src/invoke_dpe.rs
sree-revoori1: Done

test/dpe_verification/dpe_test.go
sree-revoori1: Done

runtime/src/dpe_platform.rs
sree-revoori1: Done

dpe
sree-revoori1: Done

jhand2 (approved):

PR #1296, opened 2024-01-30
FPGA JTAG using gpio

jlmahowa-amd:


.github/workflows/fpga.yml
```diff
@@ -331,6 +331,11 @@ jobs:
         run: |
           git submodule update --init dpe
 
+      - name: Install test dependencies
+        run: |
+          sudo apt install openocd
```
korran, line 336: FWIW, for this to work you would need to add -y here

Doesn't matter now that the FPGA image has these packages...

builder/src/lib.rs
```diff
@@ -183,17 +199,8 @@ pub fn build_firmware_elfs_uncached<'a>(
...
-        } else {
-            Path::new(workspace_dir).join("target")
-        };
-
         for &fwid in invocation.fwids.iter() {
-            result_map.insert(
-                fwid,
-                fs::read(target_dir.join(TARGET).join(PROFILE).join(fwid.bin_name))?,
-            );
+            result_map.insert(fwid, fs::read(get_elf_path(fwid).unwrap())?);
```
korran, line 203: Can you explain a bit more about what's going on here? I don't think we want to look at CALIPTRA_PREBUILT_FW_DIR when we're building; that's likely to cause trouble with some of my shell scripts.

jlmahowa-amd: I think I see the problem now. I had _assumed_ that this would only be run in the non-CALIPTRA_PREBUILT_FW_DIR case, not checking for the other callers. I modified this function to call get_elf_path to avoid having two places where the path is constructed. I'm debating reverting this modification or trying to do it in a different way.

korran (approved):

PR #1297, opened 2024-01-30
Add CFI instrumentation to Runtime

sree-revoori1:
RT binary size delta is + 5564

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -6,9 +6,11 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+caliptra-cfi-lib = { workspace = true, default-features = false, features = ["cfi", "cfi-counter" ] }
+caliptra-cfi-derive.workspace = true
 caliptra_common = { workspace = true, default-features = false, features = ["runtime"] }
 caliptra-cpu.workspace = true
-caliptra-drivers = { workspace = true, features = ["runtime", "no-cfi"] }
+caliptra-drivers = { workspace = true, features = ["runtime"] }
 caliptra-error = { workspace = true, default-features = false }
 caliptra-image-types = { workspace = true, default-features = false }
 caliptra-kat.workspace = true
@@ -20,7 +22,7 @@ platform.workspace = true
 ufmt.workspace = true
 zerocopy.workspace = true
 arrayvec.workspace = true
-caliptra-image-verify = { workspace = true, default-features = false, features = ["no-cfi"] }
+caliptra-image-verify = { workspace = true, default-features = false }
 zeroize.workspace = true
 
 [build-dependencies]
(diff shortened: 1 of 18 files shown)


runtime/src/drivers.rs
```diff
@@ -281,6 +288,11 @@ impl Drivers {
                     return Err(CaliptraError::RUNTIME_GLOBAL_EXCEPTION);
                 }
             }
+        } else {
+            cfi_assert_eq_12_words(
+                &<[u32; 12]>::from(latest_pcr),
+                &<[u32; 12]>::from(latest_pcr),
+            )
```
jhand2, line 295: should this be
```suggestion
            cfi_assert_eq_12_words(
                &<[u32; 12]>::from(latest_tci),
                &<[u32; 12]>::from(latest_pcr),
            )
```

runtime/src/pcr.rs
```diff
@@ -22,6 +23,7 @@ use zerocopy::FromBytes;
 
 pub struct IncrementPcrResetCounterCmd;
 impl IncrementPcrResetCounterCmd {
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
```
jhand2, line 26: I think we also probably want CFI on the following mailbox commands:

* FIPS selftest
* FIPS shutdown
* disable attestation

runtime/src/main.rs
```diff
@@ -51,6 +52,16 @@ pub extern "C" fn entry_point() -> ! {
     };
     caliptra_common::stop_wdt(&mut drivers.soc_ifc);
 
+    if !cfg!(feature = "no-cfi") {
```
jhand2, line 55: Following up on this in chat. I think there's a decent chance the linker will optimize this out to remove the if statement. But just want to be sure, since glitching past CFI initialization could cause some odd behavior.

jhand2: Confirmed

runtime/Cargo.toml
```diff
@@ -6,6 +6,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
```
jhand2, line 8: I think we want to add a `no-cfi` feature defined here and then select `no-cfi` in dependent crates based on that. Right now, CFI is still disabled on some of the dependencies (like drivers and image-verify)

runtime/src/pcr.rs
sree-revoori1: Adding CFI to any of these qill require a refactor so I was planning on doing it in the followup PR.

runtime/src/drivers.rs
sree-revoori1: Done

runtime/Cargo.toml
sree-revoori1: Done

runtime/src/pcr.rs
sree-revoori1: I went back and found that the panic test was failing due to not adding #[inline(never)] in a few places. So now CFI should be complete in RT with this PR (i.e. added to disable attestation, fips stuff, among other places).

jhand2 (dismissed):

jhand2 (approved):

PR #1314, opened 2024-02-05
Add CFI instrumentation to FMC

sree-revoori1:
FMC binary size delta is + 1604 bytes

CFI criteria: Added CFI to any functions that deal with crypto drivers, mutating persistent data, locking registers, or extending PCRs since altering control flow to these sorts of functions could cause security issues. Also added cfi_asserts to places where we check the ResetReason to make sure nothing gets optimized out and instructions don't get skipped via a fault injection. For the most part, I just followed the CFI criteria in ROM and RT.

### DIFF
--- a/fmc/src/flow/rt_alias.rs
+++ b/fmc/src/flow/rt_alias.rs
@@ -98,6 +98,7 @@ impl RtAliasLayer {
         slot == KEY_ID_RT_CDI || slot == KEY_ID_RT_PRIV_KEY || slot == KEY_ID_TMP
     }
 
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
     #[inline(never)]
     pub fn run(env: &mut FmcEnv) -> CaliptraResult<()> {
         cprintln!("[alias rt] Extend RT PCRs");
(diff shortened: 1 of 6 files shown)


fmc/src/flow/rt_alias.rs
```diff
@@ -161,12 +162,25 @@ impl RtAliasLayer {
...
+            ResetReason::UpdateReset => {
+                cfi_assert_eq(reset_reason, ResetReason::UpdateReset);
+                extend_pcr_common(env)
+            }
+            ResetReason::WarmReset => {
+                cfi_assert_eq(reset_reason, ResetReason::WarmReset);
                 cprintln!("[alias rt : skip pcr extension");
                 Ok(())
             }
+            ResetReason::Unknown => {
```
bluegate010, line 180: This is a change in FMC behavior; previously it would have just continued booting. Maybe this is fine, just calling it out.

bluegate010 (approved):

fmc/src/flow/rt_alias.rs
rusty1968: Where is this requirement coming from?

sree-revoori1: We error on UnknownReset in ROM and RT so I was thinking we should also do the same in FMC. I can change it back if it's undesired behavior though.

rusty1968: That's fine. ROM will catch the unknown reset before handing it off. The main goal is to harden the FMC.

sree-revoori1: So should we error on UnknownReset in FMC or go back to how we were handling it before?

rusty1968: It would be nice to spell out the criteria for adding the extra CFI given the penalty in code size.

rusty1968: It is ok to add the extra error handling given the overall goal is to harden the FMC.

rusty1968 (approved):

PR #1318, opened 2024-02-06
runtime: Fixup NMI and race conditions in WFI logic 

korran:
This updates the WFI logic in #1193. Please review each commit independently.

## fmc/runtime: support .data sections.

To support placing a vector table in DCCM, use PT_LOAD segments to build the image instead of explicit sections.

## runtime: Fixup NMI and race conditions in WFI logic

Previously, an unexpected NMI was firing with mcause=0xf0001002 (Machine Fast Interrupt non-DCCM region), because the firmware wasn't setting up the meivt CSR correctly.

Caliptra's VeeR instance has been configured with RV_FAST_INTERRUPT_REDIRECT. This means that the regular risc-v external interrupt dispatch is ignored, and instead the CPU is using a vector table from the meivt register. The NMI fired because we weren't setting this to a DCCM address.

This commit removes the hacks from the NMI handler, sets up a new handler for external interrupts, and ensures that mie is only set when the CPU is asleep, fixing the race condition.


runtime/src/main.rs
```diff
@@ -90,33 +90,21 @@ extern "C" fn exception_handler(trap_record: &TrapRecord) {
...
-
-    // TODO implement MMIO reg for emulator
-    if cfg!(feature = "fpga_realtime") {
-        let notif_interrupt_status = intr_block.notif_internal_intr_r();
-
-        // On wake from mailbox command, clear the interrupt status and mret to continue execution
-        if notif_interrupt_status.read().notif_cmd_avail_sts() && trap_record.mcause == NON_DCCM_NMI
-        // Machine Fast Interrupt non-DCCM region NMI
-        {
-            notif_interrupt_status.modify(|w| w.notif_cmd_avail_sts(true));
```
ArthurHeymans, line 112: Clearing the interrupt status is now never done? Does the CPU not come out of halted state even without new mailbox commands?

korran: I didn't have enough time to validate that we are going to sleep the second time around the event loop. My assumption was that this was a level triggered interrupt. If it is indeed edge triggered, we can ack the interrupt from the event loop. Will look later today

ArthurHeymans: > I didn't have enough time yesterday to validate that we are going to sleep the second time around the event loop. My assumption was that the mailbox notification was a level triggered interrupt. If it is indeed edge triggered, we can ack the interrupt from the event loop. Will look later today

It is level triggered, but it's sticky I believe. https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.intr_block_rf.error_internal_intr_r

korran: The status bit needs to be sticky, but the question is whether the actual signal itself is...

korran: Looks like this is effectively edge-triggered. The interrupt register is set with a pulse here:

https://github.com/chipsalliance/caliptra-rtl/blob/bbc9a329abb109abee5befc39419a5336672d036/src/soc_ifc/rtl/soc_ifc_top.sv#L664

And that register appears to be used to build the soc_ifc_notif_intr wire going to the cpu.

runtime/src/lib.rs
```diff
@@ -277,6 +277,12 @@ pub fn handle_mailbox_commands(drivers: &mut Drivers) -> CaliptraResult<()> {
 
         let cmd_ready = drivers.mbox.is_cmd_ready();
         if cmd_ready {
+            // We are about to handle a mailbox command; clear the sticky
+            // register so we can go back to sleep later.
+            if cfg!(feature = "fpga_realtime") {
+                drivers.soc_ifc.clear_mbox_notif_status();
+            }
```
jhand2, line 284: nit: We could probably clear this inside of `enter_idle` right after `mpmc_halt_and_enable_interrupts`, right? Just thinking it consolidates the interrupt logic a bit.

Also thinking of a case there some interrupt is fired but is_cmd_ready is false (although that's probably a bug if that happens since we only enable one interrupt..)

runtime/src/main.rs
```diff
@@ -38,6 +41,20 @@ const BANNER: &str = r#"
...
+    unsafe {
+        // Write meivt (External Interrupt Vector Table Register)
+        // VeeR has been instantiated with RV_FAST_INTERRUPT_REDIRECT,
+        // so external interrupts always bypass the standard risc-v dispatch logic
+        // and instead load the destination address from this table in DCCM.
+        core::arch::asm!(
+            "la {tmp}, _ext_intr_vector",
+            "csrw 0xbc8, {tmp}",
+            tmp = out(reg) _,
+        );
```
jhand2, line 55: nit: why do this here instead of the startup script?

runtime/src/lib.rs
korran: No, because enter_idle doesn't know for sure that you're going to process the mailbox.

If we clear the interrupt status bit, but we don't get into this branch (for example, due to a race condition or unexpected interrupt), then we risk going to sleep forever, which would be very bad.

runtime/src/main.rs
korran: startup script is shared between fmc and runtime. We want this logic in the runtime only.

korran: Also, since #1141, the FMC overflows 16k with the added 160 bytes of the lookup table....

runtime/src/main.rs
korran: Finally got my FPGA setup working again so I could validate it actually went to sleep. Turns out the interrupt was still high from FIRMWARE_LOAD, so I needed to handle that to allow it to sleep before the first runtime command.

runtime/src/ext_intr.S
```diff
@@ -0,0 +1,54 @@
...
+    csrw mscratch, t0
+    li t0, 0x80
+    csrc mstatus, t0
+    csrr t0, mscratch
+
+    // Return back to event loop
+    mret
+
+// meivt must point at an address in DCCM
+.section .data
```
ArthurHeymans, line 15: Isn't .rodata also in DCCM at this point?

korran: No, .rodata is in ICCM because it needs to be write protected (and DCCM is never write-locked).

jhand2 (approved):

jlmahowa-amd (approved):

mhatrevi (approved):

PR #1332, opened 2024-02-12
Make RT depend on caliptra-cfi and update DPE version

sree-revoori1:
Previously, we forked caliptra-sw's cfi crate into a new repo, [caliptra-cfi](https://github.com/chipsalliance/caliptra-cfi), to be able to add CFI in DPE. In this PR, we make RT use caliptra-cfi instead of caliptra-sw's cfi so that we can update DPE to the latest version containing CFI instrumentation. To do this correctly without modifying the ROM binary, we make drivers depend on caliptra-cfi only if the runtime flag is enabled. 

The plan is to make caliptra-cfi a submodule and replace the cfi crate with caliptra-cfi everywhere. We will do this for ROM 1.01. 

The RT binary size delta is + 7800 due to adding CFI to DPE.


drivers/src/sha384.rs
```diff
@@ -61,7 +59,11 @@ impl Sha384 {
     ///
     /// * `data` - Data to used to update the digest
     ///
-    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[cfg_attr(
+        not(feature = "no-cfi"),
+        cfg_attr(feature = "runtime", caliptra_cfi_derive_git::cfi_impl_fn),
```
korran, line 64: Can you try something like this in the root module?

```
#[cfg(feature = "runtime")]
use caliptra_cfi_derive_submodule as caliptra_cfi_derive;

#[cfg(not(feature = "runtime"))]
use caliptra_cfi_derive_internal as caliptra_cfi_derive;
```

sree-revoori1: Done

jhand2 (commented):
LGTM, but will need a submodule update when https://github.com/chipsalliance/caliptra-dpe/pull/310 is merged. I assume the current submodule import isn't a commit in main.

mhatrevi (approved):

jhand2 (approved):

mhatrevi (approved):

PR #1334, opened 2024-02-13
Finalize RT test coverage doc

sree-revoori1:


api/src/capabilities.rs
```diff
@@ -13,12 +13,13 @@ Abstract:
...
 bitflags::bitflags! {
+    /// First 64 bits are reserved for RT, next 32 bits are reserved for FMC, and final 32 bits are reserved for ROM
     #[derive(Default, Copy, Clone, Debug)]
     pub struct Capabilities : u128 {
         // Represents base capabilities present in Caliptra ROM v1.0
-        const ROM_BASE = 0b0001;
+        const ROM_BASE = 0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000001;
         // Represents base capabilities present in Caliptra Runtime v1.0
-        const RT_BASE = 0b0001;
+        const RT_BASE = 0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000001_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000;
```
jhand2, line 22: Like Jeff mentioned, we should use shift notation (e.g. `(1 << 64)`) to make this easier to read.

For ROM_BASE, it's probably fine to leave it as-is (no shift)

sree-revoori1: Done

jhand2 (approved):

PR #1335, opened 2024-02-13
Add capabilities mailbox command to RT

sree-revoori1:


runtime/tests/runtime_integration_tests/test_info.rs
```diff
@@ -163,3 +167,20 @@ fn test_idev_id_info() {
         .unwrap();
     GetIdevInfoResp::read_from(resp.as_slice()).unwrap();
 }
+
+#[test]
```
bluegate010, line 171: Should we also test that `ROM_BASE` is set in the response, and that it's equal to what ROM returns? (Would need to amend that test if we start adding capabilities to RT FW.)

sree-revoori1: We don't set the ROM_BASE in RT capabilities, so I didn't add this check. Should we set ROM_BASE?

bluegate010: Ah maybe we define an RT_BASE bit, to encompass all RT FW 1.0 capabilities.

jhand2 (approved):

PR #1336, opened 2024-02-13
Adding ROM version

nquarton:
- Change FW_REV_ID[0] 15:0 to be the ROM version
- Change FMC version to be 16 bits and shift to FW_REV_ID[0] 31:16

### DIFF
--- a/builder/src/version.rs
+++ b/builder/src/version.rs
@@ -1,0 +1,15 @@
+// Licensed under the Apache-2.0 license
+
+pub const ROM_VERSION_MAJOR: u16 = 1;
+pub const ROM_VERSION_MINOR: u16 = 0;
+pub const ROM_VERSION_PATCH: u16 = 1;
+
+// ROM Version - 16 bits
+// Major - 5 bits [15:11]
+// Minor - 5 bits [10:6]
+// Patch - 6 bits [5:0]
+pub fn get_rom_version() -> u16 {
+    ((ROM_VERSION_MAJOR & 0x1F) << 11)
+        | ((ROM_VERSION_MINOR & 0x1F) << 6)
+        | (ROM_VERSION_PATCH & 0x3F)
+}
--- a/rom/dev/src/main.rs
+++ b/rom/dev/src/main.rs
@@ -110,6 +110,10 @@ pub extern "C" fn rom_entry() -> ! {
         }
     );
 
+    // Set the ROM version
+    let rom_info = unsafe { &CALIPTRA_ROM_INFO };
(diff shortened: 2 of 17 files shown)


builder/src/version.rs
```diff
@@ -0,0 +1,15 @@
+// Licensed under the Apache-2.0 license
+
+pub const ROM_VERSION_MAJOR: u16 = 1;
```
korran, line 3: Do we want this in the source code, or fill this as part of the release process?

nquarton: This is something I am looking for input on. If you have an idea for a way we can populate this during a release I think that would be better. You could in theory try to pull it from a tag but I was thinking that may be fragile and requires the tag to be populated before the release is made.

builder/src/version.rs
```diff
@@ -0,0 +1,15 @@
...
+pub const ROM_VERSION_MINOR: u16 = 0;
+pub const ROM_VERSION_PATCH: u16 = 1;
+
+// ROM Version - 16 bits
```
nquarton, line 7: We mentioned the version source could be changed down the road, but just wanted to call out we won't want to change the bit allocation for these 16 bits once we merge. So we should at least make sure we are all in agreement on that.

JohnTraverAmd (approved):

builder/src/version.rs
FerralCoder: In the interest of getting this patch out quickly, I suggest we leave it like this right now, and look for a better method for the next release

FerralCoder (approved):

rom/dev/src/main.rs
```diff
@@ -110,6 +110,10 @@ pub extern "C" fn rom_entry() -> ! {
         }
     );
 
+    // Set the ROM version
+    let rom_info = unsafe { &CALIPTRA_ROM_INFO };
+    env.soc_ifc.set_rom_fw_rev_id(rom_info.version);
```
mhatrevi, line 115: Can we add a test in rom/tests (choose the appropriate module in that) to test that the version is set correctly?

mhatrevi (changes requested):

rom/dev/src/main.rs
nquarton: This is a good point. This test is missing in ROM. I'll add it.

nquarton: This has been added now

builder/src/version.rs
nquarton: It sounds like everyone is agreed this is a good format

mhatrevi (approved):

jhand2 (approved):

PR #1337, opened 2024-02-14
Adding FIPS command documentation

nquarton:
- Adding FIPS command info to runtime readme
- Updating ROM command references in readme


nquarton:
Combined with https://github.com/chipsalliance/caliptra-sw/pull/1336 (can re-separate if wanted)

PR #1338, opened 2024-02-14
runtime: Add tests for Pic

ArthurHeymans:
This adds 2 tests:
- Check that a wrongly configured ext int vector table results in an NMI
- Properly configured external interrupts wake from halt

Follow-up patches will use this to get accurate sw-emulation right.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -520,6 +520,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "caliptra-hw-model-heavy-test-fw"
+version = "0.1.0"
+dependencies = [
+ "caliptra-cpu",
+ "caliptra-drivers",
+ "caliptra-gen-linker-scripts",
+ "caliptra-registers",
+ "caliptra-runtime",
+ "caliptra-test-harness",
+ "caliptra_common",
+ "cfg-if 1.0.0",
+ "ufmt",
+]
+
+[[package]]
 name = "caliptra-hw-model-test-fw"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ members = [
   "hw/verilated",
   "hw-model",
   "hw-model/test-fw",
+  "hw-model/heavy-test-fw",
   "hw-model/types",
   "hw-model/c-binding",
   "registers",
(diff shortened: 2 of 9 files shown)


runtime/test-fw/src/pic_tests.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#![no_main]
+#![no_std]
+
+use caliptra_drivers::{cprintln, Pic, SocIfc};
+use caliptra_registers::{el2_pic_ctrl::El2PicCtrl, soc_ifc::SocIfcReg};
+use caliptra_test_harness::test_suite;
+
+#[cfg(target_arch = "riscv32")]
+core::arch::global_asm!(include_str!("ext_intr.S"));
```
korran, line 11: Can you get away with include_str!("../../extr_intr.S") and delete this file? I don't like copy-and-paste.

ArthurHeymans: Ah sure. Older iterations had similar code to NMI to vector into rust eventually. That didn't work so I had to revert it. With the new insights from the meeting I might give that another try.

ArthurHeymans: I added it back since it works with a few nop cycles after enabling global interrupts.

runtime/test-fw/src/ext_intr.S
```diff
@@ -0,0 +1,171 @@
+
+.section .init.text, "ax"
+.align 2
+_ext_intr_handler:
```
korran, line 4: I don't understand what we're testing here. What does this have to do with runtime firmware? It seems like this is more to do with testing the hardware model itself...

korran: Typically those tests go into hw-model/.

ArthurHeymans: Thanks that makes sense. I'll move it to hw-model. The goal is indeed to make the emulator more accurately match hardware.

ArthurHeymans: The runtime test firmware setup is more convenient than the test-harness rom as things like nmi is already set up. I can make the tests include another .S file that does all that ?

ArthurHeymans:
The sw-emulator patch for PIC pas this test btw.

PR #1352, opened 2024-02-21
Pull CertValidity from ImageHeader of manifest1 in DpePlatform

### LINKED ISSUE #1350: dpe: Use cert validity from the image header
We should use the cert validity from the ImageManifest for DPE certs. We can use the same logic that ROM does: https://github.com/chipsalliance/caliptra-sw/blob/2a4216a833070ba5a65463140c10dab934ba6ab3/rom/dev/src/flow/cold_reset/fw_processor.rs#L547-L570

We are currently using hard-coded values: https://github.com/chipsalliance/caliptra-sw/blob/2a4216a833070ba5a65463140c10dab934ba6ab3/runtime/src/dpe_platform.rs#L140-L145

sree-revoori1:
fixes #1350

### DIFF
--- a/runtime/src/dpe_platform.rs
+++ b/runtime/src/dpe_platform.rs
@@ -16,7 +16,8 @@ use core::cmp::min;
 
 use arrayvec::ArrayVec;
 use caliptra_drivers::cprintln;
-use caliptra_x509::{NOT_AFTER, NOT_BEFORE};
+use caliptra_image_types::{ImageHeader, ImageManifest};
+use caliptra_x509::{NotAfter, NotBefore};
 use crypto::Digest;
 use dpe::{
     x509::{CertWriter, DirectoryString, Name},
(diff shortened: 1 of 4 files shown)


runtime/src/dpe_platform.rs
```diff
@@ -34,6 +35,7 @@ pub struct DpePlatform<'a> {
     auto_init_locality: u32,
     hashed_rt_pub_key: Digest,
     cert_chain: &'a mut ArrayVec<u8, MAX_CERT_CHAIN_SIZE>,
+    image_header: ImageHeader,
```
jhand2, line 38: Can we pass this by reference to avoid copying every time?

sree-revoori1: I was trying to do this earlier but it would take a large refactor since we have both mutable and immutable borrows of persistent data so the borrow checker complains. ImageHeader is only 150 bytes so it's not too big of a copy

sree-revoori1: The option is we could extract each of the fields we need from the image header and pass it to the platform so we copy even less data.

sree-revoori1: Refactored to only pass in the nb and nf validities so that we copy 30 bytes instead of 150.

jhand2 (approved):

PR #1353, opened 2024-02-21
Removing checksum from FW_LOAD request documentation

nquarton:


jhand2 (dismissed):

runtime/README.md
```diff
@@ -533,8 +532,8 @@ Command Code: `0x494E_464F` ("INFO")
 
 ## Checksum
 
-For every command, the request and response feature a checksum. This mitigates
-glitches between clients and Caliptra.
+For every command, the request (except for FW_LOAD) and response feature a checksum. This
```
jhand2, line 535: nit
```suggestion
For every command except for FW_LOAD, the request and response feature a checksum. This
```

nquarton: Admittedly I was under the impression this only applied to the input/request args at first. That is fixed now and this comment matches your suggestion.

PR #1355, opened 2024-02-22
Add caliptra-rtl 1.1 submodule.

korran:

## Move hw-latest/{fpga,verilated} to hw/

Also, move hw-latest/caliptra-rtl to hw/latest/rtl.

This is in preparation of moving the current caliptra-rtl submodule from hw/latest/rtl to hw/1.0/rtl, and replacing hw/latest/rtl with the 1.1 caliptra-rtl branch. The fpga/ and verilated/ code is the same for both 1.0 and 1.1, so it didn't make sense to keep identical copies in the hw/latest/ and hw/1.0 directories.

## Split caliptra-registers into caliptra-registers-latest.
    
This is in preparation for a future caliptra-registers-1.0 crate.

## Create caliptra-rtl 1.0 submodule, update latest to 1.1

Everything still builds against 1.0 by default.


korran:
Pushes to my branch don't seem to update the PR; smell like some sort of internal github glitch. Opening https://github.com/chipsalliance/caliptra-sw/pull/1365 in place of this PR.

PR #1362, opened 2024-02-23
Add RT version to binary

### LINKED ISSUE #1349: Set version number in Runtime binary
For the 1.0 release, we should be setting the proper version numbers in the image manifest when building the runtime binary.

sree-revoori1:
fixes #1349

### DIFF
--- a/builder/src/version.rs
+++ b/builder/src/version.rs
@@ -4,6 +4,10 @@ pub const ROM_VERSION_MAJOR: u16 = 1;
 pub const ROM_VERSION_MINOR: u16 = 0;
 pub const ROM_VERSION_PATCH: u16 = 1;
 
+pub const RUNTIME_VERSION_MAJOR: u32 = 1;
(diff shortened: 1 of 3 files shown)


builder/src/version.rs
```diff
@@ -13,3 +17,14 @@ pub fn get_rom_version() -> u16 {
         | ((ROM_VERSION_MINOR & 0x1F) << 6)
         | (ROM_VERSION_PATCH & 0x3F)
 }
+
+// Runtime Version - 32 bits
+// [31:16] unused
```
jhand2, line 22: Should we use more of the bits here? We could probably do 16 bits for patch and then 8 bits each for major and minor.

ROM and FMC versions have to share a 32 bit register, but RT doesn't have the same constraint.

sree-revoori1: Done

jhand2 (approved):

PR #1365, opened 2024-02-26
Add caliptra-rtl 1.1 submodule.

korran:

## Move hw-latest/{fpga,verilated} to hw/

Also, move hw-latest/caliptra-rtl to hw/latest/rtl.

This is in preparation of moving the current caliptra-rtl submodule from hw/latest/rtl to hw/1.0/rtl, and replacing hw/latest/rtl with the 1.1 caliptra-rtl branch. The fpga/ and verilated/ code is the same for both 1.0 and 1.1, so it didn't make sense to keep identical copies in the hw/latest/ and hw/1.0 directories.

## Split caliptra-registers into caliptra-registers-latest.
    
This is in preparation for a future caliptra-registers-1.0 crate.

## Create caliptra-rtl 1.0 submodule, update latest to 1.1

Everything still builds against 1.0 by default.


jlmahowa-amd (dismissed):

hw-model/c-binding/examples/Makefile
```diff
@@ -16,7 +16,7 @@
 EXTRA_CARGO_CONFIG = target.'cfg(all())'.rustflags = [\"-Dwarnings\"]
 OUT = out
 TARGET = $(OUT)/smoke_test
-RTL_SOC_IFC_INCLUDE_PATH = ../../../hw/latest/rtl/src/soc_ifc/rtl
+RTL_SOC_IFC_INCLUDE_PATH = ../../../hw/1.0/rtl/src/soc_ifc/rtl
```
nquarton, line 19: Should we leave this as "latest"? This c-binding is just for testing libcaliptra against the HW model. I think we'd want to continue to test it against the latest HW for future development?

korran: I've been consistent in this PR pointing everything at 1.0 by default, as latest isn't working yet (verilator, fpga, and sw-emulator hw-models don't support latest).

The expectation is to change these once they work :)

nquarton: Ah understood. This makes sense then. Thanks!

jlmahowa-amd (approved):

PR #1366, opened 2024-02-26
FIPS test additions

nquarton:
- Adding support for use of existing ROM and FW binaries
- Support for varying expected constants between versions
- Version test and start to services tests


test/tests/fips_test_suite/common.rs
```diff
@@ -2,7 +2,107 @@
...
+    pub fmc_version: u16,
+    pub fw_version: u32,
+}
+
+const RT_EXP_CURRENT: RtExpVals = RtExpVals {
+    fmc_version: 0x0,
+    fw_version: 0x0,
+};
+
+// === Getter implementations ===
```
jlmahowa-amd, line 53: Please update the README for the test suite with how to use these variables.

nquarton: Yep I'll definitely be adding documentation for all those options

ericeilertson (commented):
This looks good to me.

PR #1371, opened 2024-02-28
[fix] Remove superfluous sections from ROM README.md

mhatrevi:
This change removes the 'DICE Layers' section since it is specified in the Caliptra Main spec. This change also removes the 'Vaults' section since it is specified in the hardware spec.

### DIFF
--- a/rom/dev/README.md
+++ b/rom/dev/README.md
@@ -60,36 +60,7 @@ following topics:
 | SVN                 | Security Version Number                                                   |
 | X509                | Digital Certificate Standard                                              |
 
-## 5. DICE Layers
-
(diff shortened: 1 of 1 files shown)


steph-morton (approved):
May be outside the scope of this PR: We removed the section numbers from other specs, so instead of renumbering, recommend removing them.

andreslagarcavilla:
awesome. need a code owner review

FerralCoder (approved):

ajisaxena (approved):

vsonims (approved):

jhand2 (approved):

PR #1375, opened 2024-02-29
Remove Alias RT CSR reference

varuns-nvidia:
Caliptra does not generate this


bluegate010:
Caliptra doesn't generate a CSR for its runtime DICE alias key, but it does use that key to sign child CSRs for DPE leaf keys.

varuns-nvidia:
> Caliptra doesn't generate a CSR for its runtime DICE alias key, but it does use that key to sign child CSRs for DPE leaf keys.

You're right. Will close this.

PR #1378, opened 2024-03-01
[fix] Removing spec opens in ROM README.md

mhatrevi:
This change removes the following spec opens from the ROM README.md as they are not applicable.

- Update the firmware image format to include the runtime configuration section
- Ability to run production signed firmware that can only run in debug mode


rom/dev/README.md
```diff
@@ -16,9 +16,7 @@
 
 
 ## 2. Spec Opens
```
steph-morton, line 18: recommend just deleting the section since it will be empty.

mhatrevi: Updated.

steph-morton (approved):

jhand2 (approved):

JohnTraverAmd (approved):

PR #1385, opened 2024-03-04
FPGA build parameter for HW_LATEST

jlmahowa-amd:


.github/workflows/nightly-release.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     uses: ./.github/workflows/fpga.yml
     with:
       artifact-suffix: -fpga-realtime-etrng-nolog
-      extra-features: slow_tests
+      extra-features: slow_tests,hw-latest
```
korran, line 86: Can we test both with hw-1.0 and hw-latest? I don't think they've released 1.0 yet...

hw-model/src/output.rs
```diff
@@ -79,6 +79,9 @@ impl OutputSink {
     pub fn set_now(&self, now: u64) {
         self.0.now.set(now);
     }
+    pub fn get_now(&self) -> u64 {
```
korran, line 82: nitpick: In idiomatic Rust, getters don't prefix get_:

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

hw-model/src/model_fpga_realtime.rs
```diff
@@ -446,9 +449,11 @@ impl HwModel for ModelFpgaRealtime {
     }
 
     fn output(&mut self) -> &mut crate::Output {
-        self.output
-            .sink()
-            .set_now(self.start_time.elapsed().as_millis().try_into().unwrap());
+        self.output.sink().set_now(
```
korran, line 452: In a future PR, we should probably consider exposing an actual cycle count register from the FPGA testbench RTL.

hw-model/src/output.rs
jlmahowa-amd: Removed

hw-model/src/model_fpga_realtime.rs
jlmahowa-amd: I agree. I don't love the assumption about the FPGA speed I had to make here.
Should be easy to do.

korran (approved):

PR #1386, opened 2024-03-04
Allow Fake ROM to work with production mode

nquarton:
- Defines bit 30 in the dbg_manuf_service_reg to enable production mode for fake ROM
- Overrides the ROM version to 0xFFFF in fake ROM

### DIFF
--- a/rom/dev/src/main.rs
+++ b/rom/dev/src/main.rs
@@ -95,9 +95,10 @@ pub extern "C" fn rom_entry() -> ! {
     cprintln!("[state] LifecycleState = {}", _lifecyle);
 
     if cfg!(feature = "fake-rom")
-        && env.soc_ifc.lifecycle() == caliptra_drivers::Lifecycle::Production
+        && (env.soc_ifc.lifecycle() == caliptra_drivers::Lifecycle::Production)
+        && !(env.soc_ifc.prod_en_in_fake_mode())
     {
-        cprintln!("Fake ROM in Production lifecycle prohibited");
+        cprintln!("Fake ROM in Production lifecycle not enabled");
         handle_fatal_error(CaliptraError::ROM_GLOBAL_FAKE_ROM_IN_PRODUCTION.into());
     }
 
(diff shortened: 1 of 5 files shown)


nquarton:
This is currently just a proposal. Plan to discuss with other workgroup members this week.

nquarton:
This was discussed in the FW Sync today with the agreement that this should be okay. Input still needed from @mhatrevi though since he wasn't able to join the meeting.

rom/dev/src/main.rs
```diff
@@ -95,9 +95,10 @@ pub extern "C" fn rom_entry() -> ! {
     cprintln!("[state] LifecycleState = {}", _lifecyle);
 
     if cfg!(feature = "fake-rom")
-        && env.soc_ifc.lifecycle() == caliptra_drivers::Lifecycle::Production
+        && (env.soc_ifc.lifecycle() == caliptra_drivers::Lifecycle::Production)
```
mhatrevi, line 98: What is the motivation for supporting the Production lifecycle mode for the fake ROM?

mhatrevi (changes requested):

vsonims:
But this means somebody can by mistake use fake RoM in prod tape out?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

nquarton:
The motivation is just that more exhaustive testing in simulation was needed than I think we originally anticipated. This includes several test lines in production mode.

A lot would have to go wrong in someone's testing procedures for fake ROM to get close to a tape-out scenario. This does partially "lower the bar" for this one catch. However, you will still get the same error about using fake ROM in production mode if the bit in the debug register is not set. This way the user has to actually acknowledge what they are doing and explicitly allow it.

Some other things that would have to be skipped in testing to get the the Fake ROM to tape in include:
1. Never testing with real FMC/FW image (real image cannot boot with the Fake ROM)
2. Never tested with production fuse recipe (the real fuses would have production key hashes which would not match the fake image)
3. Never verified Cert contents (All certs up to FMC are canned, RT cert is never generated)
4. Never attempted certain DICE flows at Caliptra runtime FW (no private keys in key vault in fake ROM and FMC)

rom/dev/src/main.rs
nquarton: The motivation is just that more exhaustive testing in simulation was needed than I think we originally anticipated. This includes several test lines in production mode. (I included some more details in my response below.)

mhatrevi (approved):

PR #1388, opened 2024-03-05
Update mio to 0.8.11

jhand2:
Resolves dependabot alert https://github.com/chipsalliance/caliptra-sw/security/dependabot/20

mio is a dependency of tokio. Updated specific version with `cargo update -p mio`.

### DIFF
--- a/ci-tools/test-matrix/Cargo.lock
+++ b/ci-tools/test-matrix/Cargo.lock
@@ -531,9 +531,9 @@ dependencies = [
 
 [[package]]
 name = "mio"
-version = "0.8.10"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


sree-revoori1 (approved):

sree-revoori1:
Why are we updating the Cargo.lock? Shouldn't we update Cargo.toml and let it propagate to the lock file?

jhand2:
> Why are we updating the Cargo.lock? Shouldn't we update Cargo.toml and let it propagate to the lock file?

mio isn't a direct dependency, but a transitive dependency. The crate that depends on it (tokio) doesn't have a newer version and just specifies that mio shouldbe >= 0.8.9. So when you do `cargo update` cargo will find the latest version of that crate that fits the requirement.

PR #1394, opened 2024-03-07
Select H/W version via  hw-latest feature 

rusty1968:
Each crate that has caliptra-drivers and/or caliptra-registers as a dependency must expose the hw-latest feature.


builder/src/firmware.rs
```diff
@@ -15,48 +15,118 @@ pub fn rom_from_env() -> &'static FwId<'static> {
...
     crate_name: "caliptra-fmc",
     bin_name: "caliptra-fmc",
     features: &["emu"],
 };
 
+#[cfg(feature = "hw-latest")]
+pub const FMC_FAKE_WITH_UART: FwId = FwId {
+    crate_name: "caliptra-fmc",
+    bin_name: "caliptra-fmc",
+    features: &["emu", "fake-fmc", "hw-latest"],
```
korran, line 78: Seeing as we're doing this to pretty much every target I wonder if it we be more maintainable to revert the changes to this file and instead decide to add the hw-latest feature to every target in build_firmware_elfs_uncached().

rusty1968: Ok.  I will make the change.

rusty1968: Done.

jlmahowa-amd (approved):

runtime/log.txt
```diff
@@ -0,0 +1,7749 @@
+   Compiling caliptra-runtime v0.1.0 (/home/ferrite/rusty1968/rev11-support/runtime)
```
korran, line 1: Delete this file?

runtime/log.txt
korran, line 1: Remove this?

runtime/log.txt
rusty1968: done. left-over from test.

runtime/log.txt
rusty1968: done.

korran (approved):

FerralCoder (approved):

jhand2 (approved):

PR #1403, opened 2024-03-12
Update CFI to latest revision

sree-revoori1:
Remove cfi directory from caliptra and depend on caliptra-cfi


error/src/lib.rs
```diff
@@ -560,6 +542,68 @@ impl CaliptraError {
...
+    pub const CFI_PANIC_TRNG_FAILURE: CaliptraError = CaliptraError::new_const(0xA000000C);
+    pub const CFI_PANIC_UNEXPECTED_MATCH_BRANCH: CaliptraError =
+        CaliptraError::new_const(0xA000000D);
+    pub const CFI_PANIC_FAKE_TRNG_USED_WITH_DEBUG_LOCK: CaliptraError =
+        CaliptraError::new_const(0xA000000E);
+}
+
+impl From<caliptra_cfi_lib::CfiPanicInfo> for crate::CaliptraError {
+    fn from(val: caliptra_cfi_lib::CfiPanicInfo) -> Self {
+        match val {
```
korran, line 567: I thought we were going to try and get rid of this remapping logic and try to make the error codes consistent.

error/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+caliptra-cfi-lib = { workspace = true, default-features = false, features = ["cfi", "cfi-counter" ] }
```
korran, line 10: I don't really like this dependency. This crate should just be a bunch of error codes.

korran, line 10: Do you actually need to set the features here? It seems to me you only need to look at the enum, which shouldn't require any features, correct?

sree-revoori1: Correct. Removed the features.

error/src/lib.rs
sree-revoori1: Yeah that was the plan initially, but I think it's better this way since each caller of the library could define their own conversion. If we have another platform that integrates DPE which also uses CFI, it's error codes would be different from the Caliptra CFI error codes.

error/Cargo.toml
sree-revoori1: Done

error/src/lib.rs
korran: Doing the conversion requires a significant lookup table, which it'd be nice to avoid, especially since this is all in the same firmware bundle. Also, the branchy nature of that code is best to avoid in a CFI error-handling context.

sree-revoori1: Reworked this to simply add the CfiPanicInfo offset to a CaliptraError base constant so the branching should be eliminated.

PR #1404, opened 2024-03-12
Experiment...

korran:


hw-model/tests/model_tests.rs
```diff
@@ -271,3 +271,117 @@ fn test_pcr_extend() {
...
+    let mut model = run_fw_elf_with_rand_puf(
+        &caliptra_builder::build_firmware_elf(&firmware::hw_model_tests::MAILBOX_RESPONDER)
+            .unwrap(),
+    );
+    let mbox = model.soc_mbox();
+    // Lock mailbox
+    assert!(!mbox.lock().read().lock());
+    assert!(mbox.lock().read().lock());
+
+    // This command writes 3 words to dint but doesn't update dlen immediately
```
jhand2, line 286: nit: s/dint/din

hw-model/tests/model_tests.rs
```diff
@@ -271,3 +271,117 @@ fn test_pcr_extend() {
...
+    mbox.cmd().write(|_| 0x1000_3000);
+    mbox.dlen().write(|_| 4);
+    mbox.datain().write(|_| 0x3a48_735b);
```
jhand2, line 289: is the goal of sending this word just to make sure it doesn't get included in the response? Might be good to comment that since otherwise it looks superfluous (the mbox responder doesn't use it).

Also I assume the choice of value is arbitrary. Is that correct?

hw-model/tests/model_tests.rs
```diff
@@ -271,3 +271,117 @@ fn test_pcr_extend() {
...
+    // Lock mailbox
+    assert!(!mbox.lock().read().lock());
+    assert!(mbox.lock().read().lock());
+
+    // This command writes two words to din but doesn't set dlen
+    mbox.cmd().write(|_| 0x1000_3001);
+    mbox.dlen().write(|_| 12);
+    mbox.datain().write(|_| 0x508b_7a98);
+    mbox.datain().write(|_| 0x9dbe_d0ec);
+    mbox.datain().write(|_| 0x3ef7_9f17);
```
jhand2, line 318: Same, let's add a comment for why the caller is writing to the mailbox

korran: Done.

hw-model/tests/model_tests.rs
korran: Yes. Fixed.

hw-model/tests/model_tests.rs
korran: Fixed.

rusty1968 (approved):

PR #1411, opened 2024-03-15
Add planned LMS mailbox API to runtime readme

nquarton:


runtime/README.md
```diff
@@ -281,6 +281,35 @@ Command Code: `0x5349_4756` ("SIGV")
...
+
+Command Code: `0x4C4D_5356` ("LMSV")
+
+*Table: `LMS_SIGNATURE_VERIFY` input arguments*
+
+| **Name**              | **Type** | **Description**
+| --------              | -------- | ---------------
+| chksum                | u32      | Checksum over other input arguments, computed by the caller. Little endian.
+| pub\_key\_tree\_type  | u8[4]    | LMS public key algorithm type. Must equal 12.
+| pub\_key\_ots\_type   | u8[4]    | LM-OTS algorithm type. Must equal 7.
```
jhand2, line 298: Should we list the enum values under this command heading so they seem less magic? Also if they're spec defined, hopefully folks will be less likely to change them in the code :)

runtime/README.md
```diff
@@ -281,6 +281,35 @@ Command Code: `0x5349_4756` ("SIGV")
...
+| pub\_key\_id          | u8[16]   | Private key identifier
+| pub\_key\_digest      | u8[24]   | Public key hash value
+| signature\_q          | u8[4]    | Leaf of the Merkle tree where the OTS public key appears
+| signature\_ots        | u8[1252] | LM-OTS signature
+| signature\_tree\_type | u8[4]    | LMS signature Algorithm type. Must equal 12.
+| signature\_tree\_path | u8[360]  | Path through the tree from the leaf associated with the LM-OTS signature to the root
```
jhand2, line 304: I wonder if we should define the structure of the signature in a separate table and then just have a flat `signature` field in the command.

I don't feel too strongly though.

runtime/README.md
```diff
@@ -281,6 +281,35 @@ Command Code: `0x5349_4756` ("SIGV")
 | chksum      | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips\_status | u32      | Indicates if the command is FIPS approved or an error.
 
+### LMS\_SIGNATURE\_VERIFY
```
jhand2, line 284: Should note somewhere here that if signature verification fails it will be returned as an fw_non_fatal error code (rather than in the response).

runtime/README.md
```diff
@@ -281,6 +281,35 @@ Command Code: `0x5349_4756` ("SIGV")
...
+
+*Table: `LMS_SIGNATURE_VERIFY` input arguments*
+
+| **Name**              | **Type** | **Description**
+| --------              | -------- | ---------------
+| chksum                | u32      | Checksum over other input arguments, computed by the caller. Little endian.
+| pub\_key\_tree\_type  | u8[4]    | LMS public key algorithm type. Must equal 12.
+| pub\_key\_ots\_type   | u8[4]    | LM-OTS algorithm type. Must equal 7.
+| pub\_key\_id          | u8[16]   | Private key identifier
+| pub\_key\_digest      | u8[24]   | Public key hash value
```
jhand2, line 300: Same as for the signature, should the public key be passed as a single buffer? The public key format is spec-defined: https://datatracker.ietf.org/doc/html/rfc8554#section-4.3

jhand2: Although to be fair, we do split apart signatures and public keys for ECDSA into x/y and r/s.

runtime/README.md
```diff
@@ -281,6 +281,35 @@ Command Code: `0x5349_4756` ("SIGV")
...
+Command Code: `0x4C4D_5356` ("LMSV")
+
+*Table: `LMS_SIGNATURE_VERIFY` input arguments*
+
+| **Name**              | **Type** | **Description**
+| --------              | -------- | ---------------
+| chksum                | u32      | Checksum over other input arguments, computed by the caller. Little endian.
+| pub\_key\_tree\_type  | u8[4]    | LMS public key algorithm type. Must equal 12.
+| pub\_key\_ots\_type   | u8[4]    | LM-OTS algorithm type. Must equal 7.
+| pub\_key\_id          | u8[16]   | Private key identifier
```
jhand2, line 299: I think the description should say "Public key identifier"

runtime/README.md
jhand2, line 284: For the parameters, we should be clear what each thing refers too (maybe referencing https://datatracker.ietf.org/doc/html/rfc8554).

For example it's not really clear which part of the public key the key id is.

runtime/README.md
nquarton: In the description? Yeah I had considered that and just didn't end up doing it. I'll make that change.

runtime/README.md
nquarton: Good idea, will do

runtime/README.md
nquarton: This is something I wasn't clear on. That document for LMS (https://datatracker.ietf.org/doc/html/rfc8554#section-5.3)  uses this description for it "The value I is the private key identifier and is the value used for all computations for the same LMS tree." But if that's inaccurate I can absolutely change it.

runtime/README.md
nquarton: This crossed my mind at one point to. The only reason I didn't do it is because I didn't think we had done multiple layers of structs anywhere else. I am okay with it either way.

runtime/README.md
nquarton: See response above, this was mainly done this way for consistency. But it is a lot of params. I am okay with flattening to just the pub_key and signature params if wanted.

runtime/README.md
nquarton: I can  include that "I" and "T[1]" notation in the description if that would make it clearer. I was building on the name/descriptions from the LMS fields in the [image format](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#firmware-image-format-and-verification).

runtime/README.md
jhand2: Ok to leave it since this aligns with how the ECDSA command works.

runtime/README.md
jhand2: In that case, it would also be fine if you just want to reference sections from the image format doc (or just replicate those format descriptions here. Maybe that's better to avoid the cross-reference).

runtime/README.md
jhand2: Ack, ok to leave as-is

runtime/README.md
jhand2: Updated descriptions look good.

runtime/README.md
```diff
@@ -281,6 +284,47 @@ Command Code: `0x5349_4756` ("SIGV")
...
+
+The supported parameter set is limited to those used for the caliptra image signature:
+*Table: LMS parameters*
+| **Param Name**        | **Value** | **Description**
+| --------------        | --------- | ---------------
+| n                     | 24        | 24 bytes output from sha256/192 hash function
+| w                     | 4         | Width (in bits) of the Winternitz coefficient
+| h                     | 15        | Height of the tree
+| LMS algorithm type    | 12        |
+| LM-OTS algorithm type | 7         |
```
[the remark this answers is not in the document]
nquarton: Absolutely, good point

jhand2 (approved):

PR #1419, opened 2024-03-20
Add support for custom extensions to the DPE leaf cert

### LINKED ISSUE #975: Add support for DMTF otherName in leaf certificate
https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf line 404
This can be provided as a command to runtime with the parameters to ingest. Runtime would update the certificate template (or support a second one) that populates TBS with the information received.

The motivation for this change is to enable Caliptra certificates to be bound to a particular device/model without solely relying on the pCA. A verifier can correlate the device/model in the leaf certificate with the intermediate certificates because they would all have the same UEID.

@jhand2 for thoughts

sree-revoori1:
The first custom extension added in this PR is DMTF otherName. 

fixes #975

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -1,5 +1,6 @@
 // Licensed under the Apache-2.0 license
 
+use bitflags::bitflags;
 use caliptra_error::{CaliptraError, CaliptraResult};
 use core::mem::size_of;
 use zerocopy::{AsBytes, FromBytes, LayoutVerified};
@@ -24,8 +25,8 @@ impl CommandId {
     pub const INCREMENT_PCR_RESET_COUNTER: Self = Self(0x50435252); // "PCRR"
     pub const QUOTE_PCRS: Self = Self(0x50435251); // "PCRQ"
     pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
-
-    pub const TEST_ONLY_HMAC384_VERIFY: Self = Self(0x484D4143); // "HMAC"
+    pub const ADD_SUBJECT_ALT_NAME: Self = Self(0x414C544E); // "ALTN"
+    pub const CERTIFY_KEY_EXTENDED: Self = Self(0x434B4558); // "CKEX"
 
     /// FIPS module commands.
     /// The status command.
@@ -137,6 +138,7 @@ pub enum MailboxResp {
     GetTaggedTci(GetTaggedTciResp),
     GetRtAliasCert(GetRtAliasCertResp),
     QuotePcrs(QuotePcrsResp),
+    CertifyKeyExtended(CertifyKeyExtendedResp),
 }
 
 impl MailboxResp {
@@ -155,6 +157,7 @@ impl MailboxResp {
             MailboxResp::GetFmcAliasCert(resp) => resp.as_bytes_partial(),
             MailboxResp::GetRtAliasCert(resp) => resp.as_bytes_partial(),
             MailboxResp::QuotePcrs(resp) => Ok(resp.as_bytes()),
+            MailboxResp::CertifyKeyExtended(resp) => Ok(resp.as_bytes()),
         }
     }
 
@@ -173,6 +176,7 @@ impl MailboxResp {
             MailboxResp::GetFmcAliasCert(resp) => resp.as_bytes_partial_mut(),
             MailboxResp::GetRtAliasCert(resp) => resp.as_bytes_partial_mut(),
             MailboxResp::QuotePcrs(resp) => Ok(resp.as_bytes_mut()),
+            MailboxResp::CertifyKeyExtended(resp) => Ok(resp.as_bytes_mut()),
         }
     }
 
@@ -228,6 +232,8 @@ pub enum MailboxReq {
     IncrementPcrResetCounter(IncrementPcrResetCounterReq),
     QuotePcrs(QuotePcrsReq),
     ExtendPcr(ExtendPcrReq),
+    AddSubjectAltName(AddSubjectAltNameReq),
+    CertifyKeyExtended(CertifyKeyExtendedReq),
 }
 
 impl MailboxReq {
@@ -248,6 +254,8 @@ impl MailboxReq {
             MailboxReq::IncrementPcrResetCounter(req) => Ok(req.as_bytes()),
             MailboxReq::QuotePcrs(req) => Ok(req.as_bytes()),
             MailboxReq::ExtendPcr(req) => Ok(req.as_bytes()),
+            MailboxReq::AddSubjectAltName(req) => req.as_bytes_partial(),
+            MailboxReq::CertifyKeyExtended(req) => Ok(req.as_bytes()),
         }
     }
 
@@ -268,6 +276,8 @@ impl MailboxReq {
             MailboxReq::IncrementPcrResetCounter(req) => Ok(req.as_bytes_mut()),
             MailboxReq::QuotePcrs(req) => Ok(req.as_bytes_mut()),
             MailboxReq::ExtendPcr(req) => Ok(req.as_bytes_mut()),
+            MailboxReq::AddSubjectAltName(req) => req.as_bytes_partial_mut(),
+            MailboxReq::CertifyKeyExtended(req) => Ok(req.as_bytes_mut()),
         }
     }
 
@@ -288,6 +298,8 @@ impl MailboxReq {
             MailboxReq::IncrementPcrResetCounter(_) => CommandId::INCREMENT_PCR_RESET_COUNTER,
             MailboxReq::QuotePcrs(_) => CommandId::QUOTE_PCRS,
             MailboxReq::ExtendPcr(_) => CommandId::EXTEND_PCR,
+            MailboxReq::AddSubjectAltName(_) => CommandId::ADD_SUBJECT_ALT_NAME,
+            MailboxReq::CertifyKeyExtended(_) => CommandId::CERTIFY_KEY_EXTENDED,
         }
     }
 
@@ -500,22 +512,6 @@ impl Request for EcdsaVerifyReq {
 }
 // No command-specific output args
 
-// TEST_ONLY_HMAC384_SIGNATURE_VERIFY
-#[repr(C)]
-#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
-pub struct HmacVerifyReq {
-    pub hdr: MailboxReqHeader,
-    pub key: [u8; 48],
-    pub tag: [u8; 48],
-    pub len: u32,
-    pub msg: [u8; 256],
-}
-impl Request for HmacVerifyReq {
-    const ID: CommandId = CommandId::TEST_ONLY_HMAC384_VERIFY;
-    type Resp = MailboxRespHeader;
-}
-// No command-specific output args
-
 // STASH_MEASUREMENT
 #[repr(C)]
 #[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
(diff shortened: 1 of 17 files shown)


sree-revoori1:
The build failure is due to some CFI version issues and is unrelated to this PR. This PR is ready for review.

varuns-nvidia (commented):
I'm not qualified to review the code, but the documentation and test case looks good to me.

runtime/README.md
```diff
@@ -604,6 +604,29 @@ Table: `SHUTDOWN` output arguments
 | chksum       | u32       | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips_status  | u32       | Indicates if the command is FIPS approved or an error
 
+### ADD\_SUBJECT\_ALT\_NAME
+
+Provides a subject alternative name otherName. The next time certifyKey is called via InvokeDpe, the
```
jhand2, line 609: This setting will persist until next reboot right? Should make that clear in the documentation.

runtime/src/dpe_platform.rs
```diff
@@ -49,13 +50,15 @@ impl<'a> DpePlatform<'a> {
         cert_chain: &'a ArrayVec<u8, 4096>,
         not_before: &'a NotBefore,
         not_after: &'a NotAfter,
+        dmtf_device_info: &'a [u8],
```
jhand2, line 53: nit: Should this be an `Option<...>`. Might make it easier to convey that it may not be popualted.

runtime/README.md
```diff
@@ -604,6 +604,29 @@ Table: `SHUTDOWN` output arguments
 | chksum       | u32       | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips_status  | u32       | Indicates if the command is FIPS approved or an error
 
+### ADD\_SUBJECT\_ALT\_NAME
```
jhand2, line 607: Would there be cases where the caller wants to set this extension for some keys but not others? e.g. SPDM responder v.s. CCA HES. I'm wondering if we should also have an "unset" mailbox command.

But also it's a non-critical extension so maybe it's not a big deal.

cc @varuns-nvidia

runtime/README.md
sree-revoori1: Done

api/src/mailbox.rs
```diff
@@ -24,6 +24,7 @@ impl CommandId {
     pub const INCREMENT_PCR_RESET_COUNTER: Self = Self(0x50435252); // "PCRR"
     pub const QUOTE_PCRS: Self = Self(0x50435251); // "PCRQ"
     pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
+    pub const ADD_SUBJECT_ALT_NAME: Self = Self(0x4153414E); // "ASAN"
```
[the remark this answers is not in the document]
sree-revoori1: Done

runtime/src/dpe_platform.rs
sree-revoori1: Done

runtime/README.md
steven-bellock: For CSRs it's probably not a big deal since the CA can pick and choose the fields it cares about. For the leaf certificate though there may be Verifiers that care, and that might be strictly the presence or absence of the field.

jhand2: Agreed. It's probably best if we give the SoC a way to request this on specific calls.

One way we could do this is via an extra boolean `add_subject_alt_name` parameter to InvokeDpe. If true, use the name provided by `ADD_SUBJECT_ALT_NAME`. Otherwise don't use it. If the command is not CertifyKey, this flag is ignored.

To avoid changing the InvokeDpe command too much for stuff like this in the future, we could reserve a u32 of flags.

jhand2: Another idea: Could have a separate CERTIFY_KEY_EXTENDED command which takes these extra parameters and calls CertifyKey internally. This is an easier mechanism to get Caliptra to add extra informational things to the leaf cert.

sree-revoori1: @steven-bellock @varuns-nvidia Any opinion on which route we should take here?

varuns-nvidia: I like Jordan's idea of CERTIFY_KEY_EXTENDED.

sree-revoori1: Done

api/src/mailbox.rs
```diff
@@ -566,6 +579,33 @@ impl Response for StashMeasurementResp {}
 // No command-specific input args
 // No command-specific output args
 
+// CERTIFY_KEY_EXTENDED
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct CertifyKeyExtendedReq {
+    pub hdr: MailboxReqHeader,
+    pub certify_key_req: [u8; CertifyKeyExtendedReq::CERTIFY_KEY_REQ_SIZE],
+    pub flags: u32,
```
jhand2, line 588: nit: Will probably make the code more readable if you use a bitflags type here. Shouldn't change the underlying type, so I assume you can still use repr C.

api/src/mailbox.rs
```diff
@@ -566,6 +579,33 @@ impl Response for StashMeasurementResp {}
 // No command-specific input args
 // No command-specific output args
 
+// CERTIFY_KEY_EXTENDED
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct CertifyKeyExtendedReq {
+    pub hdr: MailboxReqHeader,
+    pub certify_key_req: [u8; CertifyKeyExtendedReq::CERTIFY_KEY_REQ_SIZE],
```
jhand2, line 587: nit: If you put this at the end it would make it less invasive if we need to increase the size later.

runtime/README.md
```diff
@@ -601,6 +601,59 @@ Table: `SHUTDOWN` output arguments
...
+| chksum      | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
+| fips\_status | u32      | Indicates if the command is FIPS approved or an error.
+
+### CERTIFY\_KEY\_EXTENDED
+
+Produces a DPE leaf certificate or CSR containing custom extensions provided by the SoC.
+
+Command Code: `0x434B_4558` ("CKEX")
+
+*Table: `CERTIFY_KEY_EXTENDED` input arguments*
```
jhand2, line 635: nit: can you align the columns so it's easier to read in raw markdown?

runtime/README.md
```diff
@@ -601,6 +601,59 @@ Table: `SHUTDOWN` output arguments
 | chksum       | u32       | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips_status  | u32       | Indicates if the command is FIPS approved or an error
 
+### ADD\_SUBJECT\_ALT\_NAME
```
jhand2, line 604: Now that we're adding more commands, maybe we should mark the spec version as 1.1 and bump the internal version number.

The 1.0 spec will be published as a PDF, so I think it's fine for main to always be tracking the latest changes.

cc @FerralCoder @nquarton I'm not sure if you folks have already thought about this for the LMS commands.

runtime/README.md
```diff
@@ -601,6 +601,59 @@ Table: `SHUTDOWN` output arguments
...
+
+Command Code: `0x414C_544E` ("ALTN")
+
+*Table: `ADD_SUBJECT_ALT_NAME` input arguments*
+
+| **Name**     | **Type** | **Description**
+| --------     | -------- | ---------------
+| chksum       | u32      | Checksum over other input arguments, computed by the caller. Little endian.
+| dmtf\_device\_info\_size  | u32   | The size of the DMTF Device Info UTF8String.
+| dmtf\_device\_info | u8[128]   | The DMTF Device Info UTF8String.
```
jhand2, line 620: Is the max here defined by the X.509 spec or just an arbitrary choice?

runtime/src/certify_key_extended.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub struct CertifyKeyExtendedCmd;
+impl CertifyKeyExtendedCmd {
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
+        if let Some(cmd) = CertifyKeyExtendedReq::read_from(cmd_args) {
+            let hashed_rt_pub_key = drivers.compute_rt_alias_sn()?;
+            let key_id_rt_cdi = Drivers::get_key_id_rt_cdi(drivers)?;
+            let key_id_rt_priv_key = Drivers::get_key_id_rt_priv_key(drivers)?;
+            let pdata = drivers.persistent_data.get_mut();
+            let crypto = DpeCrypto::new(
```
jhand2, line 51: nit: Can we share more of the code for initializing DpeCrypto and DpePlatform? Maybe with a helper in drivers.

runtime/src/certify_key_extended.rs
```diff
@@ -0,0 +1,121 @@
...
+                    return Err(CaliptraError::RUNTIME_CERTIFY_KEY_EXTENDED_FAILED);
+                }
+            };
+
+            let certify_key_extended_resp = CertifyKeyExtendedResp {
+                hdr: MailboxRespHeader::default(),
+                certify_key_resp: certify_key_resp
+                    .as_bytes()
+                    .try_into()
+                    .map_err(|_| CaliptraError::RUNTIME_CERTIFY_KEY_EXTENDED_FAILED)?,
```
jhand2, line 113: nit: Should we use a different error code for a serialization error?

runtime/src/subject_alt_name.rs
```diff
@@ -0,0 +1,60 @@
...
+            let mut cmd = AddSubjectAltNameReq::default();
+            cmd.as_bytes_mut()[..cmd_args.len()].copy_from_slice(cmd_args);
+
+            let dmtf_device_info_size = cmd.dmtf_device_info_size as usize;
+            if dmtf_device_info_size > cmd.dmtf_device_info.len() {
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            }
+
+            // validate DMTF otherName
+            let dmtf_device_info_utf8 =
```
jhand2, line 41: nit: Might be worth moving to a private helper just to make it clear that we're just validating the input rather than doing any conversion

runtime/README.md
sree-revoori1: It is arbitrary. UTF8Strings have no max in the spec.

api/src/mailbox.rs
sree-revoori1: Done

api/src/mailbox.rs
sree-revoori1: Done

runtime/README.md
sree-revoori1: Done. I went ahead and aligned all the other columns in the readme too.

runtime/README.md
sree-revoori1: Done

runtime/src/subject_alt_name.rs
sree-revoori1: Done

runtime/src/certify_key_extended.rs
sree-revoori1: Done

runtime/src/certify_key_extended.rs
sree-revoori1: I tried doing this earlier but ran into some issues with the borrow checker. Will try again in a followup PR since this one is already big.

runtime/README.md
nquarton: This is a good point. The PR that adds LMS to this doc (https://github.com/chipsalliance/caliptra-sw/pull/1411) is still out, so I can just rebase it on this when it goes in. I'll comment to make sure it's not missed.

jhand2 (approved):

PR #1432, opened 2024-03-27
Address Runtime CFI Ordering issues

sree-revoori1:


runtime/src/main.rs
```diff
@@ -79,6 +78,12 @@ pub extern "C" fn entry_point() -> ! {
         cprintln!("[state] CFI Disabled");
     }
 
+    drivers.run_reset_flow().unwrap_or_else(|e| {
+        cprintln!("[rt] Runtime failed reset flow");
+        handle_fatal_error(e.into());
+    });
+    caliptra_common::stop_wdt(&mut drivers.soc_ifc);
```
jhand2, line 85: I'm not sure why we had this so far up before. I would think we don't really want to stop the watchdog until we're about to enter the mailbox loop: https://github.com/chipsalliance/caliptra-sw/blob/9d8206e21edc2b89d435fc86dca38e7acb835819/runtime/src/lib.rs#L265

cc @korran in case I'm missing something.

runtime/src/drivers.rs
```diff
@@ -176,6 +139,40 @@ impl Drivers {
         })
     }
 
+    pub fn run_reset_flow(&mut self) -> CaliptraResult<()> {
```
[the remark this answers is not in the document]
sree-revoori1: Done

runtime/src/main.rs
sree-revoori1: Done

jhand2 (approved):

PR #1433, opened 2024-03-27
Use normal for loop instead of map+fold for populating PCRs

jhand2:
Depending on optimizations applied by the compiler, it sometimes inserts a panic in the PCR accumulation code that depends on map/fold.

It is unclear in which cases it will insert one and which it won't, but for simplicity, replace this code with a for loop.

### DIFF
--- a/runtime/src/pcr.rs
+++ b/runtime/src/pcr.rs
@@ -50,24 +50,18 @@ impl GetPcrQuoteCmd {
             .ok_or(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
 
         let pcr_hash = drivers.sha384.gen_pcr_hash(args.nonce.into())?;
-
         let signature = drivers.ecc384.pcr_sign_flow(&mut drivers.trng)?;
-
         let raw_pcrs = drivers.pcr_bank.read_all_pcrs();
 
-        let pcrs_as_bytes = raw_pcrs
-            .into_iter()
-            .map(|raw_pcr_value| raw_pcr_value.into())
-            .enumerate()
-            .fold([[0; 48]; 32], |mut acc, (idx, pcr_value)| {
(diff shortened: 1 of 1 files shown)


runtime/src/pcr.rs
```diff
@@ -50,24 +50,18 @@ impl GetPcrQuoteCmd {
...
+        let mut pcrs = [[0u8; 48]; 32];
+        for (i, p) in raw_pcrs.iter().enumerate() {
+            pcrs[i] = p.into()
+        }
 
         Ok(MailboxResp::QuotePcrs(QuotePcrsResp {
             hdr: MailboxRespHeader::default(),
             nonce: args.nonce,
-            pcrs: pcrs_as_bytes,
+            pcrs: pcrs,
```
sree-revoori1, line 64: nit: No need to do `pcrs: pcrs`, can just do `pcrs,`

runtime/src/pcr.rs
```diff
@@ -15,7 +15,7 @@ Abstract:
 use crate::Drivers;
 use caliptra_cfi_derive_git::cfi_impl_fn;
 use caliptra_common::mailbox_api::{
-    ExtendPcrReq, IncrementPcrResetCounterReq, MailboxResp, MailboxRespHeader, QuotePcrsReq,
+    ExtendPcrReq, IncrementPcrResetCounterReq, MailboxResp, MailboxRespHeader, PcrValue, QuotePcrsReq,
```
sree-revoori1, line 18: Is this import necessary?

jhand2: No, removed

runtime/src/pcr.rs
jhand2: Done

sree-revoori1 (approved):

rusty1968 (approved):

PR #1440, opened 2024-04-01
Add most basic commands to execute_all_services_rt

jlmahowa-amd:


test/tests/fips_test_suite/services.rs
```diff
@@ -140,6 +298,103 @@ pub fn exec_cmd_stash_measurement<T: HwModel>(hw: &mut T) {
...
+    };
+    let fw_info_resp = mbx_send_and_check_resp_hdr::<_, FwInfoResp>(
+        hw,
+        u32::from(CommandId::FW_INFO),
+        payload.as_bytes(),
+    )
+    .unwrap();
+
+    // Verify command-specific response data
+    assert_eq!(fw_info_resp.fmc_revision, image_revision().unwrap());
```
nquarton, line 313: I think this revision check needs to be changed to just "contains_some_data" so it's compatible with released versions as well. The structs that are supposed to handle this won't really work since there is more than one commit/revision for ROM and FW that can correspond to a released version. (HW may be more predictable?)

jlmahowa-amd: I changed ROM to use "contains_some_data" but I added some logic to extract the fmc and rt revisions to pull from the fw_image, and tested that with released bins.

test/tests/fips_test_suite/services.rs
```diff
@@ -140,6 +298,104 @@ pub fn exec_cmd_stash_measurement<T: HwModel>(hw: &mut T) {
...
+        chksum: caliptra_common::checksum::calc_checksum(u32::from(CommandId::FW_INFO), &[]),
+    };
+    let fw_info_resp = mbx_send_and_check_resp_hdr::<_, FwInfoResp>(
+        hw,
+        u32::from(CommandId::FW_INFO),
+        payload.as_bytes(),
+    )
+    .unwrap();
+
+    let manifest = ImageManifest::read_from_prefix(&**fw_image).unwrap();
```
nquarton, line 312: I like this solution, good idea!

nquarton (approved):

PR #1446, opened 2024-04-04
Make x509 template generation optional

sree-revoori1:
Add static templates in case we don't build with the generate_templates flag.

### DIFF
--- a/x509/build/build.rs
+++ b/x509/build/build.rs
@@ -13,29 +13,40 @@ Abstract:
 
 --*/
 
+#[cfg(feature = "generate_templates")]
 mod cert;
+#[cfg(feature = "generate_templates")]
 mod code_gen;
+#[cfg(feature = "generate_templates")]
 mod csr;
+#[cfg(feature = "generate_templates")]
 mod tbs;
+#[cfg(feature = "generate_templates")]
 mod x509;
 
-use code_gen::CodeGen;
-use x509::{EcdsaSha384Algo, Fwid, FwidParam, KeyUsage};
-
-use std::env;
+#[cfg(feature = "generate_templates")]
+use {
+    code_gen::CodeGen,
+    std::env,
+    x509::{EcdsaSha384Algo, Fwid, FwidParam, KeyUsage},
+};
 
 // Main Entry point
 fn main() {
-    let out_dir_os_str = env::var_os("OUT_DIR").unwrap();
-    let out_dir = out_dir_os_str.to_str().unwrap();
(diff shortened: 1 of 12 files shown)


x509/build/build.rs
jhand2, line 1: We should keep all this code and maybe move it to a tool. Because we still want the ability to regenerate the templates if needed. We just don't need to be doing it on every build

x509/src/fmc_alias_cert.rs
```diff
@@ -13,7 +13,205 @@ Abstract:
 --*/
 
 // Note: All the necessary code is auto generated
-include!(concat!(env!("OUT_DIR"), "/fmc_alias_cert_tbs.rs"));
+pub struct FmcAliasCertTbsParams<'a> {
```
jhand2, line 16: Let's still check these in in separate files (can even keep the names). That way it'll be easier to regenerate them.

x509/build/build.rs
sree-revoori1: Done

x509/src/fmc_alias_cert.rs
sree-revoori1: Done

jhand2 (approved):

x509/build/fmc_alias_cert_tbs.rs
```diff
@@ -0,0 +1,212 @@
+/*++
```
korran, line 1: You should add a presubmit test that ensures this code is exactly the same as the generated code, similar to what we do for the generated register code:

https://github.com/chipsalliance/caliptra-sw/blob/b72c29345cb3734b7263968af16f757a58d8a505/.github/workflows/build-test-verilator.yml#L100

x509/build/fmc_alias_cert_tbs.rs
```diff
@@ -0,0 +1,212 @@
...
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    fmc_alias_cert_tbs.rs
+
+Abstract:
+
+    File contains generated FMC Alias Certificate template
```
korran, line 11: Make it clear that this is autogenerated by a particular tool, and make it clear to the reader how to regenerate.

sree-revoori1: Done

x509/build/fmc_alias_cert_tbs.rs
sree-revoori1: Done

jhand2 (approved):

PR #1447, opened 2024-04-04
Prevent mailbox buffer overflow vulnerability

sree-revoori1:


runtime/src/main.rs
```diff
@@ -91,7 +91,7 @@ pub extern "C" fn entry_point() -> ! {
     if let Err(e) = caliptra_runtime::handle_mailbox_commands(&mut drivers) {
         handle_fatal_error(e.into());
     }
-    loop {}
+    caliptra_drivers::ExitCtrl::exit(0);
```
jhand2, line 94: nit: depending on the error code, `GENERIC_OUTPUT_WIRES` will get a different value https://github.com/chipsalliance/caliptra-sw/blob/c92c83e2e7b982eb461e891752e8a0889ced171d/drivers/src/exit_ctrl.rs#L38

It seems like this will lead to specific bytes to be written to UART https://github.com/chipsalliance/caliptra-sw/blob/c92c83e2e7b982eb461e891752e8a0889ced171d/drivers/src/uart.rs#L40

From what I can tell poking around the code, I think `0xff` error code is supposed to denote a failure (see `fmc/src/main.rs`)

sree-revoori1: Done

jhand2 (approved):

PR #1454, opened 2024-04-08
Hw model cold reset support

nquarton:


hw-model/src/model_emulated.rs
```diff
@@ -262,4 +266,13 @@ impl crate::HwModel for ModelEmulated {
         self.cpu.warm_reset();
         self.step();
     }
+
+    fn cold_reset(&mut self) {
+        // TODO: Only saving ROM from init params right now
+        *self = ModelEmulated::new_unbooted(InitParams {
+            rom: &self.cold_reset_rom,
```
nquarton, line 273: I need a better solution for this. Right now this just saves the ROM. Should I just choose other params from InitParams that make sense to save? These are the options I considered and the problems with each.

**1. Store InitParams when we call new_unbooted. This is challenging because**

- InitParams contains several references, we can’t/don’t expect the objects they reference to outlive the model. We would need to store a copy of them somewhere that lives longer and reference that.
- There are several Box<dyn<…>> fields in InitParams that look messy to copy, but I think this is doable without pulling a new crate (dyn-clone) in?
- Some InitParams aren’t logical to carry over through a cold reset (ICCM, DCCM, trng nibbles/responses)

**2. Implement cold reset within the model implementation. This would actually reset reg values within the same instance**

- This looks like a large task and doesn’t seem like the right approach.

**3. Store the fresh copy of the CPU instance in model_emulated as a reset state in new_unbooted.**

- I tried to get CPU to derive clone, but eventually ended up in low-level components that were types from crates that did not or could not derive clone.

**4. Store the fields from InitParams we care about in the model_emulated struct and use them to create a new model in cold_reset.**

- Adds a maintenance cost if more fields are added to InitParams since we manually cherry pick fields to copy
- Adding an InitParam struct within ModelEmulated forces a lifetime annotation (due to the references in InitParams) this propagates all the way up to hw-model and makes a mess. Instead we’d probably want to individually duplicate the fields we want. The references in InitParams are not really useful anyway since we have to store a full copy of the data somewhere.

nquarton: Direction for now is to not support this in the SW emulator and push that decision making to the test. There can be a either be switch for different behavior in the sw-emulator instance (to create a new model) or the test can just be allowed to run on verilator and FPGA only

jlmahowa-amd (approved):

rom/dev/tests/rom_integration_tests/rv32_unit_tests.rs
```diff
@@ -6,15 +6,15 @@ use caliptra_hw_model::{BootParams, HwModel, InitParams};
 #[test]
 fn test_asm() {
     let rom = caliptra_builder::build_firmware_rom(&firmware::rom_tests::ASM_TESTS).unwrap();
-    let mut hw = caliptra_hw_model::new(BootParams {
-        init_params: InitParams {
+    let mut hw = caliptra_hw_model::new(
```
rusty1968, line 9: This seems fine, but Hardware model could benefit of using Rust's builder pattern. HwModelBuilder builds HwModel https://rust-unofficial.github.io/patterns/patterns/creational/builder.html

rusty1968 (approved):

hw-model/src/lib.rs
```diff
@@ -584,10 +597,14 @@ pub trait HwModel {
 
     /// Toggle reset pins and wait for ready_for_fuses
     fn warm_reset(&mut self) {
-        // sw-emulator lacks support: https://github.com/chipsalliance/caliptra-sw/issues/540
         panic!("warm_reset unimplemented");
     }
 
+    /// Toggle reset/pwrgood pins and wait for ready_for_fuses
+    fn cold_reset(&mut self) {
+        panic!("cold_reset unimplemented");
```
korran, line 605: This panic doesn't seem to jive with the PR title "Hw model cold reset support"

hw-model/src/lib.rs
```diff
@@ -250,6 +250,7 @@ pub struct BootParams<'a> {
     pub initial_repcnt_thresh_reg: Option<CptraItrngEntropyConfig1WriteVal>,
     pub initial_adaptp_thresh_reg: Option<CptraItrngEntropyConfig0WriteVal>,
```
korran, line 251: Did you forget to remove init_params from this struct?

hw-model/src/lib.rs
rusty1968: I think this is just re-factoring before the actual implementation. It should be named accordingly.

nquarton: Sorry for the confusion. I did just reply to Kor about this in an IM. This is the base implementation. The FPGA and verilator implementations override this. This is what you get on the SW emulator though (further explanation on that above).

I will be adding a note here to make this clearer.

hw-model/src/lib.rs
nquarton: I did forget this... I did all the work to separate them then missed the last step. Thank you for catching it!

nquarton: I have pushed a new patch that removes this

hw-model/src/lib.rs
nquarton: This has been added now

rom/dev/tests/rom_integration_tests/rv32_unit_tests.rs
nquarton: That's a good point. This could be a good place to use this in the future.

hw-model/src/lib.rs
```diff
@@ -476,77 +478,88 @@ pub trait HwModel {
     /// Create a model, and boot it to the point where CPU execution can
     /// occur. This includes programming the fuses, initializing the
     /// boot_fsm state machine, and (optionally) uploading firmware.
-    fn new(run_params: BootParams) -> Result<Self, Box<dyn Error>>
+    fn new(init_params: InitParams, boot_params: BootParams) -> Result<Self, Box<dyn Error>>
```
ArthurHeymans, line 479: Could the same be achieved without changing the function signature? That would remove the need to retouch all the tests.

nquarton: I definitely would have perferred to not change the API. The issue was the HW model combined the process of booting the HW model with the process of creating the model itself. This creates a verification coverage gap since you can't boot Caliptra multiple times without fully reinitializing the HW. This change separates those 2 processes to support that.

The alternative would be to leave init params as a member of boot params. That makes the API quite confusing since init is not part of boot anymore and those init parameters would then be ignored when calling boot().

ArthurHeymans: Thanks. that makes sense.

ArthurHeymans (approved):

jhand2 (dismissed):
Did a pass and I agree separating init an boot seems like the best way to handle modeling cold boot in the tests. The big refactor is unfortunate but I don't have any better ideas. Thanks!

JohnTraverAmd (approved):

JohnTraverAmd (approved):

FerralCoder (approved):

PR #1455, opened 2024-04-08
Add DPE commands to execute_all_services_rt

jlmahowa-amd:


nquarton (approved):

test/tests/fips_test_suite/common.rs
```diff
@@ -237,11 +237,101 @@ pub fn mbx_send_and_check_resp_hdr<T: HwModel, U: FromBytes + AsBytes>(
     //Ok(U::read_from(resp_bytes.as_bytes()).unwrap())
 }
 
-// Returns true if not all bytes are the same
+fn get_cmd_id(dpe_cmd: &mut Command) -> u32 {
```
jhand2, line 240: Should be able to use `u32::from(cmd)`: https://github.com/chipsalliance/caliptra-dpe/blob/main/dpe/src/commands/mod.rs#L89-L102

test/tests/fips_test_suite/common.rs
```diff
@@ -237,11 +237,101 @@ pub fn mbx_send_and_check_resp_hdr<T: HwModel, U: FromBytes + AsBytes>(
...
+        }
+        Command::DestroyCtx(_) => Response::DestroyCtx(ResponseHdr::read_from(resp_bytes).unwrap()),
+        Command::GetProfile => Response::GetProfile(GetProfileResp::read_from(resp_bytes).unwrap()),
+        Command::InitCtx(_) => Response::InitCtx(NewHandleResp::read_from(resp_bytes).unwrap()),
+        Command::RotateCtx(_) => Response::RotateCtx(NewHandleResp::read_from(resp_bytes).unwrap()),
+        Command::Sign(_) => Response::Sign(SignResp::read_from(resp_bytes).unwrap()),
+    }
+}
+
+pub fn execute_dpe_cmd<T: HwModel>(hw: &mut T, dpe_cmd: &mut Command) -> Response {
```
jhand2, line 284: nit: You could probably use https://github.com/chipsalliance/caliptra-sw/blob/cc31e5f5b349ba47c052648f3c529225f97962c0/runtime/tests/runtime_integration_tests/common.rs#L163

Although you might need to refactor and to share the implementation, so I'm fine if you want to defer this one.

test/tests/fips_test_suite/services.rs
jhand2, line 1: Just for my information, is the goal here to give integrators a single set of tests they can tell their FIPS auditor to run?

jhand2 (approved):

test/tests/fips_test_suite/services.rs
nquarton: That's correct. The goal is to cover all the cases in the test requirements for FIPS 140-3. Some portion of this is already covered in existing tests, but most requirements have specific flows or conditions that we don't explicitly cover. The tester/auditor can run this test suite on FPGA and eventually silicon (excluding tests that need ROM modifications) during the FIPS certification process.

test/tests/fips_test_suite/common.rs
jlmahowa-amd: From consumes the value, which we need to stick around.

jhand2: Ack. It would be pretty trivial to also implement `From<&Command>`. Something like:

```rust
impl From<&Command> for u32 {
  fn from (c: &Command) -> u32 {
    <normal implementation>
  }
}

impl From<Command> for u32 {
  fn from (c: Command) -> u32 {
    u32::from(&c)
  }
}
```

But not needed for this PR.

jhand2: Oh, especially since the change would have to be in the DPE repo.

PR #1460, opened 2024-04-10
Add RustCrypto implementation of image

sree-revoori1:


builder/src/sha256.rs
```diff
@@ -1,12 +1,16 @@
...
-use sha2::{Digest, Sha256};
+#[cfg(feature = "openssl")]
+use caliptra_image_crypto::OpensslHasher as Hasher;
+#[cfg(feature = "rustcrypto")]
+use caliptra_image_crypto::RustCryptoHasher as Hasher;
+use caliptra_image_crypto::Sha256Hasher;
 
 pub fn sha256_word_reversed(bytes: &[u8]) -> [u32; 8] {
-    let mut sha = Sha256::new();
+    let mut sha = Hasher::new();
```
jhand2, line 9: see https://github.com/chipsalliance/caliptra-sw/pull/1461. The only thing is we'll need to do `Crypto::default()` instead of `OsslCrypto::default()`

sree-revoori1: Left a comment.

jhand2 (approved):

mhatrevi (approved):

PR #1461, opened 2024-04-10
Update builder to use ImageGeneratorCrypto

jhand2:
Update builder so that all the image generation code uses the ImageGeneratorCrypto trait. This makes it easier to swap out the crypto implementation as needed for different integrations.

This ensures the whole image build can use the same crypto library.

### DIFF
--- a/image/gen/src/lib.rs
+++ b/image/gen/src/lib.rs
@@ -44,6 +44,9 @@ pub trait ImageGenratorExecutable {
 
 /// Image Gnerator Crypto Trait
 pub trait ImageGeneratorCrypto {
+    /// Calculate SHA-256 digest
+    fn sha256_digest(&self, data: &[u8]) -> anyhow::Result<[u32; SHA256_DIGEST_WORD_SIZE]>;
+
     /// Calculate SHA-384 digest
     fn sha384_digest(&self, data: &[u8]) -> anyhow::Result<ImageDigest>;
 
--- a/image/gen/src/lib.rs
+++ b/image/gen/src/lib.rs
@@ -42,10 +42,26 @@ pub trait ImageGenratorExecutable {
     fn size(&self) -> u32;
 }
 
+pub trait ImageGeneratorHasher {
+    type Output: Copy;
+
+    fn update(&mut self, data: &[u8]);
+
+    fn finish(self) -> Self::Output;
+}
+
 /// Image Gnerator Crypto Trait
 pub trait ImageGeneratorCrypto {
+    type Sha256Hasher: ImageGeneratorHasher<Output = [u32; SHA256_DIGEST_WORD_SIZE]>;
+
+    fn sha256_start(&self) -> Self::Sha256Hasher;
+
     /// Calculate SHA-256 digest
-    fn sha256_digest(&self, data: &[u8]) -> anyhow::Result<[u32; SHA256_DIGEST_WORD_SIZE]>;
+    fn sha256_digest(&self, data: &[u8]) -> anyhow::Result<[u32; SHA256_DIGEST_WORD_SIZE]> {
+        let mut hasher = self.sha256_start();
+        hasher.update(data);
+        Ok(hasher.finish())
+    }
 
     /// Calculate SHA-384 digest
     fn sha384_digest(&self, data: &[u8]) -> anyhow::Result<ImageDigest>;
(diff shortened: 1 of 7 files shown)


sree-revoori1:
I would prefer to have an update/finish interface instead of one digest function for hashing. Reason is that we use the update/finish interface in other places in image/crypto and we cannot use vector since its no_std. We could allocate a const buffer, and copy all the individual pieces we want to update the hasher with, but this would lead to a lot more messy code. 

This is the reason I moved the hashing logic to a new interface and parametrized the functions over that interface in https://github.com/chipsalliance/caliptra-sw/pull/1460.

image/gen/src/lib.rs
```diff
@@ -44,6 +44,9 @@ pub trait ImageGenratorExecutable {
 
 /// Image Gnerator Crypto Trait
 pub trait ImageGeneratorCrypto {
+    /// Calculate SHA-384 digest
```
sree-revoori1, line 47: *SHA-256

image/gen/src/lib.rs
```diff
@@ -44,6 +44,9 @@ pub trait ImageGenratorExecutable {
...
+    fn sha256_digest(&self, data: &[u8]) -> anyhow::Result<[u32; 8]>;
```
sree-revoori1, line 48: Can you make 8 a const in image/types and check that it doesn't change the ROM binary? If it does you can leave the const in this file.

jhand2: Done

image/gen/src/lib.rs
jhand2: Done

sree-revoori1 (dismissed):

korran:
> ~I would prefer to have an update/finish interface instead of one digest function for hashing. Reason is that we use the update/finish interface in other places in image/crypto and we cannot use vector since its no_std. We could allocate a const buffer, and copy all the individual pieces we want to update the hasher with, but this would lead to a lot more messy code. ~
> 
> ~This is the reason I moved the hashing logic to a new interface and parametrized the functions over that interface in #1460.~
> 
> Actually looks like it is not no_std so we can just use vector. So this LGTM

This is a performance regression however. I would prefer that we fix the "preferred abstraction" to not be a regression before forcing everyone to use it.

image/openssl/src/lib.rs
```diff
@@ -45,6 +45,12 @@ const D_LEAF: u16 = 0x8282;
 const D_INTR: u16 = 0x8383;
 
 impl ImageGeneratorCrypto for OsslCrypto {
+    fn sha256_digest(&self, data: &[u8]) -> anyhow::Result<[u32; SHA256_DIGEST_WORD_SIZE]> {
```
korran, line 48: This function can never fail. I don't understand why the output is wrapped in Result.

korran:
> > ~I would prefer to have an update/finish interface instead of one digest function for hashing. Reason is that we use the update/finish interface in other places in image/crypto and we cannot use vector since its no_std. We could allocate a const buffer, and copy all the individual pieces we want to update the hasher with, but this would lead to a lot more messy code. ~
> > ~This is the reason I moved the hashing logic to a new interface and parametrized the functions over that interface in #1460.~
> > Actually looks like it is not no_std so we can just use vector. So this LGTM
> 
> This is a significant memory-usage and performance regression, and we're hashing a fair bit of data here. I would prefer that we fix the "preferred abstraction" to not be a regression before forcing everyone to use it.

https://github.com/jhand2/caliptra-sw/pull/3/commits shows how to maintain the original semantics.

image/openssl/src/lib.rs
jhand2: I only did it this way because that's how sha384_digest was. I guess the idea is that in some implementations it might fail.

jhand2:
> > > ~I would prefer to have an update/finish interface instead of one digest function for hashing. Reason is that we use the update/finish interface in other places in image/crypto and we cannot use vector since its no_std. We could allocate a const buffer, and copy all the individual pieces we want to update the hasher with, but this would lead to a lot more messy code. ~
> > > ~This is the reason I moved the hashing logic to a new interface and parametrized the functions over that interface in #1460.~
> > > Actually looks like it is not no_std so we can just use vector. So this LGTM
> > 
> > 
> > This is a significant memory-usage and performance regression, and we're hashing a fair bit of data here. I would prefer that we fix the "preferred abstraction" to not be a regression before forcing everyone to use it.
> 
> [jhand2#3](https://github.com/jhand2/caliptra-sw/pull/3) shows how to maintain the original API semantics.

Thanks, merged your commit

jlmahowa-amd (approved):

mhatrevi (approved):

PR #1465, opened 2024-04-16
Enable shutdown test in execute_all_services_rt

jlmahowa-amd:


nquarton (approved):
The fix I was referring to was https://github.com/chipsalliance/caliptra-sw/pull/1351

Thanks for uncommenting this, I had forgotten about it.

PR #1468, opened 2024-04-17
Changing build_release to use latest RTL

nquarton:


nquarton:
The build_release script already uses the latest build for ROM, FMC, and FW. This just updates the RTL path to match.

(Originally I was considering building both a latest and 1.0 release. However, we don't have a 1.0 FMC or FW available on main, so that's not possible.)

jlmahowa-amd (approved):

PR #1475, opened 2024-04-19
ECC driver error code fixes

nquarton:


drivers/src/ecc384.rs
```diff
@@ -379,9 +382,14 @@ impl Ecc384 {
...
 
+        // Verify the signature just created
         let r = self.verify_r(pub_key, data, sig)?;
-        caliptra_cfi_lib::cfi_assert_eq_12_words(&r.0, &sig.r.0);
-        sig_result
+        if r == sig.r {
+            caliptra_cfi_lib::cfi_assert_eq_12_words(&r.0, &sig.r.0);
+            sig_result
+        } else {
+            Err(CaliptraError::DRIVER_ECC384_SIGN_VALIDATION_FAILED)
```
FerralCoder, line 391: I believe this should still be DRIVER_ECC384_KEYGEN_PAIRWISE_CONSISTENCY_FAILURE.

PCT means using the newly generated key to sign something and verify that sig. If that process fails, the PCT fails.

nquarton: So I used this error because my understanding was sign() was just the generic signing function that can be called from a few places. This error seemed like the right match for this case.

One of those uses is that key_pair() generation function. In that function, if this sign call fails, key_pair() returns DRIVER_ECC384_KEYGEN_PAIRWISE_CONSISTENCY_FAILURE instead of directly passing this failure up (see line 270). So you'll never actually see this SIGN_VALIDATION_FAILED error when generating a key pair, but you could see it from other callers.

mhatrevi (approved):

drivers/src/ecc384.rs
FerralCoder: sorry, that was my mistake when reviewing. The diff viewer was skipping lines and I thought this was still part of the key_pair() function. I have no concern here.

drivers/src/ecc384.rs
```diff
@@ -379,9 +382,14 @@ impl Ecc384 {
         let mut sig_result = self.sign_internal(priv_key, data, trng);
         let sig = okmutref(&mut sig_result)?;
 
+        // Verify the signature just created
         let r = self.verify_r(pub_key, data, sig)?;
-        caliptra_cfi_lib::cfi_assert_eq_12_words(&r.0, &sig.r.0);
```
korran, line 383: I believe this change is a CFI regression.

One of the most important things is to make sure that data derived from our private key doesn't get leaked out, which is why this CFI-robust check is the first thing we do.

The attack scenario here is somebody glitching the chip trying to interrupt the ECC engine while there is sensitive data in the register. Before your change, we had a very hard to glitch unrolled inlined check on every word of that signature, and if any of those checks were to fail, we jump directly to CFI jail. Now, we are doing a regular comparison, and if that fails, we follow the regular error-handling path, which is much more vulnerable to glitches.

Attack scenario:

1. Attacker glitches caliptra while ECC computation is occurring. The signature now contains a small amount of entropy from the private key.

2. The `r == sig.r` comparison fails, so this function returns with DRIVER_ECC384_SIGN_VALIDATION_FAILED.

3. Attacker does their first glitch of the firmware, causing it to jump from the error handling path back to the success path. The firmware copies the corrupt signature into the mailbox response, and now the attacker has a bit of info about the private key.

4. The attacker performs the above attack thousands of times, and collects enough info about the private key to reconstitute it.

nquarton: Per our conversation on slack, creating a PR to revert this part of the change now. Thanks for noticing this!

PR #1488, opened 2024-04-29
Changing RT self test failure behavior to fatal error and unlock mailbox

nquarton:


nquarton:
I believe the current behavior does satisfy FIPS requirements. A self test failure will currently report a non-fatal error and leave the mailbox locked. I feel this may not be very intuitive to an SoC FW developer though since mailbox commands will hang and no fatal error is reported.

This change makes the error a fatal error (which still guarantees the SoC cannot use the module) and unlocks the mailbox so subsequent commands (including SELF_TEST_GET_RESULTS) will return a failure with the error code reported in the self test.

jlmahowa-amd (dismissed):

jhand2 (approved):

mhatrevi (approved):

PR #1503, opened 2024-05-08
FMC Compat. check

rusty1968:


fmc/src/main.rs
```diff
@@ -60,6 +61,19 @@ pub extern "C" fn entry_point() -> ! {
         Ok(env) => env,
         Err(e) => handle_fatal_error(e.into()),
     };
+    if !cfg!(feature = "fake-fmc") {
+        let rom_version = RomVersion::from(env.soc_ifc.get_rom_fw_rev_id());
+        // FMC to be removed in Caliptra 2.0
+        if rom_version.major != 1 {
```
embediver, line 67: Wouldn't it be better to not have this hardcoded here as 1.x.x?
And how is FMC getting removed in 2.0? I can't find a reference to that.

Having the FMC version specified somewhere else or having it generated would be better imo (for forward-porting).

common/src/lib.rs
```diff
@@ -42,5 +42,35 @@ pub const FMC_SIZE: u32 = 20 * 1024;
...
+impl core::fmt::Debug for RomVersion {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("RomVersion")
+            .field("major", &self.major)
+            .field("minor", &self.minor)
+            .field("patch", &self.patch)
+            .finish()
+    }
+}
+
```
embediver, line 74: Isn't this the same as a regular `#[derive(Debug)]`?

A Dedicated `Display` impl which displays a Semver (e.g. `1.2.0`) might be nice though.

PR #1507, opened 2024-05-10
Adding LMS KAT to RT boot

nquarton:


nquarton:
This is the non-ideal approach of always running the LMS KAT at the start of RT FW (before processing commands). Ideally, this would be run exactly once as/if needed before an LMS operation is performed in runtime. This gets a little tricky in rust though as it requires static state to be shared about whether or not the KAT has been run.

sree-revoori1:
Can we add a bool to the drivers struct indicating whether or not the KAT has been run instead?

nquarton:
> Can we add a bool to the drivers struct indicating whether or not the KAT has been run instead?

I did consider that. It's an accessible place to put some state. The only issue is the drivers don't bring in KATs as a crate. So either
1. The runtime code needs to enforce this rule, check/update the state, and run the KAT. Certainly possible but feels a little disaggregated.
2. The drivers need to have a circular dependency on the KATs (is this okay/allowed?)

But if people still like this path (or I'm missing a better option) then I can absolutely do this.

nquarton:
I have updated this implementation to store state in the LMS driver.

jhand2 (approved):

mhatrevi (dismissed):

jhand2 (approved):

mhatrevi (approved):

PR #1536, opened 2024-05-23
Properly evaluate PL0 PAUSERs when the fw header bit is unset

jhand2:
Previously, if PL0 PAUSER bit was unset, all users were treated as PL0. Align with the spec and treat them all as PL1 in that case.

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -590,7 +590,8 @@ impl Drivers {
     /// * `flags` - Flags from manifest header
     /// * `locality` - Caller's locality
     pub fn is_caller_pl1(pl0_pauser: u32, flags: u32, locality: u32) -> bool {
-        flags & PL0_PAUSER_FLAG == 0 && locality != pl0_pauser
+        (flags & PL0_PAUSER_FLAG == 0) // There is no PL0 PAUSER
+            || (locality != pl0_pauser) // There is a PL0 PAUSER, but it's not the current user
     }
 
     /// Get the KeyId for the RT Alias CDI
--- a/runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
+++ b/runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
@@ -1,15 +1,23 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_builder::ImageOptions;
+use caliptra_builder::{
+    build_firmware_elf,
+    firmware::{APP_WITH_UART, FMC_WITH_UART},
+    ImageOptions,
+};
 use caliptra_common::mailbox_api::{
     CertifyKeyExtendedFlags, CertifyKeyExtendedReq, CommandId, MailboxReq, MailboxReqHeader,
     PopulateIdevCertReq, StashMeasurementReq,
 };
 use caliptra_error::CaliptraError;
-use caliptra_hw_model::HwModel;
+use caliptra_hw_model::{BootParams, Fuses, HwModel, InitParams, SecurityState};
+use caliptra_image_crypto::OsslCrypto as Crypto;
+use caliptra_image_elf::ElfExecutable;
+use caliptra_image_gen::{ImageGenerator, ImageGeneratorConfig};
 use caliptra_runtime::{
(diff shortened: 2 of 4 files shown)


runtime/src/drivers.rs
```diff
@@ -590,7 +590,8 @@ impl Drivers {
     /// * `flags` - Flags from manifest header
     /// * `locality` - Caller's locality
     pub fn is_caller_pl1(pl0_pauser: u32, flags: u32, locality: u32) -> bool {
-        flags & PL0_PAUSER_FLAG == 0 && locality != pl0_pauser
+        (flags & PL0_PAUSER_FLAG == 0) // There is no PL0 PAUSER
+            || (locality != pl0_pauser) // This is a PL0 PAUSER, but it's not the current user
```
bluegate010, line 594: s/This is/There is/

bluegate010 (dismissed):

runtime/src/drivers.rs
jhand2: Done

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
```diff
@@ -453,3 +453,46 @@ fn test_measurement_log_pl_context_threshold() {
...
+            security_state: SecurityState::from(fuses.life_cycle as u32),
+            ..Default::default()
+        },
+        fuses,
+        ..Default::default()
+    })
+    .unwrap();
+
+    let mut opts = ImageOptions::default();
+    opts.vendor_config.pl0_pauser = None;
```
jlmahowa-amd, line 473: This test, as is, does not fail when the change to drivers.rs is reverted. This covers the pl0_pauser not set, and locality != pl0_pauser case that I believe other tests here already cover.

There are two cases I think we should cover:
pl0_pauser_flag not set and locality == pl0_pauser. To hit this we would need to use model.set_apb_pauser to set the pauser to 0. Unfortunately the SW model does not implement this function at the moment.
pl0_pauser_flag set and locality != pl0_pauser. For this we would provide Some(0) (any value other than 1)

jhand2: > This test, as is, does not fail when the change to drivers.rs is reverted

Are you sure? I did check that it failed before making the driver change.

> pl0_pauser_flag not set and locality == pl0_pauser. To hit this we would need to use model.set_apb_pauser to set the pauser to 0. Unfortunately the SW model does not implement this function at the moment.
pl0_pauser_flag set and locality != pl0_pauser. For this we would provide Some(0) (any value other than 1)

Good point. I'll add another test for this case.

jhand2: Added test for the second case.

jhand2: Never mind, you're right that it doesn't fail. Looking into this.

jhand2: Alright, the test ended up being very complicated because I have to change a header bit and then re-sign. But it now fails/passes in the correct scenarios.

jlmahowa-amd (approved):

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
```diff
@@ -453,3 +458,133 @@ fn test_measurement_log_pl_context_threshold() {
...
+
+    model
+        .upload_firmware(&image_bundle.to_bytes().unwrap())
+        .unwrap();
+
+    model.step_until(|m| {
+        m.soc_ifc().cptra_boot_status().read() == u32::from(RtBootStatus::RtReadyForCommands)
+    });
+
+    // If PL0 PAUSER is unset, make sure PL0-only operation fails
```
bluegate010, line 577: Update comment?

bluegate010 (dismissed):

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
jhand2: Done

PR #1546, opened 2024-05-28
Runtime CFI check of mailbox lock causes race condition

rusty1968:
The second read of the lock register was inserted to prevent a glitch attack from bypassing the FIPS self test logic and execute a different mailbox command. However, the second read causes a race condition and the SoC can maliciously or accidentally force Caliptra to crash by releasing the lock at the time the CFI check of the lock is performed.

Given an attacker can prevent FIPS self-test from happening by just controlling the acquisition and release of the lock, the CFI check is not protecting much and can potentially be removed altogether.


sree-revoori1:
LGTM, thanks for finding this

PR #1561, opened 2024-06-07
Adjust maximum number of DPE handles.

rusty1968:
Adjust the maximum number of available DPE handles.
    
This commit contains the following changes:

-  Maximum number of DPE handles  configurable via cargo feature. 
-  Reduced DATA_SIZE  to make room for the extra DPE storage.


fdamato:
Do we have a test that verifies that the certificate doesn`t exceed the max size (2kB) ?

rusty1968:
> Do we have a test that verifies that the certificate doesn`t exceed the max size (2kB) ?

Is this test_pl0_derive_context_dpe_context_thresholds test doing what we need?

runtime/build.sh
```diff
@@ -4,7 +4,8 @@
 
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
-cargo build \
+ARBITRARY_MAX_HANDLES=32 cargo build \
```
jhand2, line 7: It might be easier to just hard code this to 32 for now and set the environment variable in the RT build script.

I think how this is written right now, all the tests are still using the default 24. And I'm not sure we really want to test both configurations.

rusty1968: Hi, Jordan. Do you mean hardcode in caliptra-dpe?

runtime/Cargo.toml
```diff
@@ -58,3 +58,4 @@ fips_self_test=[]
 no-cfi = ["caliptra-image-verify/no-cfi", "caliptra-drivers/no-cfi"]
 fpga_realtime = ["caliptra-drivers/fpga_realtime"]
 "hw-1.0" = ["caliptra-builder/hw-1.0", "caliptra-drivers/hw-1.0", "caliptra-registers/hw-1.0", "caliptra-kat/hw-1.0","caliptra-cpu/hw-1.0"]
+arbitrary_max_handles = ["dpe/arbitrary_max_handles"]
```
jhand2, line 61: I don't feel too strongly about this, but we probably don't need to define a feature on the RT crate for this. On the DPE dependency we can just add `features = ["arbitrary_max_handles"]`

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
```diff
@@ -400,6 +400,7 @@ fn test_stash_measurement_pl_context_thresholds() {
     }
 }
 
+#[ignore]
```
jhand2, line 403: Why skip this test?

drivers/src/memory_layout.rs
```diff
@@ -63,13 +68,15 @@ pub const RTALIAS_TBS_SIZE: u32 = 1024;
 pub const PCR_LOG_SIZE: u32 = 1024;
 pub const MEASUREMENT_LOG_SIZE: u32 = 1024;
 pub const FUSE_LOG_SIZE: u32 = 1024;
-pub const DPE_SIZE: u32 = 4 * 1024;
+pub const DPE_SIZE: u32 = 5 * 1024;
 pub const PCR_RESET_COUNTER_SIZE: u32 = 1024;
 pub const DATA_SIZE: u32 = 78 * 1024;
 pub const STACK_SIZE: u32 = 22 * 1024;
 pub const ROM_STACK_SIZE: u32 = 14 * 1024;
-pub const ESTACK_SIZE: u32 = 1024;
```
jhand2, line 71: I think I would prefer to shrink the data size rather than change the stack size. I don't think we're using nearly that much data.

fdamato: I agree...reducing stack size is risky...

drivers/src/memory_layout.rs
```diff
@@ -38,12 +38,17 @@ pub const PCR_LOG_ORG: u32 = 0x50004800;
...
+pub const STACK_ORG: u32 = DATA_ORG + DATA_SIZE;
 pub const ROM_STACK_ORG: u32 = 0x5001C000;
-pub const ESTACK_ORG: u32 = 0x5001F800;
-pub const NSTACK_ORG: u32 = 0x5001FC00;
+
+pub const ESTACK_ORG: u32 = STACK_ORG + STACK_SIZE;
+pub const ROM_ESTACK_ORG: u32 = 0x5001F800;
+
+pub const NSTACK_ORG: u32 = ESTACK_ORG + ESTACK_SIZE;
+pub const ROM_NSTACK_ORG: u32 = 0x5001FC00;
```
jhand2, line 51: Because of RT hitless update, it makes me a little nervous to compute these offsets dynamically. I think we may need to be careful about which things get moved across hitless update.

Do we need a concept of an RT that makes a breaking change to the memory model and can't be hitlessly update to.

jhand2 (approved):

nquarton:
Is this change supposed to be backported to 1.0? Or is this more of a behavioral change for 1.1 onward?

PR #1562, opened 2024-06-07
Changing SHA384ACC KAT to SHA512

nquarton:
- Adding support for SHA512 digest in the SHA acclerator driver
- Enabling support for SHA512 in the SW emulator SHA accelerator


nquarton:
I'm looking for input on where to draw the line on sha384acc naming vs sha512acc. This is a single driver and the old/current name is sha384acc. sha512acc is more correct, but would require touching several files.

For now, I have only changed the "externally visible" naming to sha512acc (in the KAT logs and error codes). But I'm very open to updating the driver naming everywhere if people think that is the right direction.

JohnTraverAmd:
The HW Team has named this HW the sha512_acc.  I wouldn't mind using the same name as the HW.  

The other option is to take the numbers off and just call it the sha_accelerator.

nquarton:
Per the discussion in the FW sync today, I have updated the name to sha2_512_384acc. The name change is a separate commit to make reviewing this PR easier.

mhatrevi (approved):

rusty1968 (approved):

PR #1573, opened 2024-06-14
Make caliptra-coverage optional for the emulated hw-model.

zhalvorsen:
This adds a new default feature flag for the hw-model. caliptra-coverage depends on caliptra-builder which uses cargo directly. This can make using the hw-model easier for build environments that may not have cargo accessible.

### DIFF
--- a/hw-model/src/model_emulated.rs
+++ b/hw-model/src/model_emulated.rs
@@ -9,7 +9,9 @@ use std::path::PathBuf;
 use std::rc::Rc;
 
 use caliptra_emu_bus::Clock;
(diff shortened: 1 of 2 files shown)


hw-model/src/model_emulated.rs
```diff
@@ -58,10 +60,13 @@ pub struct ModelEmulated {
     cpu_enabled: Rc<Cell<bool>>,
     trace_path: Option<PathBuf>,
 
+    #[cfg(feature = "coverage")]
```
jhand2, line 63: nit: Even though it would be unused, should we still accept the image tag as input just to make the interface consistent?

jhand2 (approved):

PR #1583, opened 2024-06-24
Implement API for requesting/retrieving  Caliptra's IDEVID CSR.

rusty1968:
This commit adds functionality to request and retrieve an IDEVID Certificate Signing Request.


libcaliptra/examples/generic/idev_csr_array.h
```diff
@@ -0,0 +1,28 @@
+// Licensed under the Apache-2.0 license
+
+#include <stdint.h>
+#define IDEV_CSR_LEN  443
+uint8_t idev_csr_bytes[IDEV_CSR_LEN] = {
+    48, 130, 1, 183, 48, 130, 1, 62, 2, 1, 0, 48, 105, 49, 28, 48, 26, 6, 
+    3, 85, 4, 3, 12, 19, 67, 97, 108, 105, 112, 116, 114, 97, 32, 49, 46, 
+    48, 32, 73, 68, 101, 118, 73, 68, 49, 73, 48, 71, 6, 3, 85, 4, 5, 19,
```
fdamato, line 8: where does this CSR array come from ?

libcaliptra/examples/generic/main.c
```diff
@@ -66,13 +74,13 @@ void dump_caliptra_error_codes()
...
 
-int boot_to_ready_for_fw()
+int boot_to_ready_for_fw(bool req_idev_csr)
 {
     int status;
 
     // Initialize FSM GO
     caliptra_bootfsm_go();
-
+
```
fdamato, line 83: NP: why adding this NL

libcaliptra/src/caliptra_api.c
```diff
@@ -348,6 +351,45 @@ bool caliptra_ready_for_firmware(void)
...
+*
+* @return bool True if ready, false otherwise    
+*/
+void caliptra_wait_for_csr_ready(void) 
+{
+    while (true) 
+    {
+        if (caliptra_is_csr_ready()) {
+            break;
+        }
```
fdamato, line 388: can we add a timeout here ?

libcaliptra/src/caliptra_api.c
```diff
@@ -1060,4 +1102,53 @@ int caliptra_capabilities(struct caliptra_capabilities_resp *resp, bool async)
     CREATE_PARCEL(p, OP_CAPABILITIES, &checksum, resp);
 
     return pack_and_execute_command(&p, async);
-}
\ No newline at end of file
+}
+
+int caliptra_retrieve_idevid_csr(struct caliptra_buffer* caliptra_idevid_csr) 
+{
+    if (!caliptra_is_idevid_csr_ready()) {
```
fdamato, line 1109: can we do a param check for caliptra_idevid_csr?

libcaliptra/src/caliptra_mbox.h
```diff
@@ -128,7 +128,7 @@ static inline bool caliptra_mbox_is_busy(void)
 
 static inline uint8_t caliptra_mbox_read_status_fsm(void)
 {
-    return (uint8_t)(caliptra_mbox_read(MBOX_CSR_MBOX_STATUS) >> 16 & MBOX_CSR_MBOX_STATUS_STATUS_MASK);
+    return (uint8_t)(caliptra_mbox_read(MBOX_CSR_MBOX_STATUS) >> 6) & 7;
```
fdamato, line 131: what are 6 and 7 ?

libcaliptra/examples/generic/main.c
```diff
@@ -263,7 +275,7 @@ int rt_test_all_commands()
 {
     int failure = 0;
     uint32_t non_fatal_error;
-    int status = boot_to_ready_for_fw();
+    int status = boot_to_ready_for_fw(false);
 
     if (status){
```
fdamato, line 280: why are we overriding the actual cause of failure with 1 ?

libcaliptra/examples/generic/main.c
```diff
@@ -609,6 +621,55 @@ int rt_test_all_commands()
...
+
+    // Check if the buffer was allocated successfully
+    if (caliptra_idevid_csr_buf.data == NULL) {
+        printf("Failed to allocate memory for IDEV CSR\n");
+        return 1;
+    }      
+ 
+    bool request_csr = true;
+    int status = boot_to_ready_for_fw(request_csr);
+
```
fdamato, line 641: why are we overriding the failure ? this function returns an int, but we are treating it as a bool

libcaliptra/examples/generic/main.c
rusty1968: I don't have the context. I was not the original implementer of the testbench.

libcaliptra/examples/generic/main.c
nquarton: Looks like failure just tracks whether or not there was a failure to report to the test caller. A bool would make more sense for it. If I wrote this, it was probably habit from a c environment that did not support the bool type.

libcaliptra/src/caliptra_api.c
nquarton: I don't think libcaliptra can have any platform-agnostic way of supporting timing. If the caller wants a timeout, I think they would need to directly use caliptra_is_csr_ready() in their own generic test-with-timeout function/macro.

libcaliptra/src/caliptra_api.c
```diff
@@ -269,6 +269,9 @@ int caliptra_init_fuses(struct caliptra_fuses *fuses)
     if (!caliptra_ready_for_fuses())
         return NOT_READY_FOR_FUSES;
 
+    // Set Manuf service reg
+    caliptra_generic_and_fuse_write(GENERIC_AND_FUSE_REG_CPTRA_DBG_MANUF_SERVICE_REG, fuses->dbg_manuf_service);
```
nquarton, line 273: I don't think this should be included with the fuses. If it doesn't exist, we could add a separate wrapper to write it.

Alternatively, there could be a function that takes the fuses and other init regs (dbg_manuf_service, wdt, trng config, pauser, etc.) that sets those and also wraps this fuse init call.

libcaliptra/src/caliptra_api.c
rusty1968: done

libcaliptra/src/caliptra_mbox.h
rusty1968: i changed it to use the #defines available in the headers.

libcaliptra/examples/generic/main.c
rusty1968: I would like to narrow down the scope of the review to the functionality  I am trying to add. We can address the testbench in another PR.

libcaliptra/inc/caliptra_api.h
```diff
@@ -171,4 +177,12 @@ int caliptra_shutdown(bool async);
 // Capabilities
 int caliptra_capabilities(struct caliptra_capabilities_resp *resp, bool async);
 
+bool caliptra_is_idevid_csr_ready();
```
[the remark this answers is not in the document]
rusty1968: We only have one now.

libcaliptra/src/caliptra_fuses.h
```diff
@@ -102,4 +109,10 @@ static inline void caliptra_set_fuse_pauser_lock()
...
 {
     caliptra_generic_and_fuse_write(GENERIC_AND_FUSE_REG_CPTRA_FUSE_VALID_PAUSER, data);
-}
\ No newline at end of file
+}
+
+static inline void caliptra_write_dbg_manuf_serv(uint32_t data) 
+{
+    // Set Manuf service reg
+    caliptra_generic_and_fuse_write(GENERIC_AND_FUSE_REG_CPTRA_DBG_MANUF_SERVICE_REG, 1);
```
nquarton, line 117: I think this is supposed to be data instead of 1

libcaliptra/inc/caliptra_types.h
```diff
@@ -37,6 +39,7 @@ struct caliptra_fuses {
     bool anti_rollback_disable;
     uint32_t idevid_cert_attr[24];
     uint32_t idevid_manuf_hsm_id[4];
+    uint32_t dbg_manuf_service;
```
nquarton, line 42: Did you mean to remove this from here too?

nquarton (commented):
Just 2 small things then I think it's good to go from my perspective

nquarton (approved):

nquarton (approved):

libcaliptra/examples/generic/main.c
rusty1968: removed

libcaliptra/src/caliptra_api.c
rusty1968: I removed this function from the public api and left it as test fixture.

libcaliptra/src/caliptra_api.c
rusty1968: done.

fdamato (dismissed):

rusty1968:
@jlmahowa-amd , @fdamato

libcaliptra/src/caliptra_mbox.h
```diff
@@ -139,4 +139,8 @@ static inline uint32_t caliptra_mbox_read_dlen(void)
 static inline void caliptra_mbox_write_dlen(uint32_t dlen)
 {
     caliptra_mbox_write(MBOX_CSR_MBOX_DLEN, dlen);
-}
\ No newline at end of file
+}
+
+
+    
+
```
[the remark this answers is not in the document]
rusty1968: done

fdamato (approved):

PR #1589, opened 2024-06-27
fuzzing: Update dependencies

benjamindoron:
Also, to avoid doing this more than necessary, specify only the major and minor version. This should be enough to avoid all breakages, while allowing the build number to update.

### DIFF
--- a/drivers/fuzz/Cargo.toml
+++ b/drivers/fuzz/Cargo.toml
@@ -10,15 +10,15 @@ edition = "2021"
 cargo-fuzz = true
 
 [dependencies]
-libfuzzer-sys = { version = "0.4.6", optional = true }
-afl = { version = "0.13.3", optional = true }
-zerocopy = "0.6.1"
-arbitrary = { version = "1.3.0", optional = true, features = ["derive"] }
-sha2 = { version = "0.10.2", default-features = false, features = ["compress"] }
+libfuzzer-sys = { version = "0.4", optional = true }
+afl = { version = "0.15", optional = true }
+zerocopy = "0.6"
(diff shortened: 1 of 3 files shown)


drivers/fuzz/Cargo.toml
```diff
@@ -10,15 +10,15 @@ edition = "2021"
...
+afl = { version = "0.15", optional = true }
+zerocopy = "0.6"
+arbitrary = { version = "1.3", optional = true, features = ["derive"] }
+sha2 = { version = "0.10", default-features = false, features = ["compress"] }
 
 [patch.crates-io]
-byteorder = { git = "https://github.com/benjamindoron/byteorder.git", branch = "struct_aware-1.4.3" }
-zerocopy = { git = "https://github.com/benjamindoron/zerocopy.git", branch = "struct_aware-v0.6.3" }
+byteorder = { git = "https://github.com/benjamindoron/byteorder.git", branch = "struct_aware" }
+zerocopy = { git = "https://github.com/benjamindoron/zerocopy.git", branch = "struct_aware" }
```
leongross, line 21: It looks like @benjamindoron used his fork with changes on branches `struct_aware` for fuzzing. I think it would make more sense if we moved the forks with their respective changes to the `chipsalliance` org, so keeping it up to date and rebased is easier.

@jhand2 @benjamindoron was there a reason this was forked to a private account instead of the org?

benjamindoron: We didn't have org permissions at the time, and I think it's as simple as nobody minding because these aren't main build targets. Yes, I agree we might want to move these to the org: I can occasionally push updates, but it would be easier if not. Lately, I've been working on other projects.

PR #1594, opened 2024-06-27
Fips test suite tests

nquarton:


jlmahowa-amd (approved):

mhatrevi (approved):

runtime/Cargo.toml
```diff
@@ -59,3 +59,4 @@ fips_self_test=[]
 no-cfi = ["caliptra-image-verify/no-cfi", "caliptra-drivers/no-cfi"]
 fpga_realtime = ["caliptra-drivers/fpga_realtime"]
 "hw-1.0" = ["caliptra-builder/hw-1.0", "caliptra-drivers/hw-1.0", "caliptra-registers/hw-1.0", "caliptra-kat/hw-1.0","caliptra-cpu/hw-1.0"]
+fips-test-hooks = ["caliptra-drivers/fips-test-hooks"]
```
jhand2, line 62: nit: should this be marked as an optional dependency?

nquarton: I think I'm missing what you mean here. This adds a feature and propagates it to the drivers crate. But there is not a separate, new dependency here that can be made optional. But please let me know if I'm misunderstanding something.

rusty1968 (approved):

runtime/Cargo.toml
jhand2: Oh my mistake, I misread this. That's what I get for reviewing on my phone :)

nquarton: I mean that's what I get for accidentally asking you to do this on vacation. Thanks!

PR #1596, opened 2024-06-28
[fix] Raise a fatal error in the ROM if max. limit of measurements is crossed

### LINKED ISSUE #1590: Error for Stash Measurement MB Command Causes Fatal Error
Within Caliptra, if the SOC sends too many Stash measurement commands, or if the Stash measurement command were to fail wihtin Caliptra for any reason, Caliptra should move to a Fatal Error state.

This will ensure that any SOC ROM/FW couldn't send a stash measurement that wasn't properly stored within Caliptra.

mhatrevi:
This change updates the logic when a stash measurement is requested beyond the max. supported. The new code raises a fatal error instead of continuing the boot sequence.

This change addresses issue# 1590.


jhand2:
Left a comment on #1590 about this. This feels more like a behavior change than a bugfix.

jlmahowa-amd (dismissed):

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -271,10 +271,14 @@ impl FirmwareProcessor {
...
-                                "[fwproc] Maximum supported number of measurements already received, ignoring."
+                                "[fwproc] Maximum supported number of measurements already received."
                             );
                             txn.complete(false)?;
-                            continue;
+
+                            // Raise a fatal error on hitting the max. limit.
+                            // This ensures that any SOC ROM/FW couldn't send a stash measurement
+                            // that wasn't properly stored within Caliptra.
+                            return Err(CaliptraError::FW_PROC_MAILBOX_STASH_MEASUREMENT_MAX_LIMIT);
```
jhand2, line 281: We should make it very clear in the spec that this will happen

nquarton:
Is it the case that this and the related, runtime behavior change are only going to affect 1.1 and beyond? Or are we backporting these to 1.0?

nquarton (dismissed):

mhatrevi:
> Is it the case that this and the related, runtime behavior change are only going to affect 1.1 and beyond? Or are we backporting these to 1.0?

Currently this is only for 1.1.

vsonims (approved):

PR #1601, opened 2024-07-03
Some nits to the smoke test.

zhalvorsen:
* Adds `extern C` to the header files in case of C++
* Removes an unused variable

### DIFF
--- a/hw-model/c-binding/build.rs
+++ b/hw-model/c-binding/build.rs
@@ -13,6 +13,8 @@ fn main() {
     let config = cbindgen::Config {
         header: Some(String::from_str("// Licensed under the Apache-2.0 license").unwrap()),
         language: cbindgen::Language::C,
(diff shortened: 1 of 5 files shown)


hw-model/c-binding/build.rs
```diff
@@ -13,6 +13,8 @@ fn main() {
     let config = cbindgen::Config {
         header: Some(String::from_str("// Licensed under the Apache-2.0 license").unwrap()),
         language: cbindgen::Language::C,
+        pragma_once: true,
```
jhand2, line 16: nit: We might want to use the `include_guard` option instead since `#pragma once` is technically compiler-specific (although it's widely supported in gcc and clang for a long time) and we don't really know which compiler various hardware integrators are going to use.

zhalvorsen: Cool, I didn't see that option initially

jhand2 (approved):

PR #1602, opened 2024-07-04
Move libcaliptra example functionality into a test.c file.

zhalvorsen:
This can make it easier to use the same test functionality in a different build system. Moves all test logic to `test.c` and only exposes a `run_tests` function that takes a parameter with the settings that will be used for the tests. The actually functionality of the test isn't changed at all.

Reading in the FW images was moved to `main.c` and an initialization function was added to `interface.c`. It gets initialized at the beginning of `run_tests`.

Some function parameters were changed to const because they weren't being changed within the function.

Also contains some `#include` cleanup.


jhand2 (approved):

jlmahowa-amd (approved):

nquarton (approved):

libcaliptra/examples/generic/main.c
```diff
@@ -36,613 +18,56 @@ static const uint32_t default_uds_seed[] = { 0x00010203, 0x04050607, 0x08090a0b,
...
-        printf("\t\tlibcaliptra test failures reported\n");
-    } else {
-        printf("\t\tAll libcaliptra tests passed\n");
-    }
-
-    return global_test_result;
+    test_info info = {
+        .rom = read_file_or_exit(ROM_PATH),
+        .image_bundle = read_file_or_exit(FW_PATH),
+        .vpk_hash = {0},
```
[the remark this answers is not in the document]
zhalvorsen: I agree this is the better approach. I replaced `vpk_hash` and `opk_hash` with a `caliptra_fuses` struct.

libcaliptra/examples/generic/test.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include <stdint.h>
+
+#include "caliptra_types.h"
+
+typedef struct test_info {
+  struct caliptra_buffer rom;
+  struct caliptra_buffer image_bundle;
+  struct caliptra_fuses fuses;
```
nquarton, line 11: Thanks for adding this. The uds_seed and field_entropy are actually in that caliptra_fuses struct though. So you probably won't want to duplicate them here.

zhalvorsen: Oh shoot, thanks! That would have been really redundant.

nquarton (approved):

jhand2 (dismissed):

zhalvorsen:
I rebased this onto origin/main and should be ready for submission again.

The update after the rebase fixes two things:

1. A memory leak
2. A `printf` for a string that was not null terminated and causing a stack overflow

jhand2 (approved):

jlmahowa-amd (approved):

Issue #1464, opened 2024-04-15
Support CI workflow with specified ROM, FMC, and RT versions


jhand2:
Revisions of ROM and FMC/RT are spread across different branches/tags. We do not currently have a way to produce a single release which will package binaries at specific revisions (currently our release builds everything from a single commit).

We should be able to build FMC/RT from head, but still bundle ROM for a specific release (e.g. 1.0.1).

This can be a Github workflow which takes as input parameters specific tags/commit hashes at which to build each firmware stage.

benjamindoron:
At the moment, we can't build FMC and RT separately: the builder does all of this, then creates the image bundle.
- We could enhance the builder to also support generating a bundle from an FMC and RT ELF

The current plan is to create a new workflow that takes ROM and FW git refs as parameters, and:
- Execute tests: Pass these in to the FPGA and sw-emulator tests. These will now build twice, and copy in the ROM ELF tests as overrides to a destination folder for cargo-nextest to consume.
  - Progress: Done, but many tests still build their own ROM. We need to refactor them to optionally consume a ROM from, for example, a path in an environment variable.
- Build release: Pass these into build_release.sh
  - Progress: Done. But we note that it's fragile, because some paths have moved over the course of releases. For example `hw-latest/` to `hw/latest/`. We might address this in rom-1.0 by symlinking the old path over to the new one. But this will not fix it for any possible git ref.

leongross:
Taking over this PR on [leongross/build_versioned_release_rom](https://github.com/chipsalliance/caliptra-sw/tree/leongross/build_versioned_release_rom)

jhand2:
Perfect, thanks! Let me know if you have any questions @leongross

clundin25:
Need to revisit this now that the 2.x release artifacts are also a part of this.

PR #1604, opened 2024-07-04
ci: Build ROM release from versioned git ref

benjamindoron:
Address https://github.com/chipsalliance/caliptra-sw/issues/1464

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -272,6 +272,27 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: 'true'
+          ref: 'rom-1.0.2'
+
+      - name: Build version-locked ROMs
+        run: |
+          rm -rf release
+
+          WORKSPACE_DIR="release/workspace"
+          mkdir -p $WORKSPACE_DIR
+
+          cargo run --manifest-path=builder/Cargo.toml --bin image -- --rom-no-log $WORKSPACE_DIR/caliptra-rom.bin
+          cp -a target/riscv32imc-unknown-none-elf/firmware/caliptra-rom $WORKSPACE_DIR/caliptra-rom.elf
+
+          cargo run --manifest-path=builder/Cargo.toml --bin image -- --rom-with-log $WORKSPACE_DIR/caliptra-rom-with-log.bin
+          cp -a target/riscv32imc-unknown-none-elf/firmware/caliptra-rom $WORKSPACE_DIR/caliptra-rom-with-log.elf
+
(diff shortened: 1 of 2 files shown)


zhalvorsen (approved):

ci-tools/release/build_release.sh
```diff
@@ -12,29 +12,16 @@ fi
...
 WORKSPACE_DIR="release/workspace"
 release_scripts_path=$(dirname "$0")
-# Generate Release Folder
-rm -rf release
-mkdir -p $WORKSPACE_DIR
 
-# Generate ROM and Image Bundle Binary
-cargo run --manifest-path=builder/Cargo.toml --bin image -- --rom-no-log $WORKSPACE_DIR/caliptra-rom.bin --fw $WORKSPACE_DIR/image-bundle.bin
-# Copy ROM ELF
-cp -a target/riscv32imc-unknown-none-elf/firmware/caliptra-rom $WORKSPACE_DIR/caliptra-rom.elf
```
jhand2, line 22: I'm not sure I understand why the things in this file need to be moved to the action definition.

I think the desire with these scripts is that someone can run them locally to build the expected release artifacts.

.github/workflows/nightly-release.yml
```diff
@@ -269,6 +269,27 @@ jobs:
       pull-requests: write
 
     steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+          ref: 'rom-1.0.2'
```
jhand2, line 275: I think we want to ability to build all the fw at a specific revision (e.g. via configuration when manually running the job). But I think we still want the automated nightly release to build from head. Otherwise we won't catch any regressions in ROM until we go to build a new release.

benjamindoron: Okay, I think that I got a little lost in all the workflow files. Shouldn't that be handled by build-test.yml? And if this workflow file doesn't generate the release, which one does? This is the only file to call build_release.sh.

ci-tools/release/build_release.sh
benjamindoron: I suppose that's valid. The intention was to build the ROM with the chosen tag first, before checking out HEAD and building some more. I wanted to avoid an actual `git checkout` command here, which would mean calling this has side effects.

Other solutions include passing git commit hashes to this script, and running `git checkout` here (if the CI is okay with it, and people won't mind the side effects that calling it will have to their working directory, it's an option). Alternatively, we could pass arguments to the script to determine which build stage to run: ROM, FMC/RT, or both. Then the CI/developer is responsible for making any changes in between.

.github/workflows/nightly-release.yml
leongross: What is the last status of this?

PR #1610, opened 2024-07-11
Adding ECC error checking during waits

nquarton:


nquarton:
I should add a test case here where we cause this error via the ECC verify service

mhatrevi (approved):

runtime/tests/runtime_integration_tests/test_ecdsa.rs
```diff
@@ -248,3 +248,32 @@ fn test_ecdsa_verify_bad_chksum() {
         resp,
     );
 }
+
+// HW errors are not supported on the SW emulator yet
+#[cfg(any(feature = "verilator", feature = "fpga_realtime"))]
```
jhand2, line 253: nit: if there isn't one already, could you open an issue for emulator support?

jhand2 (approved):

runtime/tests/runtime_integration_tests/test_ecdsa.rs
nquarton: No problem, done https://github.com/chipsalliance/caliptra-sw/issues/1615

PR #1619, opened 2024-07-25
Make the preamble non persistent

ArthurHeymans:


runtime/src/set_auth_manifest.rs
```diff
@@ -132,11 +131,12 @@ impl SetAuthManifestCmd {
...
-        let fw_preamble = &persistent_data.manifest1.preamble;
-
         // Verify the vendor ECC signature.
-        let vendor_fw_ecc_key =
-            &fw_preamble.vendor_pub_keys.ecc_pub_keys[fw_preamble.vendor_ecc_pub_key_idx as usize];
+        let vendor_fw_ecc_key = &fw_preamble
+            .vendor_pub_keys
+            .ecc_pub_keys
+            .get(fw_preamble.vendor_ecc_pub_key_idx as usize)
+            .ok_or(CaliptraError::RUNTIME_AUTH_MANIFEST_VENDOR_ECC_SIGNATURE_INVALID)?;
```
mhatrevi, line 139: This won't fail as fw_preamble is validated during ROM boot flow.

ArthurHeymans: Sure, but the runtime does not know this. A panic handler is being generated otherwise. Is there a way to make rust assume that this will be valid without checking?

mhatrevi (approved):

PR #1620, opened 2024-07-25
Libcaliptra fixes for checksum on uninitialized buffers and missing DPE structs

nquarton:


libcaliptra/inc/caliptra_types.h
```diff
@@ -365,6 +373,8 @@ struct caliptra_invoke_dpe_req {
     struct caliptra_req_header hdr;
     uint32_t data_size;
     union {
+        struct dpe_cmd_hdr                      cmd_hdr;
```
ekoohest, line 376: Why is it required in here?

libcaliptra/inc/caliptra_types.h
```diff
@@ -347,6 +351,10 @@ struct dpe_destroy_context_cmd {
     uint32_t flags;
```
ekoohest, line 351: I believe this should be removed.
Reference: https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.tcg#L584

libcaliptra/inc/caliptra_types.h
```diff
@@ -380,12 +390,14 @@ struct caliptra_invoke_dpe_resp {
...
+        struct dpe_resp_hdr                         resp_hdr;
         struct dpe_get_profile_response             get_profile_resp;
         struct dpe_initialize_context_response      initialize_context_resp;
         struct dpe_derive_context_response          derive_context_resp;
         struct dpe_certify_key_response             certify_key_resp;
         struct dpe_sign_response                    sign_resp;
         struct dpe_rotate_context_handle_response   rotate_context_handle_resp;
+        struct dpe_destroy_context_response         destroy_context_resp;
         struct dpe_get_certificate_chain_response   get_certificate_chain_resp;
         uint8_t                                     data[sizeof(struct dpe_certify_key_response)];
```
ekoohest, line 402: In fact, sizeof(struct dpe_get_certificate_chain_response) is the longest.
I suggest two separate macros for Cmd and Resp:

#define DPE_RESP_DATA_MAX_SIZE                      2176
#define DPE_REQ_DATA_MAX_SIZE                       512

ekoohest (changes requested):

libcaliptra/inc/caliptra_types.h
```diff
@@ -380,12 +390,14 @@ struct caliptra_invoke_dpe_resp {
     struct caliptra_resp_header cpl;
     uint32_t data_size;
     union {
+        struct dpe_resp_hdr                         resp_hdr;
```
ekoohest, line 393: Not sure why it is required.

libcaliptra/inc/caliptra_types.h
```diff
@@ -254,6 +254,10 @@ struct dpe_resp_hdr {
 #endif
```
ekoohest, line 255: General suggestion, the definitions that are coming from iROT profile spec being maintained in separate header file and being used in here.

ekoohest (changes requested):

libcaliptra/inc/caliptra_types.h
ekoohest: Second thought, I see usefulness of this field. Ignore this comment please.

libcaliptra/inc/caliptra_types.h
ekoohest: Second thought, I see usefulness of this field. Ignore this comment please.

libcaliptra/inc/caliptra_types.h
nquarton: That's a decent idea. In general, libcaliptra as a whole needs more automated generation. The generation of most of the structs shouldn't be manual. I'd expect there would be a separation like you are describing if/when that finally gets addressed.

libcaliptra/inc/caliptra_types.h
nquarton: I'll ping you on this one. Unfortunately I can't open that link. I do see this still in the latest version of the DPE IROT profile I can find (v0.10) though.

libcaliptra/inc/caliptra_types.h
nquarton: Okay cool, I figured it would be easier for working with a command generically

libcaliptra/inc/caliptra_types.h
nquarton: Maybe I am missing something. It looks to me like dpe_certify_key_response is still the longest as it contains the certificate of 2048 bytes as well as a few additional fields over dpe_get_certificate_chain_response.

But won't this work in C? The union should still be the size of the largest element which is the goal here.
image

ekoohest: Yes, you're right. Certify_key_response is the longest.

In the DPE client side, we use "DPE_RESP_DATA_MAX_SIZE" in a compile-time check to be ensure that size of union is matching to what it is supposed to be, regardless of which response is bigger.

ekoohest (approved):

jhand2 (dismissed):

libcaliptra/inc/caliptra_types.h
nquarton: After talking with Ehsan, this field was not part of the actual implementation and is being removed from the spec

jlmahowa-amd (approved):

nquarton:
Changing back to draft since I have a related issue I am looking into

nquarton:
This updates and adds a few DPE structs as well as fixes 2 bugs exposed by uninitialized command and response buffers:
1. The checksum field in the command request may be populated before calculating the checksum. This would get overwritten, but the new checksum value would be incorrect. This is resolved by simply setting the checksum field to 0 before calculating.
2. The response checksum was calculated over the whole response buffer instead of just the received data. Garbage in the untouched part of the buffer would change the checksum. The actual number of bytes read is recorded and propagated to the checksum calculation now.

jlmahowa-amd (approved):

libcaliptra/examples/generic/test.c
```diff
@@ -548,7 +550,7 @@ int rt_test_all_commands(const test_info* info)
 
 
     // FIPS_VERSION
-    struct caliptra_fips_version_resp version_resp = {};
```
jhand2, line 551: Is the reason for removing the initialization so that we can check that libcaliptra handles uninitialize structs properly?

Might be worth commenting that somewhere if so.

jhand2 (approved):

libcaliptra/examples/generic/test.c
nquarton: It is. There was a bug with the checksum check on the response where we were reading the max buffer size in some cases when the actual received data was smaller. That bug is fixed now, but I wanted coverage on this.

I can add a comment for it

jhand2 (approved):

jlmahowa-amd (approved):

PR #1622, opened 2024-07-30
Remove clearing SRAM by default.

zhalvorsen:
This puts clearing SRAM at startup behind the feature `clear-sram`.


rom/dev/src/start.S
```diff
@@ -104,6 +104,7 @@ _start:
     andi t1, t1, 0x3
     bne t1, x0, post_ecc_init
 
+#ifdef CARGO_FEATURE_CLEAR_SRAM
```
korran, line 107: Does this actually work? I can't find any documentation about global_asm! ifdefs...

korran: nm, looks like somebody is running this through the C preprocessor:

https://github.com/chipsalliance/caliptra-sw/blob/51ff0a89f169bbf8e06acb49b31db555e99fefb6/rom/dev/build.rs#L70

timothytrippel (approved):

PR #1623, opened 2024-07-30
sw-emulator: Add dilithium peripheral

ArthurHeymans:
This adds support for the dilithium peripheral in the sw-emulator based on preliminary support available.
Things currently unknown like how certain functionality works exactly and the MMIO offset are marked as TODO so that those can be updated once more details are available.


Cargo.toml
```diff
@@ -139,6 +139,7 @@ cms = "0.2.2"
 convert_case = "0.6.0"
 dpe = { path = "dpe/dpe", default-features = false, features = ["dpe_profile_p384_sha384"] }
 crypto = { path = "dpe/crypto", default-features = false }
+crystals-dilithium = "1.0.0"
```
korran, line 142: This crate is GPL-3.0; is that a problem for us?

https://crates.io/crates/crystals-dilithium

ArthurHeymans: Hmm I suppose that's a problem.
Maybe https://crates.io/crates/fips204 can be used, but the sizes of the keys are bit different as dilithium is not map exactly to fips204 even though it's based on it. I guess it needs to match what the HW is going to end up being.

Alternatively https://crates.io/crates/pqcrypto can be used which is a wrapper around a C library.

jhand2: The hardware should be implementing FIPS 204 ML-DSA-87. so i think the FIPS 204 crate is a good option. and the license is compatible.

sw-emulator/lib/periph/src/root_bus.rs
```diff
@@ -265,6 +266,9 @@ pub struct CaliptraRootBus {
     #[peripheral(offset = 0x1002_8000, mask = 0x0000_7fff)]
     pub sha256: HashSha256,
 
+    #[peripheral(offset = 0x1003_0000, mask = 0x0000_7fff)] // TODO update when known
+    pub dilithium: Dilithium,
```
jhand2, line 270: Minor nit: as you pointed out there are some small changes from dilithium to ML-DSA. I think we should consider naming this peripheral something like mldsa87. But I'm not sure what it's named in RTL.

cc @mhatrevi @mojtaba-bisheh

Cargo.toml
ArthurHeymans: Done

sw-emulator/lib/periph/src/helpers.rs
```diff
@@ -23,7 +23,9 @@ pub trait U32Array: AsRef<[u32]> {
 }
 pub trait U32ArrayBytes: AsRef<[u8]> {
     const WORD_LEN: usize;
-    type WordArray: AsMut<[u32]> + Default;
+    type WordArray: AsMut<[u32]>;
+    // Required because Default isn't implemented for [u32; 64 and higher]
+    fn default_result() -> Self::WordArray;
```
jhand2, line 28: nit: why not have `impl Default for [u8; $word_len * 4]` in the macro?

Not blocking for this PR. If this ends up the only comment I'll merge anyway.

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+    /// Control Register Fields
+    Control [
+        CTRL OFFSET(0) NUMBITS(2) [
+            IDLE = 0b00,
+            GEN_KEY = 0b01,
+            SIGN = 0b10,
+            VERIFY = 0b11,
+        ],
+        ZEROIZE OFFSET(2) NUMBITS(1) [],
+//        PCR_SIGN OFFSET(3) NUMBITS(1) [] TODO remove right??
```
jhand2, line 57: Will PCR sign not be supported for ML-DSA? I don't really care either way, just curious.

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+    #[register(offset = 0x0000_0004)]
+    name1: ReadOnlyRegister<u32>,
+
+    /// Version 0 register
+    #[register(offset = 0x0000_0008)]
+    version0: ReadOnlyRegister<u32>,
+
+    /// Version 1 register
+    #[register(offset = 0x0000_000C)]
+    version1: ReadOnlyRegister<u32>,
```
jhand2, line 87: nit: For these multi-word registers, is it easier to just model it as an array of registers?

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+    fn write_access_fault(
+        &self,
+        _size: RvSize,
+        _index: usize,
+        _val: RvData,
+    ) -> Result<(), BusError> {
+        Err(BusError::StoreAccessFault)
+    }
+
+    // Clear registers
```
jhand2, line 188: nit: Add todo to implement this?

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+                    .modify(Status::READY::CLEAR + Status::VALID::CLEAR);
+
+                self.op_complete_action = Some(self.timer.schedule_poll_in(ML_DSA87_OP_TICKS));
+            }
+            _ => {}
+        }
+
+        if self.control.reg.is_set(Control::ZEROIZE) {
+            self.zeroize();
+        }
```
jhand2, line 226: Should this be part of the `match`?

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+
+        self.pk = words_from_bytes_le(&pk.into_bytes());
+        self.sk_out = words_from_bytes_le(&sk.into_bytes());
+    }
+
+    fn sign(&mut self) {
+        let seed_bytes = &bytes_from_words_le(&self.seed);
+        let mut rng = StdRng::from_seed(*seed_bytes);
+
+        let secret_key_bytes = &bytes_from_words_le(&self.sk_in);
```
jhand2, line 244: I thought the peripheral was supposed to regen the key from a seed on every signing operation? I think the plan is that KV slots will only hold the seed and HW will generate the private key from the seed as-needed.

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+        let secret_key = PrivateKey::try_from_bytes(*secret_key_bytes).unwrap();
+
+        let message = &bytes_from_words_le(&self.message);
+
+        // The Ml_Dsa87 5 signature is 4595 len but the reg is one byte longer
```
jhand2, line 249: nit
```suggestion
        // The Ml_Dsa87 signature is 4595 len but the reg is one byte longer
```

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+        let message = &bytes_from_words_le(&self.message);
+
+        let public_key = {
+            let key_bytes = &bytes_from_words_le(&self.pk);
+            PublicKey::try_from_bytes(*key_bytes).unwrap()
+        };
+
+        let signature = &bytes_from_words_le(&self.signature);
+
+        let result = public_key.verify(message, &signature[..SIG_LEN].try_into().unwrap());
```
jhand2, line 270: Shouldn't signature be a fixed size? Not sure why we need to take a subslice

swenson: I think this is because `self.signature` is `[u32; 1157]` which expands to `[u8; 4628]`, but the `SIG_LEN` is 4627.

sw-emulator/lib/periph/src/ml_dsa87.rs
```diff
@@ -0,0 +1,1254 @@
...
+        // TODO: Reset registers
+    }
+
+    /// Called by Bus::update_reset() to indicate an update reset
+    fn update_reset(&mut self) {
+        // TODO: Reset registers
+    }
+}
+
+#[cfg(test)]
```
swenson, line 308: Due to the length of the tests, I'd suggest one of:

* moving them to a separate file,
* generating the keys from a seed to reduce the size, or
* moving the keys to a separate file and using `include!`.

sw-emulator/lib/periph/src/ml_dsa87.rs
jhand2: Ah, makes sense. Thanks :)

sw-emulator/lib/periph/src/ml_dsa87.rs
jhand2: I like the idea of putting kat data in a separate file and using `include!`.

sw-emulator/lib/periph/src/ml_dsa87.rs
ArthurHeymans: I think what you say is indeed the actual use case. I think the dilithium reflects the ecc384 engine which can also take a key as input, which I was told was for early validation/ testing.

sw-emulator/lib/periph/src/ml_dsa87.rs
ArthurHeymans: DOne

sw-emulator/lib/periph/src/ml_dsa87.rs
ArthurHeymans: Done

sw-emulator/lib/periph/src/ml_dsa87.rs
ArthurHeymans: Not possible as the match is on a only a part of the

sw-emulator/lib/periph/src/ml_dsa87.rs
mhatrevi: Yes, it will be supported.

mhatrevi (approved):

swenson (approved):
LGTM

PR #1625, opened 2024-07-30
Moving ROM 1.x to a new 1.0 folder

mhatrevi:


korran:
Do we really need to keep the 1.x ROM source around in the tree? We've already got binaries for regression-testing purposes:

https://github.com/chipsalliance/caliptra-sw/blob/main/hw/1.0/caliptra-rom-1.0.1-9342687.bin

jhand2 (commented):
Should this folder be named 1.x? I want to avoid confusion since the current rom in this directory is 1.1.

korran:
> Do we really need to keep the 1.x ROM source around in the main tree? We've already got binaries for regression-testing purposes:
> 
> https://github.com/chipsalliance/caliptra-sw/blob/main/hw/1.0/caliptra-rom-1.0.1-9342687.bin

I guess if we're planning to simultaneously work on a 1.2 ROM along with a very-different 2.0 ROM, maybe it makes sense to have a directory for each. But if 1.2 is runtime-fw-only, I think we can just delete the 1.x ROM source (or evolve it into 2.x).

mhatrevi:
This change is no longer needed.

PR #1629, opened 2024-08-05
rom: Implement fake fips verification using const generics

ArthurHeymans:
The fake rom codepath is largely similar to the regular codepath except certain parts get skipped. Instead of reimplementing most of the verification trait functions use a generic parameter in common code to determine whether the fake rom codepath needs to be taken.

Use the "const-gen" crate to generate a constant literal that is used as the const generics input (using cfg!() won't work).


mhatrevi:
Closing this for now. We'll revisit this with additional design thinking.

PR #1637, opened 2024-08-14
libcaliptra: add SHA accelerator API

### LINKED ISSUE #1361: Libcaliptra support needed for SHA accelerator engine
Libcaliptra should implement support for the SHA accelerator engine. This API should also support the caller streaming data (instead of providing the full buffer).

Note users can still access and use the SHA engine via the generic reg read and write functions within libcaliptra.

MDr164:
Resolves #1361 by providing an abstraction above the `caliptra_write_u32` and `caliptra_read_u32` function calls according to the info at https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.sha512_acc_csr

### DIFF
--- a/libcaliptra/src/caliptra_api.c
+++ b/libcaliptra/src/caliptra_api.c
@@ -1455,12 +1455,6 @@ int caliptra_start_sha_stream(int mode, uint8_t* in_data, uint32_t data_len) {
         return REG_ACCESS_ERROR;
     }
 
-    // Set SHA_DLEN to 0 initially
-    error = caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_DLEN, 0);
-    if (error) {
-        return REG_ACCESS_ERROR;
-    }
-
     return caliptra_update_sha_stream(in_data, data_len);
 }
 


MDr164:
TODO: Needs some test cases

nquarton (commented):
I think there are some issues when it comes to obtaining and releasing the lock. It think there should also be a test that performs these operations with a known answer to ensure this is working.

libcaliptra/src/caliptra_api.c
```diff
@@ -1264,3 +1264,130 @@ bool caliptra_is_idevid_csr_ready() {
...
+ * @return 0 on success, or an error code on failure.
+ */
+int caliptra_compute_sha(int mode, int endian, uint32_t* data, uint32_t data_len, uint32_t* hash, uint32_t mbox_start_addr) {
+    if (!data || !hash) {
+        return INVALID_PARAMS;
+    }
+
+    if (mode == CALIPTRA_SHA_ACCELERATOR_MODE_MBOX_384 || mode == CALIPTRA_SHA_ACCELERATOR_MODE_MBOX_512) {
+        // Writing 1 will clear the lock
+        caliptra_write_u32(CALIPTRA_SHA_ACCELERATOR_LOCK_ADDR, 0x1);
```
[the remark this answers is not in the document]
MDr164: I'm somewhat confused about the locking behavior. Writing 1 will clear the lock and "reading 0" will set the lock but how does "reading 0" even work? That documentation looks odd to me now that I went over it again and it also seems like the emulator is not really doing what is written in the documentation.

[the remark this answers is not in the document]
jhand2: So my understanding is that when you read the lock register, if the read returns 0, it means the lock was available and now you have the lock. Every subsequent read should return 1 (that the lock is held).

Then writing 1 will clear the lock, so that the next read will return 0.

> That documentation looks odd to me now that I went over it again and it also seems like the emulator is not really doing what is written in the documentation.

I would need to double check, but it's possible there's a mismatch. What unexpected behavior are you seeing?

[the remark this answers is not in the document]
MDr164: iirc the lock was also set when writing 0 instead of just reading from it but your explanation that the read itself should cause the lock to be held and whether that worked is explained by the response.

jhand2 (approved):

libcaliptra/inc/caliptra_api.h
```diff
@@ -225,6 +225,18 @@ void caliptra_req_idev_csr_start();
 // Clear IDEV CSR request.
 void caliptra_req_idev_csr_complete();
 
+// Computes the SHA hash using the specified mode and endianess.
+int caliptra_compute_sha(int mode, int endian, uint32_t* data, uint32_t data_len, uint32_t* hash, uint32_t mbox_start_addr);
```
nquarton, line 229: So this is a hash over data in the mailbox? Can the function name or comment clarify that?

libcaliptra/examples/generic/test.c
```diff
@@ -671,6 +670,63 @@ int rt_test_all_commands(const test_info* info)
         printf("Certify Key Extended: OK\n");
     }
 
+    // SHA Engine Tests
```
nquarton, line 673: Can there be a test for caliptra_compute_sha as well?

libcaliptra/src/caliptra_sha.h
```diff
@@ -0,0 +1,14 @@
+// Licensed under the Apache-2.0 license
```
nquarton, line 1: These should come from caliptra_top_reg.h. See the other examples in caliptra_fuses.h and caliptra_mbox.h. That file comes from the HW repo submodule (hw/latest/rtl/src/soc_ifc/rtl/caliptra_top_reg.h)

libcaliptra/examples/generic/test.c
MDr164: Sure thing, it would test the same functionality but just in a different input way but shouldn't be too hard to add another test for that as well.

libcaliptra/inc/caliptra_api.h
MDr164: I can change that yes, if I remember correctly a user would need to specify the mbox regs and then it computes the hash over it. I will also update the name of the parameters a bit then as it could help users of the function.

libcaliptra/src/caliptra_sha.h
MDr164: I must have missed those declarations, will update :+1:

MDr164: I mirrored the defines, let me know if that is a decent approach or if I should drop the header file here alltogether

nquarton: To be consistent with the rest of the code and eliminate this maintenance, you can drop this altogether. You may want some access functions like you see in caliptra_fuses.h and caliptra_,mbox.h. But the offsets themselves should be directly used from the generated top_reg file.

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1357,148 @@ bool caliptra_is_idevid_csr_ready() {
...
+ * @brief Computes the SHA hash of data inside the mailbox using the specified mode, endianess, offset and length.
+ *
+ * @param mode The SHA mode to use (e.g., CALIPTRA_SHA_ACCELERATOR_MODE_MBOX_384).
+ * @param endian The endianess to use (e.g., CALIPTRA_SHA_ACCELERATOR_ENDIANESS_BIG).
+ * @param mbox_start_addr The mailbox start address.
+ * @param data_len Length of the data to hash.
+ * @param hash Pointer to the buffer to store the resulting hash.
+ * @return 0 on success, or an error code on failure.
+ */
+int caliptra_compute_mbox_sha(int mode, int endian, uint32_t mbox_start_addr, uint32_t data_len, uint32_t* hash) {
```
sunguk-bin-fad, line 1371: I'm wondering if this works. As per IntergrationSpec(https://github.com/chipsalliance/caliptra-rtl/blob/v1.1/docs/CaliptraIntegrationSpecification.md#overview-1), it doesn't seem to support mailbox support. So I'm just asking.

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1357,148 @@ bool caliptra_is_idevid_csr_ready() {
...
+    // By reading the lock register we also start holding the lock if it was not already held
+    caliptra_read_u32(CALIPTRA_SHA_ACCELERATOR_LOCK_ADDR, &lock_status);
+    if (lock_status & 0x1) {
+        return MBX_BUSY;
+    }
+
+    // Zeroize engine registers to start fresh
+    caliptra_write_u32(CALIPTRA_SHA_ACCELERATOR_CONTROL_ADDR, 0x1);
+    // Set mode and endianess accordingly
+    uint32_t control_value = (mode & 0xFFFF) | ((endian & 0xFF) << 16);
```
sunguk-bin-fad, line 1438: I was not able to find a datasheet, but I found that SHA512_ACC_CSR_MODE_ENDIAN_TOGGLE_LOW defined as 2, not 16. (https://github.com/chipsalliance/caliptra-rtl/blob/7c953512251707d6f77d7d9a2f6be186ed013d50/src/soc_ifc/rtl/caliptra_top_reg.h#L68) And I wonder if this endian field impacts on the streaming mode. On my test, it doesn't seem to affect to hash result.

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1357,148 @@ bool caliptra_is_idevid_csr_ready() {
...
+    caliptra_write_u32(CALIPTRA_SHA_ACCELERATOR_CONTROL_ADDR, control_value);
```
sunguk-bin-fad, line 1439: CALIPTRA_SHA_ACCELERATOR_ **MODE** _ADDR, not CALIPTRA_SHA_ACCELERATOR_ **CONTROL** _ADDR

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1357,148 @@ bool caliptra_is_idevid_csr_ready() {
...
+
+    // Write initial data to the SHA accelerator
+    for (uint32_t i = 0; i < data_len; i++) {
```
sunguk-bin-fad, line 1442: I think that data_len is byte length, and we can put 4 bytes data into DATAIN register at once.

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1357,148 @@ bool caliptra_is_idevid_csr_ready() {
...
+ *
+ *        This function will write the data to the SHA accelerator and requires that caliptra_start_sha_stream()
+ *        has been called previously.
+ *
+ * @param in_data Pointer to the data to hash.
+ * @param data_len Length of the data to hash.
+ * @return 0 on success, or an error code on failure.
+ */
+int caliptra_update_sha_stream(uint32_t* in_data, uint32_t data_len) {
+    if (!data || data_len < 1) {
```
sunguk-bin-fad, line 1460: in_data, not data

sunguk-bin-fad (commented):
While searching for the history of SHA control, I came across this PR, which was very helpful. While testing, I found a few issues and left some comments. Great job! Thanks.

libcaliptra/src/caliptra_api.c
embediver: Fixed.
Also the Mask used for the `mode` didn't seemed right to me, now values are calculated using the defines from `caliptra_top_reg.h`.

Apart from that I also noticed that, according to [https://chipsalliance.github.io/caliptra-rtl/v1_1/external-regs/?p=caliptra_top_reg.sha512_acc_csr.MODE#ENDIAN_TOGGLE.desc](https://chipsalliance.github.io/caliptra-rtl/v1_1/external-regs/?p=caliptra_top_reg.sha512_acc_csr.MODE#ENDIAN_TOGGLE.desc), endianness is only relevant for mailbox mode and could thus be removed from the stream mode function, right?

sunguk-bin-fad: This line should be... uint32_t control_value = (mode & 0x3); ?

libcaliptra/src/caliptra_api.c
```diff
@@ -1357,3 +1356,112 @@ bool caliptra_is_idevid_csr_ready() {
...
+    caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_MODE, control_value);
+
+    // Write data length to SHA_DLEN
+    caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_DLEN, data_len);
+
+    // Write initial data to the SHA accelerator
+    for (uint32_t i = 0; i < data_len/4; i++) {
+        caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_DATAIN, in_data[i]);
+    }
+    // TODO: handle data_len%4 != 0
```
embediver: Currently the user has to provide a uint32_t* buffer.

uint8_t* as a parameter to support hashing data which isn't 32bit or a multiple of 4 byte aligned would be better.

I suspect that that an implementation with a uint8_t* buffer would need to support both little- and big-endian systems right?
Or do we want the user to format the data into a uint32 buffer?

embediver: I changed the implementation to u8 buffers now.

For little-endian the the buffer is just casted into u32 values, for big-endian the bytes are twizzled around.

To check the endianness of the system, the GCC  `__BYTE_ORDER__` macro is used.

The test still fails and I haven't been able to find the issue until now.

embediver:
I encountered a problem with the emulator:

It seems that on certain data length the emulator crashes.
In the test I used 127 Bytes of data which results in a panic when calculating a SHA384.
The behavior can be reproduced by changing the amount of data to 127 bytes in the test @sunguk-bin-fad wrote in his implementation.

embediver:
Rebased the branch and also fixed an oversight in the doc comments and removed / commented out some unsupported modes in the enums.

Otherwise this PR should be good to go now.

libcaliptra/examples/generic/test.c
```diff
@@ -847,6 +846,66 @@ int rt_test_all_commands(const test_info* info)
...
+    status = caliptra_finish_sha_stream(stream_hash);
+    if (status) {
+        printf("Finish SHA Stream failed: 0x%x\n", status);
+        dump_caliptra_error_codes();
+        failure = 1;
+    } else {
+        printf("Finish SHA Stream: OK\n");
+
+        // Verify the hash against the expected value
+        if (memcmp(stream_hash, expected_stream_hash, 12) == 0) {
```
zhalvorsen: This is only comparing the first 12 bytes of the hash
```suggestion
        if (memcmp(stream_hash, expected_stream_hash, sizeof(stream_hash)) == 0) {
```

libcaliptra/examples/generic/test.c
```diff
@@ -1190,6 +1249,217 @@ int upload_fw_piecewise(const test_info* info)
...
+    status = caliptra_finish_sha_stream(digest);
+    if (status) {
+        printf("SHA final failed: 0x%x\n", status);
+        dump_caliptra_error_codes();
+        failure = 1;
+    } else {
+        printf("SHA final: OK\n");
+
+        // Verify the hash against the expected value
+        if (memcmp(digest, expected_sha384_digest, 12) == 0) {
```
zhalvorsen: Same here:

```suggestion
        if (memcmp(digest, expected_sha384_digest, sizeof(expected_sha384_digest)) == 0) {
```

libcaliptra/examples/generic/test.c
```diff
@@ -1190,6 +1249,217 @@ int upload_fw_piecewise(const test_info* info)
...
+    status = caliptra_finish_sha_stream(digest);
+    if (status) {
+        printf("SHA final failed: 0x%x\n", status);
+        dump_caliptra_error_codes();
+        failure = 1;
+    } else {
+        printf("SHA final: OK\n");
+
+        // Verify the hash against the expected value
+        if (memcmp(digest, msg2_expected_sha384_digest, 12) == 0) {
```
zhalvorsen:
```suggestion
        if (memcmp(digest, msg2_expected_sha384_digest, sizeof(msg2_expected_sha384_digest)) == 0) {
```

libcaliptra/examples/generic/test.c
```diff
@@ -1190,6 +1249,217 @@ int upload_fw_piecewise(const test_info* info)
...
+    status = caliptra_finish_sha_stream(digest);
+    if (status) {
+        printf("SHA final failed: 0x%x\n", status);
+        dump_caliptra_error_codes();
+        failure = 1;
+    } else {
+        printf("SHA final: OK\n");
+
+        // Verify the hash against the expected value
+        if (memcmp(digest, expected_sha512_digest, 16) == 0) {
```
zhalvorsen:
```suggestion
        if (memcmp(digest, expected_sha512_digest, sizeof(expected_sha512_digest)) == 0) {
```

libcaliptra/src/caliptra_api.c
```diff
@@ -1415,3 +1414,170 @@ bool caliptra_is_idevid_csr_ready() {
...
+    error = caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_MODE, control_value);
+    if (error) {
+        return REG_ACCESS_ERROR;
+    }
+
+    // Set SHA_DLEN to 0 initially
+    error = caliptra_write_u32(CALIPTRA_TOP_REG_SHA512_ACC_CSR_DLEN, 0);
+    if (error) {
+        return REG_ACCESS_ERROR;
+    }
```
zhalvorsen: Does this need to be done if zeroize was called on line 1447?

libcaliptra/examples/generic/test.c
embediver: 12 words (`[u32; 12]`) is correct for sha384.
~~I will apply the proposed change since that improves readability.~~
For the other cases, a buffer of length 16 is used to also store sha512 hashes, but sizes are correct.
~~Will apply the suggested changes too.~~

Should the output be provided as uint8 buffer?

Edit:
It would have to be `sizeof(stream_hash) / sizeof(stream_hash[0])`.
For the other tests also a single buffer is used that can hold a sha512 for both sha512 and sha384....
Do you want this to be changed to a more readable solution?

zhalvorsen: The problem is the size parameter in `memcmp` is supposed to be in bytes instead of words. So you are only comparing 12 or 16 bytes of the digest

embediver: My bad, doing to little C these days...

Will update.

zhalvorsen: No worries, its hard to keep that stuff straight in C. Thanks for updating!

PR #1643, opened 2024-08-17
ci-tools: Add rtool subcommand to download artifact

korran:
This is useful to keep the runner images up-to-date.


ci-tools/github-runner/cmd/rtool/main.go
```diff
@@ -45,6 +45,24 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
+	} else if os.Args[1] == "download_artifact" {
```
swenson, line 48: I think you'll want to update the `usage()` function, and ideally, with this many arguments, I'd prefer to use something like `flag`.

swenson (dismissed):
LGTM; a minor nit and cleanup suggestion (but I understand that converting to `flag` is probably out of scope for htis PR).

ci-tools/github-runner/cmd/rtool/main.go
korran: Updated usage().

ci-tools/github-runner/artifact.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		log.Fatalf("Response failed with status code: %d and body %s\n", resp.StatusCode, body)
+	}
+	_, err = io.Copy(os.Stdout, resp.Body)
+	if err != nil {
+		return errors.New("unable to copy response body")
+	}
```
leongross, line 65: nit
```suggestion
	if err = io.Copy(os.Stdout, resp.Body); err != nil {
		return errors.New("unable to copy response body")
	}
```

ci-tools/github-runner/cmd/rtool/main.go
```diff
@@ -45,6 +93,24 @@ func main() {
...
+			log.Fatal(err)
+		}
+		client, err := hello.GithubClient(appId, installationId)
+		if err != nil {
+			log.Fatal(err)
+		}
+		err = hello.DownloadArtifact(ctx, client, os.Args[4], os.Args[5])
+		if err != nil {
+			log.Fatal(err)
+		}
```
leongross, line 112: nit
```suggestion
		if err = hello.DownloadArtifact(ctx, client, os.Args[4], os.Args[5]); err != nil {
			log.Fatal(err)
		}
```

leongross:
Is this still of interest? @korran @jhand2

leongross (approved):

PR #1644, opened 2024-08-19
[feat] Manifest Based Image Authorization - Manifest creation tool

mhatrevi:
This change contains the manifest creation tool for the Manifest Based Image Authorization feature.

Design document: https://docs.google.com/document/d/1BofLEAzeszgAGVcaEjTxM7tDxwIq9KhVXuqGUlrg1sc/edit#heading=h.z8ehugjyw5jp


auth-manifest/app/src/config.rs
```diff
@@ -0,0 +1,137 @@
...
+    if let Some(pem_file) = &config.lms_priv_key {
+        let priv_key_path = path.join(pem_file);
+        priv_keys.lms_priv_key = lms_priv_key_from_pem(&priv_key_path)?;
+        gen_config.priv_keys = Some(priv_keys);
+    }
+
+    Ok(gen_config)
+}
+
+pub(crate) fn vendor_config(
```
jhand2, line 100: nit: maybe rename these to something like `vendor_config_from_toml`. The naming of these functions is a bit confusing.

Same comment for all the `pub` functions in this file.

auth-manifest/app/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    result.unwrap();
+}
+
+pub(crate) fn run_auth_man_cmd(args: &ArgMatches) -> anyhow::Result<()> {
+    let version: &u32 = args
+        .get_one::<u32>("version")
+        .with_context(|| "version arg not specified")?;
+
+    let flags: &u32 = args
```
jhand2, line 83: nit: Could use `bitflags` crate throughout these libraries to make the code a little more readable.

auth-manifest/app/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    let config_path: &PathBuf = args
+        .get_one::<PathBuf>("config")
+        .with_context(|| "config arg not specified")?;
+
+    let key_dir: &PathBuf = args
+        .get_one::<PathBuf>("key-dir")
+        .with_context(|| "key-dir arg not specified")?;
+
+    if !key_dir.exists() {
```
jhand2, line 95: Should we do this same validation on the config path?

auth-manifest/gen/src/generator.rs
```diff
@@ -0,0 +1,166 @@
...
+            )?;
+            auth_manifest.preamble.vendor_pub_keys_signatures.ecc_sig = sig;
+
+            let lms_sig = self.crypto.lms_sign(&digest, &priv_keys.lms_priv_key)?;
+            auth_manifest.preamble.vendor_pub_keys_signatures.lms_sig = lms_sig;
+        }
+
+        // Sign the owner manifest keys.
+        if let Some(owner_fw_config) = &config.owner_fw_key_info {
+            if let Some(owner_man_config) = &config.owner_man_key_info {
```
jhand2, line 91: nit: I think you can do something like this to avoid the nested ifs
```suggestion
        if let (Some(owner_fw_config), Some(owner_man_config)) = (&config.owner_fw_key_info, &config.owner_man_key_info) {
```

auth-manifest/gen/src/generator.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        if let Some(priv_keys) = config.vendor_fw_key_info.priv_keys {
+            let sig = self.crypto.ecdsa384_sign(
+                &digest,
+                &priv_keys.ecc_priv_key,
+                &config.vendor_fw_key_info.pub_keys.ecc_pub_key,
+            )?;
+            auth_manifest.preamble.vendor_pub_keys_signatures.ecc_sig = sig;
+
+            let lms_sig = self.crypto.lms_sign(&digest, &priv_keys.lms_priv_key)?;
```
jhand2, line 85: Currently this tool requires the manifest to be signed with both ecdsa and lms keys. Do we want to support manifests that are only signed with ecdsa (since we support this config for Caliptra FW).

auth-manifest/gen/src/generator.rs
```diff
@@ -0,0 +1,166 @@
...
+                auth_manifest
+                    .preamble
+                    .vendor_image_metdata_signatures
+                    .lms_sig = lms_sig;
+            }
+        }
+
+        // Sign with the owner manifest public keys.
+        if let Some(owner_man_config) = &config.owner_man_key_info {
+            if let Some(owner_man_priv_keys) = &owner_man_config.priv_keys {
```
jhand2, line 143: Same comment to avoid nested ifs

auth-manifest/app/src/config.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    // Get the Public Keys.
+    let pub_key_path = path.join(&config.ecc_pub_key);
+    gen_config.pub_keys.ecc_pub_key = Crypto::ecc_pub_key_from_pem(&pub_key_path)?;
+
+    let pub_key_path = path.join(&config.lms_pub_key);
+    gen_config.pub_keys.lms_pub_key = lms_pub_key_from_pem(&pub_key_path)?;
+
+    // Get the Private Keys.
+    let mut priv_keys = AuthManifestPrivKeys::default();
```
swenson, line 84: Can we modify `gen_config.priv_keys` directly instead of using this variable and copying? Or avoiding the `mut` altogether and making the `AuthManifestGeneratorKeyConfig` at the end?

auth-manifest/README.md
```diff
@@ -0,0 +1,33 @@
...
+| Owner ECC Public Key | 96 | ECC P-384 public key used to verify the IMC Signature. <br> **X-Coordinate:** Public Key X-Coordinate (48 bytes) <br> **Y-Coordinate:** Public Key Y-Coordinate (48 bytes) |
+| Owner LMS Public Key | 48 | LMS public key used to verify the IMC Signature. <br> **tree_type:** LMS Algorithm Type (4 bytes) <br> **otstype:** LMS Ots Algorithm Type (4 bytes) <br> **id:**  (16 bytes) <br> **digest:**  (24 bytes) |
+| Owner ECC Signature | 96 | Owner ECDSA P-384 signature of the Version, Flags, Owner ECC and LMS public keys, hashed using SHA2-384. <br> **R-Coordinate:** Random Point (48 bytes) <br> **S-Coordinate:** Proof (48 bytes) |
+| Owner LMS Signature | 1620 | Owner LMS signature of the Version, Flags, Owner ECC and LMS public keys, hashed using SHA2-384. <br> **q:** Leaf of the Merkle tree where the OTS public key appears (4 bytes) <br> **ots:** Lmots Signature (1252 bytes) <br> **tree_type:** Lms Algorithm Type (4 bytes) <br> **tree_path:** Path through the tree from the leaf associated with the LM-OTS signature to the root. (360 bytes) |
```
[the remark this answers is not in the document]
mhatrevi: Currently, this feature is pre-2.0, so there is no support for ML-DSA. For post 2.0, it makes sense to update it to support ML-DSA signatures.

auth-manifest/gen/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+Abstract:
+
+    Caliptra Image generator
+
+--*/
+mod generator;
+
+pub use generator::AuthManifestGenerator;
+
+//use auth_man_generator::{AuthManPrivKeys, AuthManPubKeys};
```
swenson, line 18:
```suggestion
```

auth-manifest/types/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+use memoffset::span_of;
+use zerocopy::{AsBytes, FromBytes};
+use zeroize::Zeroize;
+
+pub const AUTH_MANIFEST_MARKER: u32 = 0x4154_4D4E;
+pub const AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT: usize = 16;
+pub const AUTH_MANIFEST_VENDOR_SIGNATURE_REQURIED_FLAG: u32 = 0x1;
+
+#[repr(C)]
+#[derive(AsBytes, FromBytes, Default, Debug, Clone, Copy, Zeroize)]
```
swenson, line 30: Things other than private keys don't really need to be zeroized, but I suppose it doesn't hurt?

auth-manifest/types/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub ecc_pub_key: ImageEccPubKey,
+    #[zeroize(skip)]
+    pub lms_pub_key: ImageLmsPublicKey,
+}
+
+#[repr(C)]
+#[derive(AsBytes, FromBytes, Default, Debug, Clone, Copy, Zeroize)]
+pub struct AuthManifestPrivKeys {
+    pub ecc_priv_key: ImageEccPrivKey,
+    #[zeroize(skip)]
```
swenson, line 42: Why are we skipping zeroizing the LMS private key?

auth-manifest/README.md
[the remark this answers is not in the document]
jhand2: Makes sense. We can just rev the version for 2.0.

auth-manifest/app/src/auth-man.toml
```diff
@@ -0,0 +1,37 @@
+# Licensed under the Apache-2.0 license
+
+[vendor_fw_key_config]
```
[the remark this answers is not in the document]
mhatrevi: This is a configuration file for specifying the key file names and image hashes for creating the authorization manifest file. The PEM are being reused from the caliptra-sw/image/fake-keys/src/lib.rs.

auth-manifest/app/src/config.rs
mhatrevi: Done

auth-manifest/app/src/auth-man.toml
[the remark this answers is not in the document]
jhand2: Oh, I see, they're generated by the build. Thanks.

auth-manifest/gen/src/generator.rs
mhatrevi: Updated.

auth-manifest/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+ *Note: All fields are little endian unless specified*
+
+| Field | Size (bytes) | Description|
+|-------|--------|------------|
+| Manifest Marker | 4 | Magic Number marking the start of the manifest. The value must be 0x41544D4EE (‘ATMN’ in ASCII)|
+| Manifest Size | 4 | Size of the full manifest structure |
+| Version | 4 | Manifest version |
+| Flags | 4 | Feature flags. <br> **Bit0:** - Vendor Signature Required. If set, verify the vendor public key endorsement(s) <br>**Bit1-Bit31:** Reserved |
+| Manufacturer ECC Public Key | 96 | ECC P-384 public key used to verify the IMC Signature. <br> **X-Coordinate:** Public Key X-Coordinate (48 bytes) <br> **Y-Coordinate:** Public Key Y-Coordinate (48 bytes) |
```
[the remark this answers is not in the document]
mhatrevi: Updated.

auth-manifest/app/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+            arg!(--"key-dir" <FILE> "Key files directory path")
+                .required(true)
+                .value_parser(value_parser!(PathBuf)),
+        )
+        .arg(
+            arg!(--"config" <FILE> "Configuration file")
+                .required(true)
+                .value_parser(value_parser!(PathBuf)),
+        )
+        .arg(
```
nquarton, line 53: I think this is a duplicate here

auth-manifest/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- ### **Image Metadata Collection**
+The Image Metadata Collection (IMC) is a collection of Image Metadata entries (IME). Each IME has a hash that matches a SOC images. The manifest manufacturer and owner private keys sign the IMC. The Preamble holds the IMC signatures. The manifest IMC manufacturer signatures are optional and are validated only if the FLAGS field Bit 0 = 1. Up to sixteen image hashes will be supported.
+
+| Field | Size (bytes) | Description|
+|-------|--------|------------|
+| Revision | 4 | Version of the IMC structure |
+| Reserved | 4 | Reserved |
+| Image Metadata Entry (IME) Count | 4 | Number of IME(s) in the IMC |
+| Image Metadata Entry (N) | Variable | List of Image Metadata Entry structures |
```
nquarton, line 33: Is the size of each entry fixed and "variable" here refers to the fact the entry count just varies? Would it make sense to include the "Image Metadata Entry structure" format in the readme?

auth-manifest/types/src/lib.rs
nquarton: This is a good question, it looks like it's the case for all LMS keys and signatures here.

nquarton (commented):
Would it be worthwhile to add any test for the app? Maybe even just a simple static config file with an expected output? Or do we expect some coverage of this in subsequent PRs?

auth-manifest/README.md
```diff
@@ -0,0 +1,33 @@
...
+| Owner ECC Signature | 96 | Owner ECDSA P-384 signature of the Version, Flags, Owner ECC and LMS public keys, hashed using SHA2-384. <br> **R-Coordinate:** Random Point (48 bytes) <br> **S-Coordinate:** Proof (48 bytes) |
+| Owner LMS Signature | 1620 | Owner LMS signature of the Version, Flags, Owner ECC and LMS public keys, hashed using SHA2-384. <br> **q:** Leaf of the Merkle tree where the OTS public key appears (4 bytes) <br> **ots:** Lmots Signature (1252 bytes) <br> **tree_type:** Lms Algorithm Type (4 bytes) <br> **tree_path:** Path through the tree from the leaf associated with the LM-OTS signature to the root. (360 bytes) |
+
+- ### **Image Metadata Collection**
+The Image Metadata Collection (IMC) is a collection of Image Metadata entries (IME). Each IME has a hash that matches a SOC images. The manifest manufacturer and owner private keys sign the IMC. The Preamble holds the IMC signatures. The manifest IMC manufacturer signatures are optional and are validated only if the FLAGS field Bit 0 = 1. Up to sixteen image hashes will be supported.
```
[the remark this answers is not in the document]
mhatrevi: I updated the description. This bit is for validating the vendor signatures on the image metadata collection (IMC). If Bit 0 = 0, Runtime will not verify the vendor signatures on the IMC.

auth-manifest/app/src/main.rs
mhatrevi: Updated.

auth-manifest/gen/src/generator.rs
mhatrevi: This is slightly different as the owner_man_priv_keys are obtained from owner_man_config. The (Some(), Some()) construct does not work in this case.

auth-manifest/README.md
```diff
@@ -0,0 +1,33 @@
...
+The Caliptra SOC manifest has two main components:
+
+- ### **Preamble**
+ The Preamble section contains the authorization manifest ECC and LMS public keys of the manufacturer and the owner. These public keys correspond to the private keys that sign the Image Metadata Collection (IMC) section. These signatures are included in the Preamble. The Caliptra firmware’s private keys endorse the manifest’s public keys, and these endorsements (signatures) are part of the Preamble as well.
+
+ *Note: All fields are little endian unless specified*
+
+| Field | Size (bytes) | Description|
+|-------|--------|------------|
+| Manifest Marker | 4 | Magic Number marking the start of the manifest. The value must be 0x41544D4EE (‘ATMN’ in ASCII)|
```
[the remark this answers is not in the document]
mhatrevi: Updated

auth-manifest/types/src/lib.rs
mhatrevi: This is due to a FIPS requirement for zeroizing local state.

auth-manifest/app/src/main.rs
mhatrevi: Removed.

auth-manifest/README.md
mhatrevi: That's correct. Added the IME format in the README.

mhatrevi:
> Would it be worthwhile to add any test for the app? Maybe even just a simple static config file with an expected output? Or do we expect some coverage of this in subsequent PRs?

There are tests in part 2 of this feature which utilize the generator library to create a manifest and validate it. I'll look into the possibility of adding a test to run the tool and validate the generated image as a separate change.

auth-manifest/types/src/lib.rs
mhatrevi: This is being done for the FW image as well. I don't have the context for why that was done originally. We'll review this as a separate task.

auth-manifest/app/src/config.rs
mhatrevi: Updated.

auth-manifest/gen/src/generator.rs
mhatrevi: This is modeled after the Caliptra FW generator which always takes LMS keys as part of the config. I can make the LMS keys optional as part of a separate change.

auth-manifest/README.md
```diff
@@ -0,0 +1,39 @@
...
+| Owner LMS Signature | 1620 | Owner LMS signature of the Version, Flags, Owner ECC and LMS public keys, hashed using SHA2-384. <br> **q:** Leaf of the Merkle tree where the OTS public key appears (4 bytes) <br> **ots:** Lmots Signature (1252 bytes) <br> **tree_type:** Lms Algorithm Type (4 bytes) <br> **tree_path:** Path through the tree from the leaf associated with the LM-OTS signature to the root. (360 bytes) |
+
+- ### **Image Metadata Entry**
+| Field | Size (bytes) | Description|
+|-------|--------|------------|
+| Image Hash | 48 | SHA2-384 hash of a SOC image |
+| Image Source | 4 | <TBD> |
+
+- ### **Image Metadata Collection**
+The Image Metadata Collection (IMC) is a collection of Image Metadata entries (IME). Each IME has a hash that matches a SOC images. The manifest vendor and owner private keys sign the IMC. The Preamble holds the IMC signatures. The manifest IMC vendor signatures are optional and are validated only if the FLAGS field Bit 0 = 1. Up to sixteen image hashes are supported.
```
jhand2, line 32: Will the owner signatures still be validated if Bit 0 = 0?

auth-manifest/gen/src/generator.rs
jhand2: > I can make the LMS keys optional as part of a separate change.

Wouldn't the spec need to change to allow marking LMS keys as optional? Like an extra flag bit or something?

mhatrevi: What I meant was while creating the authorization manifest, O'll make LMS keys optional and if not provided, won't perform any signing operations with those (obvious). Currently, both Caliptra FW and Auth Manifest generation code expect LMS keys to be provided. During validation of the authorization manifest in Runtime, the value from 'fuse_lms_verify' will be used to validate with the LMS keys. Not specifying LMS keys in the authorization manifest but setting fuse_lms_verify to true should be considered a configuration error.

auth-manifest/README.md
mhatrevi: Yes, owner signatures will still be validated. Bit0 is specifically for validation of vendor signatures over the image metadata entries.

nquarton (dismissed):

auth-manifest/gen/src/generator.rs
jhand2: Got it. Updating the tooling in a future PR sounds good.

> During validation of the authorization manifest in Runtime, the value from 'fuse_lms_verify' will be used to validate with the LMS keys.

Can you add this to the md file?

Thanks!

jhand2 (dismissed):

auth-manifest/gen/src/generator.rs
mhatrevi: Added guidance to the README file. I'll make the LMS keys optional in a follow-up PR.

nquarton (approved):

PR #1651, opened 2024-08-27
Libcaliptra piecewise FW load

nquarton:


jlmahowa-amd (dismissed):

libcaliptra/examples/generic/test.c
```diff
@@ -668,6 +668,91 @@ int rom_test_devid_csr(const test_info* info)
     return failure;
 }
 
+// Issue FW load commands repeatedly
+// Coverage for piecewise FW load and runtime FW updates
+int upload_fw_piecewise(const test_info* info)
+{
+    int failure = 0;
+    int status = boot_to_ready_for_fw(info, false);
```
fdamato, line 676: shall we sanity_check info ?

nquarton: If anything I would say to check it at run_tests() instead of each individual one. But I think either way the test is going to fail if someone breaks something.

libcaliptra/examples/generic/test.c
```diff
@@ -668,6 +668,96 @@ int rom_test_devid_csr(const test_info* info)
...
+
+    if (status){
+        printf("Failed to boot to ready for FW: 0x%x\n", status);
+        failure = 1;
+    }
+
+    // Some "random" size to split up the FW load into chunks
+    // These represent the first two chunks, the third chunk is the remainder of the image
+    // Sizes of 0 are ignored (meaning one fewer chunk is sent)
+    uint32_t chunk_sizes[][2] =    {
```
fdamato, line 686: Ideally the chunk should be randomly generated, and make sure they don`t add up for more space than the total image size...but I understand this is non production code...so it is not a sticky point

libcaliptra/examples/generic/test.c
```diff
@@ -668,6 +668,96 @@ int rom_test_devid_csr(const test_info* info)
...
+        {
+            printf("FW Load %d End Failed: 0x%x\n", i, status);
+            dump_caliptra_error_codes();
+            failure = 1;
+        } else {
+            printf("FW Load %d End: OK\n", i);
+        }
+    }
+
+    return failure;
```
fdamato, line 759: np: why carrying another flag here ? can`t we just return status ?

fdamato (approved):

libcaliptra/examples/generic/test.c
nquarton: These are written so the test will continue after a single failure but still report the result. So failure is sticky whereas status gets overwritten on each call.

libcaliptra/examples/generic/test.c
nquarton: I think the methodology I have seen in Caliptra is for deterministic testing over randomization which can cause flakeyness. I do agree that would be better coverage overall tough. (So long as the corner cases are for the code being tested and not corner cases in the test itself.)

fdamato (approved):

PR #1655, opened 2024-08-29
Deterministic ecc key pair gen kat

nquarton:


kat/src/ecc384_kat.rs
```diff
@@ -73,12 +90,44 @@ impl Ecc384Kat {
...
         }
         Ok(())
     }
+
+    fn kat_deterministic_key_pair_gen(
+        &self,
+        ecc: &mut Ecc384,
+        trng: &mut Trng,
+    ) -> CaliptraResult<()> {
+        let seed = Array4x12::new([0u32; 12]);
```
nquarton, line 112: TODO: Need to confirm seed==nonce==0 is right right test case for this

rom/dev/src/wdt.rs
```diff
@@ -35,17 +35,12 @@ pub fn start_wdt(soc_ifc: &mut SocIfc) {
         if wdt_timeout_cycles == 0 {
             wdt_timeout_cycles = 1;
         }
-        cprintln!(
-            "[state] Starting the Watchdog Timer {} cycles",
-            wdt_timeout_cycles
-        );
+        cprintln!("[state] WDT start {} cycles", wdt_timeout_cycles);
```
nquarton, line 38: These logs were made more concise to squeeze ROM back into it's size limit with the addition of the new KAT.

kat/src/ecc384_kat.rs
```diff
@@ -44,6 +44,22 @@ const SIGNATURE: Ecc384Signature = Ecc384Signature {
     ]),
 };
 
+const KEY_GEN_PRIV_KEY: Array4x12 = Array4x12::new([
```
swenson, line 47: If space is such a concern, couldn't we use these keys for the signature verification tests as well, and delete `SIGN_PRIV_KEY` and `SIGN_PUB_KEY`?

swenson (approved):
LGTM

kat/src/ecc384_kat.rs
nquarton: Thanks for the feedback. I am actually making the change of combining these two tests into one. For this reason and because of the time this test takes in general.

kat/src/ecc384_kat.rs
```diff
@@ -51,7 +51,7 @@ impl Ecc384Kat {
     /// This function executes the Known Answer Tests (aka KAT) for ECC384.
     ///
     /// Test vector source:
-    /// Generated using MbedTLS library.
+    /// TODO: Verify these vectors against some third party tool
```
swenson, line 54: I have verified they are correct using the Python `cryptography` library (which uses OpenSSL):

```
from cryptography.hazmat.primitives.asymmetric import ec
from cryptography.hazmat.primitives.hashes import SHA384
from cryptography.hazmat.primitives.asymmetric.utils import decode_dss_signature
from cryptography.hazmat.primitives.asymmetric.utils import Prehashed


def to_num(l):
    return int("".join(["%08x" % x for x in l]), 16)

priv_key = to_num([0xfeeef554, 0x4a765649, 0x90128ad1, 0x89e873f2, 0x1f0dfd5a, 0xd7e2fa86, 0x1127ee6e, 0x394ca784, 0x871c1aec, 0x032c7a8b, 0x10b93e0e, 0xab8946d6])
public_x = to_num([0xd7dd94e0, 0xbffc4cad, 0xe9902b7f, 0xdb154260, 0xd5ec5dfd, 0x57950e83, 0x59015a30, 0x2c8bf7bb, 0xa7e5f6df, 0xfc168516, 0x2bdd35f9, 0xf5c1b0ff])
public_y = to_num([0xbb9c3a2f, 0x061e8d70, 0x14278dd5, 0x1e66a918, 0xa6b6f9f1, 0xc1937312, 0xd4e7a921, 0xb18ef0f4, 0x1fdd401d, 0x9e771850, 0x9f8731e9, 0xeec9c31d])
r = to_num([0x93799D55, 0x12263628, 0x34F60F7B, 0x945290B7, 0xCCE6E996, 0x01FB7EBD, 0x026C2E3C, 0x445D3CD9, 0xB65068DA, 0xC0A848BE, 0x9F0560AA, 0x758FDA27])
s = to_num([0xE548E535, 0xA1CC600E, 0x133B5591, 0xAEBAAD78, 0x054006D7, 0x52D0E1DF, 0x94FBFA95, 0xD78F0B3F, 0x8E81B911, 0x9C2BE008, 0xBF6D6F4E, 0x4185F87D])

priv = ec.derive_private_key(priv_key, ec.SECP384R1())
pub = priv.public_key()


assert pub.public_numbers().x == public_x
assert pub.public_numbers().y == public_y

sign_data = b"\x00" * 48
signature = decode_dss_signature(
    priv.sign(sign_data, ec.ECDSA(Prehashed(SHA384()), deterministic_signing=True))
)

assert r == signature[0]
assert s == signature[1]
```

swenson (approved):

kat/src/ecc384_kat.rs
nquarton: I really appreciate you doing this! This python wrapper looks like a great way to leverage OpenSSL for this kind of stuff. I'll update this todo before it merges.

error/src/lib.rs
```diff
@@ -578,10 +578,14 @@ impl CaliptraError {
     pub const KAT_HMAC384_TAG_MISMATCH: CaliptraError = CaliptraError::new_const(0x90030002);
 
     pub const KAT_ECC384_SIGNATURE_GENERATE_FAILURE: CaliptraError =
-        CaliptraError::new_const(0x90040001);
+        CaliptraError::new_const(0x90040001); // Deprecated
```
swenson, line 581: You might move these to `///` comments above so that it shows up nicer in an IDE or documentation.

swenson (approved):
LGTM, one comment suggestion.

swenson (dismissed):

drivers/src/ecc384.rs
```diff
@@ -228,6 +228,59 @@ impl Ecc384 {
...
+            Some(pct_sig),
+        )
+    }
+
+    /// Private base function to generate ECC-384 Key Pair
+    /// pct_sig should only be provided in the KAT use case
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[inline(never)]
+    fn key_pair_base(
+        // TODO: Is there a naming convention for this situation?
```
mhatrevi, line 269: Don't think there is one. Let's remove the comment before merging.

error/src/lib.rs
```diff
@@ -577,11 +577,17 @@ impl CaliptraError {
     pub const KAT_HMAC384_FAILURE: CaliptraError = CaliptraError::new_const(0x90030001);
     pub const KAT_HMAC384_TAG_MISMATCH: CaliptraError = CaliptraError::new_const(0x90030002);
 
+    /// Deprecated
     pub const KAT_ECC384_SIGNATURE_GENERATE_FAILURE: CaliptraError =
```
mhatrevi, line 581: If these are not used, let's remove them.

mhatrevi (dismissed):

error/src/lib.rs
nquarton: I can do that. I just left them in case someone was referencing this documentation in in main when they are working on an older version. You do agree with "retiring" those values though, right? I don't think we want to reuse an old error code if we deprecate it.

bluegate010: In [this](https://github.com/chipsalliance/caliptra-sw/pull/1767/files#diff-009b2077ef4a38a7715447e771702e14e3f522198f0a8cc09f073a08bd02eaa8) file from https://github.com/chipsalliance/caliptra-sw/pull/1767 I removed error codes but left comments to indicate what the codes used to be used for, to prevent reuse.

nquarton: I think that's a good idea. I'll make that change.

error/src/lib.rs
```diff
@@ -581,11 +581,13 @@ impl CaliptraError {
     pub const KAT_HMAC384_FAILURE: CaliptraError = CaliptraError::new_const(0x90030001);
     pub const KAT_HMAC384_TAG_MISMATCH: CaliptraError = CaliptraError::new_const(0x90030002);
 
-    pub const KAT_ECC384_SIGNATURE_GENERATE_FAILURE: CaliptraError =
-        CaliptraError::new_const(0x90040001);
-    pub const KAT_ECC384_SIGNATURE_VERIFY_FAILURE: CaliptraError =
-        CaliptraError::new_const(0x90040002);
+    // 0x90040001 was KAT_ECC384_SIGNATURE_GENERATE_FAILURE
+    // 0x90040002 was KAT_ECC384_SIGNATURE_VERIFY_FAILURE
```
bluegate010, line 585: Wondering if we should also mark this as deprecated in error/README.md.

nquarton: That's a good question. It doesn't look like that file has really been maintained unfortunately. It's very out-of-sync with the lib.rs source. We should probably either remove that or auto-generate it.

nquarton: This table has been removed now in https://github.com/chipsalliance/caliptra-sw/pull/1779 after discussion in the FW sync today

mhatrevi (approved):

JohnTraverAmd (approved):

PR #1657, opened 2024-08-30
man-auth-rt: Address review comments

ArthurHeymans:


jhand2:
I think this was merged with Vishal's PR. Should we close, or are there still things in this PR that should be merged?

mhatrevi:
> I think this was merged with Vishal's PR. Should we close, or are there still things in this PR that should be merged?

That's correct.

PR #1659, opened 2024-08-31
fix git submodule update command given in README

yu-zou:
If following the git steps in the main README to clone repo
```bash
git clone https://github.com/chipsalliance/caliptra-sw \
    --config submodule.recurse=true \
    --recurse-submodules=dpe
cd caliptra-sw
cargo build
```
it requires adding submodule path to make the git submodule update command, given in the verilator README, works.
```bash
git submodule update --init hw/1.0/rtl/
```


hw/verilated/README.md
```diff
@@ -21,7 +21,7 @@ sudo make install
 ## Rust integration tests:
 
 ```shell
-git submodule update --init"  # Needed the first time
+git submodule update --init hw/1.0/rtl/"  # Needed the first time
```
jhand2, line 24: This will only pull the HW for 1.0, but depending on what the person wants to do, they may want the latest hw revision.

In CI we checkout both the `1.0` and `latest` submodules. I think we at least need to have a note that tells the reader that there are different RTL submodules.

yu-zou: This fix is only related to this verilator related README. The Makefile used to compile verilated hardware does not use latest, only 1.0.

Maybe latest version is used by some other Makefile, but for this specific Makefile, I think probably it's fine to change README in this way to make it consistent with Makefile.

A better way I think might be putting this submodule update command inside Makefile, so that there will be no version inconsistency issue in the future, e.g., Makefile has changed to use latest version but README is not updated together.

jhand2: Ok good point. This looks good. If we change it to add the ability to use a different hw version, we can update the readme at that time.

jhand2 (approved):

PR #1666, opened 2024-09-07
Get runtime using recovery register interface

ArthurHeymans:


mhatrevi:
Closing as this is no longer needed.

PR #1667, opened 2024-09-09
Enhance PAUSER tests

rusty1968:
- Modify HwModel  trait so it boots with a set of  five valid pausers for all supported model types.
- Add a negative test to ascertain that , when FpgaRealTime is selected, a  SIGBUS is raised if an invalid
   PAUSER tries to access the mailbox.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -81,6 +81,15 @@ pub type DefaultHwModel = ModelVerilated;
 #[cfg(feature = "fpga_realtime")]
 pub type DefaultHwModel = ModelFpgaRealtime;
 
+pub enum ValidPaUsers {
+    Pl0 = 0x01,
+    Pl1 = 0x02,
+    Pl2 = 0x03,
+    Pl3 = 0x04,
+    Pl4 = 0x05,
+}
+pub const DEFAULT_APB_PAUSER: u32 = ValidPaUsers::Pl0 as u32;
+
 /// Constructs an HwModel based on the cargo features and environment
 /// variables. Most test cases that need to construct a HwModel should use this
 /// function over HwModel::new_unbooted().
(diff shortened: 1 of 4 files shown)


hw-model/src/lib.rs
```diff
@@ -531,15 +543,31 @@ pub trait HwModel {
...
-        self.soc_ifc()
-            .cptra_mbox_valid_pauser()
-            .at(0)
-            .write(|_| boot_params.valid_pauser);
-        self.soc_ifc()
-            .cptra_mbox_pauser_lock()
-            .at(0)
-            .write(|w| w.lock(true));
+        // Set up two valid PAUSERs  for the mailbox
+        #[cfg(not(feature = "fpga_realtime"))]
```
korran, line 547: Why not support multiple pausers for all implementations?

hw-model/src/lib.rs
```diff
@@ -248,6 +254,9 @@ pub struct BootParams<'a> {
     pub initial_dbg_manuf_service_reg: u32,
     pub initial_repcnt_thresh_reg: Option<CptraItrngEntropyConfig1WriteVal>,
     pub initial_adaptp_thresh_reg: Option<CptraItrngEntropyConfig0WriteVal>,
+    #[cfg(feature = "fpga_realtime")]
```
korran, line 257: Can we just make this a vector of valid pausers? (no conditional computation?)

rusty1968: Makes sense. Will do.

hw-model/src/lib.rs
rusty1968: Do we have any restrictions on the verilated model?

korran: I think this will work fine there too

hw-model/src/lib.rs
rusty1968: Done

rusty1968:
@korran , is there anything that would delete a crate binary i am using for integration tests? CARGO_BIN_EXE_fpga_realtime_mbox_pauser?

hw-model/src/lib.rs
rusty1968: done

jlmahowa-amd (approved):

PR #1684, opened 2024-09-20
registers/systemrdl: add I3C registers and RDL

swenson:
And update parser and generator to support the more advanced formats.


swenson:
Superseded by https://github.com/chipsalliance/caliptra-sw/pull/1697

PR #1691, opened 2024-09-25
[DUPLICATE] ci: Build ROM release from versioned git ref

benjamindoron:
Reopening to attempt executing CI inside organisation


jhand2:
Idea for making this work:

* Send a PR to main with a dummy workflow of this name & with all the permissions this workflow needs
* Submit that to main
* Then we should be able to test the full workflow in your branch with a manual invocation

PR #1693, opened 2024-10-01
fix verilator git steps

mhatrevi:
If following the git steps in the main README to clone repo, it requires adding submodule path to make the git submodule update command, given in the verilator README, works.


mhatrevi:
Abandoning this PR and creating a new one with original commit reference.

PR #1699, opened 2024-10-04
Update rtl

ArthurHeymans:
This adds SRAM and the AXI DMA engine to the register access.


swenson:
If you apply this patch, the build should pass:

```diff
diff --git a/ureg/lib/codegen/src/lib.rs b/ureg/lib/codegen/src/lib.rs
index b8e4166e..f696427c 100644
--- a/ureg/lib/codegen/src/lib.rs
+++ b/ureg/lib/codegen/src/lib.rs
@@ -874,6 +874,7 @@ pub fn generate_code(block: &ValidatedRegisterBlock, options: Options) -> TokenS
         }
         reg_tokens.extend(anon_type_tokens);
         block_tokens = quote! {
+            #[allow(dead_code)]
             #[derive(Clone, Copy)]
             pub struct RegisterBlock<TMmio: ureg::Mmio + core::borrow::Borrow<TMmio>>{
                 ptr: *mut #raw_ptr_type,
```

swenson (approved):
LGTM.

.github/workflows/build-test-verilator.yml
```diff
@@ -34,96 +34,95 @@ jobs:
...
-          export RUSTC_WRAPPER=~/.cargo/bin/sccache
-          export CXX="sccache g++"
-          (cd hw-model && cargo --config "$EXTRA_CARGO_CONFIG" test --locked --release --features verilator)
-          (cd hw/verilated && cargo --config "$EXTRA_CARGO_CONFIG" test --locked --release --features verilator)
-
-      - name: Check source-code formatting (run "cargo fmt" if this fails)
-        run: |
-          (cd hw-model && cargo fmt --check)
-
+      # - name: Restore sccache binary
```
jhand2, line 37: Why is this removed? If you're having an issue with the cache being incorrect you can change the cache buster value: https://github.com/chipsalliance/caliptra-sw/blob/470e9f9844d8d3c52149592986bf367ad08b44ea/.github/workflows/build-test-verilator.yml#L26

jhand2 (changes requested):

.github/workflows/build-test-verilator.yml
ArthurHeymans: So the whole verilator flow is not working on the 2.0 RTL and needs some work which I was told will be taken care of in a few weeks. I move forward commenting out most of this action was done. There isn't anything wrong with sscache afaik.

jhand2: Got it, thanks for clarifying

mhatrevi (dismissed):

swenson (approved):

jhand2 (approved):

PR #1700, opened 2024-10-04
Add i3c-core RTL code

ArthurHeymans:


swenson (dismissed):
LGTM, though I'd prefer it were loaded somewhere else in memory that is closer to our other peripherals (like in the `0x200x` range) so that we leave a cleaner footprint for integrators.

ArthurHeymans:
> LGTM, though I'd prefer it were loaded somewhere else in memory that is closer to our other peripherals (like in the `0x200x` range) so that we leave a cleaner footprint for integrators.

SW-emulator sets it to 0x1003_8000. I can update that.

registers/bin/generator/src/main.rs
```diff
@@ -41,6 +41,8 @@ static CALIPTRA_RDL_FILES: &[&str] = &[
     "src/integration/rtl/caliptra_reg.rdl",
 ];
 
+static I3C_CORE_RDL_FILES: &[&str] = &["src/rdl/registers.rdl"];
```
jhand2, line 44: I was under the impression that Caliptra Core would just see the recovery interface and not bus-specific details. Could you point me at any spec for what this i3c RTL in Caliptra Core is for?

ArthurHeymans: So the recovery interface is part of the i3c periph. The intention is indeed to only use that bus abstracted part.
SecFwRecoveryIf in https://github.com/chipsalliance/i3c-core/tree/main/src/rdl is that interface we'll be targetting.

I think the other capabilities like TTI (I3C Target) will be used by MCU.

mhatrevi (dismissed):

registers/bin/generator/src/main.rs
jhand2: Ok, thanks for the details. @bharatpillilli would it make sense to factor the recovery interface RTL to be independent from the i3c peripheral RTL?

Not a huge deal, just thinking it makes the organization more clear for what Caliptra Core depends on.

jhand2 (dismissed):

swenson (approved):

Issue #1678, opened 2024-09-17
Runtime FW SHALL refuse any commands coming from PAUSER 0xFFFFFFFF


fdamato:
Currently Runtime FW is accepting stash measurement from PAUSER 0xFFFFFFFF, which is reserved by Caliptra as its own locality for DPE (e.g. RTJM uses it) to indicate the evidence has been measured by Caliptra itself.

jhand2:
I think the final decision here is that we will reject any STASH_MEASUREMENT commands in Runtime that don't come from the PL0 PAUSER.

PR #1704, opened 2024-10-07
Limit STASH_MEASUREMENT command to PL0 callers.

clundin25:
Fixes https://github.com/chipsalliance/caliptra-sw/issues/1678.

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -61,6 +61,12 @@ use core::cmp::Ordering::{Equal, Greater};
 use crypto::{AlgLen, Crypto, CryptoBuf, Hasher};
 use zerocopy::AsBytes;
 
+#[derive(PartialEq)]
+pub enum PauserPrivileges {
+    PL0,
+    PL1,
+}
+
 pub struct Drivers {
     pub mbox: Mailbox,
     pub sha_acc: Sha512AccCsr,
@@ -237,16 +243,8 @@ impl Drivers {
             }
         } else {
             let pl0_pauser = drivers.persistent_data.get().manifest1.header.pl0_pauser;
-            let flags = drivers.persistent_data.get().manifest1.header.flags;
-            let locality = drivers.mbox.user();
             // check that DPE used context limits are not exceeded
-            let dpe_context_threshold_exceeded = Self::is_dpe_context_threshold_exceeded(
-                pl0_pauser,
-                flags,
-                locality,
-                &drivers.persistent_data.get().dpe,
-                true,
-            );
+            let dpe_context_threshold_exceeded = drivers.is_dpe_context_threshold_exceeded();
             if cfi_launder(dpe_context_threshold_exceeded.is_ok()) {
                 cfi_assert!(dpe_context_threshold_exceeded.is_ok());
             } else {
@@ -364,6 +362,7 @@ impl Drivers {
         let caliptra_locality = 0xFFFFFFFF;
         let pl0_pauser_locality = drivers.persistent_data.get().manifest1.header.pl0_pauser;
         let hashed_rt_pub_key = drivers.compute_rt_alias_sn()?;
+        let privilege_level = drivers.caller_privilege_level();
 
         // create a hash of all the mailbox valid pausers
         const PAUSER_COUNT: usize = 5;
@@ -446,13 +445,12 @@ impl Drivers {
         for measurement_log_entry in measurement_log.iter().take(num_measurements) {
             // Check that adding this measurement to DPE doesn't cause
             // the PL0 context threshold to be exceeded.
-            let flags = pdata.manifest1.header.flags;
-            Self::is_dpe_context_threshold_exceeded(
+            //
+            // Use the helper method here because the DPE instance holds a mutable reference to driver
+            Self::is_dpe_context_threshold_exceeded_helper(
                 pl0_pauser_locality,
-                flags,
-                pl0_pauser_locality,
+                &privilege_level,
                 &dpe,
-                false,
             )?;
 
             let measurement_data = measurement_log_entry.pcr_entry.measured_data();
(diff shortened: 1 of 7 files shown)


runtime/src/drivers.rs
```diff
@@ -582,6 +582,22 @@ impl Drivers {
...
     }
 
+    /// Checks if the caller is privilege level 0
+    ///
+    /// # Arguments
+    ///
+    /// * `pl0_pauser` - Value of PL0 PAuser
+    /// * `flags` - Flags from manifest header
+    /// * `locality` - Caller's locality
+    pub fn is_caller_pl0(pl0_pauser: u32, flags: u32, locality: u32) -> bool {
```
clundin25, line 592: I think refactoring this and `is_caller_pl1` to return a privilege level enum would be easier to reason about.

jhand2: I think that's reasonable. Feel free to either do that in a this PR or in a follow-up

jhand2, line 592: If you make this a member function on the drivers struct (`pub fn is_caller_pl0(&self, ...)`) are you able to remove all the parameters? I think all of this information comes from various drivers. So we can probably hide all these details from the caller.

runtime/src/drivers.rs
```diff
@@ -537,20 +535,18 @@ impl Drivers {
...
+        )
+    }
+
+    fn is_dpe_context_threshold_exceeded_helper(
         pl0_pauser: u32,
-        flags: u32,
-        locality: u32,
+        caller_privilege_level: &PauserPrivileges,
         dpe: &DpeInstance,
-        check_already_exceeded: bool,
```
clundin25, line 553: Looks like this was never used.

runtime/src/drivers.rs
clundin25: I was able to make the public interface a member function. The refactor also bubbled up to `is_dpe_context_threshold_exceeded` since it calls `is_caller_pl1`. I went ahead and also refactored this method, but because of some logic in `initialize_dpe` I had to keep a private non member function.

There may be a better way, but I didn't find an obvious way to de-couple the newly created DPE instance from the driver, as said instance is holding a mutable reference to the driver by the time the second call to `DeriveContextCmd` happens. 

To work around this I kept a private helper method for `is_dpe_context_threshold_exceeded`.

runtime/src/drivers.rs
```diff
@@ -446,13 +445,12 @@ impl Drivers {
...
             // the PL0 context threshold to be exceeded.
-            let flags = pdata.manifest1.header.flags;
-            Self::is_dpe_context_threshold_exceeded(
-                pl0_pauser_locality,
-                flags,
+            //
+            // Use the helper method here because the DPE instance holds a mutable reference to driver
+            Self::is_dpe_context_threshold_exceeded_helper(
                 pl0_pauser_locality,
+                &privilege_level,
```
jhand2, line 452: nit: Since the enum is small, can probably just pass by value instead of borrowing

runtime/src/drivers.rs
jhand2: lgtm

runtime/src/drivers.rs
```diff
@@ -537,20 +535,18 @@ impl Drivers {
...
-    pub fn is_dpe_context_threshold_exceeded(
+    pub fn is_dpe_context_threshold_exceeded(&self) -> CaliptraResult<()> {
+        Self::is_dpe_context_threshold_exceeded_helper(
+            self.persistent_data.get().manifest1.header.pl0_pauser,
+            &self.caller_privilege_level(),
+            &self.persistent_data.get().dpe,
+        )
+    }
+
+    fn is_dpe_context_threshold_exceeded_helper(
```
jhand2, line 546: I assume this is necessary because of borrow-checker issuses? It's hard to tell from reading the code what the problem might be.

jhand2 (commented):
Great work! Using the enum for priv levels makes the code a lot clearer.

runtime/src/drivers.rs
clundin25: Exactly. Here is a short snippet extracted from `initialize_dpe`

```rust
        let pdata = drivers.persistent_data.get_mut(); 
        let mut crypto = DpeCrypto::new( // <=== Holds Mutable reference to drivers until dropped.
            &mut drivers.sha384,
            &mut drivers.trng,
            &mut drivers.ecc384,
            &mut drivers.hmac384,
            &mut drivers.key_vault,
            &mut pdata.fht.rt_dice_pub_key,
            key_id_rt_cdi,
            key_id_rt_priv_key,
        );

        let (nb, nf) = Self::get_cert_validity_info(&pdata.manifest1);
        let mut env = DpeEnv::<CptraDpeTypes> {
            crypto,
           . . .
        };
        let mut dpe = DpeInstance::new_auto_init(
            &mut env,
            . . .
        )
        .map_err(|_| CaliptraError::RUNTIME_INITIALIZE_DPE_FAILED)?;
```

The env / dpe objects are used to execute `DeriveContextCmd`, so it's difficult to grab a reference to drivers in between calls.

runtime/src/drivers.rs
clundin25: I had to be careful to instantiate the enum before a mutable reference to the `driver` variable is grabbed. In this case moving the enum would make the object invalid in the next loop iteration.

Let me know if you have any concerns about this. A deeper refactor or an alternative approach could be cleaner.

jhand2: If you derive the Copy trait on the enum does that solve the problem?

jhand2: I can't remember if you also need to derive Clone or not

clundin25: I've added a clone and now the enum is passed by value

runtime/src/drivers.rs
jhand2: Oh I understand now. I thought this helper was only used in the function right above it. But it's used elsewhere as well. Ok thanks for the explanation!

runtime/src/drivers.rs
jhand2: Looks good, thanks!

jhand2 (approved):

PR #1705, opened 2024-10-08
[feat] Manifest based authorization - Authorization manifest and image hash separation

mhatrevi:


auth-manifest/gen/src/lib.rs
```diff
@@ -39,6 +41,24 @@ pub struct AuthManifestGeneratorConfig {
...
 
     pub owner_man_key_info: Option<AuthManifestGeneratorKeyConfig>,
+}
+
+// /// Image Metadata Collection Image Generator Key Configuration
+// #[derive(Default, Clone)]
+// pub struct ImcGeneratorKeyConfig {
+//     pub priv_keys: Option<AuthManifestPrivKeys>,
+// }
+
```
swenson, line 51: Accidentally left in?

```suggestion
```

runtime/src/authorize_and_stash.rs
```diff
@@ -62,20 +63,52 @@ impl AuthorizeAndStashCmd {
...
+            let mut auth_result = IMAGE_NOT_AUTHORIZED;
+            let mut fw_id_found = false;
+            for metadata_entry in auth_manifest_image_metadata_col
+                .image_metadata
+                .image_metadata_list
+                .iter()
+            {
+                if cmd.fw_id == metadata_entry.fw_id {
+                    fw_id_found = true;
+                    if cfi_launder(metadata_entry.digest) == cmd.measurement {
```
swenson, line 95: Are we not doing constant-time comparisons for digests?

swenson, line 95: We should use constant-time comparisons for user-provided hashes when possible, e.g., `constant_time_eq::constant_time_eq`.

swenson (approved):
LGTM, though I am a little concerned about constant-time comparisons. I'll do a separate investigation on that though.

api/src/mailbox.rs
```diff
@@ -49,6 +49,9 @@ impl CommandId {
     // The authorization manifest set command.
     pub const SET_AUTH_MANIFEST: Self = Self(0x4154_4D4E); // "ATMN"
 
+    // The authorization manifest set command.
```
swenson, line 52:
```suggestion
    // The image metadata set command.
```

runtime/src/set_image_metadata.rs
```diff
@@ -0,0 +1,344 @@
...
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    set_image_metadata.rs
+
+Abstract:
+
+    File contains AuthManifest mailbox command.
```
swenson, line 11:
```suggestion
    File contains Set Image Metadata mailbox command.
```

nquarton (commented):
This looks pretty good to me. I think this mainly needs an update in the runtime readme with the new command and its usage. Is there (or will there be) an update to the original "Measurement Manifest" doc that describes this feature?

PR #1711, opened 2024-10-10
Remove ROM freeze policy

mhatrevi:


swenson (dismissed):
LGTM.

In the future, we might want to bring this back, and limit it to the right target branches.

swenson (approved):
LGTM

mhatrevi:
This change is no longer necessary as the ROM freeze policy is made optional for the main-2.x branch.

PR #1714, opened 2024-10-10
cfi: Simpler launder implementation for common types

swenson:
`cfi_launder` is meant to prevent the Rust compiler from optimizing a value away.

Our current implementation uses `core::hint::black_box()`, which is the recommended way in Rust. The problem is, this appears to often force the argument to spill into memory and to be reloaded, which can be a lot of extra instructions.

The original inspiration for this function is from, I believe, [OpenTitan's launder* functions](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/base/hardened.h#L193). There, they use an LLVM-specific trick of a blank inline assembly block to force the compiler to keep the argument in a register.

After reviewing our code and speaking with @vsonims, it sounds like the intention of the launder in our code is to prevent the compiler from optimizing the value away (as the comments suggest), so the simpler inline assembly trick may be sufficient (since we use the official Rust compiler, which uses LLVM).

The biggest problem is that we launder many types of values in our code and not all of them fit into a register.

So, this PR represents an incremental change: for `u32`s and similar small types, we implement `cfi_launder` using the inline assembly trick from OpenTitan. For any other types, we have a trait that can be derived that will call `core::hint::black_box` in the same way as today.

We can do future follow-up PRs to try to try to clean up some of those other uses of `cfi_launder` to hopefully shrink the code more.

I also slipped in avoid a few extra copies in the verifier by using references instead of copies (this saves ~80 bytes of instruction space).

This PR appears to shrink the ROM code size by 1232 bytes and the runtime firmware by 700 bytes.


cfi/lib/src/cfi.rs
```diff
@@ -94,16 +94,75 @@ impl From<CfiPanicInfo> for CaliptraError {
...
-        // Note: The black box seems to be disabling more optimization
-        // than necessary and results in larger binary size
-        core::hint::black_box(val)
+        Launder { _val: PhantomData }.launder(val)
     } else {
         val
     }
 }
 
+pub trait LaunderTrait<T> {
```
swenson, line 108: These Rust type acrobatics are so that we can have special, smaller implementations for register-sized but default to `core::hint::black_box` for everything else.

I'm open to suggestions on how to make this simpler or avoid a `derive`.

cfi/lib/src/cfi.rs
```diff
@@ -94,16 +94,75 @@ impl From<CfiPanicInfo> for CaliptraError {
...
+pub struct Launder<T> {
+    _val: PhantomData<T>,
+}
+
+impl LaunderTrait<u32> for Launder<u32> {
+    #[allow(asm_sub_register)]
+    fn launder(&self, val: u32) -> u32 {
+        let mut val = val;
+        unsafe {
+            core::arch::asm!(
```
jhand2, line 123: I think this is magic enough to be worth a comment about why it works 😄 Also can we be confidence that this will continue to work in future versions of LLVM? I wonder if we can write a test that proves the compiled binary didn't optimizer out values passed to cfi_launder.

swenson: I'll definitely write a comment.

It's quite difficult to test this, I think, and it certainly is not guaranteed by LLVM in the future (nor would be basically any trick, probably, since all of these are meant to fool the compiler without actually doing anything). I verified it by loading code into Compiler Explorer for now.

There are some ways to test this by invoking the compiler in a test and checking the emitted assembly and, for example, counting instructions with and without this function being applied.

If this stops working in the future, this doesn't stop the code using it from working, it just means that one layer of protection is removed. And theoretically we'd notice because the code size would change.

swenson: I added a test that invokes `rustc --emit asm` and does a simple check if a functions is optimized enough (with and without laundering).

jhand2: Nice :) Thanks!

I talked with the OT folks and they do manual pre-TO validation for things like this. Might be worth doing something at ROM release time (or adding a note to pre-TO guidance for integrators) to spot check important sections.

jhand2 (approved):

cfi/lib/src/cfi.rs
swenson: I think we're mostly safe here, since we pin the version of Cargo (and presumably rustc and LLVM). So we mostly need to validate when we update our toolchain.

swenson:
(I've also looked at removing `Copy` traits from a few other types and trying to use more references so that the laundering could be more effective without copying, but I was surprised to find that this increased the code size.

There are more likely a few more types though that would be worth streamlining so we can save more code space, but I leave that to a future PR.)

jhand2 (approved):

rusty1968 (approved):

nquarton (dismissed):

swenson:
@jhand2, @nquarton, @rusty1968 -- approvals were wiped when I rebased and updated the SHAs. No other changes were made. Please re-review and I'll get someone to merge this ASAP for the 1.2 ROM release.

nquarton (approved):

mhatrevi (dismissed):

cfi/lib/src/cfi.rs
```diff
@@ -184,10 +276,28 @@ cfi_assert_macro!(cfi_assert_lt, <, Ord, PartialOrd, AssertLtFail);
 cfi_assert_macro!(cfi_assert_ge, >=, Ord, PartialOrd, AssertGeFail);
 cfi_assert_macro!(cfi_assert_le, <=, Ord, PartialOrd, AssertLeFail);
 
+// special case for bool assert
+#[inline(always)]
+#[allow(unused)]
```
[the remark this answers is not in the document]
swenson: It is used, but not in this package, so it was generating warnings for me.

mhatrevi (approved):

PR #1624, opened 2024-07-30
Dilithium driver

ArthurHeymans:
This add preliminary driver support for the dilithium peripheral.
The code follows the ecc384 driver structure.

PR #1717, opened 2024-10-14
[feat] MLDSA-87 driver interface

mhatrevi:
This change contains the interface for the MLDSA-87 driver.


ArthurHeymans (approved):
Is the idea to pull in code from https://github.com/chipsalliance/caliptra-sw/pull/1624 ?

swenson (approved):
LGTM

mhatrevi:
> Is the idea to pull in code from #1624 ?

Yes; this change is to unblock the DICE derivation changes. The driver implementation will happen when the Dilithium Key Vault hw integration is available.

PR #1720, opened 2024-10-15
Reject mbx commands from 0xFFFFFFFF pauser in both ROM and runtime

nquarton:


jhand2 (dismissed):
This looks reasonable. Thanks!

jhand2 (approved):

Issue #1662, opened 2024-09-03
Consider shrinking DATA_ORG section to allow for expanding DCCM usage


jhand2:
See https://github.com/chipsalliance/caliptra-sw/pull/1649#discussion_r1733430746 for details

PR #1722, opened 2024-10-17
Refactor memory_layout and persistent_data

clundin25:
* Reserve a portion of DCCM for future changes to PersistentData.
    * Reserved 19 KiB.
* Remove most hard coded addresses in memory layout.
* Left 8K in the Data section. Currently less than 4KiB is being used.

This resolves https://github.com/chipsalliance/caliptra-sw/issues/1662.

### DIFF
--- a/drivers/src/memory_layout.rs
+++ b/drivers/src/memory_layout.rs
@@ -12,12 +12,6 @@ Abstract:
 
 --*/
 
-#[cfg(test)]
-use crate::FirmwareHandoffTable;
-
-#[cfg(test)]
-use caliptra_image_types::ImageManifest;
-
 //
 // Memory Addresses
 //
@@ -28,20 +22,9 @@ pub const DCCM_ORG: u32 = 0x50000000;
 pub const ROM_DATA_ORG: u32 = 0x50000000;
 pub const CFI_STATE_ORG: u32 = 0x500003E4; // size = 6 words
 pub const BOOT_STATUS_ORG: u32 = 0x500003FC;
-pub const MAN1_ORG: u32 = 0x50000400;
-pub const MAN2_ORG: u32 = 0x50001C00;
-pub const FHT_ORG: u32 = 0x50003400;
-pub const LDEVID_TBS_ORG: u32 = 0x50003C00;
-pub const FMCALIAS_TBS_ORG: u32 = 0x50004000;
-pub const RTALIAS_TBS_ORG: u32 = 0x50004400;
-pub const PCR_LOG_ORG: u32 = 0x50004800;
-pub const MEASUREMENT_LOG_ORG: u32 = 0x50004C00;
-pub const FUSE_LOG_ORG: u32 = 0x50005000;
-pub const DPE_ORG: u32 = 0x50005400;
-pub const PCR_RESET_COUNTER_ORG: u32 = 0x50006800;
-pub const AUTH_MAN_IMAGE_METADATA_LIST_ORG: u32 = 0x50006C00;
-pub const IDEVID_CSR_ORG: u32 = 0x50008800;
-pub const DATA_ORG: u32 = 0x50008C00;
+pub const PERSISTENT_DATA_ORG: u32 = 0x50000400;
+
+pub const DATA_ORG: u32 = 0x5000D800;
 
 pub const STACK_ORG: u32 = 0x5000f800;
 pub const ROM_STACK_ORG: u32 = 0x5001C000;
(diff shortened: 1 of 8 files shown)


jhand2 (dismissed):

drivers/src/memory_layout.rs
```diff
@@ -28,20 +22,12 @@ pub const DCCM_ORG: u32 = 0x50000000;
 pub const ROM_DATA_ORG: u32 = 0x50000000;
 pub const CFI_STATE_ORG: u32 = 0x500003E4; // size = 6 words
 pub const BOOT_STATUS_ORG: u32 = 0x500003FC;
-pub const MAN1_ORG: u32 = 0x50000400;
-pub const MAN2_ORG: u32 = 0x50001C00;
-pub const FHT_ORG: u32 = 0x50003400;
 pub const LDEVID_TBS_ORG: u32 = 0x50003C00;
 pub const FMCALIAS_TBS_ORG: u32 = 0x50004000;
 pub const RTALIAS_TBS_ORG: u32 = 0x50004400;
```
jhand2, line 27: Aren't these addresses part of `PERSISTENT_DATA` as well?

drivers/src/persistent.rs
```diff
@@ -102,35 +119,47 @@ impl IdevIdCsr {
...
+const _: () = assert!(size_of::<PcrLogArray>() <= PCR_LOG_SIZE as usize);
+const _: () = assert!(size_of::<StashMeasurementArray>() <= MEASUREMENT_LOG_SIZE as usize);
+const _: () = assert!(size_of::<FuseLogArray>() <= FUSE_LOG_SIZE as usize);
+#[cfg(feature = "runtime")]
+const _: () = assert!(size_of::<PcrResetCounter>() <= PCR_RESET_COUNTER_SIZE as usize);
+#[cfg(feature = "runtime")]
+const _: () = assert!(
+    size_of::<AuthManifestImageMetadataCollection>() <= AUTH_MAN_IMAGE_METADATA_MAX_SIZE as usize
+);
+const _: () = assert!(size_of::<IdevIdCsr>() <= IDEVID_CSR_SIZE as usize);
```
jhand2, line 135: Are these asserts necessary? I would expect the assert on line 122 to be sufficient. I think we mostly just care that weird alignment hasn't changed the overall size of the struct.

drivers/src/memory_layout.rs
clundin25: They are, but they are also used in the generated linker script here: https://github.com/chipsalliance/caliptra-sw/blob/main/cpu/gen/src/lib.rs#L12-L14.

I've gone ahead and removed them. Let me know if there is any concern, I think it's okay to remove them from the linker script, but I am not sure.

drivers/src/persistent.rs
clundin25: There is padding between certain sections / reserved memory for different compilation targets, so if a struct grows beyond the hard coded sizes weird things might happen.

I think it's a good sanity check in case some of these types grow in the future, but could be a bit redundant.

What's your preference?

jhand2: Ok, makes sense. Let's go ahead and leave this.

drivers/src/memory_layout.rs
jhand2: lgtm, thanks!

jhand2 (approved):

jhand2 (dismissed):

drivers/src/memory_layout.rs
```diff
@@ -28,20 +22,9 @@ pub const DCCM_ORG: u32 = 0x50000000;
...
-pub const MEASUREMENT_LOG_ORG: u32 = 0x50004C00;
-pub const FUSE_LOG_ORG: u32 = 0x50005000;
-pub const DPE_ORG: u32 = 0x50005400;
-pub const PCR_RESET_COUNTER_ORG: u32 = 0x50006800;
-pub const AUTH_MAN_IMAGE_METADATA_LIST_ORG: u32 = 0x50006C00;
-pub const IDEVID_CSR_ORG: u32 = 0x50008800;
-pub const DATA_ORG: u32 = 0x50008C00;
+pub const PERSISTENT_DATA_ORG: u32 = 0x50000400;
+
+pub const DATA_ORG: u32 = 0x5000D800;
```
mhatrevi, line 27: Is the DATA_ORG really needed?

drivers/src/memory_layout.rs
```diff
@@ -55,26 +38,15 @@ pub const ROM_NSTACK_ORG: u32 = 0x5001FC00;
 //
 // Memory Sizes In Bytes
 //
+pub const PERSISTENT_DATA_SIZE: u32 = 53 * 1024;
```
mhatrevi, line 41: Can you put a comment explaining what how PERSISTENT_DATA_SIZE is computed?

drivers/src/memory_layout.rs
```diff
@@ -55,26 +38,15 @@ pub const ROM_NSTACK_ORG: u32 = 0x5001FC00;
...
-pub const RTALIAS_TBS_SIZE: u32 = 1024;
-pub const PCR_LOG_SIZE: u32 = 1024;
-pub const MEASUREMENT_LOG_SIZE: u32 = 1024;
-pub const FUSE_LOG_SIZE: u32 = 1024;
-pub const DPE_SIZE: u32 = 5 * 1024;
-pub const PCR_RESET_COUNTER_SIZE: u32 = 1024;
-pub const AUTH_MAN_IMAGE_METADATA_MAX_SIZE: u32 = 7 * 1024;
-pub const IDEVID_CSR_SIZE: u32 = 1024;
-pub const DATA_SIZE: u32 = 27 * 1024;
+pub const DATA_SIZE: u32 = 8 * 1024;
```
mhatrevi, line 49: Do we need a DATA_SIZE separately defined?

clundin25: It's used by the linker script generator in https://github.com/chipsalliance/caliptra-sw/blob/main/cpu/gen/src/lib.rs#L24. 

IIRC there is some data that exists in the data section, e.g. the external vector interrupt table.

drivers/src/memory_layout.rs
clundin25: Similar comment to the data size, this is used by the linker script generator

drivers/src/memory_layout.rs
clundin25: Added a comment. Let me know if any more details are needed.

mhatrevi (approved):

jhand2 (approved):

PR #1724, opened 2024-10-18
Add DMA driver

ArthurHeymans:
This adds a few emulator fixes and implements a DMA driver


drivers/src/mailbox.rs
```diff
@@ -86,6 +86,13 @@ impl Mailbox {
         }
     }
 
+    pub fn raw_recv_txn(&mut self) -> MailboxRecvTxn {
```
mhatrevi, line 89: Remove this for now.

sw-emulator/app/src/main.rs
```diff
@@ -269,7 +269,6 @@ fn main() -> io::Result<()> {
 
     let bus_args = CaliptraRootBusArgs {
```
mhatrevi, line 270: Where do you take the argument that ROM is executing in ACTIVE vs PASSIVE mode?

drivers/src/dma.rs
```diff
@@ -0,0 +1,253 @@
...
+        self.flush();
+
+        let read_transaction = DmaReadTransaction {
+            read_addr,
+            fixed_addr,
+            length: payload_len_bytes,
+            target: DmaReadTarget::Mbox,
+        };
+        self.setup_dma_read(read_transaction);
+        self.dma
```
mhatrevi, line 246: We should add a description comment to the function stating that the client needs to take the mailbox lock before calling this function.

drivers/test-fw/src/bin/dma_tests.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // Get mailbox lock
+    let mut mbox_driver = unsafe { Mailbox::new(MboxCsr::new()) };
+    let mut txn = mbox_driver.try_start_send_txn().unwrap();
+    txn.send_request(0xdead_beef, b"").unwrap();
+
+    dma.transfer_payload_to_mbox(rri_regs, test_image.len() as u32, true, block_size)
+        .unwrap();
+
+    // TODO need a way to get DLEN and raw mailbox content from txn
```
mhatrevi, line 64: You should directly access the mailbox sram and compare that to the test_image.

sw-emulator/lib/periph/src/dma.rs
```diff
@@ -261,11 +263,11 @@ impl DmaRegs {
             .write_dlen(RvSize::Word, self.byte_count.reg.get())
             .unwrap();
 
-        assert_eq!(data.len(), self.byte_count.reg.get() as usize);
+        //        assert_eq!(data.len(), self.byte_count.reg.get() as usize);
```
mhatrevi, line 266: Why is this commented out?

mhatrevi (changes requested):
Please address the PR feedback.

sw-emulator/app/src/main.rs
ArthurHeymans: I remove it from the bus itself and in hw-model I have a separate way of uploading it. It's also possible to do this in sw-emulator app. I thought it would be appropriate to add the ACTIVE/PASSIVE things in a separate PR in there as it's not strictly related to a DMA driver

sw-emulator/lib/periph/src/dma.rs
ArthurHeymans: That's probably an accident.

drivers/test-fw/src/bin/dma_tests.rs
ArthurHeymans: This not possible with the current mailbox driver API. I added a function that would allow it but you previously requested to remove it.

drivers/src/dma.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    fn flush(&mut self) {
+        let dma = self.dma.regs_mut();
+
+        dma.ctrl().write(|c| c.flush(true));
+
+        // Wait till we're not busy and have no errors
+        while {
+            let status0 = dma.status0().read();
+            status0.busy() || status0.error()
```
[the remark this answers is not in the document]
ArthurHeymans: Done

drivers/test-fw/src/bin/dma_tests.rs
ArthurHeymans: Oh I see. It can ofc also be done without a driver but with raw reg access. I'll do that

mhatrevi: Exactly.

drivers/test-fw/src/bin/dma_tests.rs
```diff
@@ -0,0 +1,83 @@
...
+    let mut txn = mbox_driver.try_start_send_txn().unwrap();
+    txn.send_request(0xdead_beef, b"").unwrap();
+
+    dma.transfer_payload_to_mbox(rri_regs, test_image.len() as u32, true, block_size)
+        .unwrap();
+
+    let mbox = unsafe { MboxCsr::new() };
+    let dlen = mbox.regs().dlen().read();
+
+    assert_eq!(dlen, test_image.len() as u32);
```
mhatrevi, line 68: How is this assert succeeding? AFAIK, the mapx.dlen register shouldn't not be getting set when the payload is DMAed into the sram.

mhatrevi (changes requested):

drivers/test-fw/src/bin/dma_tests.rs
ArthurHeymans: The programming flowchart in the subsystem spec does not mention that about DLEN with regards to having the mailbox as target. Only the lock needs to be aquired.

ArthurHeymans: I assumed that this would be the case. I'll update the emulator and the test.

ArthurHeymans: So the mailbox emulator state machine makes this necessary (first command and dlen). I assumed something similar for the hardware.

ArthurHeymans: I removed this from the test. We don't need this.

mhatrevi (approved):

swenson (approved):
LGTM

PR #1725, opened 2024-10-18
Cert tweaks to names and TCB info

nquarton:


rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -252,6 +252,6 @@ impl FmcAliasLayer {
             flags |= dice::FLAG_BIT_DEBUG;
         }
 
-        flags.to_be_bytes()
+        flags.reverse_bits().to_be_bytes()
```
jhand2, line 255: Why do the bits need to be reversed? The constants seem to match the spec.

test/tests/caliptra_integration_tests/smoke_testdata/rt_alias_cert_redacted.txt
```diff
@@ -30,7 +30,7 @@ Certificate:
             2.23.133.5.4.4: 
                 0....................
             2.23.133.5.4.1: 
-                DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD
+                DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD
```
jhand2, line 33: What's the change that leads to the TCBInfo extension being shorter?

rom/dev/src/flow/cold_reset/fmc_alias.rs
nquarton: Our understanding for a DER encoded ASN1 bit string is that for the actual value (the flags) the most-significant-bit of the first byte transferred corresponds to bit 0 of the flags.

See this example: https://learn.microsoft.com/en-us/windows/win32/seccertenroll/about-bit-string

test/tests/caliptra_integration_tests/smoke_testdata/rt_alias_cert_redacted.txt
nquarton: This is the removal of Vendor="Caliptra" and Model="RT" from this info.

fdamato (approved):

rom/dev/src/flow/cold_reset/fmc_alias.rs
jhand2: Ah, you're right. Copying from our chat discussion

Looking at the ASN.1 spec: https://www.itu.int/ITU-T/studygroups/com17/languages/X.680-0207.pdf

Section 21.2
> The first bit in a bit string is called the leading bit. The final bit in a bit string is called the trailing bit.

Section 21.4
> The leading bit of the bit string is identified by the "number" zero, with succeeding bits having successive values.

jhand2 (approved):

rom/dev/src/flow/cold_reset/fmc_alias.rs
nquarton: Oh nice this is the point of truth we needed. I was only finding references that implied this. Thanks for getting to the bottom of this!

mhatrevi (approved):

JohnTraverAmd (approved):

PR #1727, opened 2024-10-21
Cherry pick commits from main [9/25-10/21]

mhatrevi:


rom/dev/tests/rom_integration_tests/test_mailbox_errors.rs
```diff
@@ -151,3 +151,27 @@ fn test_mailbox_invalid_req_size_zero() {
...
+
+#[test]
+// Changing PAUSER not supported on sw emulator
+#[cfg(any(feature = "verilator", feature = "fpga_realtime"))]
+fn test_mailbox_reserved_pauser() {
+    let (mut hw, _image_bundle) =
+        helpers::build_hw_model_and_image_bundle(Fuses::default(), ImageOptions::default());
+
+    // Set pauser to the reserved value
+    hw.set_apb_pauser(0xffffffff);
```
ArthurHeymans, line 163: set_axi_id

ArthurHeymans (approved):

swenson (approved):
LGTM

rusty1968 (approved):

error/src/lib.rs
```diff
@@ -500,6 +501,7 @@ impl CaliptraError {
     pub const FW_PROC_MAILBOX_PROCESS_FAILURE: CaliptraError = CaliptraError::new_const(0x01020007);
     pub const FW_PROC_MAILBOX_STASH_MEASUREMENT_MAX_LIMIT: CaliptraError =
         CaliptraError::new_const(0x01020008);
+    pub const FW_PROC_MAILBOX_RESERVED_PAUSER: CaliptraError = CaliptraError::new_const(0x01020009);
```
nquarton, line 504: We have the potential for some divergence on these error codes if we add them in each branch at the same time. Are we okay with the same error potentially having a slightly different code on 1.x vs 2.0? Or do we need to ensure they are consistent?

nquarton (approved):

jhand2 (approved):

zhalvorsen (approved):

error/src/lib.rs
korran: As I've mentioned before, branching many of these crates is a mistake.

korran:
Are we sure we want to do manual cherry-picks? If we're really going to do this sort of thing, maybe we should turn off "linear history" and then use standard Git merges from main into this branch. At least then it can be tracked automatically...

error/src/lib.rs
mhatrevi: @nquarton We'll can do a sweep on main-2.x at a later point for the error codes and ensure they are in sync with main.

PR #1728, opened 2024-10-21
[feat] ROM changes to retrieve firmware image from Recovery Interface

mhatrevi:


mhatrevi:
This change is no longer needed as the updated changes are in https://github.com/chipsalliance/caliptra-sw/pull/1867

PR #1730, opened 2024-10-22
[doc] ROM README update for MLDSA DICE derivation

mhatrevi:


rom/dev/README.md
```diff
@@ -219,16 +222,25 @@ Both UDS and Field Entropy are available only during cold reset of Caliptra.
 - Keys Slot 0 - 31 are empty and Usage Bits are all cleared
 - PCR 0 - 31 are all cleared
 - Data Vault is all cleared
+- Dice Certificate mode is set to either “ECDSA-P-384 & ML-DSA-87” (aka Dual-Cert Mode) or “ECDSA-P-384 only” in the CPTRA_DICE_CERT_MODE register.
```
jhand2, line 225: nit: Let's name this `CPTRA_DICE_MODE` (drop the "cert"). Because this will also impact what DPE support is available.

rom/dev/README.md
```diff
@@ -242,50 +254,85 @@ Both UDS and Field Entropy are available only during cold reset of Caliptra.
...
+
+    `IDevIdTbsDigestMldsa = sha512_digest(IDevIdTbsMldsa)`
+
+    `IDevIdCertSigMldsa = mldsa87_sign(KvSlot10, IDevIdTbsDigestMldsa)`
+
+10. Verify the signature of IDevID `To Be Signed` Blob
+
+    `Result = mldsa87_verify(IDevIdPubKeyMldsa, IDevIdTbsDigestMldsa, IDevIdCertSigMldsa)`
+
+11. Upload the CSR(s) to mailbox and wait for JTAG to read the CSR out of the mailbox.
```
jhand2, line 335: nit: Don't need to solve in this PR, but we should add a structure for outputting multiple CSRs so the caller doesn't need to do ASN.1 parsing to separate them.

jhand2 (dismissed):
Looks good, just a couple small comments. Thanks!

rom/dev/README.md
```diff
@@ -162,6 +162,9 @@ The following sections define the various cryptographic primitives used by Calip
...
 |   | `ecc384_sign(priv_kv_slot, data) -> sig` | ECC384 signing operation<br>**Input**:<br>***priv_kv_slot*** - key vault slot to use a private key from<br>***data*** - data to sign<br>**Output**:<br>***sig*** - signature |
 | | `ecc384_verify(pub_key, data, sig) -> CaliptraResult<Array4xN<12, 48>>` | ECC384 verify operation<br>**Input**:<br>***pub-key*** -public key<br>data - data to verify<br>sig - signature<br>**Output**:<br>***Ecc384Result*** - verify.r value on success, else an error |
+| Module-Lattice-Based Digital Signature Algorithm | `mldsa87_keygen(seed_kv_slot) -> pub_key` | Generate MLDSA87 Key Pair.<br>**Input**:<br>***seed_key_slot*** - key vault slot to use as seed for key generation<br>**Output**:<br>***pub-key*** - public key associated with the private key |
+|   | `mldsa87_sign(seed_kv_slot, data) -> sig` | MLDSA87 signing operation<br>**Input**:<br>***seed_kv_slot*** - key vault slot to use a generate a private key from for signing<br>***data*** - data to sign<br>**Output**:<br>***sig*** - signature |
```
bluegate010, line 166: Grammar: "key vault slot to use _to_ generate a private key"

Though I suggest we just make this row consistent with the one prior: "key vault slot to use as a seed for key generation", since that's what's being done under the hood.

rom/dev/README.md
```diff
@@ -242,50 +254,85 @@ Both UDS and Field Entropy are available only during cold reset of Caliptra.
 
 | Slot | Key Vault | PCR Bank | Data Vault 48 Byte (Sticky) | Data Vault 4 Byte (Sticky) |
 |------|-----------|----------|-----------------------------|----------------------------|
-| 0 | UDS (48 bytes) | | | |
-| 1 | Field Entropy (32 bytes) | | | |
+| 0 | UDS for ECDSA (48 bytes) | | | |
+| 1 | Field Entropy for ECDSA (32 bytes) | | | |
+| 10 | Conditioned UDS for MLDSA | | | |
```
bluegate010, line 259: I don't think we need to allocate extra KV slots for MLDSA-specific seeds. Every time we use a seed to generate an ECDSA key, we first condition it with a KDF and a context string. For example see [here](https://github.com/chipsalliance/caliptra-sw/blob/f102017b28c27642c50ad04a748d13e05dc35655/rom/dev/src/flow/cold_reset/idev_id.rs#L189) for IDevID generation. The conditioned values are stored in `KEY_ID_TMP` and then used to derive the asymmetric key.

We can simply pass a different context string when conditioning the UDS/CDIs for MLDSA generation.

rom/dev/README.md
mhatrevi: Updated.

rom/dev/README.md
```diff
@@ -242,50 +254,85 @@ Both UDS and Field Entropy are available only during cold reset of Caliptra.
...
 **Pre-conditions:**
 
-- UDS is loaded in Key Vault Slot 0
+- UDS for ECDSA is in Key Vault Slot 0
+- If Dual-Cert Mode is enabled, conditioned UDS for MLDSA is in Key Vault Slot 10
 
 **Actions:**
 
-1. Derive the CDI using ROM specified label and UDS in Slot 0 as data and store the resultant mac in KeySlot6
+1. Derive the ECDSA CDI using ROM specified label and UDS in Key Vault Slot 0 as data and store the resultant mac in Key Vault Slot 6
```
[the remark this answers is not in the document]
mhatrevi: Updated

rom/dev/README.md
```diff
@@ -242,50 +254,85 @@ Both UDS and Field Entropy are available only during cold reset of Caliptra.
...
 
     `hmac384_kdf(KvSlot0, b"idevid_cdi", KvSlot6)`
 
-2. Clear the UDS in key vault
+    If Dual-Cert Mode is enabled, use the conditioned UDS to derive the MLDSA CDI store the resultant mac in Key Vault Slot 12
```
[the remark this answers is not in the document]
mhatrevi: Fixed

rom/dev/README.md
mhatrevi: We do pass a different string when creating the seed for MLDSA key generation as noted below:

"Derive the ECDSA CDI using ROM specified label and UDS in Key Vault Slot 0 as data and store the resultant MAC in Key Vault Slot 6

hmac384_kdf(KvSlot0, b"idevid_cdi", KvSlot6)

If Dual-Cert Mode is enabled, use the conditioned UDS to derive the MLDSA CDI and store the resultant MAC in Key Vault Slot 12

hmac512_kdf(KvSlot10, b"idevid_mldsa_cdi", KvSlot12)"

Have a reconditioned UDS/FE was something that was discussed and agreed upon on the FW sync. Do you think using the same UDS/FE and just having separate strings for CDI (seed) generation (via KDF) is sufficient from a security standpoint?

rom/dev/README.md
mhatrevi: Ack. I'll document the CSR format as part of the CSR envelop signing doc update PR.

rom/dev/README.md
mhatrevi: Updated.

rom/dev/README.md
bluegate010: > Have a reconditioned UDS/FE was something that was discussed and agreed upon on the FW sync.

Ah sorry I wasn't around for that discussion.

> Do you think using the same UDS/FE and just having separate strings for CDI (seed) generation (via KDF) is sufficient from a security standpoint?

Yes. The CDI tree is algorithm-agnostic; it's just a sequence of secret seeds. Before any such seed is used for any purpose such as ECC keygen, the seed is run through a KDF. So we can keep the CDI tree the same and just tweak the KDF labels used at the leaves when creating asymmetric keys.

mhatrevi: Ok, I'll address this as a separate update to the document.

vsonims (approved):

Issue #1687, opened 2024-09-20
RFC: If requested, persist IDevID CSR for later retrieval


jhand2:
Targeted Release: ROM 1.3
Impacted FW stages: ROM only

# Background

If the [CPTRA_DBG_MANUF_SERVICE_REG](https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_DBG_MANUF_SERVICE_REG) `MFG_FLAG_GEN_IDEV_CSR` bit is set, Caliptra ROM will generate/sign an IDevID CSR and write it to the mailbox. The SoC Manager must retrieve the CSR from the Caliptra mailbox before it can do anything else (e.g. call `STASH_MEASUREMENT`).

# Proposal

The general expectation is that this flow is being done by SoC ROM. Instead of making SoC ROM deal with exporting the IDevID CSR, we should provide an option to retrieve the CSR at a later point. Proposal to achieve this:

1. If `MFG_FLAG_GEN_IDEV_CSR` is set, write IDevID CSR to temporary DCCM (can be cleared before ROM jumps to FMC)
2. Add ROM mailbox command to retrieve IDevID CSR

This allows SoC manager to retrieve the CSR at any point until it boots Caliptra mutable FW. Notably, it can stash measurements and load its mutable manufacturing fw which can deal with the CSR.

If CSR request bit was not set prior to ROM boot, this mailbox command will fail.

# (Optional) Variation

There is a small change we could make where we persist the IDevID CSR in DCCM, add it's offset/size to the FHT, and add a mailbox command to RT to retrieve it. This is a bit more flexible and allows the SoC to retrieve the CSR at any time. But it seems unnecessary. If SoC manager is fetching/booting Caliptra firmware, surely it can handle storing the CSR somewhere.

The goal of this proposal is just to allow SoC ROM to be as simple as possible.

jhand2:
cc @mhatrevi @nquarton @varuns-nvidia let me know if you have any thoughts

PR #1732, opened 2024-10-23
Add command to retrieve IDevID CSR from persistent storage.

clundin25:
* This resolves https://github.com/chipsalliance/caliptra-sw/issues/1687, for ROM.

### ROM Backwards compatibility

If the runtime is used with a ROM that does not support CSR generation, `RUNTIME_GET_IDEV_ID_UNSUPPORTED_ROM` is returned as an error.

This was tested manually by pulling in the ROM checked into https://github.com/chipsalliance/caliptra-sw/releases/tag/release_v20240719_1 and running the `get_idev_csr` test suite.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -51,6 +51,9 @@ impl CommandId {
 
     // The authorize and stash command.
     pub const AUTHORIZE_AND_STASH: Self = Self(0x4154_5348); // "ATSH"
+
+    // The get IDevID CSR command.
+    pub const GET_IDEV_CSR: Self = Self(0x4944_4352); // "IDCR"
 }
 
 impl From<u32> for CommandId {
@@ -151,6 +154,7 @@ pub enum MailboxResp {
     QuotePcrs(QuotePcrsResp),
     CertifyKeyExtended(CertifyKeyExtendedResp),
     AuthorizeAndStash(AuthorizeAndStashResp),
+    GetIdevIdCsr(GetIdevIdCsrResp),
 }
 
 impl MailboxResp {
@@ -171,6 +175,7 @@ impl MailboxResp {
             MailboxResp::QuotePcrs(resp) => Ok(resp.as_bytes()),
             MailboxResp::CertifyKeyExtended(resp) => Ok(resp.as_bytes()),
             MailboxResp::AuthorizeAndStash(resp) => Ok(resp.as_bytes()),
+            MailboxResp::GetIdevIdCsr(resp) => Ok(resp.as_bytes()),
         }
     }
 
@@ -191,6 +196,7 @@ impl MailboxResp {
             MailboxResp::QuotePcrs(resp) => Ok(resp.as_bytes_mut()),
             MailboxResp::CertifyKeyExtended(resp) => Ok(resp.as_bytes_mut()),
             MailboxResp::AuthorizeAndStash(resp) => Ok(resp.as_bytes_mut()),
+            MailboxResp::GetIdevIdCsr(resp) => Ok(resp.as_bytes_mut()),
         }
     }
 
@@ -458,6 +464,7 @@ pub struct GetIdevInfoResp {
 pub struct GetLdevCertReq {
     header: MailboxReqHeader,
 }
+
 impl Request for GetLdevCertReq {
     const ID: CommandId = CommandId::GET_LDEV_CERT;
     type Resp = GetLdevCertResp;
@@ -972,6 +979,40 @@ impl Default for SetAuthManifestReq {
             hdr: MailboxReqHeader::default(),
             manifest_size: 0,
             manifest: [0u8; SetAuthManifestReq::MAX_MAN_SIZE],
+        }
+    }
+}
+
+// GET_IDEVID_CSR
+#[repr(C)]
+#[derive(Default, Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIdevIdCsrReq {
+    pub hdr: MailboxReqHeader,
+}
+
+impl Request for GetIdevIdCsrReq {
+    const ID: CommandId = CommandId::GET_IDEV_CSR;
+    type Resp = GetIdevIdCsrResp;
+}
+
+#[repr(C)]
+#[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIdevIdCsrResp {
+    pub hdr: MailboxRespHeader,
+    pub data_size: u32,
+    pub data: [u8; Self::DATA_MAX_SIZE],
+}
+impl GetIdevIdCsrResp {
+    pub const DATA_MAX_SIZE: usize = 512;
+}
+impl ResponseVarSize for GetIdevIdCsrResp {}
+
+impl Default for GetIdevIdCsrResp {
+    fn default() -> Self {
+        Self {
+            hdr: MailboxRespHeader::default(),
+            data_size: 0,
+            data: [0u8; Self::DATA_MAX_SIZE],
         }
     }
 }
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -7,6 +7,7 @@ use caliptra_auth_man_types::{
     AuthManifestImageMetadata, AuthManifestImageMetadataCollection,
     AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT,
 };
+use caliptra_error::{CaliptraError, CaliptraResult};
 use caliptra_image_types::ImageManifest;
 #[cfg(feature = "runtime")]
 use dpe::{DpeInstance, U8Bool, MAX_HANDLES};
@@ -23,6 +24,7 @@ use crate::{
 #[cfg(feature = "runtime")]
 use crate::pcr_reset::PcrResetCounter;
 
+pub const MAX_CSR_SIZE: usize = 512;
 pub const PCR_LOG_MAX_COUNT: usize = 17;
 pub const FUSE_LOG_MAX_COUNT: usize = 62;
 pub const MEASUREMENT_MAX_COUNT: usize = 8;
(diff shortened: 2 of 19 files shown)


runtime/README.md
```diff
@@ -794,6 +794,25 @@ Command Code: `0x4154_5348` ("ATSH")
...
+| chksum      | u32      | Checksum over other input arguments, computed by the caller. Little endian.  |
+
+*Table: `GET_IDEVID_CSR` output arguments*
+| **Name**      | **Type** | **Description**
+| --------      | -------- | ---------------
+| chksum        | u32      | Checksum over other output arguments, computed by Caliptra. Little endian. |
+| data\_size    | u32      | Length in bytes of the valid data in the data field.                       |
+| data          | u8[...]  | DER-encoded IDevID certificate signing request.                            |
+
+**Note**: The `GENERATE_IDEVID_CSR` Manufacturing State Flag **MUST** have been set to generate a CSR. Otherwise this command will return an empty data buffer.
```
clundin25, line 814: @jhand2 do you prefer an error code to an empty data buffer? The  flag gets cleared once the CSR is sent over the mailbox, so we either need to persist that the flag was set or inspect the CSR buffer, and map an empty buffer to an error code. 

Let me know what your preference is.

clundin25:
@jhand2 I have a runtime implementation as well, but I need to refactor the runtime integration tests to allow setting the `GENERATE_CSR` flag.

mhatrevi (changes requested):
Discussion pending on whether a ROM change is necessary.

jhand2:
> Discussion pending on whether a ROM change is necessary.

https://github.com/chipsalliance/caliptra-sw/issues/1687 can't be done without a ROM change because no other code can wield IDevID to sign the CSR. But anyway, we can discuss the use case and if anyone is uncomfortable with the solution.

api/src/mailbox.rs
```diff
@@ -976,6 +983,40 @@ impl Default for SetAuthManifestReq {
     }
 }
 
+// GET_IDEVID_CSR
+#[repr(C)]
+#[derive(Default, Debug, AsBytes, FromBytes, PartialEq, Eq)]
+pub struct GetIDevIDCSRReq {
```
jhand2, line 989: minor nit: The other APIs don't capitalize the acronyms. e.g. `GetIdevIdCsrReq`

drivers/src/idevid_csr.rs
```diff
@@ -0,0 +1,49 @@
...
+use zerocopy::{AsBytes, FromBytes};
+use zeroize::Zeroize;
+
+use crate::memory_layout;
+
+pub const MAX_CSR_SIZE: usize = 512;
+
+#[derive(FromBytes, AsBytes, Zeroize)]
+#[repr(C)]
+pub struct IDevIDCsr {
```
jhand2, line 13: nit: Should this struct go in the persistent data structure since it's really only intended to be used there?

drivers/src/idevid_csr.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub csr: [u8; MAX_CSR_SIZE],
+    pub csr_len: u32,
```
jhand2, line 15: I'm not positive, but it might be better for alignment to put `csr_len` as the first field.

drivers/src/idevid_csr.rs
```diff
@@ -0,0 +1,49 @@
...
+use zeroize::Zeroize;
+
+use crate::memory_layout;
+
+pub const MAX_CSR_SIZE: usize = 512;
+
+#[derive(FromBytes, AsBytes, Zeroize)]
+#[repr(C)]
+pub struct IDevIDCsr {
+    pub csr: [u8; MAX_CSR_SIZE],
```
jhand2, line 14: If we're going to have a get function, should the fields be non-public?

hw-model/src/lib.rs
```diff
@@ -585,6 +585,15 @@ pub trait HwModel: SocManager {
...
             let mut cycles = 0;
             while !self.ready_for_fw() {
+                // If GENERATE_IDEVID_CSR was set then we need to clear cptra_dbg_manuf_service_reg
+                // once the CSR is ready to continue making progress.
+                //
+                // Generally the CSR should be read from the mailbox at this point, but to
+                // accommodate test cases that ignore the CSR mailbox, we will ignore it here.
+                if self.soc_ifc().cptra_flow_status().read().idevid_csr_ready() {
+                    self.soc_ifc().cptra_dbg_manuf_service_reg().write(|_| 0);
+                }
```
jhand2, line 595: For tests which set the bit, I wonder if they should be responsible for clearing it (even if they're not going to read the CSR). I think it makes the expected behavior when reading the test a bit less magic.

rom/dev/src/fht.rs
```diff
@@ -151,10 +151,7 @@ impl FhtDataStore {
 pub fn initialize_fht(env: &mut RomEnv) {
     let pdata = &env.persistent_data.get();
 
-    cprintln!(
-        "[fht] Storing FHT @ 0x{:08X}",
-        &pdata.fht as *const _ as usize
-    );
+    cprintln!("[fht] FHT @ 0x{:08X}", &pdata.fht as *const _ as usize);
```
jhand2, line 154: nit: Should we do a separate PR for cleaning up the log strings?

Another option would be to squash the actual feature into one commit and then split the log changes into a separate commit.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -306,6 +304,32 @@ impl FirmwareProcessor {
...
+                            /* CSR was never written to DCCM. This most likely means the
+                             * gen_idev_id_csr manufacturing flag was not set before booting into
+                             * ROM. */
+                            return Err(
+                                CaliptraError::FW_PROC_MAILBOX_GET_IDEV_CSR_UNPROVISIONED_CSR,
+                            );
+                        }
+
+                        resp.data_size = csr_persistent_mem.csr_len;
+                        resp.data[..resp.data_size as usize].copy_from_slice(csr);
```
jhand2, line 328: nit: Would be good to add a comment about why this is panic-free

rom/dev/src/flow/cold_reset/x509.rs
```diff
@@ -16,29 +16,6 @@ use crate::cprintln;
 use crate::rom_env::RomEnv;
 use caliptra_drivers::*;
 
-/// Wrapper to hold certificate buffer and length
-pub struct Certificate<'a, const LEN: usize> {
```
jhand2, line 20: Was this only used for the CSR

jhand2 (commented):
Make it about half way throught. Will take another look later today or tomorrow.

As a general comment, I think moving the log stuff to a separate PR would help make this one a lot smaller.

runtime/tests/runtime_integration_tests/test_lms.rs
```diff
@@ -838,7 +838,7 @@ fn test_lms_verify_cmd() {
 
 #[test]
 fn test_lms_verify_failure() {
-    let mut model = run_rt_test(None, None, None);
+    let mut model = run_rt_test(RuntimeTestArgs::default());
```
jhand2, line 841: Also moving this refactor to a separate PR might be a good idea. Since the CSR change touches ROM, I think it would be good to keep this PR a bit smaller.

clundin25: Opened https://github.com/chipsalliance/caliptra-sw/pull/1776

api/src/mailbox.rs
clundin25: Done. I'll wait to push these changes until the UART logs PR is merged.

drivers/src/idevid_csr.rs
clundin25: Done. I'll wait to push these changes until the UART logs PR is merged.

drivers/src/idevid_csr.rs
clundin25: I think so too (at least similar structs do). I've made this change and will include it in the next rev.

hw-model/src/lib.rs
clundin25: We chatted offline and decided to keep this. 

It's unfortunate that this coupled logic is buried but the other options involve a log of duplicate logic or large refactor.

rom/dev/src/flow/cold_reset/x509.rs
clundin25: That's right.

drivers/src/idevid_csr.rs
clundin25: Refactored this, but I will need to remove some more log statements to re-coop the bytes.

rom/dev/src/flow/cold_reset/fw_processor.rs
clundin25: Added! Will be in next rev

drivers/src/persistent.rs
```diff
@@ -44,6 +47,65 @@ pub type StashMeasurementArray = [MeasurementLogEntry; MEASUREMENT_MAX_COUNT];
 pub type AuthManifestImageMetadataList =
     [AuthManifestImageMetadata; AUTH_MANIFEST_IMAGE_METADATA_LIST_MAX_COUNT];
 
+#[derive(FromBytes, AsBytes, Zeroize)]
+#[repr(C)]
+pub struct IDevIDCsr {
```
jhand2, line 52: nit: Change capitalization to match the other struct

error/src/lib.rs
```diff
@@ -443,6 +443,10 @@ impl CaliptraError {
     pub const RUNTIME_CMD_RESERVED_PAUSER: CaliptraError = CaliptraError::new_const(0x000E004F);
     pub const RUNTIME_AUTH_AND_STASH_MEASUREMENT_DPE_ERROR: CaliptraError =
         CaliptraError::new_const(0x000E0050);
+    pub const RUNTIME_GET_IDEV_ID_UNPROVISIONED: CaliptraError =
+        CaliptraError::new_const(0x000E0051);
+    pub const RUNTIME_GET_IDEV_ID_UNSUPPORTED_ROM: CaliptraError =
```
jhand2, line 448: Maybe no action required, but can you double check that there's no overlap with the error codes in the `main-2.x` branch? Sam comment for all the errors in this file.

Ultimately I would like to pull the errors crate into a separate repo, but for now we may need to manually keep them in sync.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -304,6 +304,32 @@ impl FirmwareProcessor {
...
                     }
+                    CommandId::GET_IDV_CSR => {
+                        let mut request = MailboxReqHeader::default();
+                        Self::copy_req_verify_chksum(&mut txn, request.as_bytes_mut())?;
+
+                        let csr_persistent_mem = &persistent_data.idevid_csr;
+                        let mut resp = GetIdevIdCsrResp::default();
+
+                        if csr_persistent_mem.is_unprovisioned() {
+                            /* CSR was never written to DCCM. This most likely means the
```
jhand2, line 315: nit: are there other reasons it wouldn't be populated? if not we may want to make the language in the comment stronger.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -264,37 +262,49 @@ impl InitDevIdLayer {
...
+        let mut result = Self::send_csr(env, &dev_id_csr);
+        if result.is_ok() {
+            result = Self::write_csr_to_peristent_storage(env, dev_id_csr);
+        }
+        csr_buf.zeroize();
 
         result
     }
 
+    fn write_csr_to_peristent_storage(env: &mut RomEnv, csr: IDevIDCsr) -> CaliptraResult<()> {
```
jhand2, line 294: nit: are we confident this won't make a copy on the stack?

runtime/src/get_idv_csr.rs
jhand2, line 1: nit: let's spell out `idev` in the file name

runtime/src/lib.rs
```diff
@@ -226,6 +228,7 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         CommandId::SHUTDOWN => FipsShutdownCmd::execute(drivers),
         CommandId::SET_AUTH_MANIFEST => SetAuthManifestCmd::execute(drivers, cmd_bytes),
         CommandId::AUTHORIZE_AND_STASH => AuthorizeAndStashCmd::execute(drivers, cmd_bytes),
+        CommandId::GET_IDV_CSR => GetIdevIdCsrCmd::execute(drivers, cmd_bytes),
```
jhand2, line 231: Same, I think it would be clearer to spell out idev

runtime/tests/runtime_integration_tests/test_get_idev_csr.rs
```diff
@@ -0,0 +1,61 @@
...
+use caliptra_runtime::RtBootStatus;
+use openssl::x509::X509Req;
+use zerocopy::{AsBytes, FromBytes};
+
+use crate::common::{run_rt_test, RuntimeTestArgs};
+
+#[test]
+fn test_get_csr() {
+    let args = RuntimeTestArgs {
+        test_mfg_flags: Some(MfgFlags::GENERATE_IDEVID_CSR),
```
jhand2, line 17: nit: might be worth a comment that `run_rt_test` will clear this flag to allow RT to boot.

error/src/lib.rs
clundin25: These error messages are not yet taken in `main-2.x`, so it should be okay assuming there is no open PR using them.

jhand2 (approved):

rom/dev/src/flow/cold_reset/idev_id.rs
clundin25: Changed to a Ref + Clone instead of the move.

We can follow up with a PR to assembly the CSR directly in the PersistentData memory slice if we want to save the stack space.

jhand2 (dismissed):

mhatrevi (approved):

jhand2 (approved):

PR #1733, opened 2024-10-23
[feat] ROM IDEVID DICE changes for MLDSA support

mhatrevi:
This change adds MLDSA support to the IDEVID DICE derviation. Generation of MLDSA CSR is not included in this change and will come later.


drivers/src/memory_layout.rs
```diff
@@ -63,6 +67,7 @@ pub const ROM_DATA_SIZE: u32 = 996;
 pub const MAN1_SIZE: u32 = 6 * 1024;
 pub const MAN2_SIZE: u32 = 6 * 1024;
 pub const FHT_SIZE: u32 = 2 * 1024;
+pub const IDEVID_MLDSA_PUB_KEY_MAX_SIZE: u32 = 3 * 1024;
```
swenson, line 70: Are we rounding to 3K for alignment? I think the key size is always exactly 2592 bytes, which is 32-byte aligned at least.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -36,6 +40,15 @@ const DOE_IV: Array4x4 = Array4xN::<4, 16>([0xfb10365b, 0xa1179741, 0xfba193a1,
 /// Maximum Certificate Signing Request Size
 const MAX_CSR_SIZE: usize = 512;
 
+const DUAL_CERT_MODE: bool = true;
+
+struct MLdsaKeyVaultInfo {
```
swenson, line 45: Nit: I think we have been using `Mldsa` in most other places (with the `L` lowercase), though for the public key we use `MlDsa`. It would be nice if we were consistent.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -36,6 +40,15 @@ const DOE_IV: Array4x4 = Array4xN::<4, 16>([0xfb10365b, 0xa1179741, 0xfba193a1,
 /// Maximum Certificate Signing Request Size
 const MAX_CSR_SIZE: usize = 512;
 
+const DUAL_CERT_MODE: bool = true;
```
swenson, line 43: Do we need to add a TODO for this to be read from a fuse or elsewhere?

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -118,9 +167,20 @@ impl InitDevIdLayer {
...
+        uds_mldsa: Option<KeyId>,
+    ) -> CaliptraResult<()> {
         // Engage the Deobfuscation Engine to decrypt the UDS
-        env.doe.decrypt_uds(&DOE_IV, uds)?;
+        env.doe.decrypt_uds(&DOE_IV, uds_ecdsa)?;
+
+        // If Dual-Cert Mode is enabled, recondition the UDS for MLDSA.
+        // [TODO] Use hmac512; set tag usage as seed for mldsa key pair generation.
+        if let Some(uds_mldsa) = uds_mldsa {
+            Crypto::hmac384_mac(env, uds_ecdsa, &b"uds_recon".into(), uds_mldsa)?;
```
swenson, line 181: Is there a reason we use HMAC instead of HKDF for this and the FE below?

rom/dev/src/flow/cold_reset/crypto.rs
```diff
@@ -227,4 +227,35 @@ impl Crypto {
         digest.0.zeroize();
         result
     }
+
+    /// Generate ECC Key Pair
```
mhatrevi, line 231: Update this to MLDSA Public Key.

common/src/keyids.rs
```diff
@@ -14,17 +14,25 @@ Abstract:
 use caliptra_drivers::KeyId;
 
 #[cfg(feature = "rom")]
-pub const KEY_ID_UDS: KeyId = KeyId::KeyId0;
+pub const KEY_ID_ECDSA_UDS: KeyId = KeyId::KeyId0;
```
bluegate010, line 17: As noted in https://github.com/chipsalliance/caliptra-sw/pull/1730, we do not need to allocate any extra key IDs for algorithm agility. We can condition the CDI differently when we turn it into a key, because we always pass it through a KDF first.

drivers/src/hand_off.rs
```diff
@@ -280,8 +280,11 @@ pub struct FirmwareHandoffTable {
     /// Index of LdevId Certificate Signature S Component in the Data Vault.
     pub ldevid_cert_sig_s_dv_hdl: HandOffDataHandle,
 
-    /// IDevID public key
-    pub idev_dice_pub_key: Ecc384PubKey,
+    /// IDevID ECDSA public key
+    pub idev_dice_ecdsa_pub_key: Ecc384PubKey,
+
+    /// IDevID MLDSA public key address in DCCM
+    pub idev_dice_mldsa_pub_key_load_addr: u32,
```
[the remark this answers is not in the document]
jhand2: Since the FHT is a fixed structure that's shared between fw layers, I think we should stick with fixed-size types. If we need to change to u64 in the future we'll have to make that an explicit decision and make sure all fw layers handle it properly.

drivers/src/hand_off.rs
```diff
@@ -470,14 +478,14 @@ mod tests {
     use super::*;
     use core::mem;
     const FHT_SIZE: usize = 2048;
-    const KEY_ID_FMC_PRIV_KEY: KeyId = KeyId::KeyId5;
+    const KEY_ID_FMC_ECDSA_PRIV_KEY: KeyId = KeyId::KeyId5;
```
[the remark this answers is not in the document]
jhand2: Probably easier to just use the keyids.rs definitions in the tests

drivers/src/memory_layout.rs
```diff
@@ -29,27 +32,28 @@ pub const ROM_DATA_ORG: u32 = 0x50000000;
...
+pub const IDEVID_MLDSA_PUB_KEY_ORG: u32 = FHT_ORG + FHT_SIZE;
+pub const LDEVID_TBS_ORG: u32 = IDEVID_MLDSA_PUB_KEY_ORG + IDEVID_MLDSA_PUB_KEY_MAX_SIZE;
+pub const FMCALIAS_TBS_ORG: u32 = LDEVID_TBS_ORG + LDEVID_TBS_SIZE;
+pub const RTALIAS_TBS_ORG: u32 = FMCALIAS_TBS_ORG + FMCALIAS_TBS_SIZE;
+pub const PCR_LOG_ORG: u32 = RTALIAS_TBS_ORG + RTALIAS_TBS_SIZE;
+pub const MEASUREMENT_LOG_ORG: u32 = PCR_LOG_ORG + PCR_LOG_SIZE;
+pub const FUSE_LOG_ORG: u32 = MEASUREMENT_LOG_ORG + MEASUREMENT_LOG_SIZE;
+pub const DPE_ORG: u32 = FUSE_LOG_ORG + FUSE_LOG_SIZE;
+pub const PCR_RESET_COUNTER_ORG: u32 = DPE_ORG + DPE_SIZE;
+pub const AUTH_MAN_IMAGE_METADATA_LIST_ORG: u32 = PCR_RESET_COUNTER_ORG + PCR_RESET_COUNTER_SIZE;
```
jhand2, line 46: (Not for this PR) Just as an FYI, we're planning to make an update in main to just have a `PERSISTENT_ORG` address and reserve a big chunk of memory. And then it will up to the PersistentData driver to decide how to allocate that memory. I think that will be easier to maintain than needing to update it both there and in this file.

drivers/src/memory_layout.rs
jhand2: See comment above. I think the 1K alignment was just to make maintaining this file easier. But I think it's prefferable to just use the PersistentData struct. And then we can just keep things word-aligned and call it good.

But we haven't merged that change in `main` yet because we're running into a couple test failures. So I'm ok if we want to just do this for now and change it to the new model in a future PR.

fmc/README.md
```diff
@@ -128,7 +128,8 @@ fields may not be changed or removed). Table revisions with different Major Vers
 | rt_dice_sign          | 96           | FMC        | RT Alias DICE signature.                                                                                 |
 | ldevid_cert_sig_r_dv_hdl | 4         | ROM        | Handle of LDevId Certificate Signature R Component in the Data Vault.                                    |
 | ldevid_cert_sig_s_dv_hdl | 4         | ROM        | Handle of LDevId Certificate Signature S Component in the Data Vault.                                    |
-| idev_dice_pub_key     | 96           | ROM        | Initial Device ID Public Key.                                                                            |
+| idev_dice_ecdsa_pub_key     | 96           | ROM        | Initial Device ID ECDSA Public Key.                                                                            |
+| idev_dice_mlddsa_pub_key_addr     | 4        | ROM        | Address of Initial Device ID MLDSA Public Key.                                                                            |
```
jhand2, line 132: I'm not sure how necessary it is to pass the address in the FHT if we're going to use the PersistentData driver.

I think that driver really needs to be kept in backwards compatible with all same-version ROMs anyway (e.g. 2.2 RT cannot overwrite some region that was used by 2.0 ROM).

rom/dev/src/flow/cold_reset/idev_id.rs
jhand2: I think we'll want it to be read from a register, but +1

mhatrevi: Resolving as this no longer applies.

drivers/src/memory_layout.rs
```diff
@@ -29,27 +32,28 @@ pub const ROM_DATA_ORG: u32 = 0x50000000;
...
-
-pub const STACK_ORG: u32 = 0x5001A000;
-pub const ROM_STACK_ORG: u32 = 0x5001C000;
-
-pub const ESTACK_ORG: u32 = 0x5001F800;
-pub const ROM_ESTACK_ORG: u32 = 0x5001F800;
-
-pub const NSTACK_ORG: u32 = 0x5001FC00;
-pub const ROM_NSTACK_ORG: u32 = 0x5001FC00;
+pub const MAN2_ORG: u32 = MAN1_ORG + MAN1_SIZE;
```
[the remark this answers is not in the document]
mhatrevi: Every time there is a change to one of the structs, we have to manually adjust the addresses. This way, it's more robust. I don't see a point in having tests that are checking against a hard-coded numeric value.

drivers/src/memory_layout.rs
mhatrevi: @swenson 3K alignment is for easier maintenance. We'll switch to the new model with Jordan's change.

common/src/keyids.rs
mhatrevi: With the new design, there is a single UDS for deriving the CDI.

drivers/src/memory_layout.rs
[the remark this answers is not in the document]
mhatrevi: Correct.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -75,22 +81,35 @@ impl InitDevIdLayer {
...
         let output = DiceOutput {
-            subj_key_pair: key_pair,
-            subj_sn,
-            subj_key_id,
+            ecc_subj_key_pair: ecc_key_pair,
+            ecc_subj_sn,
+            ecc_subj_key_id,
+            mldsa_subj_key_id: [0; 20],
+            mldsa_subj_key_pair: MlDsaKeyPair {
+                key_pair_seed: KEY_ID_IDEVID_MLDSA_KEYPAIR_SEED,
```
bluegate010, line 109: Nit: it's a bit inconsistent that `Self::derive_key_pair()` returns a tuple where the ECC key is a complete key pair, but the ML-DSA key is only the public portion, requiring us to make sure that the key ID passed here is the same key ID passed to `Self::derive_key_pair()` above. Instead of that, we could just have `Self::derive_key_pair()` returned a tuple of type `(Ecc384KeyPair, MlDsaKeyPair)`.

(And similarly, `mldsa_key_gen()` could be updated to return a full `MlDsaKeyPair` instead of just the public portion.)

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -185,16 +208,22 @@ impl InitDevIdLayer {
...
             cfi_assert!(result.is_ok());
             report_boot_status(IDevIdKeyPairDerivationComplete.into());
         } else {
             cfi_assert!(result.is_err());
         }
-        result
+        let ecc_keypair = result?;
+
+        // Derive the MLDSA Key Pair.
+        let mldsa_pub_key =
```
bluegate010, line 224: For consistency we should move the ML-DSA keygen to be before `report_boot_status(IDevIdKeyPairDerivationComplete.into())`, since IDevID keypair derivation is not complete until we've done both.

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -62,23 +66,30 @@ impl LocalDevIdLayer {
...
-            subj_key_pair: key_pair,
-            subj_sn,
-            subj_key_id,
+            ecc_subj_key_pair: key_pair,
+            ecc_subj_sn,
+            ecc_subj_key_id,
+            mldsa_subj_key_id: [0; 20],
+            mldsa_subj_key_pair: MlDsaKeyPair {
+                key_pair_seed: KEY_ID_LDEVID_MLDSA_KEYPAIR_SEED,
+                pub_key: Default::default(),
```
bluegate010, line 90: We should actually generate an ML-DSA key and place its public portion here, right?

test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -215,8 +215,8 @@ fn smoke_test() {
     let ldev_cert_txt = String::from_utf8(ldev_cert.to_text().unwrap()).unwrap();
 
     // To update the ldev cert testdata:
-    // std::fs::write("tests/smoke_testdata/ldevid_cert.txt", &ldev_cert_txt).unwrap();
-    // std::fs::write("tests/smoke_testdata/ldevid_cert.der", ldev_cert_der).unwrap();
+    // std::fs::write("tests/caliptra_integration_tests/smoke_testdata/ldevid_cert.txt", &ldev_cert_txt).unwrap();
+    // std::fs::write("tests/caliptra_integration_tests/smoke_testdata/ldevid_cert.der", ldev_cert_der).unwrap();
 
     assert_eq!(
```
bluegate010, line 221: We'll also want new .txt / .der files to let us do a smoke test on the ML-DSA certificates.

rom/dev/src/flow/cold_reset/dice.rs
```diff
@@ -32,12 +32,21 @@ pub struct DiceInput<'a> {
...
+    /// ECC Subject Serial Number
+    pub ecc_subj_sn: [u8; 64],
 
-    /// Subject Key Identifier
-    pub subj_key_id: [u8; 20],
+    /// ECC Subject Key Identifier
+    pub ecc_subj_key_id: [u8; 20],
+
+    /// MLDSA Subject key pair for this layer
+    pub mldsa_subj_key_pair: MlDsaKeyPair,
```
bluegate010, line 45: I gather that code needed to generate certs for these keys will come in a follow-up PR.

test/tests/caliptra_integration_tests/smoke_test.rs
mhatrevi: That's correct. We'll be working on that this/next week.

drivers/src/hand_off.rs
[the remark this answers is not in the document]
mhatrevi: The definitions are behind a conditional feature in keyid.rs. I updated the test value.

drivers/src/memory_layout.rs
mhatrevi: Ack.

fmc/README.md
mhatrevi: Let's keep this consistent with how other info is shared (e.g. TBS). We can discuss removing as a separate change.

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Udpated.

rom/dev/src/flow/cold_reset/ldev_id.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: UDS/FE reconditioning is removed and this no longer applies.

drivers/src/memory_layout.rs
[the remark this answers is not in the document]
korran: The reason the tests were comparing to hard-coded values was to confirm that a previous refactoring didn't change the values and break binary compatibility. Since 2.x is moving everything around anyways, I think this is ok. In fact, we can probably get rid of all these constants and just write a goldendata test to confirm that things don't move around unexpectedly during future refactorings.

bluegate010 (dismissed):

drivers/src/memory_layout.rs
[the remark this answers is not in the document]
mhatrevi: We'll address this as a separate change.

bluegate010 (approved):

vsonims (approved):

PR #1734, opened 2024-10-24
[feat] x509 MLDSA Certificate and CSR 

ArthurHeymans:
Add generic functions to encode DER unsigned integers.

### DIFF
--- a/x509/src/der_helper.rs
+++ b/x509/src/der_helper.rs
@@ -1,0 +1,101 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    der_helper.rs
+
+Abstract:
+
+    Helpers for encoding DER unsigned integers
+
+--*/
+
+/// DER Integer Tag
+const DER_INTEGER_TAG: u8 = 0x02;
+
+#[inline(never)]
+fn trim_leading_zeros(val: &[u8]) -> &[u8] {
+    // Count the leading zeros
+    for i in 0..val.len() {
+        if val[i] != 0 {
+            return &val[i..];
+        }
+    }
+    // If everything is 0, then we need len 1, and 0 as value
+    &val[0..1] // single 0
+}
+
+#[inline(never)]
+fn encode_length(val: &[u8]) -> usize {
+    for i in 0..val.len() {
+        if val[i] != 0 {
+            return val.len() - i + (val[i] >> 7) as usize;
+        }
+    }
+    return 1;
+}
+
+/// Compute len of DER encoding of an unsinged integer
+#[inline(never)]
+pub fn der_uint_len(val: &[u8]) -> Option<usize> {
+    let encode_length = encode_length(val);
+
+    let len_field_size = match encode_length {
+        0..=127 => 1,
+        128.. => trim_leading_zeros(&encode_length.to_be_bytes()).len(),
+        _ => None?,
+    };
+
+    // Tag + len + int
+    Some(1 + len_field_size + encode_length)
+}
+
+/// Encode a DER length
+#[inline(never)]
+pub fn der_encode_len(len: usize, buf: &mut [u8]) -> Option<usize> {
+    match len {
+        1..=127 => {
+            *buf.get_mut(0)? = len as u8;
+            Some(1)
+        }
+        128.. => {
+            let encode_len_be_bytes = len.to_be_bytes();
+            let len_in_be_bytes = trim_leading_zeros(&encode_len_be_bytes);
+            let len = len_in_be_bytes.len();
+            *buf.get_mut(0)? = 0x80 | (len as u8);
+            buf.get_mut(1..)?
+                .get_mut(..len)?
+                .copy_from_slice(len_in_be_bytes);
+            Some(len + 1)
+        }
+        _ => None?,
+    }
+}
+
+/// DER Encode unsigned integer
+#[inline(never)]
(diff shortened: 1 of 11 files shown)


swenson:
I have a patch to reduce this down another ~80 bytes or so by being careful with bounds checking and inlining, but it still doesn't come back down under 128 KB total:

```patch
​diff --git a/x509/src/der_helper.rs b/x509/src/der_helper.rs
index 841d2b50..629b7dbe 100644
--- a/x509/src/der_helper.rs
+++ b/x509/src/der_helper.rs
@@ -16,33 +16,34 @@ Abstract:
const DER_INTEGER_TAG: u8 = 0x02;

fn subslice_without_zero(val: &[u8]) -> &[u8] {
-    // Count the leading zeros
-    let clz = val
-        .iter()
-        .enumerate()
-        .find(|(_, byte)| **byte != 0)
-        .map_or(val.len(), |(idx, _)| idx);
-
+    for i in 0..val.len() {
+        if val[i] != 0 {
+            return &val[i..];
+        }
+    }
     // If everything is 0, then we need len 1, and 0 as value
-    val.get(clz..).unwrap_or(&[0])
+    &val[0..1] // single 0
}

+#[inline(never)]
fn encode_lenght(val: &[u8]) -> usize {
-    // Remove leading 0
-    let sub_val = subslice_without_zero(val);
-
-    let msb_set = *sub_val.get(0).unwrap_or(&0) > 127;
-
-    sub_val.len() + if msb_set { 1 } else { 0 }
+    for i in 0..val.len() {
+        if val[i] != 0 {
+            return val.len() - i + (val[i] >> 7) as usize;
+        }
+    }
+    return 1;
}

/// Length of DER encoded integer
+#[inline(never)]
pub fn der_uint_len(val: &[u8]) -> Option<usize> {
     let encode_length = encode_lenght(val);

     let len_field_size = match encode_length {
         0..=127 => 1,
-        128.. => subslice_without_zero(&encode_length.to_be_bytes()).len(),
+        128..=255 => 2,
+        256.. => 3,
         _ => None?,
     };

@@ -51,27 +52,42 @@ pub fn der_uint_len(val: &[u8]) -> Option<usize> {
}

/// Encode a DER length
+/// Limitation: this only works with lengths <= 65535, which should be sufficient
+/// for all of our certificates.
+#[inline(never)]
pub fn der_encode_len(len: usize, buf: &mut [u8]) -> Option<usize> {
+    debug_assert!(len <= 65535);
+    if buf.len() < 1 {
+        return None;
+    }
     match len {
         1..=127 => {
-            *buf.get_mut(0)? = len as u8;
+            buf[0] = len as u8;
             Some(1)
         }
-        128.. => {
-            let encode_len_be_bytes = len.to_be_bytes();
-            let len_in_be_bytes = subslice_without_zero(&encode_len_be_bytes);
-            let len = len_in_be_bytes.len();
-            *buf.get_mut(0)? = 0x80 | (len as u8);
-            buf.get_mut(1..)?
-                .get_mut(..len)?
-                .copy_from_slice(len_in_be_bytes);
-            Some(len + 1)
+        128..=255 => {
+            if buf.len() < 2 {
+                return None;
+            }
+            buf[0] = 0x81;
+            buf[1] = len as u8;
+            Some(2)
+        }
+        256.. => {
+            if buf.len() < 3 {
+                return None;
+            }
+            buf[0] = 0x82;
+            buf[1] = (len >> 8) as u8;
+            buf[2] = (len & 0xff) as u8;
+            Some(3)
         }
         _ => None?,
     }
}

/// DER Encode unsigned integer
+#[inline(never)]
pub fn der_encode_uint(val: &[u8], buf: &mut [u8]) -> Option<usize> {
     let mut pos = 0;
````

File: x509/src/der_helper.rs [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+fn subslice_without_zero(val: &[u8]) -> &[u8] {
+    // Count the leading zeros
+    let clz = val
+        .iter()
+        .enumerate()
+        .find(|(_, byte)| **byte != 0)
+        .map_or(val.len(), |(idx, _)| idx);
+
+    // If everything is 0, then we need len 1, and 0 as value
+    val.get(clz..).unwrap_or(&[0])
```
jhand2, line 28: Should this be `unwrap_or(&val[0]))`

Also of `val` is empty we may want to fail.

x509/src/der_helper.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/// Length of DER encoded integer
+#[inline(never)]
+pub fn der_uint_len(val: &[u8]) -> Option<usize> {
+    let encode_length = encode_lenght(val);
+
+    let len_field_size = match encode_length {
+        0..=127 => 1,
+        128.. => subslice_without_zero(&encode_length.to_be_bytes()).len(),
```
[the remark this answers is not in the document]
ArthurHeymans: This function is there to get the encoded size from a non encoded uint.

x509/src/der_helper.rs
ArthurHeymans: This is not new behavior. I suppose we don't have a use case for '0' encoding?

ArthurHeymans: test_rt_wdt_timeout fails if we don't encode 0 correctly.

jhand2 (approved):

mhatrevi (approved):

PR #1739, opened 2024-10-25
rdl/ureg: Offset and name fixes for I3C RDL files

swenson:
Automatically generated offsets were not being calculated correctly. If an offset for a register is missing, the RDL spec states that it should round to the next multiple of the register size (which is rounded up to the next power of 2). With this fix, the I3C register offsets match other RDL compiler's output.

This also fixes some TODOs in the code where registers were not being correctly enumerated if they were in sub-sub-blocks, and removes a related workaround (which was creating anonymous register types on-the-fly).

Finally, there are several register name collisions in the I3C Target Transaction Interface, which cause our RDL parser to panic. So, we add the ability to patch the RDL files when they are loaded so that we can rename the registers.

Since these problems only occur in I3C, this is only needed for Caliptra 2.0+.


hw/latest/registers/src/i3ccsr.rs
```diff
@@ -87,7 +88,7 @@ impl<TMmio: ureg::Mmio> RegisterBlock<TMmio> {
         }
     }
     #[inline(always)]
-    pub fn i3_c_ec(&self) -> I3cEcBlock<&TMmio> {
+    pub fn i3c_ec(&self) -> I3cEcBlock<&TMmio> {
```
ArthurHeymans, line 91: The sec_fw_recovery_if is still at 0 instead of i3c_ec + 0 offset.

swenson: Fixed!

mhatrevi (approved):

swenson:
Thanks!

PR #1745, opened 2024-10-28
[doc] Cryptographic Mailbox

swenson:
This adds sections describing the Cryptographic Mailbox system, its storage system, and related commands.

I cleaned up a little bit of the overview sections for what is new in the 1.1 and 2.0 runtime firmware to help keep things organized.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -163,6 +163,9 @@ All mailbox command codes are little endian.
 | `BAD_SIG`        | `0x4253_4947` ("BSIG") | Generic signature check failure (for crypto offload)
 | `BAD_IMAGE`      | `0x4249_4D47` ("BIMG") | Malformed input image
 | `BAD_CHKSUM`     | `0x4243_484B` ("BCHK") | Checksum check failed on input arguments
+| `CME_BAD_CMID`   | `0x434D_4249` ("CMBI") | Invalid CMID
+| `CME_BAD_CTXT`   | `0x434D_4243` ("CMBC") | Bad context
+| `CME_FULL`       | `0x434D_4546` ("CMEF") | Cryptographic Mailbox Full
 
 Relevant registers:
 
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -4,18 +4,18 @@ This specification describes the Caliptra Runtime Firmware.
 
 ## New in v1.1
 
-* [LMS Signature Verification](#lms_signature_verify-11)
+* [LMS Signature Verification](#lms_signature_verify-new-in-11)
 
 ## New in v1.2
 
-* [Manifest-Based Image Authorization](#manifest-based-image-authorization-12)
+* [Manifest-Based Image Authorization](#manifest-based-image-authorization-new-in-12)
 
 ## New in v2.0
 
 Caliptra 2.0 firmware supports passive mode (same as 1.x) and active (or susbystem) mode.
 
 * [MCU Runtime loading](#boot-and-initialization) (Active mode)
-* [Cryptographic mailbox commands](#cryptographic-mailbox-commands-20)
+* [Cryptographic mailbox commands](#cryptographic-mailbox-commands-new-in-20)
 
 ## Runtime Firmware environment
 
@@ -74,7 +74,178 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
 possible; however, it will have its own copies of all of these drivers linked into
 the Runtime Firmware binary.
 
-## Maibox commands
+## Manifest-Based Image Authorization (new in 1.2)
+
+Caliptra's goal is to enable integrators and vendors to meet standard security requirements for creating cryptographic identity and securely reporting measurements through DICE and DPE Certificate chains and Caliptra-owned private-public key pairs. In addition, Caliptra 1.0 provides an `ECDSA384_SIGNATURE_VERIFY` command to enable an SoC RoT to verify its own FW signatures so that it can develop an SoC secure boot using Caliptra cryptography. Caliptra 1.1 expanded the verify command to a PQC-safe `LMS_SIGNATURE_VERIFY` command. In each of these cases, it is left up to the vendor to ensure that they build a secure environment for introducing and verifying FW integrity and authenticity and then executing mutable FW.
+
+The Caliptra Measurement manifest feature expands on Caliptra-provided secure verifier abilities. The Measurement Manifest feature provides a standard Caliptra-supported definition to enable the following use cases for integrators, vendors, and owners.
+
+* Caliptra-Endorsed Aggregated Secure Boot
+* Caliptra-Endorsed Local Verifier
+
+Each of these abilities are tied to Caliptra Vendor and Owner FW signing keys and should be independent of any SoC RoT FW signing keys.
+
+Manifest-based image authorization is implemented via three mailbox commands: [`SET_AUTH_MANIFEST`](#set-auth-manifest), [`SET_IMAGE_METADATA`](#set-image-metadata), and [`AUTHORIZE_AND_STASH`](#authorize-and-stash).
+
+### Caliptra-Endorsed Aggregated Secure Boot
+
+Aggregated Secure Boot is a verified boot where one signed manifest attests to FW integrity of many different FW measurements. The authenticity of the FW is tied to the trust in the public key signing the measurement manifest, which is endorsed by the Caliptra Vendor and/or Owner FW Keys.
+
+### Caliptra-Endorsed Local Verifier
+
+A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest. In this case, the SoC RoT still has its own FW public-key chain that is verified by the SoC RoT, but in addition the SoC RoT introduces the Caliptra Measurement Manifest, which is endorsed by the Caliptra FW key pair. Caliptra provides approval or disapproval of the measurement of any FW back to the SoC RoT. This effectively provides a multi-factor authentication of SoC FW.
+
+The Caliptra-Endorsed Local Verify could be required by the owner only or both the vendor and the owner.
+
+The main difference between Caliptra-Endorsed Aggregated Secure Boot and Caliptra-Endorsed Local Verifier is if the SoC RoT is relying on the Measurement Manifest for SoC Secure Boot services as opposed as using it as an additional verification.
+
+### SoC RoT Enforcement of Measurement Manifest
+
+In both use cases, the SoC RoT chooses to provide the Caliptra Measurement Manifest and to enforce the result of the authorization. Caliptra 1.x is not capable of providing any enforcement of measurements for SoC FW execution.
+
+### Caliptra Measurement Manifest Signing Keys Authenticity
+
+Caliptra 1.0 and 1.1 do not put any requirements on how the SoC RoT ensures integrity and authenticity of SoC FW other than requiring the SoC RoT to provide a measurement to Caliptra of any SoC FW before execution. Caliptra Measurement Manifest enables the SoC RoT to perform the integrity check through Caliptra-authorized FW signing keys.
+
+### Unique Measurement Manifest Signing Keys
+
+In order to reduce usage of the Caliptra FW Signing keys, the measurement manifest will be signed by new key pairs: one for the owner and possibly one for the vendor. These new key pairs are endorsed once using a single signature within the Measurement Manifest, thus allowing the measurement manifest keys to be used independently of the Caliptra FW signing keys.
+
+### Caliptra Measurement Manifest Vendor Public Key Authenticity
+
+The Measurement Manifest MUST have an endorsement by the Caliptra Vendor Public Key. In order to fulfill this requirement, the Vendor has 2 options:
+
+* Vendor signing required: The Vendor creates a new Measurement keypair which will sign the measurement manifest and endorses the new public key with the Caliptra FW Vendor Private Key. The signature covers both the new public key as well as the flags field which indicates that the new Measurement Key Pair will be enforced.
+* Vendor signing **not** required: Vendor leaves the Vendor public key as all zeros, and clears the flag which enforces vendor signing and then endorses these fields with a signature in the Measurement Manifest. In this case, the Vendor releases ownership of enforcing any specific FW in execution.
+
+### Caliptra Measurement Manifest Owner Public Key Authenticity
+
+Caliptra will always verify the endorsement of the Measurement Manifest Owner Public key and require that it signed the measurement manifest.
+
+This feature is accomplished by having the SoC send a manifest to Caliptra Runtime through the `SET_AUTH_MANIFEST` mailbox command. The manifest will include a set of hashes for the different SoC images. Later, the SOC will ask for authorization for its images from the Caliptra Runtime through the `AUTHORIZE_AND_STASH` new mailbox commands. Caliptra Runtime will authorize the image based on whether its hash was contained in the manifest.
+
+#### Preamble
+
+The manifest begins with the Preamble section, which contains new manifest ECC and LMS public keys of the vendor and the owner. These public keys correspond to the private keys that sign the Image Metadata Collection (IMC) section. These signatures are included in the Preamble. The Caliptra firmware's private keys endorse the manifest's public keys and these endorsements (i.e., signatures) are part of the Preamble as well.
+
+#### Image Metadata Collection (IMC)
+
+The IMC is a collection of Image Metadata entries (IME). Each IME has a hash that matches one of the multiple SoC images. The manifest vendor and owner private keys sign the IMC. The Preamble holds the IMC signatures. The manifest IMC vendor signatures are optional and are validated only if the Flags field Bit 0 is set to 1. Up to 16 image hashes will be supported.
+
+#### Caliptra Measurement Manifest Keys Endorsement Verification Steps
+
+When Caliptra receives the Measurement Manifest, Caliptra will:
+
+* Verify the vendor endorsement using the Caliptra Vendor FW Public Key and compare with the vendor endorsement signature.
+* If the vendor endorsement is invalid, the `SET_AUTH_MANIFEST` command will be rejected.
+* If the vendor endorsement is valid, Caliptra will check if a vendor manifest measurement key is required:
+    * If the key is required, Caliptra will trust the Vendor Public key that was just endorsed.
+    * If the key is not required, Caliptra will not perform any more vendor verifications on this measurement manifest.
+* Verify the owner endorsement using the Caliptra owner public key and compare with the owner endorsement signature.
+    * If the owner endorsement is invalid, the `SET_AUTH_MANIFEST` command will be rejected.
+    * Otherwise, the owner public key will be trusted and Caliptra will use it to verify the overall measurement manifest.
+
+#### Measurement Manifest Version Number
+
+A Measurement Manifest VN is used to ensure that some enforcement is possible if a progression of measurements is required. 32 bits of the existing unused `IDEVID_MANUF_IDENTIFIER` fuse (128 bits) can be repurposed for this. This can be accomplished by updating Caliptra's main specification to redefine the fuse definition and its usage from "Programming time" to "Field Programmable".
+
+### Image Authorization Sequence
+
+The diagram below illustrates how this feature is part of the Caliptra boot flow, and the order of operations needed to use the feature.
+
+```mermaid
+sequenceDiagram
+    ROM->>FMC: Launch FMC
+    FMC->>Runtime: Launch RT
+    Runtime->>SOC: RDY_FOR_RT
+    Note over Runtime,SOC: Manifest Load
+    SOC->>Runtime: SET_MANIFEST
+    Runtime-->>SOC: Success/Failure
+    Note over Runtime,SOC: Image Authorization
+    loop n times
+        SOC->>Runtime: AUTHORIZE_AND_STASH
+        Runtime-->>SOC: Success/Failure
+    end
+
+    Note over Runtime,SOC: DPE Attestation
+    SOC->>Runtime: DPE Attestation
+```
+
+## Cryptographic Mailbox Commands (new in 2.0)
+
+Cryptographic mailbox (CM) commands are a flexible set of mailbox commands that provide access to Caliptra's cryptographic cabilities.
+This is meant for key storage and use, supporting protocols like SPDM and OCP LOCK.
+
+These commands are not meant to be high-performance as they are accessed via mailbox commands.
+
+These mailbox commands provide SHA, HMAC, HKDF, AES, and RNG services.
+Asymmetric cryptographic services are currently only provided through DPE and the `ECDSA384_SIGNATURE_VERIFY` and `LMS_SIGNATURE_VERIFY` mailbox commands, which do not use the CM storage system.
+
+### References
+
+* [SPDM 1.3.1 (DSP0274)](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.1.pdf), dated 2024-07-01.
+* [OCP Attestation v1.1](https://docs.google.com/document/d/1wA0hbJdtCpcQ1NvsVsYr2IeCkwQbgC7e/edit)
+* [RFC 5869 (HKDF)](https://www.rfc-editor.org/rfc/rfc5869.html)
+* [RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446) Section 7.4.2 & IEEE 1363 (TLS ECDH secret derivation)
+
+### Contexts
+
+Several of the methods, such as SHA and AES, support contexts, so that multiple users can have in-flight requests at the same time.
+
+The contexts contain the internal structures necessary to resume operations to support data that may exceed the size of a single mailbox command.
+
+These contexts are intended to be opaque to the user, and will be encrypted and authenticated if they contain sensitive internal data.
+
+### Data Storage
+
+The CM system has its own dedicated storage in DCCM.
+This storage is protected with authenticated encryption using internal keys.
+
+This storage is exposed to users through opaque 256-bit handles, called cryptographic mailbox IDs (CMIDs). These are mapped internally into DCCM storage blocks.
+
+Keys can be imported using `CM_IMPORT` and deleted via `CM_DELETE`.
+
+The entire contents can be cleared with `CM_CLEAR`.
+
+### Storage Design
+
+Data is stored in 16-byte blocks.
+
+Each block's status is tracked in a bitfield, 1 bit per block, with a value of 1 indicating that block is in use.
+
+For example, if we have 32 KB of storage for CM, this requires 256 bytes of additional storage.
+
+### CMIDs
+
+CMIDs require no additional storage.
+
+Unecrypted CMIDs have the following structure:
+
+| **Name**     | **Bits** | **Description**                             |
+| ------------ | -------- | ------------------------------------------- |
+| padding      | 80       | 0                                           |
+| block offset | 24       | Starting block number                       |
+| size         | 24       | Size (in bytes) of the data stored.         |
+|              |          | Does not have to be a multiple of 16 bytes. |
+
+Hence, data for a single CMID must be stored contiguously.
+
+Padding is kept for additional future features.
+
+The CMIDs are AES-256-GCM encrypted, with a 128-bit ciphertext concatenated with the 128-bit AES-GCM tag.
+
+### Storage Keys
+
+The key for the data referenced by a CMID is unique and derived via HKDF from the CDI<sub>RT</sub> with the info being the concatenation of `"CMID" || block offset || size`.
+
+The IV is the concatenation of `block offset || size`.
+
+This scheme avoids having to store the IV while avoiding IV reuse problems.
+
+GCM prevents an adversary from changing internal CMID parameters, such as changing an AES key to an ECDSA key, which has sensitive internal structure (although ECDSA is not currently supported).
+
+
+## Mailbox commands
 
 All mailbox command codes are little endian.
 
@@ -300,7 +471,7 @@ Command Code: `0x5349_4756` ("SIGV")
 | chksum        | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
 | fips\_status  | u32      | Indicates if the command is FIPS approved or an error.
 
-### LMS\_SIGNATURE\_VERIFY (1.1)
+### LMS\_SIGNATURE\_VERIFY (new in 1.1)
 
 Verifies an LMS signature. The hash to be verified is taken from
 Caliptra's SHA384 accelerator peripheral.
@@ -349,7 +520,7 @@ callers who update infrequently and cannot tolerate a changing DPE API surface.
 * Call the DPE DeriveContext command with the DefaultContext in the locality of
   the PL0 PAUSER.
 * Extend the measurement into PCR31 (`PCR_ID_STASH_MEASUREMENT`).
-* **Note**: This command can only be called in the locality of the PL0 PAUSER. 
+* **Note**: This command can only be called in the locality of the PL0 PAUSER.
 
 Command Code: `0x4D45_4153` ("MEAS")
 
@@ -677,10 +848,10 @@ Table: `SHUTDOWN` output arguments
 
 ### ADD\_SUBJECT\_ALT\_NAME
 
-Provides a subject alternative name otherName. Whenever CERTIFY_KEY_EXTENDED is called with the 
-DMTF_OTHER_NAME flag after ADD_SUBJECT_ALT_NAME is called, the resulting DPE CSR or leaf certificate 
-will contain a subject alternative name extension containing the provided otherName, which must be a 
-DMTF device info. All such certificates produced by CERTIFY_KEY_EXTENDED will continue to have the 
+Provides a subject alternative name otherName. Whenever CERTIFY_KEY_EXTENDED is called with the
+DMTF_OTHER_NAME flag after ADD_SUBJECT_ALT_NAME is called, the resulting DPE CSR or leaf certificate
+will contain a subject alternative name extension containing the provided otherName, which must be a
+DMTF device info. All such certificates produced by CERTIFY_KEY_EXTENDED will continue to have the
 DMTF otherName subject alternative name extension until reset.
 
 Command Code: `0x414C_544E` ("ALTN")
@@ -844,7 +1015,7 @@ Command Code: `0x5349_4D44` ("SIMD")
 
 ## Mailbox commands: Cryptographic Mailbox (2.0)
 
-These commands are used by the [Cryptograhic Mailbox](#cryptographic-mailbox-commands-20) system.
+These commands are used by the [Cryptograhic Mailbox](#cryptographic-mailbox-commands-new-in-20) system.
 
 ### CM_SHA
 
@@ -1000,9 +1171,9 @@ The encrypted and authenticated context's internal structure will be:
 | --------------- | -------- | --------------- |
 | tag size        | u32      |                 |
 | last length     | u32      |                 |
-| last counter    | u[16]    |                 |
-| last GHASH      | u[16]    |                 |
-| last ciphertext | u[16]    |                 |
+| last counter    | u8[16]    |                 |
+| last GHASH      | u8[16]    |                 |
+| last ciphertext | u8[16]    |                 |
 
 ### CM_AES_GCM_DECRYPT
 
@@ -1047,9 +1218,9 @@ The encrypted and authenticated context's internal structure will be:
 | --------------- | -------- | --------------- |
 | tag size        | u32      |                 |
 | last length     | u32      |                 |
-| last iv         | u[16]    |                 |
-| last GHASH      | u[16]    |                 |
-| last ciphertext | u[16]    |                 |
+| last iv         | u8[16]    |                 |
+| last GHASH      | u8[16]    |                 |
+| last ciphertext | u8[16]    |                 |
 
 
 ### CM_ECDH_GENERATE
@@ -1196,11 +1367,11 @@ Command Code: `0x434D_444C` ("CMDL")
 
 The entire contents of the CM storage is wiped.
 
-Command Code: `0x434D_5750` ("CMWP")
+Command Code: `0x434D_434C` ("CMCL")
 
-`CM_WIPE` takes no input arguments.
+`CM_CLEAR` takes no input arguments.
 
-*Table: `CM_WIPE` output arguments*
+*Table: `CM_CLEAR` output arguments*
 | **Name**    | **Type** | **Description**              |
 | ----------- | -------- | ---------------------------- |
 | chksum      | u32      |                              |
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -1,7 +1,22 @@
...
+
+* [LMS Signature Verification](#lms_signature_verify-11)
+
+## New in v1.2
+
+* [Manifest-Based Image Authorization](#manifest-based-image-authorization-12)
+
+## New in v2.0
+
+Caliptra 2.0 firmware supports passive mode (same as 1.x) and active (or susbystem) mode.
```
varuns-nvidia, line 15: suggest replacing active with subsystem throughout the document to parallel the main spec. subsystem is misspelled here.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+Usage information is required so that the data can be verified and used appropriately.
+
+Command Code: `0x434D_494D` ("CMIM")
+
+*Table: `CM_IMPORT` input arguments*
+
+| **Name**   | **Type** | **Description**                            |
+| ---------- | -------- | ------------------------------------------ |
+| chksum     | u32      |                                            |
+| usage      | u32      | Tag to specify how the data can be used    |
```
bluegate010, line 1163: These tags are equal to the mailbox command codes?

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+|            |          | `CMGE` - AES-256-GCM key, must be 32 bytes |
+|            |          | `HMAC` - Can be used for HMAC, HKDF        |
```
bluegate010, line 1165: Can I import a key of any size and use it as an HMAC key?

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
-| fips\_status  | u32      | Indicates if the command is FIPS approved or an error.
-
+| **Name**     | **Type** | **Description**                                                            |
+| ------------ | -------- | -------------------------------------------------------------------------- |
+| chksum       | u32      | Checksum over other output arguments, computed by Caliptra. Little endian. |
+| fips\_status | u32      | Indicates if the command is FIPS approved or an error.                     |
 
 ### AUTHORIZE_AND_STASH
 
+The SoC uses this command to request authorization of its various SoC images. This command has the option to receive the image hash directly from SoC or from an external source (e.g., SHA Acc).
```
varuns-nvidia, line 791: SHA Acc is deprecated in passive mode because the SHA is no longer expose on the AXI bus

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| chksum       | u32      |                                           |
+| fips_status  | u32      | FIPS approved or an error                 |
+| context size | u32      | Can be 0 if this is the final message     |
+| context      | u8[...]  |                                           |
+| hash size    | u32      | Can be 0 if this is not the final message |
+| hash         | u8[...]  |                                           |
+
+### CM_HMAC
+
+Computes an HMAC with
```
varuns-nvidia, line 883: unfinished sentence

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+### Storage Keys
+
+The key for the data referenced by a CMID is unique and derived via HKDF from the CDI<sub>RT</sub> with the info being the concatenation of `"CMID" || block offset || size`.
+
+The IV is the concatenation of `block offset || size`.
+
+This scheme avoids having to store the IV while avoiding IV reuse problems.
+
+The use of GCM is to prevent the CMID from being modified outside of Caliptra.
+Although this concern is minor in the current design, where all of the keys are uniformly random bits, if the CMID ever uses additional bits to designate key usage (or other parameters), then GCM prevents an adversary from changing those paramters, changing an AES key to an ECDSA key, which has sensitive internal structure.
```
bluegate010, line 1643: > if the CMID ever uses additional bits to designate key usage

`CM_IMPORT` allows the user to specify the usage for the key they are importing. Seems like we should reflect that usage constraint in the CMID.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Each block's status is tracked in a bitfield, 1 bit per block, with a value of 1 indicating that block is in use.
+
+For example, if we have 32 KB of storage for CM, this requires 256 bytes of additional storage.
+
+### CMIDs
+
+CMIDs require no additional storage.
+
+Unecrypted CMIDs have the following structure:
```
varuns-nvidia, line 1619: typo unencrypted

runtime/README.md
```diff
@@ -1,7 +1,22 @@
-# Caliptra Runtime Firmware v1.1
+# Caliptra Runtime Firmware v2.0
 
 This specification describes the Caliptra Runtime Firmware.
 
+## New in v1.1
```
jhand2, line 5: nit: Might be nice to format these headings as a "Changelog" table

runtime/README.md
```diff
@@ -58,20 +74,20 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
...
 
-| **Name**         | **Value**              | Description
-| -------          | -----                  | -----------
-| `SUCCESS`        | `0x0000_0000`          | Mailbox command succeeded
-| `BAD_VENDOR_SIG` | `0x5653_4947` ("VSIG") | Vendor signature check failed
-| `BAD_OWNER_SIG`  | `0x4F53_4947` ("OSIG") | Owner signature check failed
-| `BAD_SIG`        | `0x4253_4947` ("BSIG") | Generic signature check failure (for crypto offload)
-| `BAD_IMAGE`      | `0x4249_4D47` ("BIMG") | Malformed input image
-| `BAD_CHKSUM`     | `0x4243_484B` ("BCHK") | Checksum check failed on input arguments
+| **Name**         | **Value**              | Description                                          |
```
jhand2, line 83: nit: I would mildly prefer dropping these table formatting changes from this PR (or splitting them into a separate commit) to make it a bit easier to review the diff from a content perspective. I think they'll render in github the same either way.

But I understand if it's too much trouble to undo.

runtime/README.md
jhand2: I think we still need to sort this out. This is a pretty annoying hw regression to deal in fw with since it impacts how the API surface works.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
-| fips_status      | u32      | Indicates if the command is FIPS approved or an error.                     |
-| auth_req_result | u32      | AUTHORIZE_IMAGE: 0xDEADC0DE and DENY_IMAGE_AUTHORIZATION: 0x21523F21    |
+| **Name**        | **Type** | **Description**                                                            |
+| --------------- | -------- | -------------------------------------------------------------------------- |
+| chksum          | u32      | Checksum over other output arguments, computed by Caliptra. Little endian. |
+| fips_status     | u32      | Indicates if the command is FIPS approved or an error.                     |
+| auth_req_result | u32      | Authorized: `0xDEADC0DE` |
+|                 |          | Denied: `0x21523F21`     |
+
+### SET_IMAGE_METADATA
```
jhand2, line 826: Can you add a command description for this one? I'm not clear on what it's for.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| -------------- | -------- | --------------------------------- |
+| chksum         | u32      |                                   |
+| flags          | u32      | Bit 0 = this is the final message |
+| hash algorithm | u32      | Enum.                             |
+|                |          | Value 0 = reserved                |
+|                |          | Value 1 = SHA2-256                |
+|                |          | Value 2 = SHA2-384                |
+|                |          | Value 3 = SHA2-512                |
+| context size   | u32      | Size of the context               |
+| context        | u8[...]  |                                   |
```
jhand2, line 867: Can the context be fixed-size? Since we use zerocopy for command serialization/deserialization, it's quite a bit simpler to handle if the interior of the request is fixed-size.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| **Name**       | **Type** | **Description**                   |
+| -------------- | -------- | --------------------------------- |
+| chksum         | u32      |                                   |
+| flags          | u32      | Bit 0 = this is the final message |
+| hash algorithm | u32      | Enum.                             |
+|                |          | Value 0 = reserved                |
+|                |          | Value 1 = SHA2-256                |
+|                |          | Value 2 = SHA2-384                |
+|                |          | Value 3 = SHA2-512                |
+| context size   | u32      | Size of the context               |
```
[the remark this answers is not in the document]
jhand2: This API (and the hmac one) would probably be more clear using init/update/finish semantics.

runtime/README.md
```diff
@@ -1,7 +1,22 @@
-# Caliptra Runtime Firmware v1.1
+# Caliptra Runtime Firmware v2.0
```
jhand2, line 1: We should talk more about code size & ICCM size. I think we may have a hard time fitting all of this in 128KiB of SRAM.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| chksum      | u32      |                              |
+| fips_status | u32      | FIPS approved or an error    |
+
+### CM_CLEAR
+
+The entire contents of the CM storage is wiped.
+
+Command Code: `0x434D_5750` ("CMWP")
+
+`CM_WIPE` takes no input arguments.
```
jhand2, line 1201: nit: Command name is header is `CM_CLEAR`

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| chksum   | u32      |                            |
+| CMID     | u8[32]   | CMID of the data to delete |
+
+*Table: `CM_DELETE` output arguments*
+| **Name**    | **Type** | **Description**              |
+| ----------- | -------- | ---------------------------- |
+| chksum      | u32      |                              |
+| fips_status | u32      | FIPS approved or an error    |
+
+### CM_CLEAR
```
jhand2, line 1195: I wonder if it would be helpful to report how much CM storage is in use in some status command.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+|                                | Type         | 4-byte TYPE field of TCI node                                       |
+|                                | VendorInfo   | Locality of the caller (analog for PAUSER)                          |
 
 \*MultiTcbInfo contains one TcbInfo for each TCI Node in the path from the
 current TCI Node to the root. Max of 32.
+
+
+## Manifest-Based Image Authorization (1.2)
+
+Caliptra's goal is to enable integrators and vendors to meet standard security requirements for creating cryptographic identity and securely reporting measurements through DICE and DPE Certificate chains and Caliptra-owned private-public key pairs. In addition, Caliptra 1.0 provides an `ECDSA384_SIGNATURE_VERIFY` command to enable an SoC RoT to verify its own FW signatures so that it can develop an SoC secure boot using Caliptra cryptography. Caliptra 1.1 expanded the verify command to a PQC-safe `LMS_SIGNATURE_VERIFY` command. In each of these cases, it is left up to the vendor to ensure that they build a secure environment for introducing and verifying FW integrity and authenticity and then executing mutable FW.
```
jhand2, line 1476: nit: integrator and vendor generally refer to the same entity in Caliptra documentation.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+| tcg-dice-MultiTcbInfo\*        | FWIDs        | [0] "Journey" TCI Value                                             |
+|                                |              | [1] "Current" TCI Value. Latest `INPUT_DATA` made by DeriveContext. |
+|                                | Type         | 4-byte TYPE field of TCI node                                       |
+|                                | VendorInfo   | Locality of the caller (analog for PAUSER)                          |
 
 \*MultiTcbInfo contains one TcbInfo for each TCI Node in the path from the
 current TCI Node to the root. Max of 32.
+
+
+## Manifest-Based Image Authorization (1.2)
```
jhand2, line 1474: nit: Since this is a pretty sizable PR, it would be helpful to split the CM changes and the Image Manifest documentation into separate PRs (or separate commits within this PR).

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+    end
+
+    Note over Runtime,SOC: DPE Attestation
+    SOC->>Runtime: DPE Attestation
+```
+
+## Cryptographic Mailbox Commands (2.0)
+
+Cryptographic mailbox (CM) commands are a flexible set of mailbox commands that provide access to Caliptra's cryptographic cabilities.
+This is meant for key storage and use, supporting protocols like SPDM and OCP LOCK.
```
jhand2, line 1574: @bluegate010 IIRC, L.O.C.K. specifies specific commands that KMB (Caliptra) is supposed to implment. Is that correct?

If so, we may want to clarify that Caliptra L.O.C.K. firmware would implement the KMB APIs.

Also, because of the code size issue I mentioned above, we may have to think about whether L.O.C.K. integrations can get away with only implementing the KMB functions, or if they also need Caliptra Classic (TM) commands.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+Several of the methods, such as SHA and AES, support contexts, so that multiple users can have in-inflight requests at the same time.
+
+The contexts contain the internal structures necessary to resume operations to support data that may exceed the size of a single mailbox command.
+
+These contexts are intended to be opaque to the user, and will be encrypted and authenticated if they contain sensitive internal data.
+
+### Data Storage
+
+The CM system has its own dedicated storage in DCCM.
+This storage is protected with authenticated encryption using internal keys.
```
jhand2, line 1599: Are the encryption keys generated randomly on boot? Or derived deterministically?

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Hence, data for a single CMID must be stored contiguously.
+
+Padding is kept for additional future features.
+
+The CMIDs are AES-256-GCM encrypted, with a 128-bit ciphertext concatenated with the 128-bit AES-GCM tag.
+
+### Storage Keys
+
+The key for the data referenced by a CMID is unique and derived via HKDF from the CDI<sub>RT</sub> with the info being the concatenation of `"CMID" || block offset || size`.
```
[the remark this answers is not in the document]
jhand2: Could we generate a random number and put it in a keyvault slot at boot? Then whenever we need an AES key we can ask the KV slot of hmac the random seed + CMID metadata

Assuming we don't need these keys to live across cold reset.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Data is stored in 16-byte blocks.
+
+Each block's status is tracked in a bitfield, 1 bit per block, with a value of 1 indicating that block is in use.
+
+For example, if we have 32 KB of storage for CM, this requires 256 bytes of additional storage.
+
+### CMIDs
+
+CMIDs require no additional storage.
```
bluegate010, line 1617: Consider something like, "CMIDs offload storage of key metadata to mailbox callers. They therefore impose no additional storage within Caliptra Core."

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Several of the methods, such as SHA and AES, support contexts, so that multiple users can have in-inflight requests at the same time.
+
+The contexts contain the internal structures necessary to resume operations to support data that may exceed the size of a single mailbox command.
+
+These contexts are intended to be opaque to the user, and will be encrypted and authenticated if they contain sensitive internal data.
+
+### Data Storage
+
+The CM system has its own dedicated storage in DCCM.
```
bluegate010, line 1598: Does this mean that every key in CM storage is visible to Caliptra firmware? Seems like we should allow for keys that are not visible, such as shared secrets obtained from `CM_ECDH_FINISH`.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| context                     | u8[...]  |                               |
+| incoming exchange data size | u32      |                               |
+| incoming exchange data      | u8[...]  | the other side's public point |
+
+*Table: `CM_ECDH_FINISH` output arguments*
+| **Name**    | **Type** | **Description**                  |
+| ----------- | -------- | -------------------------------- |
+| chksum      | u32      |                                  |
+| fips_status | u32      | FIPS approved or an error        |
+| output CMID | u8[32]   | Output CMID of the shared secret |
```
bluegate010, line 1105: What kind of key is this, behind the CMID? Seems like it should be something that I can only pass to `CM_HKDF`.

runtime/README.md
fdamato: +1 on this

runtime/README.md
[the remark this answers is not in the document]
bluegate010: Yes, that would be a preferred way of doing this, since durability is not required across cold boot.

runtime/README.md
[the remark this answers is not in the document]
fdamato: I would agree to split, and keep a single command for single operations (no context would be required to be passed in this case)

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| tag size        | u32      |                 |
+| last length     | u32      |                 |
+| last iv         | u[16]    |                 |
+| last GHASH      | u[16]    |                 |
+| last ciphertext | u[16]    |                 |
+
+
+### CM_ECDH_GENERATE
+
+This computes the first half of an Elliptic Curve Diffie-Hellman exchange.
```
bluegate010, line 1057: Probably worth specifying that this generates an ephemeral ECDH keypair, and the private portion is represented in `context`.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+The Caliptra Measurement manifest feature expands on Caliptra-provided secure verifier abilities. The Measurement Manifest feature provides a standard Caliptra-supported definition to enable the following use cases for integrators, vendors, and owners.
+
+* Caliptra-Endorsed Aggregated Secure Boot
+* Caliptra-Endorsed Local Verifier
+
+Each of these abilities are tied to Caliptra Vendor and Owner FW signing keys and should be independent of any SoC RoT FW signing keys.
+
+Manifest-based image authorization is implemented via three mailbox commands: [`SET_AUTH_MANIFEST`](#set-auth-manifest), [`SET_IMAGE_METADATA`](#set-image-metadata), and [`AUTHORIZE_AND_STASH`](#authorize-and-stash).
+
+### Caliptra-Endorsed Aggregated Secure Boot
```
fdamato, line 1487: wondering if this should be called "Aggregated Measured Boot"

runtime/README.md
bluegate010: L.O.C.K. calls for more integrated flows, rather than purely being built in terms of crypto helpers. The integrated flows will need more firmware work, beyond what's offered in this interface. That being said, L.O.C.K. integrators _are_ likely going to want to leverage crypto helpers as part of their own firmware, so I'm not opposed to how the sentence is currently structured.

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Aggregated Secure Boot is a verified boot where one signed manifest attests to FW integrity of many different FW measurements. The authenticity of the FW is tied to the trust in the public key signing the measurement manifest, which is endorsed by the Caliptra Vendor and/or Owner FW Keys.
+
+### Caliptra-Endorsed Local Verifier
+
+A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest. In this case, the SoC RoT still has its own FW public-key chain that is verified by the SoC RoT, but in addition the SoC RoT introduces the Caliptra Measurement Manifest, which is endorsed by the Caliptra FW key pair. Caliptra provides approval or disapproval of the measurement of any FW back to the SoC RoT. This effectively provides a multi-factor authentication of SoC FW.
+
+The Caliptra-Endorsed Local Verify could be required by the owner only or both the vendor and the owner.
```
fdamato, line 1495: typo: verifier ?

runtime/README.md
```diff
@@ -1030,29 +1443,201 @@ The DPE `GET_CERTIFICATE_CHAIN` command shall return the following certificates:
...
+
+Manifest-based image authorization is implemented via three mailbox commands: [`SET_AUTH_MANIFEST`](#set-auth-manifest), [`SET_IMAGE_METADATA`](#set-image-metadata), and [`AUTHORIZE_AND_STASH`](#authorize-and-stash).
+
+### Caliptra-Endorsed Aggregated Secure Boot
+
+Aggregated Secure Boot is a verified boot where one signed manifest attests to FW integrity of many different FW measurements. The authenticity of the FW is tied to the trust in the public key signing the measurement manifest, which is endorsed by the Caliptra Vendor and/or Owner FW Keys.
+
+### Caliptra-Endorsed Local Verifier
+
+A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest. In this case, the SoC RoT still has its own FW public-key chain that is verified by the SoC RoT, but in addition the SoC RoT introduces the Caliptra Measurement Manifest, which is endorsed by the Caliptra FW key pair. Caliptra provides approval or disapproval of the measurement of any FW back to the SoC RoT. This effectively provides a multi-factor authentication of SoC FW.
```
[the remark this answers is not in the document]
fdamato: Should the manifest be endorsed by the Owner Key Pair ?

runtime/README.md
fdamato: The confusion here in calling it "Secure Boot" is that when Caliptra is Passive, it has no "power" to enforce what it is running

runtime/README.md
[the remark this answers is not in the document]
swenson: Using a random number in a keyvault slot was my first though, but that doesn't work for OCP LOCK, which needs persistence.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+The encrypted and authenticated context's internal structure will be:
+
+*Table: `CM_AES_GCM_ENCRYPT` internal context*
+| **Name**        | **Type** | **Description** |
+| --------------- | -------- | --------------- |
+| tag size        | u32      |                 |
+| last length     | u32      |                 |
+| last counter    | u[16]    |                 |
+| last GHASH      | u[16]    |                 |
+| last ciphertext | u[16]    |                 |
```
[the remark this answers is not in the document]
swenson: The last counter is the last plaintext.

But honestly this is just an implementation detail and we can probably just delete it.

runtime/README.md
swenson: Oops! I had reverted this in one of my revisions, but I must have accidentally turned back on table autoformatting. I'll undo these unrelated formatting changes.

runtime/README.md
swenson: Agreed.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| data           | u8[...]  | Data to hash                      |
+
+*Table: `CM_SHA` output arguments*
+| **Name**     | **Type** | **Description**                           |
+| ------------ | -------- | ----------------------------------------- |
+| chksum       | u32      |                                           |
+| fips_status  | u32      | FIPS approved or an error                 |
+| context size | u32      | Can be 0 if this is the final message     |
+| context      | u8[...]  |                                           |
+| hash size    | u32      | Can be 0 if this is not the final message |
```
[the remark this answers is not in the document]
swenson: I don't love having magic values, even if all 0s. Though it is probably safe here.

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| **Name**       | **Type** | **Description**    |
+| -------------- | -------- | ------------------ |
+| chksum         | u32      |                    |
+| Key CMID       | u8[32]   | CMID of key to use |
+| hash algorithm | u32      | Enum.              |
+|                |          | 0 = reserved       |
+|                |          | 1 = SHA2-256       |
+|                |          | 2 = SHA2-384       |
+|                |          | 3 = SHA2-512       |
+| data size      | u32      |                    |
```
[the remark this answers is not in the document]
swenson: Good point. I'll add that in.

runtime/README.md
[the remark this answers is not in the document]
swenson: That's fair. I'll split them up.

runtime/README.md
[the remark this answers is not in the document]
swenson: Just the last 16 bytes of the ciphertext, which the next invocation will use.

runtime/README.md
swenson: I forgot to change this in `CM_IMPORT`. I initially had the usage in the CMID, but I removed it for now since we don't have ECDSA and other sensitive keys. I am slightly inclined to keep this in a table in DCCM instead just to prevent this kind of attack completely.

Which, yes, would require augmenting the HKDF with a way to tag the output.

runtime/README.md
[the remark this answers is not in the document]
swenson: Do we zeroize DCCM on power on? (It's SRAM, so I thought it would be persistent.)

But in any case, if LOCK can re-import the key after power on every time, then we can eliminate the persistence. (Though it begs the question of where is the LOCK key being stored if not Caliptra that it can be re-imported any time.)

runtime/README.md
```diff
@@ -692,107 +708,504 @@ Command Code: `0x434B_4558` ("CKEX")
...
+Computes an HMAC with
+
+Command Code: `0x434D_484D` ("CMHM")
+
+*Table: `CM_HMAC` input arguments*
+
+| **Name**       | **Type** | **Description**    |
+| -------------- | -------- | ------------------ |
+| chksum         | u32      |                    |
+| Key CMID       | u8[32]   | CMID of key to use |
```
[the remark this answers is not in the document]
swenson: Almost every use of the CMID is slightly different (e.g., IKM, key), so I thought it best to spell it out each time.

runtime/README.md
swenson: For this particular one, yes, but I'd prefer to be flexible so that we don't have to change the format of the command if we decide to support other algorithms.

runtime/README.md
swenson: That's a good idea. I'll add that.

swenson:
I believe I have addressed the comments relevant to the cryptographic mailbox commands. I'm creating a separate PR for the manifest-based image auth doc updates, as suggested.

PTAL.

runtime/README.md
swenson: This will create a 384-bit shared secret. Theoretically it should only be used by HKDF, but other protocols could use the shared secret directly for AES.

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+*Table: `CM_ECDH_GENERATE` output arguments*
+| **Name**           | **Type** | **Description**                       |
+| ------------------ | -------- | ------------------------------------- |
+| chksum             | u32      |                                       |
+| fips_status        | u32      | FIPS approved or an error             |
+| context size       | u32      | size of context                       |
+| context            | u8[...]  | Used as the input to `CM_ECDH_FINISH` |
+| exchange data size | u32      | size of data to send in the exchange  |
+| exchange data      | u8[...]  | i.e., the public point                |
```
bluegate010, line 1262: Seems like we should document the format of this point? Is it 0x04 || x || y?

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+Command Code: `0x434D_4546` ("CMEF")
+
+*Table: `CM_ECDH_FINISH` input arguments*
+| **Name**                    | **Type** | **Description**                                          |
+| --------------------------- | -------- | -------------------------------------------------------- |
+| chksum                      | u32      |                                                          |
+| context size                | u32      | size of context                                          |
+| context                     | u8[...]  | This MUST come from the output of the `CM_ECDH_GENERATE` |
+| incoming exchange data size | u32      |                                                          |
+| incoming exchange data      | u8[...]  | the other side's public point                            |
```
bluegate010, line 1286: Ditto.

runtime/README.md
bluegate010: We should document the secret size. And document that the AES commands will implicitly truncate the CMID key if it is too large.

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| PRK CMID    | u8[32]   | CMID that refers to the output (PRK) to use |
+|             |          | with HKDF-Expand                            |
+
+### CM_HKDF_EXPAND
+
+Implements HKDF-Expand as specified in [RFC 5869](https://www.rfc-editor.org/rfc/rfc5869.html).
+
+Command Code: `0x434D_4B50` ("CMKP")
+
+*Table: `CM_HKDF` input arguments*
```
bluegate010, line 1125: s/CM_HKDF/CM_HKDF_EXPAND/?

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+Command Code: `0x434D_494D` ("CMIM")
+
+*Table: `CM_IMPORT` input arguments*
+
+| **Name**   | **Type** | **Description**                            |
+| ---------- | -------- | ------------------------------------------ |
+| chksum     | u32      |                                            |
+| usage      | u32      | Tag to specify how the data can be used    |
+|            |          | 0 - Can be used for HMAC, HKDF, AES        |
```
bluegate010, line 1352: I feel like we could be more prescriptive. Something like,

```
0 - HMAC-SHA2-256
1 - HMAC-SHA2-384
2 - HMAC-SHA2-512
3 - HKDF-SHA2-256
4 - HKDF-SHA2-384
5 - HKDF-SHA2-512
6 - AES-256-GCM
```
Then each key can only be used for a specific purpose, promoting good hygiene. CMIDs returned from `CM_ECDH_FINISH` could be implicitly tagged with type `HKDF-SHA2-384`. Callers could explicitly pass a value from this list as an arg to `CM_HKDF_EXPAND` (which could potentially replace the "Output length" arg).

runtime/README.md
```diff
@@ -58,7 +73,84 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
...
+
+Data is stored in 16-byte blocks.
+
+Each block's status is tracked in a bitfield, 1 bit per block, with a value of 1 indicating that block is in use.
+
+For example, if we have 32 KB of storage for CM, this requires 256 bytes of additional storage.
+
+### CMIDs
+
+CMIDs offload storage of key metadata to mailbox callers. They therefore impose no additional storage within Caliptra Core.
```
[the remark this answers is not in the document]
swenson: I thought about this, and the main problem is the size of systems like ML-KEM, which we will almost certainly be supporting in 2.1+.

runtime/README.md
swenson: I'm about 50% agreed, except that it is really hard to differentiate between keys for HMAC and HKDF, and TLS isn't required to use HKDF (it's only a `SHOULD`).

runtime/README.md
swenson: I'm not even sure that `0x04` is necessary. I don't see it in the TLS exchange data at least. I'll document that.

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+| plaintext      | u8[...]  | Data to encrypt                          |
+
+*Table: `CM_AES_GCM_ENCRYPT` output arguments*
+| **Name**       | **Type** | **Description**                           |
+| -------------- | -------- | ----------------------------------------- |
+| chksum         | u32      |                                           |
+| fips_status    | u32      | FIPS approved or an error                 |
+| context size   | u32      | SHALL be 0 if this is the final message   |
+| context        | u8[...]  |                                           |
+| cipertext size | u32      | MAY be 0 if this is not the final message |
```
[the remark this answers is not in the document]
swenson: It depends on the size of the input plaintext. If they send less than 16 bytes or our buffer is not full, we may not have output.

But we aren't going to buffer the entire output until the final call as that would not allow us to do large requests.

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+Command Code: `0x434D_5346` ("CMSF")
+
+*Table: `CM_SHA_FINAL` input arguments*
+| **Name**       | **Type** | **Description**                      |
+| -------------- | -------- | ------------------------------------ |
+| chksum         | u32      |                                      |
+| context size   | u32      | Size of the context.                 |
+| context        | u8[...]  | From `CM_SHA_INIT` / `CM_SHA_UPDATE` |
+| data size      | u32      | May be 0                             |
```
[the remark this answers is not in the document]
swenson: We could. But then we get back to where we before. Because if we have an empty data, then we could really skip to the update (since the init wouldn't actually be doing anything other than calculating the starting value). And then we might as well combine the final and update steps. :)

runtime/README.md
```diff
@@ -692,107 +787,631 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+*Table: `CM_AES_GCM_ENCRYPT` output arguments*
+| **Name**       | **Type** | **Description**                           |
+| -------------- | -------- | ----------------------------------------- |
+| chksum         | u32      |                                           |
+| fips_status    | u32      | FIPS approved or an error                 |
+| context size   | u32      | SHALL be 0 if this is the final message   |
+| context        | u8[...]  |                                           |
+| cipertext size | u32      | MAY be 0 if this is not the final message |
+| ciphertext     | u8[...]  |                                           |
```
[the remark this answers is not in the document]
swenson: Therein lies danger. :) FIPS generally doesn't certify implementations that have user-generated IVs. I think it is better to be prescriptive and say the IV is always random. We could revisit this if OCP Lock or another use case needs, say, a deterministic IV.

vsonims (approved):

runtime/README.md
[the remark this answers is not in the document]
swenson: Done

swenson:
Apologies for merging a little early, but we wanted to have this in before the deadline. I'm happy to do further revisions and have more discussions.

runtime/README.md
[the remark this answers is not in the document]
bluegate010: ML-KEM private keys are stored as seeds. Not sure why that's motivating consuming limited Caliptra memory, vs the caller's.

runtime/README.md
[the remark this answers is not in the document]
bluegate010: `CM_AES_GCM_DECRYPT` requires the caller to pass an IV. Where is the IV supposed to come from, if it's not returned from `CM_AES_GCM_ENCRYPT`?

runtime/README.md
bluegate010: A TLS implementation would know whether a given key is about to be used for HMAC or HKDF, so the caller should be able to specify at each step. To support either case for a key that comes out of `CM_HKDF_EXPAND`, we could allow the caller to specify the purpose as an input arg to that function.

runtime/README.md
```diff
@@ -692,107 +791,628 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+### CM_ECDH_FINISH
+
+This computes the second half of an Elliptic Curve Diffie-Hellman exchange.
+
+Currently only supports the NIST P-384 curve.
+
+The context must be passed from the `CM_ECDH_GENERATE` command.
+
+The incoming exchange data MUST be the concatenation of the x- and y- coordinates of the other side's public point.
```
bluegate010, line 1275: How are the coordinates encoded? Big-endian?

swenson: Yes, big-endian.

runtime/README.md
```diff
@@ -58,7 +77,84 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
...
+* [RFC 5869 (HKDF)](https://www.rfc-editor.org/rfc/rfc5869.html)
+* [RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446) Section 7.4.2 & IEEE 1363 (TLS ECDH secret derivation)
+
+### Contexts
+
+Several of the methods, such as SHA and AES, support contexts so that multiple users can have in-flight requests at the same time.
+
+The contexts contain the internal structures necessary to resume operations to support data that may exceed the size of a single mailbox command.
+
+These contexts are intended to be opaque to the user, and SHALL be encrypted and authenticated if they contain sensitive internal data.
```
raghuncstate, line 103: where is it stored? and why not replay protected?

raghuncstate, line 103: what threat is this intended to solve/mitigate? it is defense in depth mostly?
Asked another way, what happens if you leave it unencrypted? presumably, a runtime attack on FW has access to the key (indirectly?) used to encrypt? or is that not hte case?

runtime/README.md
```diff
@@ -58,7 +77,84 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
...
+
+The CMIDs are AES-256-GCM encrypted, with a 128-bit ciphertext concatenated with the 128-bit AES-GCM tag.
+
+### Storage Keys
+
+The key for the data referenced by a CMID is unique and derived via HKDF from the CDI<sub>RT</sub> with the info being the concatenation of `"CMID" || block offset || size`.
+
+The IV is the concatenation of `block offset || size`.
+
+This scheme avoids having to store the IV while avoiding IV reuse problems.
```
raghuncstate, line 152: can you elaborate on why IV reuse is not a problem? if you store the different data twice at the same offset with same size, would you not generate the same key and same IV again to encrypt the new data?

runtime/README.md
swenson: The context might have intermediate state that is sensitive. For example, intermediate GHASH computations in AES could be used to reveal the additional authenticated data, which may or may not be sensitive.

The authenticated encryption is more defense in depth.

runtime/README.md
swenson: I noticed that after I wrote it. It's definitely true (if the data being stored is the same length).

I'm debating a few different ways to fix this, but essentially involve storing the IV or a part of the IV either in DCCM or with the CMID.

bluegate010: > if you store the different data twice at the same offset with same size, would you not generate the same key and same IV again to encrypt the new data?

I don't think you would, right? Because the CMID does not hold any data related to the key itself, just the offset and size. So if you store two different keys with the same size at the same offset, the same key will encrypt the same CMID.

runtime/README.md
raghuncstate: thx, agree on the intermediate state being sensitive, but was wondering HOW you envision an attacker getting access to it, mostly to try and answer the question: if the attacker can get access to context data, can the attacker also not get access to the key?
What makes getting access to the key (or its handle) more or less difficult than having access to the context data?

Note, not asking for change, trying to better understand thought process behind decision to do so.

runtime/README.md
raghuncstate: okay, maybe i misunderstood, what this is encrypting. I thought this is a key, being derived, to encrypt context data. The key being derived, look like a deterministic key HKDF(CDI_RT, "CMID||block offset||size"), and that key would AES GCM the context data (which can be different data bits).
I did not think this is the key used to encrypt the CMID itself, given to the user.

runtime/README.md
```diff
@@ -692,107 +791,628 @@ Command Code: `0x434B_4558` ("CKEX")
...
+
+Command Code: `0x434D_5355` ("CMSU")
+
+*Table: `CM_SHA_UPDATE` input arguments*
+| **Name**       | **Type** | **Description**                      |
+| -------------- | -------- | ------------------------------------ |
+| chksum         | u32      |                                      |
+| context size   | u32      | Size of the context.                 |
+| context        | u8[...]  | From `CM_SHA_INIT` / `CM_SHA_UPDATE` |
+| data size      | u32      |                                      |
```
raghuncstate, line 958: do we care that data size field may become an unaligned access depending on context size?

bluegate010: I think we could define a fixed-size context struct, sized to hold the largest context data possible, and make sure it's aligned properly.

runtime/README.md
raghuncstate: i think i may have misread/misunderstood. I thought contexts are stored in the data storage, but i see that you provide it as output in commands below, so the question above may be moot, but would appreciate confirmation that data storage below, is for the KEYS, not contexts :)
Do we care about replay protection of the contexts?

swenson: Contexts could be stored insecurely in the SoC, so they are more vulnerable to being intercepted.

In this version, the keys are only accessible through handles, but the keys themselves are stored only in DCCM in Caliptra. (This may change so that the keys are also returned to the SoC in encrypted form to minimize storage requirements in Caliptra.)

Contexts are also encrypted and authenticated. For many of the commands, I don't think we're too worried about replay attacks (e.g., SHA). But, I think you're right that some commands could use replay attack protection. For instance, with AES-GCM, if you have a context and an associated ciphertext block, you could replay the context with a new plaintext and use that to recover the other plaintext.

Theoretically this is all over a trusted channel, so these attacks may not be part of the threat model, but we should be able to apply some internal counters in DCCM to help with this. I've filed an issue to address this: https://github.com/chipsalliance/caliptra-sw/issues/1761

runtime/README.md
swenson: I worry about a fixed-size struct, as it means that if we define a new command or want to redefine the context for one command, we break compatibility with all commands that use contexts. With the variable-sized contexts, we wouldn't need to break compatibility if we needed to add a field to any context.

But we can stipulate that all contexts have sizes that are multiples of 32 bits.

runtime/README.md
swenson: The scheme here was `K = HKDF(CDI_RT, block offset || size)`, which would then be used to encrypt that portion of DCCM that holds the key associated with the CMID.

But I think we're going to move away from storing keys in DCCM shortly, so this shouldn't matter. Though we will still have to be very careful about IV reuse.

PR #1749, opened 2024-10-29
[doc] Manifest-Based Image Auth

swenson:
This adds sections describing the Manifest-Based Image Authorization system.

I cleaned up a little bit of the manifest-based image auth commands and added some overview sections for what is new in the 1.1 and 1.2 runtime firmware to help keep things organized.

I've addressed most of the comments from https://github.com/chipsalliance/caliptra-sw/pull/1745 that I could. I don't necessarily know all of the details, and not everything has been merged yet (e.g., https://github.com/chipsalliance/caliptra-sw/pull/1705), but I've tried to sync everything up with the latest version of the docs and code.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -10,6 +10,10 @@ v1.1:
 
 * [LMS Signature Verification](#lms_signature_verify-new-in-11)
 
+v1.2:
+
+* [Manifest-Based Image Authorization](#manifest-based-image-authorization-new-in-12)
+
 v2.0:
 
 * Add support for passive mode (same as 1.x) and subsystem (or active) mode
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -58,7 +66,104 @@ Caliptra Runtime Firmware will share driver code with ROM and FMC where
...
+
+### Caliptra-Endorsed Local Verifier
+
+A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest. In this case, the SoC RoT still has its own FW public-key chain that is verified by the SoC RoT, but in addition the SoC RoT introduces the Caliptra Measurement Manifest, which is endorsed by the Caliptra FW key pair. Caliptra provides approval or disapproval of the measurement of any FW back to the SoC RoT. This effectively provides a multi-factor authentication of SoC FW.
+
+The Caliptra-Endorsed Local Verifier could be required by the owner only or both the vendor and the owner.
+
+The main difference between Caliptra-Endorsed Aggregated Measured Boot and Caliptra-Endorsed Local Verifier is if the SoC RoT is relying on the Measurement Manifest for SoC Secure Boot services as opposed as using it as an additional verification.
```
fdamato, line 92: it is still not clear to me the difference between the 2 usage models.
    - In case of "Caliptra-Endorsed Local Verifier" what is the remediation action SOC Manager should take in case of failure ?  Assuming boot is not stopped.
 
Also being Calipta the RTM for the SOC, I would encourage Caliptra RT FW to attest to this failure in the DPE (How ???).

swenson: It's not 100% clear to me either. Maybe @mhatrevi can weigh in?

vsonims (approved):

runtime/README.md
fdamato: Thanks Chris. I really would like to understand better the difference between usage models.

runtime/README.md
```diff
@@ -153,6 +157,102 @@ This scheme avoids having to store the IV while avoiding IV reuse problems.
...
+
+Manifest-based image authorization is implemented via three mailbox commands: [`SET_AUTH_MANIFEST`](#set-auth-manifest), [`SET_IMAGE_METADATA`](#set-image-metadata), and [`AUTHORIZE_AND_STASH`](#authorize-and-stash).
+
+### Caliptra-Endorsed Aggregated Measured Boot
+
+Aggregated Measured Boot is a verified boot where one signed manifest attests to FW integrity of many different FW measurements. The authenticity of the FW is tied to the trust in the public key signing the measurement manifest, which is endorsed by the Caliptra Vendor and/or Owner FW Keys.
+
+### Caliptra-Endorsed Local Verifier
+
+A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest. In this case, the SoC RoT still has its own FW public-key chain that is verified by the SoC RoT, but in addition the SoC RoT introduces the Caliptra Measurement Manifest, which is endorsed by the Caliptra FW key pair. Caliptra provides approval or disapproval of the measurement of any FW back to the SoC RoT. This effectively provides a multi-factor authentication of SoC FW.
```
bluegate010, line 179: > FW public-key chain

It's ambiguous whether this is a firmware signing key chain or a firmware attestation key chain. I presume this is the former.

bluegate010, line 179: > Caliptra FW key pair

Caliptra FW _signing_ key pair I presume, to disambiguate from DICE.

Is this the vendor signing key or owner signing key?

bluegate010, line 179: > multi-factor authentication

Bike-shedding here but I feel like "joint authentication" might be a better term, to avoid connotations of human-driven multi-factor authentication flows.

bluegate010, line 179: > A local verifier provides an authentication of SoC FW by matching SoC FW measurements with measurements from the Caliptra measurement manifest.

So what _is_ a "local verifier"? Is it a piece of logic living inside Caliptra? Is it a role that is shared with the SoC?

runtime/README.md
```diff
@@ -153,6 +157,102 @@ This scheme avoids having to store the IV while avoiding IV reuse problems.
...
+
+The Caliptra-Endorsed Local Verifier could be required by the owner only or both the vendor and the owner.
```
bluegate010, line 181: What does it mean for a local verifier to be "required" by a vendor or owner? Is the functionality enabled or disabled based on whether it is required?

runtime/README.md
bluegate010: Unresolving as I have the same question.

- The "Caliptra-Endorsed Aggregated Measured Boot" section states that one signed manifest includes measurements for a collection of different components.
- The "Caliptra-Endorsed Local Verifier" section first paragraph states that measurements of components are compared to a manifest.

Those sound like the same thing.

> The main difference between Caliptra-Endorsed Aggregated Measured Boot and Caliptra-Endorsed Local Verifier is if the SoC RoT is relying on the Measurement Manifest for SoC Secure Boot services as opposed as using it as an additional verification.

So in both models the SoC is checking in with Caliptra whenever it boots code. In the second model the SoC is doing its own verification on top of that. If that's the case, it seems like we could just describe a single model of manifest flows. We can just say that this is a feature that SoCs can rely on, and they can optionally also do their own verification on top of that.

bluegate010: Have filed as https://github.com/chipsalliance/caliptra-sw/issues/1757

PR #1750, opened 2024-10-29
Decrease size of ModelEmulated from ~50k to 192 bytes.

korran:
This should cut back on stack consumption and reduce unnecessary memcpy bloat when running tests.


hw-model/src/model_emulated.rs
```diff
@@ -54,7 +54,7 @@ impl<'a> Bus for EmulatedApbBus<'a> {
 
 /// Emulated model
 pub struct ModelEmulated {
-    cpu: Cpu<BusLogger<CaliptraRootBus>>,
+    cpu: Box<Cpu<BusLogger<CaliptraRootBus>>>,
```
swenson, line 57: To avoid this problem in the future or for other types, would it be better to modify the `HwModel` trait to return a `Box`? I'm not sure how easy that is without knowing the concrete type of `Self` though...

sw-emulator/lib/cpu/src/csr_file.rs
```diff
@@ -102,7 +102,7 @@ impl Csr {
 /// Configuration and status register file
 pub struct CsrFile {
     /// CSRS
-    csrs: [Csr; CsrFile::CSR_COUNT],
+    csrs: Box<[Csr; CsrFile::CSR_COUNT]>,
```
swenson, line 105: I'm okay with this, but if the CPU is boxed then I'm not sure this does much?

And I know it is the emulator, but a 4096-len array seems silly :) If we're pulling in standard library anyway, this feels like it should be a `HashMap` or a `Box<HashMap>`, or some other sparse data structure.

swenson (approved):
LGTM. A few questions but I'm okay with this cleanup regardless.

hw-model/src/model_emulated.rs
korran: ModelEmulated is a fairly light wrapper around Cpu<CaliptraRootBus>, which does all the heavy lifting. Returning a Box would arguably be a bit better, but I don't want to change the HwModel's public API, which has hundreds of tests written against it today and is also used in vendor code.

If we did want to go that route, I would do something like this and then forward all the functions:

pub struct ModelEmulated {
  imp: Box<ModelEmulatedImpl>,
}

sw-emulator/lib/cpu/src/csr_file.rs
korran: This is all a bit of dance with the optimizer. Theoretically LLVM *should* be able to RVO all these non-result constructors, but it's not guaranteed so I decided to box some of the larger types to increase the chance that it does the right thing *somewhere*.

ArthurHeymans (approved):

jhand2 (approved):

mhatrevi (approved):

PR #1765, opened 2024-11-01
Update zerocopy from 0.6.6 to 0.8.8.

clundin25:
This PR does not optimize out a large portion of copies. Ideally it will be done in follow up PRs.

There are slight changes made to optimize panics out of the final binary.

This resolves https://github.com/chipsalliance/caliptra-sw/issues/1721

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,7 +183,7 @@ ureg-schema = { path = "ureg/lib/schema" }
 ureg-systemrdl = { path = "ureg/lib/systemrdl" }
 wycheproof = "0.5.1"
 x509-parser = "0.15.0"
-zerocopy = "0.6.6"
+zerocopy = { version = "0.8.8", features = ["derive"] }
 serial_test = "2.0.0"
 nix = "0.26.2"
 libc = "0.2"
--- a/drivers/src/bounded_address.rs
+++ b/drivers/src/bounded_address.rs
@@ -4,7 +4,7 @@ use core::fmt::Debug;
 use core::marker::PhantomData;
 
 use caliptra_error::CaliptraError;
-use zerocopy::{AsBytes, FromBytes};
+use zerocopy::{FromBytes, Immutable, IntoBytes, TryFromBytes, Unalign};
 use zeroize::Zeroize;
 
 use crate::memory_layout;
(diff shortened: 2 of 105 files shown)


drivers/src/bounded_address.rs
```diff
@@ -24,18 +24,19 @@ impl MemBounds for RomBounds {
...
+#[derive(Zeroize, TryFromBytes, Immutable)]
+pub struct BoundedAddr<T: IntoBytes + FromBytes, B: MemBounds> {
     addr: u32,
     _phantom: PhantomData<(T, B)>,
 }
-unsafe impl<T: AsBytes + FromBytes, B: MemBounds> FromBytes for BoundedAddr<T, B> {
+// Unaligned is not implemented for `u32`, so `BoundedAddr` cannot currently derive `IntoBytes`.
+// Potentially `u32` can be swapped with `U32` from the zerocopy crate, but this type does not
+// implement Zeroize.
+// We could potentially wrap `U32` with the `Zeroizing` type to resolve this.
```
jhand2, line 35: What's the conclusion here? It seems like it's maybe working now.

clundin25: I haven't followed up on this, but for this case I think we will not be able to use the derive feature.

I believe the orphan rule will block the workaround (Cannot implement `Zeroizing` (zeroize trait) on `U32` (Unaligned u32 struct from zerocopy).

clundin25: Looked at this again with fresh eyes and cleaned it up. This type can now just derive the zerocopy traits directly.

zhalvorsen (dismissed):
We should cherry pick this onto the 2.x branch as soon as we can because it touches so many files and tests.

jhand2 (approved):

jhand2 (approved):

runtime/src/get_idev_csr.rs
```diff
@@ -13,44 +13,40 @@ use caliptra_error::{CaliptraError, CaliptraResult};
...
-                    // `resp.data_size` by the `IDevIDCsr::get` API.
-                    //
-                    // A valid `IDevIDCsr` cannot be larger than `MAX_CSR_SIZE`, which is the max
-                    // size of the buffer in `GetIdevCsrResp`
-                    resp.data[..resp.data_size as usize].copy_from_slice(csr);
-
-                    Ok(MailboxResp::GetIdevCsr(resp))
-                }
+        let cmd = GetIdevCsrReq::ref_from_bytes(cmd_args)
+            .map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
```
swenson, line 24: Before wasn't this `CaliptraError::RUNTIME_INSUFFICIENT_MEMORY`?

swenson (commented):
Phew! This looked like it was a big pain to write.

Looks good; just a couple of questions

Cargo.toml
```diff
@@ -183,7 +183,7 @@ ureg-schema = { path = "ureg/lib/schema" }
 ureg-systemrdl = { path = "ureg/lib/systemrdl" }
 wycheproof = "0.5.1"
 x509-parser = "0.15.0"
-zerocopy = "0.6.6"
+zerocopy = { version = "0.8.8", features = ["derive"] }
```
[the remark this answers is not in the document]
clundin25: I'll do a fast-follow up PR to bump the version, which (hopefully) will be a smaller PR.

That way I avoid dismissing approvals! 😄

runtime/src/get_idev_csr.rs
clundin25: This no longer is a copy, so `CaliptraError::RUNTIME_INSUFFICIENT_MEMORY` is not the failure scenario.

drivers/src/bounded_address.rs
```diff
@@ -24,31 +24,41 @@ impl MemBounds for RomBounds {
 pub type RomAddr<T> = BoundedAddr<T, RomBounds>;
 
 #[repr(C)]
-#[derive(Zeroize)]
-pub struct BoundedAddr<T: AsBytes + FromBytes, B: MemBounds> {
-    addr: u32,
+#[derive(TryFromBytes, IntoBytes, Immutable)]
+pub struct BoundedAddr<T: IntoBytes + FromBytes, B: MemBounds> {
+    addr: Unalign<u32>,
```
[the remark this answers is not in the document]
clundin25: Previously, the zerocopy marker traits `FromBytes` and `AsBytes` where manually implemented.

Now I have modified this type to use the zerocopy derive macros, which enforce a set of [safety conditions](https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#safety).

I have modified the `addr` field from a `u32` to `Unalign<u32>` to satisfy the safety condition _"if the type is repr(C), its fields must be [Unaligned](https://docs.rs/zerocopy/latest/zerocopy/trait.Unaligned.html)"._

swenson (approved):

Cargo.toml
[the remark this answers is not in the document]
clundin25: Looks like the build currently fails on 0.8.13.

```
error[E0446]: private type `dpe_instance::_::__Zerocopy_Field_reserved` in public interface
  --> dpe/dpe/src/dpe_instance.rs:45:35
   |
45 | #[derive(IntoBytes, TryFromBytes, KnownLayout, Immutable, Zeroize)]
   |                                   ^^^^^^^^^^^
   |                                   |
   |                                   can't leak private type
   |                                   `dpe_instance::_::__Zerocopy_Field_reserved` declared as private
   |
   = note: this error originates in the derive macro `KnownLayout` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The DPE crate has a private field in a public type, which is causing a zerocopy macro to trip a compiler error. I don't think that is an incorrect use case, but I need to dig deeper.

Issue #1703, opened 2024-10-07
Deprecate FMC SVN


bluegate010:
Firmware is tagged with two SVNs: FMC SVN and RT SVN. I don't think we need both.

FMC and RT FW is monolithically signed, and monolithically runtime-updated[1]. It can also be monolithically associated with an SVN.

For backwards compatibility, we could continue to add an SVN to the FMC alias key. But we can make that SVN equal the SVN in the RT FW alias key.

Without distinct SVNs, we would expect that a vulnerability in either FMC or RT FW would motivate an increment of the singular SVN that describes them both.

This will also be important in allowing ROM to implement a hash chain to support stable identity features. ROM can only generate a hash chain from a single-dimensional SVN. If there are two SVNs, stable identity will need to ignore one of the SVNs.

[1] While FMC cannot be runtime-updated, Caliptra [enforces](https://github.com/chipsalliance/caliptra-sw/blob/0d5ad377138e5aa461d3300b0b9398e462c8289b/image/verify/src/verifier.rs#L800-L801) that the FMC in the new bundle equals the FMC in the current bundle. Therefore, when an (FMC, RT FW) pair is signed and associated with an SVN, the reported SVN will describe the current FMC and RT FW. In other words, Caliptra does not allow drift or disassociation between monolithically-signed FMC/RT FW images.

PR #1767, opened 2024-11-02
Unify separate FMC and RT SVNs into a single firmware SVN.

bluegate010:
Apologies for the size of this PR.

With this change, there is now a singular SVN representing the security state of both FMC and Runtime. This is done as part of enabling Stable Identity in ROM, which can only easily be implemented in terms of a single SVN.

This does not alter the external API for Caliptra, with the sole exception that what was previously reported as the FMC's SVN is now the value of the (FMC+RT) FW SVN as it was during cold-boot.

- The FMC and Runtime images still each carry an SVN, but this is done only for backwards compatibility in the build tooling, and ROM ensures the two SVNs are equal.
- The FMC and Runtime alias certificates still each carry an SVN.
- The FMC alias certificate's SVN is the value that the firmware's SVN was at cold-boot.
- The Runtime alias certificate's SVN is the value that the firmware's SVN was at the last update-reset.

See additional commentary in https://github.com/chipsalliance/caliptra-sw/issues/1703.


rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -186,8 +186,8 @@ impl FmcAliasLayer {
             tcb_info_fmc_tci: &(&env.data_vault.fmc_tci()).into(),
             tcb_info_device_info_hash: &fuse_info_digest.into(),
             tcb_info_flags: &flags,
-            tcb_info_fmc_svn: &svn.to_be_bytes(),
```
vsonims, line 189: If we remove FMC SVN how do we prevent FMC rollbacks?

vsonims (changes requested):

rom/dev/src/flow/cold_reset/fmc_alias.rs
bluegate010: For update-resets we don't allow FMC to change at all, either upgrades or downgrades.

For cold-boot resets, we prevent FMC rollbacks the same way that we prevent RT FW rollbacks: increment the SVN, which (with this change) describes the security state of both portions of the FW image. The implication is that if there's a security vulnerability identified in either FMC or RT FW, the SVN for the combined FW image would be incremented.

mhatrevi: There were some discussions of enabling FMC updates via the update-reset flow. I think this question was in context of that.

mhatrevi (dismissed):

rom/dev/src/flow/cold_reset/fmc_alias.rs
bluegate010: If we enable FMC updates via the update-reset flow, then we'll just do the same <= check for the new firmware's FMC. In other words, we can avoid thinking about FMC and RT as two different images. They're signed together, they boot together, they update together; they have a single SVN for anti-rollback.

mhatrevi (approved):

drivers/src/fuse_log.rs
```diff
@@ -21,12 +21,12 @@ pub enum FuseLogEntryId {
     Invalid = 0,
     VendorEccPubKeyIndex = 1,       // 4 bytes  (From Manifest)
     VendorEccPubKeyRevocation = 2,  // 4 bytes  (From Fuse)
-    ManifestFmcSvn = 3,             // 4 bytes
+    DeprecatedManifestFmcSvn = 3,   // 4 bytes
```
jhand2, line 24: nit: Should the deprecated values have a `_` prefix to denote it isn't used? Also you can mark fields as `#[deprecated]` to get the compiler to emit a warning if they're used.

If you still want the `From` impl to be exhaustive, you can add `#[allow(deprecated)]`. Or you could just make the deprecated entries return `Invalid`

error/README.md
```diff
@@ -88,16 +88,16 @@
 | Image Verifier | Verifier | FmcLoadAddrUnaligned |0x000B0022 |
 | Image Verifier | Verifier | FmcEntryPointInvalid |0x000B0023 |
 | Image Verifier | Verifier | FmcEntryPointUnaligned |0x000B0024 |
-| Image Verifier | Verifier | FmcSvnGreaterThanMaxSupported |0x000B0025 |
-| Image Verifier | Verifier | FmcSvnLessThanMinSupported |0x000B0026 |
-| Image Verifier | Verifier | FmcSvnLessThanFuse |0x000B0027 |
+| Image Verifier | Verifier | (Deprecated) FmcSvnGreaterThanMaxSupported |0x000B0025 |
```
jhand2, line 91: Can rebase to pick up https://github.com/chipsalliance/caliptra-sw/pull/1779 and drop changes in this file.

error/src/lib.rs
```diff
@@ -229,11 +226,10 @@ impl CaliptraError {
         CaliptraError::new_const(0x000b002a);
     pub const IMAGE_VERIFIER_ERR_RUNTIME_ENTRY_POINT_UNALIGNED: CaliptraError =
         CaliptraError::new_const(0x000b002b);
-    pub const IMAGE_VERIFIER_ERR_RUNTIME_SVN_GREATER_THAN_MAX_SUPPORTED: CaliptraError =
+    pub const IMAGE_VERIFIER_ERR_FIRMWARE_SVN_GREATER_THAN_MAX_SUPPORTED: CaliptraError =
         CaliptraError::new_const(0x000b002c);
-    pub const IMAGE_VERIFIER_ERR_RUNTIME_SVN_LESS_THAN_MIN_SUPPORTED: CaliptraError =
-        CaliptraError::new_const(0x000b002d);
-    pub const IMAGE_VERIFIER_ERR_RUNTIME_SVN_LESS_THAN_FUSE: CaliptraError =
+    // 0x000b002d was IMAGE_VERIFIER_ERR_RUNTIME_SVN_LESS_THAN_MIN_SUPPORTED
```
jhand2, line 231: Why is this one removed? I would expect this to still be a possible error case.

error/src/lib.rs
```diff
@@ -436,7 +435,7 @@ impl CaliptraError {
         CaliptraError::new_const(0x000E0039);
     pub const RUNTIME_PRIV_KEY_KV_HDL_HANDOFF_FAILED: CaliptraError =
         CaliptraError::new_const(0x000E003A);
-    pub const RUNTIME_HASH_CHAIN_HANDOFF_FAILED: CaliptraError =
+    pub const FIRMWARE_HASH_CHAIN_HANDOFF_FAILED: CaliptraError =
```
jhand2, line 438: I think this should still have the `RUNTIME` prefix since it's a runtime error.

hw-model/src/lib.rs
```diff
@@ -633,10 +633,8 @@ pub trait HwModel: SocManager {
         self.soc_ifc()
             .fuse_owner_pk_hash()
             .write(&fuses.owner_pk_hash);
-        self.soc_ifc()
-            .fuse_fmc_key_manifest_svn()
-            .write(|_| fuses.fmc_key_manifest_svn);
-        self.soc_ifc().fuse_runtime_svn().write(&fuses.runtime_svn);
+        self.soc_ifc().fuse_fmc_key_manifest_svn().write(|_| 0); // FMC SVN deprecated
```
jhand2, line 636: For a future PR to caliptra-rtl, in the RDL we should maybe update this to be unpopulated so we don't generate any code for it.

image/verify/src/verifier.rs
```diff
@@ -110,10 +110,30 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
+        let fmc_info = self.verify_fmc(image_info.fmc, reason)?;
 
         // Verify Runtime
-        let (runtime_info, rt_log_info) = self.verify_runtime(image_info.runtime)?;
+        let runtime_info = self.verify_runtime(image_info.runtime)?;
+
+        // Verify SVN
+        if self.svn_check_required() {
+            if image_info.fmc.svn != image_info.runtime.svn {
+                Err(CaliptraError::IMAGE_VERIFIER_ERR_RUNTIME_FMC_SVNS_UNEQUAL)?;
```
jhand2, line 121: Not sure what to make of this. Shouldn't we just have one SVN now?

rom/dev/doc/error-attribution.md
```diff
@@ -91,23 +91,21 @@ Verifier Library | IMAGE_VERIFIER_ERR_FMC_LOAD_ADDR_INVALID
...
-Verifier Library | IMAGE_VERIFIER_ERR_FMC_SVN_LESS_THAN_MIN_SUPPORTED                  | 0x000b0026
-Verifier Library | IMAGE_VERIFIER_ERR_FMC_SVN_LESS_THAN_FUSE                           | 0x000b0027
 Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_LOAD_ADDR_INVALID                        | 0x000b0028
 Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_LOAD_ADDR_UNALIGNED                      | 0x000b0029
 Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_ENTRY_POINT_INVALID                      | 0x000b002a
 Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_ENTRY_POINT_UNALIGNED                    | 0x000b002b
-Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_SVN_GREATER_THAN_MAX_SUPPORTED           | 0x000b002c
-Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_SVN_LESS_THAN_MIN_SUPPORTED              | 0x000b002d
-Verifier Library | IMAGE_VERIFIER_ERR_RUNTIME_SVN_LESS_THAN_FUSE                       | 0x000b002e
+Verifier Library | IMAGE_VERIFIER_ERR_FIRMWARE_SVN_GREATER_THAN_MAX_SUPPORTED          | 0x000b002c
```
jhand2, line 98: For a separate PR: I would be in favor of just pointing to the source code for this file as well.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -540,7 +540,9 @@ impl FirmwareProcessor {
 
         data_vault.write_warm_reset_entry48(WarmResetEntry48::RtTci, &info.runtime.digest.into());
 
-        data_vault.write_warm_reset_entry4(WarmResetEntry4::RtSvn, info.runtime.svn);
+        data_vault.write_warm_reset_entry4(WarmResetEntry4::FwSvn, info.fw_svn);
+        // TODO: write min-svn
+        // data_vault.write_warm_reset_entry4(WarmResetEntry4::FwMinSvn, info.fw_svn); // At cold-boot, min_svn == curr_svn
```
jhand2, line 545: Why is this a TODO? Does this code not work?

rom/dev/src/flow/update_reset.rs
```diff
@@ -211,7 +211,12 @@ impl UpdateResetFlow {
     fn populate_data_vault(data_vault: &mut DataVault, info: &ImageVerificationInfo) {
         data_vault.write_warm_reset_entry48(WarmResetEntry48::RtTci, &info.runtime.digest.into());
 
-        data_vault.write_warm_reset_entry4(WarmResetEntry4::RtSvn, info.runtime.svn);
+        // TODO: calculate min-SVN
```
jhand2, line 214: Same question, why todo

rom/dev/src/lock.rs
```diff
@@ -95,11 +95,14 @@ fn lock_common_reg_set(env: &mut RomEnv) {
...
         .lock_warm_reset_entry48(WarmResetEntry48::RtTci);
 
-    // Lock the Runtime SVN in data vault until next reset
+    // Lock the Firmware SVN in data vault until next reset
     env.data_vault
-        .lock_warm_reset_entry4(WarmResetEntry4::RtSvn);
+        .lock_warm_reset_entry4(WarmResetEntry4::FwSvn);
 
     // Do not lock Runtime minimum SVN; FMC will manage this.
+    // TODO: Lock the Firmware Min-SVN in data vault until next reset
```
jhand2, line 103: Same

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
```diff
@@ -331,14 +333,14 @@ fn test_pcr_log_fmc_fuse_svn() {
...
-    const FMC_FUSE_SVN: u32 = 2;
+    const FW_SVN: u32 = 3;
+    const FW_FUSE_SVN: u32 = 2;
 
     let fuses = Fuses {
         anti_rollback_disable: false,
         key_manifest_pk_hash: vendor_pubkey_digest,
         owner_pk_hash: owner_pubkey_digest,
-        fmc_key_manifest_svn: FMC_FUSE_SVN,
+        fw_svn: [2, 0, 0, 0], // Value of FW_FUSE_SVN
```
jhand2, line 343: nit: Would be nice to have a function do this instead of keeping them in sync manually

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
```diff
@@ -472,12 +476,11 @@ fn test_pcr_log_across_update_reset() {
...
         .unwrap();
 
-    const FMC_SVN: u32 = 2;
-    const FMC_FUSE_SVN: u32 = 1;
+    const FW_SVN: u32 = 2;
 
     let fuses = Fuses {
         anti_rollback_disable: false,
-        fmc_key_manifest_svn: FMC_FUSE_SVN,
+        fw_svn: [1, 0, 0, 0],
```
jhand2, line 483: same comment about just defining the constant in one place.

jhand2, line 483: Is it intentional that the array mismatches the u32 constant? in the test above this they match

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
```diff
@@ -573,12 +578,11 @@ fn test_pcr_log_across_update_reset() {
...
 fn test_fuse_log() {
-    const FMC_SVN: u32 = 4;
+    const FW_SVN: u32 = 4;
+    const FW_SVN_FUSE: u32 = 3;
 
     let fuses = Fuses {
-        anti_rollback_disable: true,
-        fmc_key_manifest_svn: 0x0F,  // Value of FMC_SVN
-        runtime_svn: [0xF, 0, 0, 0], // Value of RT_SVN
+        fw_svn: [0x7, 0, 0, 0], // Value of FW_SVN_FUSE
```
jhand2, line 585: Same, would be helpful to have code convert this so it's more obvious where this value comes from.

test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -461,7 +465,8 @@ fn smoke_test() {
         Some(DiceTcbInfo {
             vendor: Some("Caliptra".into()),
             model: Some("RT".into()),
-            svn: Some(0x100),
+            // This is from the SVN in the image (9)
+            svn: Some(0x109),
```
jhand2, line 469: Why is there an extra byte?

drivers/src/fuse_log.rs
bluegate010: Done.

error/README.md
bluegate010: That's in main, not main-2.x. Will drop changes in this file anyways.

error/src/lib.rs
bluegate010: The error code was defined but there was never any code that emitted it. (Noted [here](https://github.com/chipsalliance/caliptra-sw/blob/926ad4ae45b3baf1e935b80f7b4dfca178014a76/test/tests/fips_test_suite/fw_load.rs#L817) in existing test code.)

error/src/lib.rs
bluegate010: Fixed.

hw-model/src/lib.rs
bluegate010: Ack.

image/verify/src/verifier.rs
bluegate010: I was trying to avoid changing the image layout, for backwards compatibility. The FMC and RT FW images within the bundle each separately carry an SVN term. Tooling now writes the same value down when generating an image, and ROM ensures that the two are equal.

Have added a note to that effect.

rom/dev/doc/error-attribution.md
bluegate010: Agreed. In the meantime though will try and avoid making the problem worse.

rom/dev/src/flow/cold_reset/fw_processor.rs
bluegate010: The code works, but I was trying to avoid more functional changes in this PR. I intend to send a follow-up to make this functional change (along with the other TODOs.)

rom/dev/src/flow/update_reset.rs
bluegate010: Was trying to avoid more functional changes in this PR.

rom/dev/src/lock.rs
bluegate010: Ditto.

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
bluegate010: Have noted as a TODO.

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
bluegate010: Ditto.

bluegate010: Ah no, that was a mistake in transcription. Fixed.

rom/dev/tests/rom_integration_tests/test_fmcalias_derivation.rs
bluegate010: Agreed. Will add in a follow-up.

test/tests/caliptra_integration_tests/smoke_test.rs
bluegate010: Looks like it's to keep the encoded width constant.

PR #1768, opened 2024-11-03
[update] Make image verification with PQC algorithms mandatory

mhatrevi:
This change makes image verifiation with PQC algorithms (LMS for now, MLDSA in future) mandatory.


drivers/src/fuse_bank.rs
```diff
@@ -63,10 +63,21 @@ impl From<IdevidCertAttr> for usize {
...
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum RomVerifyConfig {
-    EcdsaOnly = 0,
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum RomPqcVerifyConfig {
+    #[default]
     EcdsaAndLms = 1,
+    EcdsaAndMldsa = 2,
```
ArthurHeymans, line 70: I'm a bit confused. Lms is made non optional throughout most of this patch, however this dicates that it's still optional if MlDsa is used?

mhatrevi: Yeah, PQC i.e. either LMS or MLDSA FW validation is made mandatory with this change.

swenson (approved):
LGTM.

I assume that there will be a separate code path for ML-DSA signatures.

PR #1770, opened 2024-11-04
rdl: Mount external memory as a register array

swenson:
This makes external memory look like an array of registers. 64-bit and 128-bit memory arrays are converted to 32-bit, since we don't have a great story for accessing more than 32 bits at a time anyway from RISC-V.
    
This only affects MLDSA and I3C (which uses 128-bit memory access).


swenson:
Thanks!

PR #1771, opened 2024-11-04
emulator: Tock registers update

swenson:
MCU will be running the Caliptra emulator but also has other Tock libraries. The version of `tock-registers` is currently out of date, and the latest published crate is as well, so we use the git commit dependency as MCU will be using.

This only affects the emulator.


ArthurHeymans (approved):

leongross (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #1774, opened 2024-11-05
Add alignment check to PersistentData structure.

clundin25:
We should also check that the address after the `PersistentData` struct is `DATA_ORG`. Currently we are mostly checking that the struct sizes add up.


jhand2 (approved):

clundin25:
Closing this in favor of https://github.com/chipsalliance/caliptra-sw/pull/1722, which includes this check (amongst other changes).

PR #1775, opened 2024-11-05
Tweak UART messages to buy some RO memory in ROM.

clundin25:
https://github.com/chipsalliance/caliptra-sw/pull/1732 pushes the code size too far. Re-coup some of the code-size for the UART image by compressing some of the existing messages.

### DIFF
--- a/rom/dev/src/wdt.rs
+++ b/rom/dev/src/wdt.rs
@@ -36,7 +36,7 @@ pub fn start_wdt(soc_ifc: &mut SocIfc) {
             wdt_timeout_cycles = 1;
         }
         cprintln!(
-            "[state] Starting the Watchdog Timer {} cycles",
+            "[state] Starting the WD Timer {} cycles",
             wdt_timeout_cycles
         );
         caliptra_common::wdt::start_wdt(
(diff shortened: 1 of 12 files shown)


jhand2 (approved):

jhand2 (approved):

rom/dev/src/wdt.rs
```diff
@@ -35,17 +33,9 @@ pub fn start_wdt(soc_ifc: &mut SocIfc) {
...
-            wdt_timeout_cycles
-        );
         caliptra_common::wdt::start_wdt(
             soc_ifc,
             WdtTimeout::from(core::num::NonZeroU64::new(wdt_timeout_cycles).unwrap()),
         );
-    } else {
-        cprintln!(
-            "[state] Watchdog Timer is not started because the device is not locked for debugging"
-        );
```
jhand2, line 49: These might be worth leaving in (and shortening) since I think they help quite a bit when debugging-related test failures (which happen from time-to-time if people add code that makes boot take longer in some area)

clundin25: Sounds good. I'll revert this. This saved quite a bit of space, so the next commit may touch more places.

clundin25: @jhand2 I double checked and I think I can make it work without this change. If I need more space I'll open a separate PR.

mhatrevi (approved):

jhand2 (approved):

PR #1780, opened 2024-11-08
ci: add dummy workflow for building ROM release from versioned git ref

leongross:
Add a dummy workflow as suggested by @jhand2 in #1691 to shift execution to the appropriate runners.


jhand2 (approved):

benjamindoron:
Thanks for getting to this!

PR #1784, opened 2024-11-12
Increase runtime stack size to 64KiB

clundin25:


jhand2 (approved):

MDr164 (approved):

mhatrevi (changes requested):
What is the motivation for this much increase in stack size? Reducing the DATA_SIZE to 22 K will be insufficient for storing Caliptra 2.0 MLDSA related artifacts (TBSes, public keys, signatures etc)

clundin25:
> What is the motivation for this much increase in stack size? Reducing the DATA_SIZE to 22 K will be insufficient for storing Caliptra 2.0 MLDSA related artifacts (TBSes, public keys, signatures etc)

The current maximum observed stack usage is ~54,208 bytes. 

After this series of PRs is merged, I will add https://github.com/chipsalliance/caliptra-sw/pull/1785 to prevent increases in stack usage.

With that in place we can work on reducing the stack usage. I think 48KiB can be obtainable relatively quickly.

jhand2:
> What is the motivation for this much increase in stack size? Reducing the DATA_SIZE to 22 K will be insufficient for storing Caliptra 2.0 MLDSA related artifacts (TBSes, public keys, signatures etc)

This is more of a reflection of reality than an actual increase. But ya, as Carl said, we will continue work to reduce stack usage.

mhatrevi (approved):

Issue #1735, opened 2024-10-24
Add stack overflow detection to SW emulator


clundin25:
https://github.com/chipsalliance/caliptra-sw/pull/1722 uncovered a latent stack overflow. We need to add tests / checks / ci to make sure we are not hitting any stack overflows.

ArthurHeymans:
> https://github.com/chipsalliance/caliptra-sw/pull/1722 uncovered a latent stack overflow. We need to add tests / checks / ci to make sure we are not hitting any stack overflows.

https://github.com/rust-embedded/cortex-m-rt/issues/34 this explains a zero cost abstraction to achieve it

clundin25:
That looks like a good approach (cross-linking the tracking bug https://github.com/chipsalliance/caliptra-sw/issues/1665).

Adding stack monitoring in the simulator has a few advantages:
* Doesn't touch any of the prod artifacts.
* Allows for data collection, so we can monitor for code paths that are close to overflowing.

I think it makes sense to implement both.

ArthurHeymans:
> That looks like a good approach (cross-linking the tracking bug #1665).
> 
> Adding stack monitoring in the simulator has a few advantages:
> 
> * Doesn't touch any of the prod artifacts.
> * Allows for data collection, so we can monitor for code paths that are close to overflowing.
> 
> I think it makes sense to implement both.

That makes sense.
Also I think hitless update might be impossible to achieve when changing the stack location. Although I think it's a good idea to do that in 2.0 where the rom is not yet finalized.

PR #1785, opened 2024-11-12
Add stack overflow detection to sw-emulator

clundin25:
Added stack monitoring and overflow detection to the SW emulator

If a stack overflow is detected, the emulator will panic and cause test failure.

This resolves https://github.com/chipsalliance/caliptra-sw/issues/1735

### DIFF
--- a/sw-emulator/lib/cpu/src/cpu.rs
+++ b/sw-emulator/lib/cpu/src/cpu.rs
@@ -22,6 +22,126 @@ use caliptra_emu_types::{RvAddr, RvData, RvException, RvSize};
 
 pub type InstrTracer<'a> = dyn FnMut(u32, RvInstr) + 'a;
 
+/// Describes a Caliptra stack memory region
+pub struct StackRange(u32, u32);
+impl StackRange {
+    /// **Note:** `stack_start` MUST be greater than `stack_end`. Caliptra's stack grows
+    /// to a lower address.
+    pub fn new(stack_start: u32, stack_end: u32) -> Self {
+        if stack_start < stack_end {
+            panic!("Caliptra's stack grows to a lower address");
+        }
+        Self(stack_start, stack_end)
+    }
+}
+
+/// Describes a Caliptra code region
+pub struct CodeRange(u32, u32);
+impl CodeRange {
+    pub fn new(code_start: u32, code_end: u32) -> Self {
+        if code_start > code_end {
+            panic!("code grows to a higher address");
+        }
+        Self(code_start, code_end)
+    }
+}
+
+/// Contains metadata describing a Caliptra image
+pub struct ImageInfo {
+    stack_range: StackRange,
+    code_range: CodeRange,
+}
+
+impl ImageInfo {
+    pub fn new(stack_range: StackRange, code_range: CodeRange) -> Self {
+        Self {
+            stack_range,
+            code_range,
+        }
+    }
+
+    /// Checks if the program counter is contained in `self`
+    ///
+    /// returns `true` if the pc is in the image. `false` otherwise.
+    fn contains_pc(&self, pc: u32) -> bool {
+        self.code_range.0 <= pc && pc <= self.code_range.1
+    }
+
+    /// Checks if the stack pointer has overflowed.
+    ///
+    /// Returns `Some(u32)` if an overflow has occurred. The `u32` is set
+    /// to the overflow amount.
+    ///
+    /// Returns `None` if no overflow has occurred.
+    fn check_overflow(&self, sp: u32) -> Option<u32> {
+        let stack_end = self.stack_range.1;
+
+        // Stack grows to a lower address
+        if sp < stack_end {
+            let current_overflow = stack_end - sp;
+            Some(current_overflow)
+        } else {
+            None
+        }
+    }
+}
+
+/// Describes the shape of Caliptra's stacks.
+///
+/// Used to monitor stack usage and check for overflows.
+pub struct StackInfo {
+    images: Vec<ImageInfo>,
+    max_stack_overflow: u32,
+    has_overflowed: bool,
+}
+
+impl StackInfo {
+    /// Create a new `StackInfo` by describing the start and end of the stack and code segment for each
+    /// Caliptra image.
+    pub fn new(images: Vec<ImageInfo>) -> Self {
+        Self {
+            images,
+            max_stack_overflow: 0,
+            has_overflowed: false,
+        }
+    }
+}
+
+impl StackInfo {
+    /// Fetch the largest stack overflow.
+    ///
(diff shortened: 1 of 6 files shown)


sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -262,7 +339,27 @@ impl<TBus: Bus> Cpu<TBus> {
...
+        Ok(())
+    }
+
+    // Check if the stack overflows at the requested address.
+    //
+    // Panics if the stack will overflow.
+    fn check_stack(&mut self, val: RvData) {
+        if let Some(stack_info) = &mut self.stack_info {
+            if stack_info.check_overflow(val) {
+                panic!(
```
clundin25, line 356: Alternatively, this could be a print statement and the panic ocurring in `Cpu::drop`.

This would allow the max stack overflow to be observed.

clundin25: I'm making this change. It makes it a lot easier to root cause stack regressions.

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -22,6 +22,77 @@ use caliptra_emu_types::{RvAddr, RvData, RvException, RvSize};
...
+            Some(self.stack_end - self.lowest_stack_address)
+        } else {
+            None
+        }
+    }
+
+    /// Checks if the stack will overflow when pushed to `stack_address`.
+    ///
+    /// Returns `true` if the stack will overflow, `false` if it will not overflow.
+    pub fn check_overflow(&mut self, stack_address: u32) -> bool {
```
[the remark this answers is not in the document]
clundin25: I'll work on adding that change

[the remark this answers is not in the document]
clundin25: Updated to check multiple code ranges / stacks

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -22,6 +22,126 @@ use caliptra_emu_types::{RvAddr, RvData, RvException, RvSize};
...
+    }
+}
+
+/// Describes the shape of Caliptra's stacks.
+///
+/// Used to monitor stack usage and check for overflows.
+pub struct StackInfo {
+    images: Vec<ImageInfo>,
+    max_stack_overflow: u32,
+    has_overflowed: bool,
```
jhand2, line 95: Should these also save which boot stage overflowed?

clundin25: The overflowing address will be logged to stderr on each overflow.

We can definitely save that information but I think it may be redundant with the existing logs.

Here is an example of what the log might look like:

```
[EMU] Caliptra's stack overflowed by 4032 bytes at pc 0x4000b476.
[EMU] Caliptra's stack overflowed by 4144 bytes at pc 0x4000b9e0.
[EMU] Caliptra's stack overflowed by 4032 bytes at pc 0x4000ba5a.
[EMU] Caliptra's stack overflowed by 4080 bytes at pc 0x400153d0.
[EMU] Caliptra's stack overflowed by 4032 bytes at pc 0x4001542a.
[EMU] Caliptra's stack overflowed by 4528 bytes at pc 0x4000b7da.
[EMU] Caliptra's stack overflowed by 4560 bytes at pc 0x40008c2e.
[EMU] Caliptra's stack overflowed by 4528 bytes at pc 0x40008cde.
[EMU] Caliptra's stack overflowed by 4560 bytes at pc 0x40008c2e.
[EMU] Caliptra's stack overflowed by 4528 bytes at pc 0x40008cde.
[EMU] Caliptra's stack overflowed by 4736 bytes at pc 0x4000b976.
[EMU] Caliptra's stack overflowed by 4752 bytes at pc 0x40008ce2.
[EMU] Caliptra's stack overflowed by 4912 bytes at pc 0x40008ac8.
[EMU] Caliptra's stack overflowed by 5056 bytes at pc 0x40008b66.
[EMU] Caliptra's stack overflowed by 4912 bytes at pc 0x40008bba.
[EMU] Caliptra's stack overflowed by 4752 bytes at pc 0x40008b62.
[EMU] Caliptra's stack overflowed by 4736 bytes at pc 0x40008d48.
[EMU] Caliptra's stack overflowed by 4528 bytes at pc 0x4000b9dc.
[EMU] Caliptra's stack overflowed by 4032 bytes at pc 0x4000b8de.
thread 'test_idevid_derivation::test_generate_csr' panicked at '[EMU] Fatal: Caliptra's stack overflowed by 5056 bytes!', /usr/local/google/home/clundin/code/stack-overflow-check/sw-emulator/lib/cpu/src/cpu.rs:287:17
```

jhand2 (approved):

mhatrevi (approved):

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -262,7 +401,24 @@ impl<TBus: Bus> Cpu<TBus> {
     ///
     /// * `RvException` - Exception with cause `RvExceptionCause::IllegalRegister`
     pub fn write_xreg(&mut self, reg: XReg, val: RvData) -> Result<(), RvException> {
-        self.xregs.write(reg, val)
+        // XReg::X2 is the sp register.
+        if reg == XReg::X2 {
```
swenson, line 405: Theoretically, we can set the `sp` to another value as a temporary, or when initializing (we set it to 0 at the beginning of the ROM), and this isn't necessarily a stack overflow, so we might have to be a little careful here.

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -262,7 +401,24 @@ impl<TBus: Bus> Cpu<TBus> {
     ///
     /// * `RvException` - Exception with cause `RvExceptionCause::IllegalRegister`
     pub fn write_xreg(&mut self, reg: XReg, val: RvData) -> Result<(), RvException> {
-        self.xregs.write(reg, val)
+        // XReg::X2 is the sp register.
```
swenson, line 404: It might be good to document why this works: LLVM appears to always pre-allocate the stack by subtracting, and then using only positive relative offsets to SP. (Is that guaranteed though?)

sw-emulator/lib/cpu/src/cpu.rs
clundin25: That's a good point. An actionable change would be adding a bounds check ensuring `sp` is not greater than the start of the stack region.

With that change edge cases will at least result in a failure.

sw-emulator/lib/cpu/src/cpu.rs
clundin25: I'll take a closer look at this.

PR #1797, opened 2024-11-16
Update SOC Manifest for Caliptra 2.0

mlvisaya:
- Remove Image Source
- Add Image Identifier
- Add ComponentInfo fields from PLDM
- Add Load Address and Entry point


auth-manifest/README.md
```diff
@@ -3,37 +3,50 @@
...
+| Image Load Address Low  | 4            | The lower 4-bytes of the 64-bit AXI address where the image will be loaded for verification and execution or staging                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
swenson, line 39: nit: "4-byte" (with hyphen) is for when it is an adjective, but "4 bytes" is for when it is a noun.

```suggestion
| Image Load Address High | 4            | The higher 4 bytes of the 64-bit AXI address where the image will be loaded for verification and execution or staging.                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| Image Load Address Low  | 4            | The lower 4 bytes of the 64-bit AXI address where the image will be loaded for verification and execution or staging                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```

swenson (commented):
Minor nits only

auth-manifest/README.md
```diff
@@ -3,37 +3,50 @@
...
+| Image Identifier        | 4            | Unique value selected by the vendor to distinguish between component images.<br />IMEs should be sorted by an ascending order of Image Identifier within an IMC.                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
mhatrevi, line 36: This is not a requirement. Runtime will sort the entries internally for faster lookup. The only requirement is that these Ids should be unique.

auth-manifest/README.md
```diff
@@ -3,37 +3,50 @@
...
+| ImageSize               | 4            | Size in bytes of the image.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
 
 - ### **Image Metadata Collection**
+
 The Image Metadata Collection (IMC) is a collection of Image Metadata entries (IME). Each IME has a hash that matches a SOC images. The manifest vendor and owner private keys sign the IMC. The Preamble holds the IMC signatures. The manifest IMC vendor signatures are optional and are validated only if the FLAGS field Bit 0 = 1. Up to sixteen image hashes are supported.
```
mhatrevi, line 47: This should be 127.

mhatrevi (changes requested):
Please see the feedback.

auth-manifest/README.md
mlvisaya: got it. ill update

mhatrevi (approved):

PR #1802, opened 2024-11-19
[ROM] Only inspect RT FW's SVN during image validation.

bluegate010:
This is done as part of Stable Identity, which requires Caliptra to have a single SVN. As part of this work, FMC's SVN is being deprecated. Instead, the runtime SVN now describes the security state of the entire firmware image. This is safe because the FMC and runtime images are always signed, validated, and updated in tandem.

Future changes will update build tooling to prevent callers from setting FMC's SVN.


rom/dev/README.md
```diff
@@ -62,7 +62,7 @@ Following are the main FUSE & Architectural Registers used by the Caliptra ROM f
 | FUSE_LMS_REVOCATION             | 32           | Manufacturer LMS Public Key Revocation Mask             |
 | FUSE_MLDSA_REVOCATION           | 32           | Manufacturer MLDSA Public Key Revocation Mask           |
 | FUSE_OWNER_PK_HASH              | 384          | Owner ECC and LMS or MLDSA Public Key Hash              |
-| FUSE_FMC_KEY_MANIFEST_SVN       | 32           | FMC Security Version Number                             |
+| FUSE_FMC_KEY_MANIFEST_SVN       | 32           | FMC Security Version Number (Deprecated)                |
```
jhand2, line 65: nit: Do we want to reclaim these fuses for something else in 2.0? If so, we may want to mark it as reserved or something.

jhand2 (dismissed):

rom/dev/README.md
bluegate010: Should I just delete the row from here?

jhand2: Ya I think that makes sense

bluegate010: Done.

image/verify/src/verifier.rs
```diff
@@ -112,10 +112,32 @@ impl<Env: ImageVerificationEnv> ImageVerifier<Env> {
...
 
         // Verify FMC
-        let (fmc_info, fmc_log_info) = self.verify_fmc(image_info.fmc, reason)?;
+        let fmc_info = self.verify_fmc(image_info.fmc, reason)?;
 
         // Verify Runtime
-        let (runtime_info, rt_log_info) = self.verify_runtime(image_info.runtime)?;
+        let runtime_info = self.verify_runtime(image_info.runtime)?;
+
+        // The FMC and RT FW images within the bundle each include an SVN, for
```
mhatrevi, line 120: Suggest putting this logic in a function (eg verify_svn) that can return the effective fuse svn on success.

mhatrevi (approved):

image/verify/src/verifier.rs
bluegate010: Have factored this into a helper function. I think it's best to leave the effective fuse SVN calculation as its own helper, since those are different tasks.

mhatrevi (approved):

jhand2 (approved):

PR #1809, opened 2024-11-21
Fixing release CI failures and adding frozen 1.1 ROM

nquarton:


.github/workflows/fpga.yml
```diff
@@ -420,11 +420,6 @@ jobs:
           TEST_BIN=/tmp/caliptra-test-binaries
           VARS="CPTRA_UIO_NUM=4 CALIPTRA_PREBUILT_FW_DIR=/tmp/caliptra-test-firmware CALIPTRA_IMAGE_NO_GIT_REVISION=1"
 
-          if [[ "${{ inputs.workflow_call }}" && "${{ inputs.hw-version }}" != "latest" ]]; then
```
nquarton, line 423: This is now chosen when we use the ci-rom-1.0 or ci-rom1.1 feature in the fips testing

runtime/Cargo.toml
```diff
@@ -65,3 +65,5 @@ no-cfi = ["caliptra-image-verify/no-cfi", "caliptra-drivers/no-cfi"]
 fpga_realtime = ["caliptra-drivers/fpga_realtime"]
 "hw-1.0" = ["caliptra-builder/hw-1.0", "caliptra-drivers/hw-1.0", "caliptra-registers/hw-1.0", "caliptra-kat/hw-1.0","caliptra-cpu/hw-1.0"]
 fips-test-hooks = ["caliptra-drivers/fips-test-hooks"]
+"ci-rom-1.0" = ["caliptra-builder/ci-rom-1.0"]
```
korran, line 68: I'm really nervous adding features to the entire crate, as it will be tempting for the runtime firmware to look at these and change it's behavior (which will then break in prod when these features aren't set).

Alternatives:

* use environment variables to tell the test which logic to execute (my preferred)
* Move the tests to a separate crate that only has tests, and add this feature there.

korran: Thinking about this a bit more, maybe this is not a problem as long as caliptra-builder NEVER passes this feature when it builds the firmware.

That said, the existing rom_from_env() function uses an environment variable to decide between UART and non-UART, maybe it could do something similar for the ROM version....

nquarton: This change has been made. Instead of features, it uses the environment variable "CPTRA_CI_ROM_VERSION". get_ci_rom_version() provides the version to both the tests and the function that provides the ROM binary (rom_for_fw_integration_tests).

nquarton:
This run shows it solving all the test failures in the nightly. (The final publish step fails as expected since this is not the main branch).
https://github.com/chipsalliance/caliptra-sw/actions/runs/11976900101

jlmahowa-amd (approved):

mhatrevi (approved):

PR #1811, opened 2024-11-21
Changing hw/latest to track patch_v1.1 branch in RTL repo

nquarton:


jlmahowa-amd (approved):

jhand2 (dismissed):

.gitmodules
```diff
@@ -4,7 +4,7 @@
 [submodule "hw/latest/rtl"]
 	path = hw/latest/rtl
 	url = https://github.com/chipsalliance/caliptra-rtl
-	branch = main
+	branch = patch_v1.1
```
jhand2, line 7: lgtm, but I think you may need to commit the branch change so that git knows which commit the submodule is at.

nquarton: Can you elaborate on that? The branch does exist in the RTL repo. But I may be misunderstanding something about the submodule behavior here.

jlmahowa-amd (dismissed):

PR #1820, opened 2024-11-26
Add a new crypto helper and move the module up a layer.

bluegate010:
The move will allow crypto helpers to be used from runtime-update flows.

The new helper will be used in cases where borrowing RomEnv is not allowed.


swenson (approved):
LGTM

mhatrevi (approved):

jhand2 (approved):

PR #1819, opened 2024-11-26
[feat] Updating HMAC KDF from 384 to 512 for ROM

mhatrevi:

PR #1823, opened 2024-11-26
Change runtime's HMAC helper from a simple HMAC to a KDF.

bluegate010:
In order to use the resulting key in FIPS, each operation will need to be an approved KDF.

### DIFF
--- a/runtime/src/hmac.rs
+++ b/runtime/src/hmac.rs
@@ -82,20 +82,29 @@ fn ecc384_key_gen(
 pub enum Hmac {}
 
 impl Hmac {
-    /// "Hash" the data in the provided KV slot by HMACing it with an empty slice.
-    /// This mechanism is necessary because the hardware does not directly support
-    /// hashing data in KV slots.
+    /// Calculate HMAC-384 KDF
     ///
     /// # Arguments
     ///
     /// * `drivers` - Drivers
-    /// * `input` - KeyId containing the input data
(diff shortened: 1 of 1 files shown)


runtime/src/hmac.rs
```diff
@@ -81,20 +81,28 @@ fn ecc384_key_gen(
...
+    /// * `key` - HMAC384 key slot
+    /// * `label` - Input label
+    /// * `context` - Input context
+    /// * `output` - Key slot to store the output
     #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
-    pub fn hmac384_hash(drivers: &mut Drivers, input: KeyId, output: KeyId) -> CaliptraResult<()> {
-        drivers.hmac384.hmac(
-            &KeyReadArgs::new(input).into(),
-            &HmacData::Slice(&[]),
+    pub fn hmac384_kdf(
```
swenson, line 94: Is this code even used, or plan on being used?

With #1819, will this need to be updated to HMAC-SHA-512?

bluegate010: This function was added a bit ago in anticipation of using it to use/test the hash chain feature. However the hash chain implementation never landed. Once the outstanding PRs are merged I can send the implementation for review, where this will be used.

And yeah we may as well make the hash chain use SHA512.

jhand2 (approved):

PR #1824, opened 2024-11-27
emulator: Add Zba,Zbc,Zbs instruction sets

swenson:
And enables support in code generation to use them. This saves 44, 8, and 168 bytes in the ROM, FMC, and Runtime, respectively.

This has the negative side effect of making tests a little slower to run due to the quantity of macros used in generating all of the test cases.

This work extends the previous Zbb bit manipulation instruction set we supported to include the rest of the b family, all of which are supported by our VeeR CPU.

We also add missing tests for the Zbb instruction set.

(This work was done to support the MCU firmware, which needs the space and performance benefits more. So there's no reason not to bring it back here for "free".)


ArthurHeymans (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #1826, opened 2024-12-01
[feat] Update SHA2 driver to support SHA2-512 digests

mhatrevi:


swenson (dismissed):
LGTM

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -65,8 +65,8 @@ impl FirmwareProcessor {
             // sha256
             sha256: &mut env.sha256,
 
-            // SHA2-384 Engine
-            sha384: &mut env.sha384,
+            // SHA2-512/384 Engine
+            sha2_512_384: &mut env.sha2_512_384,
```
jhand2, line 69: minor nit: I would probably just call this driver sha2_512 and then also include sha384 functions on it. Since SHA384 effectively is just a variant of sha512 (different IV + truncated). But not a huge deal.

drivers/src/sha2_512_384.rs
```diff
@@ -4,50 +4,65 @@ Licensed under the Apache-2.0 license.
...
 use caliptra_cfi_derive::cfi_impl_fn;
 use caliptra_error::{CaliptraError, CaliptraResult};
 use caliptra_registers::sha512::Sha512Reg;
 
-const SHA384_BLOCK_BYTE_SIZE: usize = 128;
-const SHA384_BLOCK_LEN_OFFSET: usize = 112;
-const SHA384_MAX_DATA_SIZE: usize = 1024 * 1024;
+const SHA2_BLOCK_BYTE_SIZE: usize = 128;
+const SHA2_BLOCK_LEN_OFFSET: usize = 112;
+const SHA2_MAX_DATA_SIZE: usize = 1024 * 1024;
```
jhand2, line 27: naming nit: Different SHA2 modes has different block sizes, it's just that 384 and 512 have the same block size. Similar to but other comment, I would probably just name these as `SHA512_*`

drivers/src/sha2_512_384.rs
```diff
@@ -94,18 +174,19 @@ impl Sha384 {
...
                     // cannot reason about `offset` parameter to optimize out
                     // the panic call.
-                    if let Some(slice) = buf.get(offset..offset + SHA384_BLOCK_BYTE_SIZE) {
-                        let block = <&[u8; SHA384_BLOCK_BYTE_SIZE]>::try_from(slice).unwrap();
-                        self.digest_block(block, first, false)?;
-                        bytes_remaining -= SHA384_BLOCK_BYTE_SIZE;
+                    if let Some(slice) = buf.get(offset..offset + SHA2_BLOCK_BYTE_SIZE) {
+                        let block = <&[u8; SHA2_BLOCK_BYTE_SIZE]>::try_from(slice).unwrap();
+                        self.digest_block(ShaMode::Sha512, block, first, false)?;
+                        bytes_remaining -= SHA2_BLOCK_BYTE_SIZE;
```
jhand2, line 180: `sha512_digest` and `sha384_digest` are very similar functions, they just have different return types and different modes. Can have a single core implementation to avoid duplicating all this code?

jhand2: Probably can do it similarly to how you parameterized the Sha2DigestOp type.

drivers/src/sha2_512_384.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: I named as such it to keep the name consistent with sha sha2_512_384_acc.

drivers/src/sha2_512_384.rs
mhatrevi: Created a helper function to separate the common code.

jhand2 (approved):

swenson (approved):

PR #1828, opened 2024-12-02
[update] Manifest Based Authorization feature update.

mhatrevi:
This PR  makes the following changes to the feature:

1. Adds a 'firmware id' to the image metadata entry (IME) structure.
2. Reduces max IME limit from 128 to 127 to account for size increase due to the firmware id addition.
3. SET_AUTH_MANIFEST command validates that the IME collection does not contain duplicate firmware ids.
4. SET_AUTH_MANIFEST command sorts the IME on the firmware ids.
5. Adds a new bit to the IME flags named 'SKIP AUTHORIZATION'. If set, AUTHORIZE_AND_STASH command authorizes an IME with a matching firmware id ignoring the image hash.
6. If a firmware id is not found, AUTHORIZE_AND_STASH command returns IMAGE_NOT_AUTHORIZED status code in the command output.
7. If a firmware id is found but the image hash does not match, AUTHORIZE_AND_STASH returns IMAGE_HASH_MISMATCH status code in the command output.
8. If a firmware id is found and the image hash matches or if the 'SKIP AUTHORIZATION' flag is set for the firmware id, AUTHORIZE_AND_STASH command returns IMAGE_AUTHORIZED status code in the command output.
9. AUTHORIZE_AND_STASH command uses binary search on the firmware id to lookup an IME.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -1059,8 +1059,8 @@ impl AuthAndStashFlags {
 #[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
 pub struct AuthorizeAndStashReq {
     pub hdr: MailboxReqHeader,
-    pub metadata: [u8; 4],
-    pub measurement: [u8; 48],
+    pub fw_id: [u8; 4],
+    pub measurement: [u8; 48], // Image digest.
     pub context: [u8; 48],
     pub svn: u32,
     pub flags: u32,
@@ -1070,7 +1070,7 @@ impl Default for AuthorizeAndStashReq {
     fn default() -> Self {
         Self {
             hdr: Default::default(),
-            metadata: Default::default(),
+            fw_id: Default::default(),
             measurement: [0u8; 48],
             context: [0u8; 48],
             svn: Default::default(),
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -725,7 +725,7 @@ Command Code: `0x4154_4D4E` ("ATMN")
 | preamble\_marker              | u32          | Marker needs to be 0x4154_4D4E for the preamble to be valid                 |
 | preamble\_size                | u32          | Size of the preamble                                                        |
 | preamble\_version             | u32          | Version of the preamble                                                     |
-| preamble\_flags               | u32          | Preamble flags                                                              |
+| preamble\_flags               | u32          | Manifest flags. See AUTH_MANIFEST_FLAGS below |
 | preamble\_vendor\_ecc384\_key | u32[24]      | Vendor ECC384 key with X and Y coordinates in that order                    |
 | preamble\_vendor\_lms\_key    | u32[6]       | Vendor LMS-SHA192-H15 key                                                   |
 | preamble\_vendor\_ecc384\_sig | u32[24]      | Vendor ECC384 signature                                                     |
@@ -739,7 +739,7 @@ Command Code: `0x4154_4D4E` ("ATMN")
 | metadata\_owner\_ecc384\_sig  | u32[24]      | Metadata Owner ECC384 signature                                             |
 | metadata\_owner\_LMS\_sig     | u32[1344]    | Metadata Owner LMOTS-SHA192-W4 signature                                    |
 | metadata\_entry\_entry\_count | u32          | number of metadata entries                                                  |
-| metadata\_entries             | MetaData[128] | The max number of metadata entries is 128 but less can be used             |
+| metadata\_entries             | Metadata[127] | The max number of metadata entries is 127 but less can be used             |
 
 
 *Table: `AUTH_MANIFEST_FLAGS` input flags*
@@ -750,10 +750,19 @@ Command Code: `0x4154_4D4E` ("ATMN")
 
 *Table: `AUTH_MANIFEST_METADATA_ENTRY` digest entries*
 
-| **Name**      | **Type** | **Description**        |
-|---------------|----------|------------------------|
-| digest        | u32[48]  | Digest of the metadata |
-| image\_source | u32      | Image source           |
+| **Name**      | **Type** | **Description**                |
+|---------------|----------|--------------------------------|
+| fw\_id        | u32      | Id of the image                |
+| flags         | u32      | See METADATA_ENTRY_FLAGS below |
+| digest        | u32[48]  | Digest of the image            |
+
+
+*Table: `METADATA_ENTRY_FLAGS` input flags*
+
+| **Name**            | **Size (Bits)** | **Description** |
+|---------------------|-----------------|-----------------|
+| image\_source       | 2               | 1: InRequest    |
+| ignore\_auth\_check | 1               | If set, the image digest is not compared for the firmware id |
 
 *Table: `SET_AUTH_MANIFEST` output arguments*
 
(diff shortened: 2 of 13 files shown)


mlvisaya:
Should we also add the changes to the README.md doc?

nquarton (commented):
Overall implementation looks good to me. I think the runtime readme will need some updates to reflect these changes.

You might also add a couple more test cases for the sorting and binary search. They look simple and correct to me. But I know corner cases can bite you on that. I'll still approve without it, just something to consider.

runtime/src/authorize_and_stash.rs
```diff
@@ -100,4 +112,46 @@ impl AuthorizeAndStashCmd {
...
+    ///
+    #[inline(never)]
+    fn find_metadata_entry<'a>(
+        auth_manifest_image_metadata_col: &'a AuthManifestImageMetadataCollection,
+        cmd_fw_id_bytes: &[u8; 4],
+    ) -> Option<&'a AuthManifestImageMetadata> {
+        let mut left = 0;
+        let mut right = auth_manifest_image_metadata_col.entry_count as usize;
+
+        while left < right {
```
swenson, line 139: Can we use `auth_manifest_image_metadata_col.image_metadata_list.binary_search_by_key()`?

Either way, we should add unit tests for this function.

runtime/src/set_auth_manifest.rs
```diff
@@ -401,6 +415,34 @@ impl SetAuthManifestCmd {
             soc_ifc,
         )?;
 
+        // Sort the image metadata list by firmware ID in place. Also check for duplicate firmware IDs.
+        let slice =
+            &mut metadata_mailbox.image_metadata_list[..metadata_mailbox.entry_count as usize];
```
swenson, line 420: Why can't we use `metadata_mailbox.image_metadata_list.sort_unstable_by_key()`?

If we need to check for duplicates, we can do it after sorting.

runtime/src/authorize_and_stash.rs
mhatrevi: Updated.

runtime/src/set_auth_manifest.rs
mhatrevi: sort_unstable_by_key is increasing the Runtime size by 2+KB.

runtime/README.md
```diff
@@ -790,7 +799,7 @@ Command Code: `0x4154_5348` ("ATSH")
 | --------      | -------- | ---------------
 | chksum            | u32      | Checksum over other output arguments, computed by Caliptra. Little endian. |
 | fips_status      | u32      | Indicates if the command is FIPS approved or an error.                     |
-| auth_req_result | u32      | AUTHORIZE_IMAGE: 0xDEADC0DE and DENY_IMAGE_AUTHORIZATION: 0x21523F21    |
+| auth_req_result | u32      | AUTHORIZE_IMAGE (0xDEADC0DE), IMAGE_NOT_AUTHORIZED (0x21523F21) or IMAGE_HASH_MISMATCH (0x8BFB95CB) |
```
nquarton, line 802: For most other commands (like signature verify commands) for a negative results, we report a mailbox command failure and put the reason in the non-fatal error code reg. Just want to call that out to be sure we want this to behave differently.

On the other hand, DPE does report a result that can be a negative result with a "mailbox complete". But that's maybe a bit of an exception since it's a whole separate submodule.

nquarton (dismissed):

api/src/mailbox.rs
```diff
@@ -1059,8 +1059,8 @@ impl AuthAndStashFlags {
 #[derive(Debug, AsBytes, FromBytes, PartialEq, Eq)]
 pub struct AuthorizeAndStashReq {
     pub hdr: MailboxReqHeader,
-    pub metadata: [u8; 4],
-    pub measurement: [u8; 48],
+    pub metadata: [u8; 4],     // Firmware Id.
```
mlvisaya, line 1062: show we just rename this as fw_id? so as not to confuse with metadata in "image metadata entry"

runtime/src/set_auth_manifest.rs
swenson: in that case, we should add some tests for this, and we can make it more efficient later

swenson (commented):
Looks good, but definitely I would want some tests on the sorting routine.

runtime/src/set_auth_manifest.rs
```diff
@@ -401,6 +415,34 @@ impl SetAuthManifestCmd {
...
+        for i in 0..slice.len() {
+            for j in 0..(slice.len() - 1 - i) {
+                // This check is needed to avoid out of bounds access panic.
+                if (j + 1) >= slice.len() {
+                    break;
+                }
+
+                match slice[j].fw_id.cmp(&slice[j + 1].fw_id) {
+                    core::cmp::Ordering::Greater => {
+                        slice.swap(j, j + 1);
```
swenson, line 430: If you use insertion sort instead (just changing the indices a little), then this runs something like 4x faster.

mhatrevi: Updated. Also added unit tests for this.

api/src/mailbox.rs
mhatrevi: Makes sense, updated.

runtime/README.md
mhatrevi: Ack.

mhatrevi:
> Looks good, but definitely I would want some tests on the sorting routine.

Added unit tests for the same.

nquarton (approved):

swenson (approved):
LGTM

runtime/src/authorize_and_stash.rs
```diff
@@ -100,4 +112,31 @@ impl AuthorizeAndStashCmd {
...
+    ///   otherwise returns `None`.
+    ///
+    #[inline(never)]
+    fn find_metadata_entry(
+        auth_manifest_image_metadata_col: &AuthManifestImageMetadataCollection,
+        cmd_fw_id: u32,
+    ) -> Option<&AuthManifestImageMetadata> {
+        auth_manifest_image_metadata_col
+            .image_metadata_list
+            .binary_search_by(|metadata| metadata.fw_id.cmp(&cmd_fw_id))
```
jhand2, line 138: Are the entries sorted? I would expect them not to be.

runtime/src/set_auth_manifest.rs
```diff
@@ -401,6 +415,45 @@ impl SetAuthManifestCmd {
             soc_ifc,
         )?;
 
+        // Sort the image metadata list by firmware ID in place. Also check for duplicate firmware IDs.        let slice =
+        let slice =
+            &mut metadata_mailbox.image_metadata_list[..metadata_mailbox.entry_count as usize];
+
+        Self::sort_and_check_duplicate_fwid(slice)?;
```
jhand2, line 422: This seems unnecessarily complicated. I can't imagine linear search being noticeably slow here.

runtime/src/authorize_and_stash.rs
mhatrevi: They are sorted in the SET_AUTH_MANIFEST command processing.

runtime/src/set_auth_manifest.rs
mhatrevi: Given we had 128 (now 127) hashes, there was some concern about the search runtimes.

jhand2 (approved):

PR #1834, opened 2024-12-04
Add documentation for the key ladder feature in ROM.

bluegate010:


rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
...
+
+When the hash chain is initialized at cold-boot, it is bound to the lifecycle state, debug-locked state, and the firmware's "epoch" from the image header. This ensures that across lifecycle or debug state transtions, or across intentional epoch changes, the keys in the hash chain will change.
+
+Across update-resets, ROM tracks the minimum SVN that has run since cold-boot. It ensures that the chain's length always corresponds to that minimum SVN. The hash chain can only be shortened (and thereby give access to newer SVNs' keys) by cold-booting into firmware with a newer SVN and re-initializing the chain.
+
+#### Cold-boot
+
+ROM initializes a hash chain for the firmware. LDevID CDI in Key Vault Slot6 is used as an HMAC Key, and the data is a fixed string. The resultant MAC is stored in Slot 2.
+
+    `HashChainContext = lifecycle state || debug_locked state || firmware epoch`
```
swenson, line 687: The backticks appear in the rendered output. Since this is indented, I think it is being monospaced anyway, even the "Loop" part.

rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
...
+
+    `hmac512_kdf(KvSlot6, label: b"hash_chain", context: HashChainContext, KvSlot2)`
```
swenson, line 689: nit: Maybe `stable_identity` instead of `hash_chain`, so that the intent rather than the mechanism is used?

rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
...
+
+    Loop (MAX_FIRMWARE_SVN - (current firmware SVN)) times:
+
+        `hmac512_kdf(KvSlot2, label: b"", context: None, KvSlot2)`
```
swenson, line 693: I think we should always put something in the label, even though it is probably not strictly necessary here.

rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
 
 See Firmware [Image Validation Process](#firmware-image-validation-process).
 
+### Derivation of the hash chain for Stable Identity
```
swenson, line 671: I think it would be more accurately called a KDF ratchet, or a KDF chain.

rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
...
+
+    `hmac512_kdf(KvSlot6, label: b"hash_chain", context: HashChainContext, KvSlot2)`
+
+    Loop (MAX_FIRMWARE_SVN - (current firmware SVN)) times:
+
+        `hmac512_kdf(KvSlot2, label: b"", context: None, KvSlot2)`
+
+#### Update-reset
+
+During update-reset, the chain initialized at cold boot is lengthened if necessary, such that its length always corresponds with the minimum SVN since cold boot.
```
swenson, line 697: So we can only rollback, and never roll forward, during update reset?

rom/dev/README.md
bluegate010: Yes, I've been thinking I should rename that. Changed to "key ladder".

rom/dev/README.md
bluegate010: Ah, I was mirroring style from elsewhere in this spec, such as [here](https://github.com/chipsalliance/caliptra-sw/blob/cbd59bfabb9c9e1a4e8ea1dbc859741fec05965b/rom/dev/README.md?plain=1#L445). Have removed though.

rom/dev/README.md
bluegate010: Done.

rom/dev/README.md
bluegate010: Done, though I think what we really want is just to ensure that any time a key is branched off from the chain, the label is different than what was used to extend it (in this case just make sure it's non-empty).

Have gone with "si_init" and "si_extend" for the two hash chain operations.

rom/dev/README.md
bluegate010: If an update-reset rolls the SVN forward from X to X+1, that will function, and that new firmware will boot. However, the new firmware won't have access to the keys for SVN X+1 until it cold-boots into that firmware (or any other firmware with SVN >= X+1).

I think the only way to allow the new firmware to have access to the new keys immediately upon update-reset is for ROM to regain access to the LDevID CDI on update-reset. As-is it only has access to AliasFMC CDI, and that CDI is not stable across FMC updates and so is not suitable for Stable Identity.

bluegate010: Have filed https://github.com/chipsalliance/caliptra-rtl/issues/668 to see if we can make that change.

swenson (approved):
LGTM

rom/dev/README.md
```diff
@@ -668,6 +668,42 @@ Following is the sequence of steps that are performed to download the firmware i
...
 
+### Derivation of the key ladder for Stable Identity
+
+Stable Identity calls for a secret that remains stable across firmware updates, but which can ratchet forward when major firmware vulnerabilities are fixed. Caliptra ROM implements this feature in terms of a "key ladder".
+
+The key ladder is initialized from the LDevID CDI during cold-boot. The key ladder length is inversely related to the firmware's SVN. Each step of the ladder is an SVN-unique key. The key for SVN X can be obtained by applying a one-way cryptographic operation to the key for SVN X+1. In this manner, firmware with a given SVN can wield keys bound to its SVN or older, but cannot wield keys bound to newer SVNs.
+
+To comply with FIPS, the one-way cryptographic operation used to compute keys is an SP 800-108 KDF.
+
+When the key ladder is initialized at cold-boot, it is bound to the lifecycle state, debug-locked state, and the firmware's "epoch" from the image header. This ensures that across lifecycle or debug state transtions, or across intentional epoch changes, the keys of the ladder will change.
```
jhand2, line 679: To clarify, epoch will need to be a new field in the image, correct? Is there already an issue to track this?

jhand2 (approved):

rom/dev/README.md
jhand2: nvm, it's already there.

PR #1845, opened 2024-12-09
[feat] IDEVID CSR Envelope Signing

mhatrevi:
This change creates the CSR envelope containing the ECC and MLDSA CSRs.


drivers/src/hmac.rs
```diff
@@ -492,12 +558,20 @@ impl Hmac {
 
         if first {
             // Submit the first block
-            hmac.hmac512_ctrl()
-                .write(|w| w.init(true).next(false).mode(mode == HmacMode::Hmac512));
+            hmac.hmac512_ctrl().write(|w| {
+                w.init(true)
+                    .next(false)
+                    .mode(hmac_mode == HmacMode::Hmac512)
```
swenson, line 564: Should we do some validation that the mode is 384 if `csr_mode` is true?

drivers/src/persistent.rs
```diff
@@ -47,62 +48,86 @@ pub type AuthManifestImageMetadataList =
...
-    pub fn get_csr_len(&self) -> u32 {
-        self.csr_len
     }
+}
 
-    /// Check if the CSR was unprovisioned
-    pub fn is_unprovisioned(&self) -> bool {
-        self.csr_len == Self::UNPROVISIONED_CSR
-    }
+macro_rules! impl_idevid_csr {
```
swenson, line 81: Maybe use generics instead of macros? It makes it easier to debug at least.

mhatrevi:
This draft PR is replaced by https://github.com/chipsalliance/caliptra-sw/pull/1868

PR #1846, opened 2024-12-10
Adding command to get owner pub key hash received with image bundle

nquarton:


JohnTraverAmd (approved):

nquarton:
Replaced with https://github.com/chipsalliance/caliptra-sw/pull/1854 which adds the hash to the FW_INFO command instead of creating a new command

PR #1850, opened 2024-12-13
[feat] FMC DICE changes for adding MLDSA support

mhatrevi:


common/src/crypto.rs
```diff
@@ -11,13 +11,103 @@ Abstract:
...
+/// # Arguments
+///
+/// * `sha256` - SHA256 driver
+/// * `data`   - Input data to hash
+///
+/// # Returns
+///
+/// * `Array4x8` - Digest
+pub fn sha256_digest(sha256: &mut Sha256, data: &[u8]) -> CaliptraResult<Array4x8> {
+    sha256.digest(data)
```
swenson, line 59: I'm not sure why we need this function, which just wraps a simple method call? (ditto for sha384)

common/src/x509.rs
```diff
@@ -0,0 +1,166 @@
...
+use crate::crypto::{self, PubKey};
+use caliptra_drivers::*;
+use core::mem::size_of;
+use core::usize;
+use zerocopy::AsBytes;
+
+/// X509 API
+pub enum X509 {}
+
+impl X509 {
```
swenson, line 23: Since these are already in a separate `mod`, it feels redundant to have an empty `enum` to contain. We could just have these be functions in this file without the `enum` and `impl`.

That way later we don't have to have the triply redundant `caliptra_x509::x509::X509::method_name()`.

rom/dev/src/flow/cold_reset/x509.rs
```diff
@@ -12,80 +12,19 @@ Abstract:
...
+    crypto::{self, PubKey},
+    x509,
+};
 use caliptra_drivers::*;
 use core::mem::size_of;
-use core::usize;
-use zerocopy::AsBytes;
 
 /// X509 API
 pub enum X509 {}
```
swenson, line 25: Ditto here about eliminating this empty `enum`.

common/src/x509.rs
mhatrevi: Updated.

common/src/crypto.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/x509.rs
mhatrevi: Udpated.

swenson (dismissed):
LGTM

nquarton (dismissed):
Can't say I got to look at every line in detail but the changes definitely make sense. The chunks of code reorganization are also an improvement.

nquarton (approved):

vsonims (approved):

PR #1854, opened 2024-12-14
Adding owner pub key hash to FW_INFO cmd

nquarton:
A method of obtaining the Owner public key hash from Caliptra is needed to assist the SoC with setting the owner pub key hash fuses. This is to eliminate any dependency by the SoC on the Caliptra image format and guarantee correct hashing of the keys.


rusty1968 (approved):

jhand2 (approved):
Thanks

JohnTraverAmd (approved):

PR #1855, opened 2024-12-16
[feat] UDS Programming Flow

mhatrevi:
This change adds the UDS programming flow to the ROM.


ArthurHeymans (dismissed):

drivers/src/soc_ifc.rs
```diff
@@ -365,6 +365,59 @@ impl SocIfc {
...
+
+    pub fn uds_fuse_row_granularity_64(&self) -> bool {
+        let config_val: u32 = self.soc_ifc.regs().cptra_hw_config().read().into();
+        ((config_val >> 6) & 1) != 0
+    }
+
+    pub fn fuse_controller_base_addr(&self) -> u64 {
+        let low = self.soc_ifc.regs().ss_otp_fc_base_addr_l().read();
+        let high = self.soc_ifc.regs().ss_otp_fc_base_addr_h().read();
+        (high as u64) << 32 | low as u64
```
ArthurHeymans, line 419: Maybe return AxiAddr instead of u64?

drivers/src/soc_ifc.rs
```diff
@@ -365,6 +365,59 @@ impl SocIfc {
...
+                } else {
+                    w.uds_program_success(false);
+                    w.uds_program_fail(true)
+                }
+            });
+    }
+
+    pub fn uds_seed_dest_base_addr_low(&self) -> u32 {
+        self.soc_ifc.regs().ss_uds_seed_base_addr_l().read()
+    }
```
ArthurHeymans, line 401: It's a bit strange that the upper part is unused?

nit: maybe also return an AxiAddr type and use .lo inside the UDS programming flow?

swenson (dismissed):
LGTM

drivers/src/soc_ifc.rs
mhatrevi: Let's maintain the return type as u64 since the caller needs to add an offset to the address. If the return type were AxiAddr, the caller would need to convert it back to u64 before adding the offset.

vsonims (approved):

PR #1858, opened 2024-12-18
emulator: Decode PMP NAPOT addresses correctly

swenson:
Decoding NAPOT (natural power of 2) PMP regions involves some bit twiddly math, and the current code did it incorrectly, which I discovered when testing against a Tock library.

I added some tests for the function as well.

I also extended the PMP registers out to 64, which should be how many we support in HW.

The spec has the following diagram, which is sort of helpful:

image


mhatrevi (approved):

swenson:
Thanks!

PR #1861, opened 2024-12-18
Fix driver/dma: Operate on 64bit addresses

ArthurHeymans:
This fixes the driver side to operate on 64bit AXI addresses.

Currently the sw-emulator DMA engine operates on the root_bus which is 32bit.
Is that 32bit root bus that caliptra has access to a subset of the 64bit AXI bus which only the DMA engine has full access to?

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -22,11 +22,26 @@ use zerocopy::AsBytes;
 pub enum DmaReadTarget {
     Mbox,
     AhbFifo,
-    AxiWr(usize),
+    AxiWr(AxiAddr),
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct AxiAddr {
+    pub lo: u32,
+    pub hi: u32,
+}
+
+impl From<u64> for AxiAddr {
+    fn from(addr: u64) -> Self {
+        Self {
+            lo: addr as u32,
+            hi: (addr >> 32) as u32,
+        }
+    }
 }
 
 pub struct DmaReadTransaction {
-    pub read_addr: usize,
+    pub read_addr: AxiAddr,
     pub fixed_addr: bool,
     pub length: u32,
     pub target: DmaReadTarget,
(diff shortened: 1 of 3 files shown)


drivers/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub use bounded_address::{BoundedAddr, MemBounds, RomAddr};
 pub use caliptra_error::{CaliptraError, CaliptraResult};
 pub use csrng::{Csrng, HealthFailCounts as CsrngHealthFailCounts, Seed as CsrngSeed};
 pub use data_vault::{ColdResetEntries, DataVault, WarmResetEntries};
-pub use dma::{Dma, DmaReadTarget, DmaReadTransaction, DmaWriteOrigin, DmaWriteTransaction};
+pub use dma::{AxiAddr Dma, DmaReadTarget, DmaReadTransaction, DmaWriteOrigin, DmaWriteTransaction};
```
swenson, line 63: I think this is a typo.

swenson (commented):
AXI is a separate bus from the main memory bus. Importantly, AXI does *NOT* have access to the main memory bus or DCCM in any way.

Instead, AXI maps specific addresses to external peripherals (and maybe to their RAM as well).

This code seems okay for the driver, but we might need to rethink how the emulator works.

drivers/src/dma.rs
```diff
@@ -78,16 +93,13 @@ impl Dma {
     fn setup_dma_read(&mut self, read_transaction: DmaReadTransaction) {
         let dma = self.dma.regs_mut();
 
-        let read_addr: usize = read_transaction.read_addr;
-        #[cfg(target_pointer_width = "64")]
-        dma.src_addr_h().write(|_| (read_addr >> 32) as u32);
-        dma.src_addr_l().write(|_| (read_addr & 0xffff_ffff) as u32);
+        let read_addr = read_transaction.read_addr;
+        dma.src_addr_h().write(|_| read_addr.lo);
```
mhatrevi, line 97: The addresses are swapped.

drivers/src/lib.rs
mhatrevi: How is this even compiling?

mhatrevi (changes requested):

mhatrevi (approved):

drivers/src/lib.rs
mhatrevi: Updated.

PR #1862, opened 2024-12-18
Rust 1.83

ArthurHeymans:
The goal of updating the rust library is to be able to use Mldsa from
rustCrypto repo: https://github.com/RustCrypto/signatures/pull/877

The goal of using this code is to replace openssl to generate TBS and
CSR tempaltes. Openssl does not yet support Mldsa

The minimum required rust version is 1.81.

It looks like this toolchain needs a lot more stack than 1.70.

Rust 1.83 warns about unused stuff a lot more often which is addressed
in this commit.

This also updates DPE to compile without warnings and errors.
The changes are on github.com/caliptra-dpe but is not in the main
branch.


dpe
jhand2, line 1: This commit isn't upstream

jhand2:
Can you update the PR description with the motivation for doing this?

In general, all PRs should have descriptions.

rom/dev/src/rom.ld
```diff
@@ -29,14 +29,14 @@ CFI_STATE_ORG      = 0x500003E4;
...
- * ensure that the sections fit within the true 48k ROM size.
+ * ensure that the sections fit within the true 96k ROM size.
  */
 ROM_RELAXATION_PADDING = 8k;
 ROM_SIZE          = 96K;
 ICCM_SIZE         = 128K;
 DCCM_SIZE         = 256K;
 DATA_SIZE         = 996;
-STACK_SIZE        = 40K;
+STACK_SIZE        = 46K;
```
swenson, line 39: Why does the stack and ROM size need to change in this PR?

rom/dev/src/rom_env.rs
```diff
@@ -65,9 +64,6 @@ pub struct RomEnv {
     /// PCR Bank
     pub pcr_bank: PcrBank,
 
-    /// FHT Data Store
-    pub fht_data_store: FhtDataStore,
```
swenson, line 69: This also seems unrelated to upgrading to Rust 1.83

swenson (commented):
There seem to be some unrelated changes in this PR related to the manifest-based auth and ROM size changes.

dpe
ArthurHeymans: I added a comment in the PR description. Let me know how you think this should be dealt with.

ArthurHeymans:
> Can you update the PR description with the motivation for doing this?
> 
> In general, all PRs should have descriptions.
> 
> Also, this PR seems to have 2 classes of changes:
> 
> 1. Things required because the compiler version changed (e.g. formatting, new clippy annotations)
> 2. Things that leverage new rust features that are now available.
> 
> Ideally all the things in (2) would be in a separate PR.

This PR only attempts to fix warnings and errors with the newer toolchain. I added some motivation documentation to the PR.

rom/dev/src/rom_env.rs
ArthurHeymans: It's unused and rust 1.83 complains about it.

rom/dev/src/rom.ld
ArthurHeymans: So it looks like with the newer toolchain significantly more stack is used. The stack overflow detection code in the emulator tripped on it and this was the minimal increase necessary.

swenson:
This update appears to add a panic because Rust can't prove the indexes are correct in a `ufmt` macro. This should fix it I think:

```patch
diff --git a/drivers/src/printer.rs b/drivers/src/printer.rs
index 4b6d86c9..45f47e23 100644
--- a/drivers/src/printer.rs
+++ b/drivers/src/printer.rs
@@ -68,8 +68,19 @@ impl uDisplay for HexBytes<'_> {
     where
         W: uWrite + ?Sized,
     {
-        for byte in self.0.iter() {
-            ufmt::uwrite!(f, "{:02X}", *byte)?;
+        for &x in self.0.iter() {
+            let c = x >> 4;
+            if c < 10 {
+                f.write_char((c + b'0') as char)?;
+            } else {
+                f.write_char((c - 10 + b'A') as char)?;
+            }
+            let c = x & 0xf;
+            if c < 10 {
+                f.write_char((c + b'0') as char)?;
+            } else {
+                f.write_char((c - 10 + b'A') as char)?;
+            }
         }
         Ok(())
     }
```

dpe
jhand2: Let's merge the caliptra-dpe PR first and then use the upstream commit. It looks like there's an open comment from zach on that one.

ArthurHeymans:
dropped in favor of https://github.com/chipsalliance/caliptra-sw/pull/1956

PR #1867, opened 2024-12-25
[feat] Firmware image download from Recovery Interface

mhatrevi:
This change adds the ROM flow to download the firmware image from the recovery interface.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor.rs
@@ -35,7 +35,7 @@ use caliptra_image_types::{ImageManifest, IMAGE_BYTE_SIZE};
 use caliptra_image_verify::{ImageVerificationInfo, ImageVerificationLogInfo, ImageVerifier};
 use caliptra_kat::KatsEnv;
 use caliptra_x509::{NotAfter, NotBefore};
-use core::mem::ManuallyDrop;
+use core::mem::{size_of, ManuallyDrop};
 use zerocopy::{AsBytes, LayoutVerified};
 use zeroize::Zeroize;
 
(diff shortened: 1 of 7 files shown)


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -735,4 +809,78 @@ impl FirmwareProcessor {
...
+    /// * `soc_ifc` - SOC Interface
+    ///
+    /// # Returns
+    /// * `()` - Ok
+    ///   Error code on failure.
+    fn retrieve_image_from_recovery_interface(
+        dma: &mut Dma,
+        soc_ifc: &mut SocIfc,
+    ) -> CaliptraResult<u32> {
+        let rri_base_addr: u64 = soc_ifc.recovery_interface_base_addr();
```
swenson, line 826: I'm confused. I thought the recovery interface was accessible via the I3C registers (so didn't need DMA): https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/hw/latest/registers/src/i3ccsr.rs#L624

swenson: Nevermind, we discussed this, and it sounds like this will only be accessible via DMA.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -735,4 +809,78 @@ impl FirmwareProcessor {
...
+        const PROT_CAP_2_OFFSET: u32 = 0xC;
+        const DEVICE_STATUS_0: u32 = 0x30;
```
swenson, line 828: If this is only accessible via DMA, I'm tempted to make a DMA generator, along the same lines as the RDL / Registers autogenerator. But for now, this is okay.

swenson (dismissed):
LGTM

swenson (approved):
LGTM

PR #1869, opened 2024-12-31
[update] Add MLDSA key revocation to the verifier

mhatrevi:
This change adds the MLDSA key revocation logic to the verifier.


swenson (approved):
LGTM

PR #1870, opened 2025-01-01
[update] Implement IDEVID Subject Key ID for MLDSA

mhatrevi:
This change queries the correct algorithm type for generating the IDEVID subject key id for MLDSA from the fuse_idevid_cert_attr fuse.

### DIFF
--- a/drivers/src/fuse_bank.rs
+++ b/drivers/src/fuse_bank.rs
@@ -164,33 +164,51 @@ impl FuseBank<'_> {
     /// Get the subject key identifier.
     ///
     /// # Arguments
-    /// * None
+    /// * `ecc_subject_key_id` - Whether to get ECC or MLDSA subject key identifier
     ///
     /// # Returns
     ///     subject key identifier
     ///
-    pub fn subject_key_id(&self) -> [u8; 20] {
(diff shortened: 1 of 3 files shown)


drivers/src/fuse_bank.rs
```diff
@@ -85,25 +91,31 @@ impl FuseBank<'_> {
...
+        let mut flags = soc_ifc_regs
             .fuse_idevid_cert_attr()
             .at(IdevidCertAttr::Flags.into())
             .read();
 
-        match flags & 0x3 {
+        if !ecc_key_id_algo {
+            // ECC Key Id Algo is in Bits 0-2.
+            // MLDSA Key Id Algo is in Bits 3-5.
+            flags >>= 3
```
swenson, line 110: missing semicolon?

swenson (dismissed):
LGTM but I'm not sure how that line compiles

drivers/src/fuse_bank.rs
mhatrevi: It's strictly not necessary as the flags will get updated. That said, I added the semicolon so there is no confusion.

swenson (approved):
LGTM

PR #1874, opened 2025-01-03
Implement key ladder in ROM and RT FW.

bluegate010:
The key ladder is initialized in cold-boot from LDevID CDI. It is lengthened during update-reset flows to match min-SVN-since-cold-boot.

This PR does not include a smoke test, because the production application firmware doesn't yet include any APIs for exercising the key ladder.

The additional cycles from this feature depends on the firmware's SVN, and is roughly 2.7k cycles * (128 - fw_svn). In the worst case, when fw_svn = 0, the cycle count is ~350k.


runtime/src/key_ladder.rs
```diff
@@ -0,0 +1,100 @@
...
+            )
+        };
+
+        let secret_source = if num_iters == 0 {
+            key_ladder_kv
+        } else {
+            extend_ladder(key_ladder_kv, KEY_ID_TMP)?;
+            for _ in 1..num_iters {
+                extend_ladder(KEY_ID_TMP, KEY_ID_TMP)?;
+            }
```
jhand2, line 81: nit: I think you could avoid the lambda and extra case outside the loop by doing

```rust
let mut in_idx = key_ladder_kv
for _ in 0..num_iters {
    Hmac::hmac_kdf(
        drivers,
        in_idx,
        b"si_extend",
        None,
        HmacMode::Hmac512,
        KEY_ID_TMP,
    )
    in_idx = KEY_ID_TMP;
}
```

runtime/tests/runtime_integration_tests/test_update_reset.rs
```diff
@@ -390,3 +397,297 @@ fn test_pcr_reset_counter_persistence() {
...
+        test_image_options: Some(get_image_opts(&epoch_a, 1)),
+        ..Default::default()
+    });
+
+    let ladder_a = get_ladder_digest(&mut model, 1);
+
+    runtime_update_to_svn(&mut model, &epoch_b, 1);
+    let ladder_b = get_ladder_digest(&mut model, 1);
+
+    // A runtime update of the epoch does not affect the key ladder.
```
jhand2, line 566: I wonder if we should fail runtime updates that change the epoch. Since that may end up with some unexpected behavior where the operator intended to update the epoch but they didn't.

runtime/tests/runtime_integration_tests/test_update_reset.rs
```diff
@@ -390,3 +397,297 @@ fn test_pcr_reset_counter_persistence() {
...
+
+    // Ask FW for a secret available to SVN 1.
+    let ladder_1_from_max = get_ladder_digest(&mut model, 1);
+
+    assert_ne!(ladder_1_from_max, ladder_max);
+    assert_eq!(ladder_1_from_max, ladder_1);
+}
+
+#[test]
+fn test_key_ladder_runtime_update() {
```
jhand2, line 481: Should a runtime update with an SVN > MAX fail? Should maybe have a test for that.

runtime/tests/runtime_integration_tests/test_update_reset.rs
```diff
@@ -390,3 +397,297 @@ fn test_pcr_reset_counter_persistence() {
...
+        ..Default::default()
+    });
+
+    let resp = model.mailbox_execute(0xE000_0000, &[]).unwrap().unwrap();
+
+    let max = u16::from_le_bytes(resp.try_into().unwrap());
+    assert_eq!(max as u32, MAX_SVN);
+}
+
+fn make_model_with_security_state(
```
jhand2, line 604: could these be added to `RuntimeTestArgs` and then just use the normal test setup function? https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/tests/runtime_integration_tests/common.rs#L52-L57

runtime/tests/runtime_integration_tests/test_update_reset.rs
bluegate010: There's a test for that [here](https://github.com/chipsalliance/caliptra-sw/blob/37dc9b19ce994ec85b23a1eeb67ca0985fc985c6/rom/dev/tests/rom_integration_tests/test_image_validation.rs#L1752)

runtime/tests/runtime_integration_tests/test_update_reset.rs
bluegate010: Done.

runtime/src/key_ladder.rs
bluegate010: Done.

runtime/tests/runtime_integration_tests/test_update_reset.rs
bluegate010: Can address that as a follow-on PR after discussion with the group.

bluegate010: Update: we'll now just ignore the epoch field and remove it in a separate PR.

jhand2 (dismissed):

swenson (approved):
LGTM. Are you planning to update the README with with the key ladder spec in a separate PR?

bluegate010:
> LGTM. Are you planning to update the README with with the key ladder spec in a separate PR?

Yeah, will do so in a followup.

PR #1876, opened 2025-01-03
Remove reference to main branch in nightly release

nquarton:
Removing a reference to main in the nightly release that would force some aspects of the release to reference the main branch instead of the branch the action was run against.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -23,7 +23,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: 'true'
-          ref: 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -23,7 +23,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: 'true'
-          ref: 'main'
```
nquarton, line 26: This would cause the actual release to reference the main branch when all the testing was done on a different branch (in this case, Caliptra-1.1). This has been fixed in other branches already.

jlmahowa-amd (approved):

PR #1878, opened 2025-01-04
Add Axi bus behind the DMA engine

ArthurHeymans:
This rewrites the DMA engine to operate on the Axi bus which is a 64bit bus.
The recovery register interface is placed on that bus.
An Axi address offset is placed in soc_ifc.


sw-emulator/lib/types/src/axi_bus.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Store address misaligned exception
+    StoreAddrMisaligned,
+
+    /// Store access fault exception
+    StoreAccessFault,
+}
+
+/// Represents an abstract AXI memory bus. Used to read and write from RAM and
+/// peripheral addresses.
+pub trait AxiBus {
```
korran, line 37: I absolutely hate this copy-paste. Why didn't we just increase the address size to 64-bit everywhere?

korran: (assuming the only reason for this change is the increased word size): my preference:

1. Use 64-bit addresses and data in Bus trait everywhere; the 32-bit parts could ignore the high bits and error when size is 64-bit.
2. If that's not workable for some reason we could add associated types to the existing bus trait:

```rs
trait Bus {
  type Addr: num_traits::Num + num_traits::Unsigned + From<u32>;
  type Data: num_traits::Num + num_traits::Unsigned + From<u32>;

  fn read(&mut self, addr: Self::Addr) -> Result<Self::Data, ()>;
  // ...
}
```

korran: poll and reset() could be moved to their own separate trait if necessary.

swenson: I agree I don't love the copy-pasta, especially for the derive macros and the register types.

I like the idea of 64-bit everywhere, maybe either with a type parameter or a function that specifies whether 64-bit is allowed.

Another idea that I think would be okay: keep the separate `AxiBus` trait for read64/write64, but don't support derive. Instead, use autogeneration to automatically create the `AxiBus` implementation. (This is the approach we are mostly using in the MCU, for example: https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/registers/generated-emulator/src/root_bus.rs).

Or only support the dynamic bus.

Since the AXI bus doesn't need to be quite as flexible as the main bus, I think it might be okay to go the autogenerated route.

ArthurHeymans: @korran both these approaches require a lot of code changes all across the emulator code.

I think the simplest things is to implement the toplevel axi_root_bus by hand, which dispatches to Bus.

drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -1133,22 +1133,22 @@ fn test_mailbox_txn_drop() {
...
-    let boot_params = BootParams {
-        ..Default::default()
-    };
-
-    let mut model = caliptra_hw_model::new_unbooted(init_params).unwrap();
-    model.put_firmware_in_rri(recovery_image).unwrap();
-    model.boot(boot_params).unwrap();
-    model.step_until_exit_success().unwrap();
-}
+// #[test]
```
mhatrevi, line 1136: Remove the commented out code.

swenson (commented):
I would like to keep the I3C RDL and autogenerated code. I have a PR I am working on that adapts the DMA driver to implement the `ureg::Mmio` and `ureg::MmioMut` interfaces, so that we can use the autogenerated I3C code and not write everything manually.

sw-emulator/lib/periph/src/dma.rs
```diff
@@ -253,133 +251,223 @@ impl DmaRegs {
...
-            }
-            Some(Control::WRITE_ROUTE::Value::AXI_WR) => todo!(),
+    fn op_complete(&mut self) {
+        let read_target = self.control.reg.read_as_enum(Control::READ_ROUTE).unwrap();
+        let write_origin = self.control.reg.read_as_enum(Control::WRITE_ROUTE).unwrap();
 
-            _ => {}
+        match (read_target, write_origin) {
+            (Control::READ_ROUTE::Value::MAILBOX, Control::WRITE_ROUTE::Value::DISABLE) => {
+                self.read_to_mailbox()
```
mhatrevi, line 351: The naming seems confusing. Typically it is 'read from something' and 'write to something'. Consider naming as such.

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
```diff
@@ -0,0 +1,73 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    lib.rs
```
mhatrevi, line 7: Incorrect file name.

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl Default for AxiRootBus {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl AxiRootBus {
+    const TEST_REG_OFFSET: AxiAddr = 0xaa00;
+    pub const RECOVERY_REGISTER_INTERFACE_OFFSET: AxiAddr = 0xf_00000000;
```
mhatrevi, line 38: What is this offset? Shouldn't the recovery interface be located on the AXI bus at the address specified in ss_recovery_ifc_base_addr_l and ss_recovery_ifc_base_addr_h in the SOC IFC registers?

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
```diff
@@ -0,0 +1,73 @@
...
+            reg: 0xaabbccdd,
+            recovery: RecoveryRegisterInterface::new(),
+        }
+    }
+
+    pub fn read(&mut self, size: RvSize, addr: AxiAddr) -> Result<RvData, BusError> {
+        match addr {
+            Self::TEST_REG_OFFSET => return Register::read(&self.reg, size),
+            Self::RECOVERY_REGISTER_INTERFACE_OFFSET..=Self::RECOVERY_REGISTER_INTERFACE_END => {
+                let addr = (addr - Self::RECOVERY_REGISTER_INTERFACE_OFFSET) as RvAddr;
```
mhatrevi, line 52: Why is the address subtracted from Self::RECOVERY_REGISTER_INTERFACE_OFFSET?

mhatrevi (changes requested):
Please see the feedback.

sw-emulator/lib/periph/src/dma.rs
ArthurHeymans: Fixed

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
ArthurHeymans: That's how busses work in the sw-emulator: All peripherals gets load and stores values that are relative to their own base.

ArthurHeymans:
> Please see the feedback.

Addressed it all.

I also added some driver and other fixes that should make your test branch get further.

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
ArthurHeymans: Btw derive Bus macro results in very similar code.

drivers/src/dma.rs
```diff
@@ -164,13 +181,23 @@ impl Dma {
...
+            chunk.copy_from_slice(&value.to_le_bytes());
+        }
+
+        // Process remaining bytes (1, 2, or 3)
+        let remainder = read_data.len() % 4;
+        if remainder > 0 {
+            let start = read_data.len() - remainder;
+            let last_chunk = &mut read_data[start..];
+            for byte in last_chunk.iter_mut() {
+                *byte = unsafe { read_volatile(read_data_ptr) };
```
swenson, line 198: Is this how the hardware works?

Since it is a 32-bit register underneath mapping to a 32-bit LSU, I would have expected we would do one last 32-bit read and then discard leading or trailing zeros to get the remainder bytes.

Similar for writing.

sw-emulator/lib/periph/src/dma/axi_root_bus.rs
swenson: Yep! That is the expected way that the `Bus` trait handles offsets.

drivers/src/dma.rs
ArthurHeymans: I don't know what the hardware does. The emulator does panic on fifo sizes that aren't a multiple of 4 so the driver might as well bail out as I don't think we need this.

swenson (approved):
LGTM

mhatrevi (approved):

swenson (approved):
LGTM

PR #1882, opened 2025-01-09
Support populating SVNs in image builder

jhand2:
FMC and RT SVNs are controlled by the vendor. Add an option to the image builder to specify these SVNs from the command-line.

### DIFF
--- a/builder/bin/image_gen.rs
+++ b/builder/bin/image_gen.rs
@@ -20,6 +20,14 @@ fn main() {
         )
         .arg(arg!(--"fw" [FILE] "FW bundle image").value_parser(value_parser!(PathBuf)))
         .arg(
+            arg!(--"fmc-svn" [VALUE] "Security Version Number of FMC firmware image")
+                .value_parser(value_parser!(u32)),
+        )
+        .arg(
+            arg!(--"rt-svn" [VALUE] "Security Version Number of RT firmware image")
+                .value_parser(value_parser!(u32)),
+        )
+        .arg(
             arg!(--"all_elfs" [DIR] "Build all firmware elf files")
                 .value_parser(value_parser!(PathBuf)),
         )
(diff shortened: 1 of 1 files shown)


builder/bin/image_gen.rs
```diff
@@ -19,6 +19,14 @@ fn main() {
                 .value_parser(value_parser!(PathBuf)),
         )
         .arg(arg!(--"fw" [FILE] "FW bundle image").value_parser(value_parser!(PathBuf)))
+        .arg(
+            arg!(--"fmc-svn" [FILE] "Security Version Number of FMC firmware image")
```
clundin25, line 23:
```suggestion
            arg!(--"fmc-svn" [VALUE] "Security Version Number of FMC firmware image")
```

nit: I think `[FILE]` could cause confusion when reading the help text.

builder/bin/image_gen.rs
```diff
@@ -19,6 +19,14 @@ fn main() {
...
+                .value_parser(value_parser!(u32)),
+        )
+        .arg(
+            arg!(--"rt-svn" [FILE] "Security Version Number of RT firmware image")
```
clundin25, line 27:
```suggestion
            arg!(--"rt-svn" [VALUE] "Security Version Number of RT firmware image")
```

nit: Similar to the above comment

clundin25 (dismissed):

builder/bin/image_gen.rs
jhand2: Done

builder/bin/image_gen.rs
jhand2: Done

clundin25 (approved):

builder/bin/image_gen.rs
```diff
@@ -41,6 +49,18 @@ fn main() {
         std::fs::write(path, rom).unwrap();
     }
 
+    let fmc_svn = if let Some(fmc) = args.get_one::<u32>("fmc-svn") {
```
zhalvorsen, line 52: optional: what do you think of using `or_default` on these to be more concise?

jhand2: Going to leave this because `get_one::<u32>` returns `Option<&u32>` and Default isn't implemented for `&u32`.

PR #1884, opened 2025-01-09
Update image-builder to output a hashes file.

zhalvorsen:
This can be useful for offline/external signing to extract the digests for an image bundle to be passed to the signing interface.

The output is a JSON file with the vendor and owner hashes tagged with their respective names.

This was tested by printing the hash from the ROM in the smoke tests to ensure the calculated hash is the same for both the firmware and tool.

### DIFF
--- a/builder/bin/image_gen.rs
+++ b/builder/bin/image_gen.rs
@@ -3,7 +3,12 @@
 use caliptra_builder::firmware;
 use caliptra_builder::version;
 use caliptra_builder::ImageOptions;
+use caliptra_image_types::ImageHeader;
+use caliptra_image_types::ImageManifest;
 use clap::{arg, value_parser, Command};
+use memoffset::{offset_of, span_of};
+use serde_json::{json, to_string_pretty};
+use sha2::{Digest, Sha384};
 use std::collections::HashSet;
 use std::path::PathBuf;
 
@@ -33,6 +38,10 @@ fn main() {
         )
         .arg(arg!(--"fake-rom" [FILE] "Fake ROM").value_parser(value_parser!(PathBuf)))
         .arg(arg!(--"fake-fw" [FILE] "Fake FW bundle image").value_parser(value_parser!(PathBuf)))
+        .arg(
+            arg!(--"hashes" [FILE] "File path for output JSON file containing image bundle header hashes for external signing tools")
+                .value_parser(value_parser!(PathBuf)),
+        )
         .get_matches();
 
     if let Some(path) = args.get_one::<PathBuf>("rom-no-log") {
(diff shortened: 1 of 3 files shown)


builder/bin/image_gen.rs
```diff
@@ -25,6 +30,10 @@ fn main() {
         )
         .arg(arg!(--"fake-rom" [FILE] "Fake ROM").value_parser(value_parser!(PathBuf)))
         .arg(arg!(--"fake-fw" [FILE] "Fake FW bundle image").value_parser(value_parser!(PathBuf)))
+        .arg(
+            arg!(--"hashes" [FILE] "JSON file with image bundle header hashes for external signing tools")
+                .value_parser(value_parser!(PathBuf)),
```
clundin25, line 35: The current help string doesn't make it clear to me that this is an "out" parameter

I can't think of a great way to phrase this but here is an attempt
```suggestion
            arg!(--"hashes" [FILE] "File path for output JSON file containing image bundle header hashes for external signing tools")
                .value_parser(value_parser!(PathBuf)),
```

builder/bin/image_gen.rs
```diff
@@ -53,7 +62,27 @@ fn main() {
...
+            let vendor_header_len = offset_of!(ImageHeader, owner_data);
+            let vendor_range = header_range.start..header_range.start + vendor_header_len;
+            let vendor_digest = Sha384::digest(&contents[vendor_range]);
+
+            // Get the owner digest which is the full header
+            let owner_digest = Sha384::digest(&contents[header_range]);
+
+            let json = json!({
+                "vendor": vendor_digest.to_vec(),
+                "owner": owner_digest.to_vec(),
```
clundin25, line 82: Should these vec's be encoded into a string representation, like base64?

zhalvorsen: I don't have a really strong opinion. I could see a hexstring being an easily readable and consumable type as well. @jhand2 what do you think?

builder/bin/image_gen.rs
zhalvorsen: I like your wording better. Thanks!

clundin25 (approved):

builder/bin/image_gen.rs
jhand2: I think my preference would be hex string

zhalvorsen: The output file now looks like this:

```json
{
  "owner": "a506d220c5f9023a3fdeed8ca89d2da35219e20cef6a1d5c0143bff6e763059b62a4a43582dfb6903d50a98d0c99660e",
  "vendor": "0a4a24ab60381a5052daffdfab3fceaa47f793ccab62fe30841e842121b5da917e95b8ae5c08924498308c75f87882bb"
}
```

clundin25 (approved):

jhand2 (approved):

PR #1894, opened 2025-01-15
Create an all zeros fw image

jhand2:
Creates an image bundle with FMC and Runtime populated with zeros. 0 is an ilegal opcode in risc-v, so the image will immediately NMI.

This is useful for testing production signing infrastructure. A production key can be used to sign this image before an official release is ready. The prod-signed zeros binary is benign but can be used to ensure that the prod key was loaded into fuses correctly.

### DIFF
--- a/builder/bin/image_gen.rs
+++ b/builder/bin/image_gen.rs
@@ -40,8 +40,9 @@ fn main() {
         .arg(arg!(--"fake-fw" [FILE] "Fake FW bundle image").value_parser(value_parser!(PathBuf)))
         .arg(
             arg!(--"hashes" [FILE] "File path for output JSON file containing image bundle header hashes for external signing tools")
-                .value_parser(value_parser!(PathBuf)),
+                .value_parser(value_parser!(PathBuf))
         )
+        .arg(arg!(--"zeros" "Build an image bundle with zero'd FMC and RT. This will NMI immediately."))
         .get_matches();
 
     if let Some(path) = args.get_one::<PathBuf>("rom-no-log") {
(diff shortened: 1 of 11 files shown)


nquarton (approved):
Definitely should be useful to have this. Thanks!

zero_bin/tests/test_zeros.rs
```diff
@@ -0,0 +1,38 @@
...
+        },
+    )
+    .unwrap();
+
+    // 0 is an ilegal instruction in risc-v. Image should immediately NMI.
+    model.step_until(|m| m.soc_ifc().cptra_fw_error_fatal().read() != 0);
+    assert_eq!(
+        model.soc_ifc().cptra_fw_error_fatal().read(),
+        u32::from(CaliptraError::ROM_GLOBAL_EXCEPTION)
+    );
```
zhalvorsen, line 37: Like we discussed offline, lets check that PC and `mcause` have the expected values here

zhalvorsen (commented):
This is awesome, great work!

builder/bin/image_gen.rs
```diff
@@ -72,18 +73,33 @@ fn main() {
...
+        let image = if args.contains_id("zeros") {
+            caliptra_builder::build_and_sign_image(
+                &firmware::FMC_ZEROS,
+                &firmware::APP_ZEROS,
+                ImageOptions {
+                    fmc_version: 0,
+                    app_version: 0,
+                    fmc_svn: 0,
+                    app_svn: 0,
+                    ..Default::default()
```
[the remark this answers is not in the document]
jhand2: tbh I just copy/pasted from somewhere else and then changed the fields. Changed to use ImageOptions::default

zero_bin/tests/test_zeros.rs
jhand2: Done

zhalvorsen (approved):

MDr164 (approved):

nquarton (approved):

PR #1896, opened 2025-01-16
[feat] Increase ICCM size to 256 KB

mhatrevi:
This change updates the ICCM size from 128 KB to 256 KB.


MDr164 (commented):
I see no issue in doing so but could you maybe add a rationale on why this becomes necessary now to the commit or PR text?

mhatrevi:
Abandoing this PR as https://github.com/chipsalliance/caliptra-sw/pull/1901 covers the ICCM size update.

Please see https://github.com/chipsalliance/caliptra-rtl/issues/646 for the justification of the size increase.

PR #1901, opened 2025-01-21
Update RTL: MBOX, DCCM and ICCM to 256K

ArthurHeymans:
The main functional differences are the MBOX, DCCM and ICCM are incraesed to 256K.

The RTL RDL points to an AES block which does not exist. It's commented out which results in a dirty commit.

A few other differences that are not yet fully addressed in this PR:
- CSRNG adds new register to reseed
- Generate PCR digest hash can happen with sha384 and sha512.


swenson:
The `aes` type is defined in `src/aes/data/aes.rdl`.

ArthurHeymans:
> The `aes` type is defined in `src/aes/data/aes.rdl`.

thx. done

swenson (dismissed):
LGTM

mhatrevi (approved):

swenson (commented):
LGTM

PR #1902, opened 2025-01-21
[fix] MLDSA driver fixes

mhatrevi:
This change contains the following fixes:

1. Updates the MLDSA driver to wait for the zeroize operation to be complete before further interacting with the MLDSA hardware peripheral.
2. Updates the MLDSA driver to use the last 64 bytes of the signature for comparison against the 'verify result' value from the MLDSA peripheral.


ArthurHeymans (approved):

swenson (approved):
LGTM

PR #1904, opened 2025-01-23
[runtime] Start recovery flow in the runtime

swenson:
This stubs out the recovery flow in the runtime.
    
We also update the DMA peripheral so that it can be used as a
`ureg::Mmio`, which will allow us to use the autogenerated I3C recovery
register interface instead of hardcoding offsets.
    
We also had to modify `ureg` slightly to remove the unused `impl Uint`
for `u64` (because we don't don't support 64-bit AXI transfers) and add
a `from_u32()` method for the (unused) `u8` and `u16` implementations.
   
This builds on #1867 and rewrites it to use the updated DMA interface.


drivers/src/dma.rs
```diff
@@ -62,7 +106,7 @@ pub struct DmaWriteTransaction {
 
 /// Dma Widget
 pub struct Dma {
-    dma: AxiDmaReg,
+    dma: Cell<Option<AxiDmaReg>>,
```
swenson, line 109: We have to switch to a `Cell` because the `Mmio` and `MmioMut` interfaces require that the methods use `&self` instead of `&mut self`, so we need interior mutability.

For safety, I used a `with_dma()` to ensure that the `AxiDmaReg` is only used by one function at a time, though this is probably overkill since Caliptra is single-threaded.

drivers/src/dma.rs
```diff
@@ -89,78 +124,96 @@ impl Dma {
...
+    pub fn with_dma<T>(
+        &self,
+        f: impl FnOnce(RegisterBlock<RealMmioMut>) -> T,
+    ) -> CaliptraResult<T> {
+        if let Some(mut dma) = self.dma.take() {
+            let result = f(dma.regs_mut());
+            self.dma.set(Some(dma));
+            Ok(result)
+        } else {
+            Err(CaliptraError::RUNTIME_INTERNAL) // should never happen
```
mhatrevi, line 141: Suggest renaming this to something like 'DRIVERS_INTERNAL'  to disambiguate from a Runtime app error.

drivers/src/dma.rs
mhatrevi: Suggest simplifying the code (if possible) with the single-threaded execution assumption.

runtime/src/recovery_flow.rs
```diff
@@ -0,0 +1,67 @@
...
+        };
+        // [TODO][CAP2]: authenticate SoC manifest using keys available through Caliptra Image
+        // TODO: switch to ref_from method when we upgrade zerocopy
+        drivers
+            .persistent_data
+            .get_mut()
+            .auth_manifest_image_metadata_col = image_metadata_collection;
+        // [TODO][CAP2]: capture measurement of Soc manifest?
+        // [TODO][CAP2]: this should be writing to MCU SRAM directly via AXI
+        let _mcu_size_bytes = dma_recovery.download_image_to_mbox(SOC_MANIFEST_INDEX)?;
```
mhatrevi, line 60: This should use the MCU_FIRMWARE_INDEX.

mhatrevi (dismissed):
Please see the feedback.

runtime/src/drivers.rs
```diff
@@ -171,6 +176,9 @@ impl Drivers {
             ResetReason::ColdReset => {
                 cfi_assert_eq(self.soc_ifc.reset_reason(), ResetReason::ColdReset);
                 Self::initialize_dpe(self)?;
+                if self.soc_ifc.active_mode() {
```
jhand2, line 179: I wonder if this should be a build-time flag since for a given integration it's either active or not. Otoh, it's probably easier to have a less complicated test matrix for RT.

runtime/src/recovery_flow.rs
```diff
@@ -0,0 +1,67 @@
...
+            DmaRecovery::new(drivers.soc_ifc.recovery_interface_base_addr().into(), dma);
+
+        // // download SoC manifest
+        let _soc_size_bytes = dma_recovery.download_image_to_mbox(SOC_MANIFEST_INDEX)?;
+        let Some(preamble) = AuthManifestPreamble::read_from_prefix(drivers.mbox.raw_mailbox_contents()) else {
+            return Err(CaliptraError::IMAGE_VERIFIER_ERR_MANIFEST_SIZE_MISMATCH);
+        };
+        let Some(image_metadata_collection) = AuthManifestImageMetadataCollection::read_from_prefix(&drivers.mbox.raw_mailbox_contents()[AUTH_MANIFEST_PREAMBLE_SIZE..]) else {
+            return Err(CaliptraError::IMAGE_VERIFIER_ERR_MANIFEST_SIZE_MISMATCH);
+        };
```
jhand2, line 51: nit: can probably simplify this by making a single stuct that contains `AuthManifestPreamble` and `AuthManifestImageMetadataCollection` and then doing a single read.

ureg/src/lib.rs
```diff
@@ -32,24 +32,28 @@ mod private {
...
 
 impl Uint for u32 {
     const TYPE: UintType = UintType::U32;
+    fn from_u32(val: u32) -> Self {
+        val
+    }
 }
 impl private::Sealed for u32 {}
 
-impl Uint for u64 {
```
jhand2, line 48: Why is u64 removed here?

ureg/src/lib.rs
```diff
@@ -32,24 +32,28 @@ mod private {
 
 impl Uint for u8 {
     const TYPE: UintType = UintType::U8;
+    fn from_u32(val: u32) -> Self {
+        (val & 0xff) as u8
+    }
```
jhand2, line 37: Is this needed for multiple types? I think I would prefer to the the truncation closer to the specific register that needs it so that we can review specific usages as not dropping any meaningful bytes.

ureg/src/lib.rs
swenson: We never use 64-bit registers, and I don't know if we have the semantics for it super well-defined given our 32-bit memory bus. (Things like, which word is written first.) So, it seemed easier to just to remove 64-bit support altogether.

(Side note, I'd also be a fan of removing 8- and 16-bit register support as well, since it also is a little murky semantically, and refactoring out the `Uint` type altogether to just be `u32`.)

ureg/src/lib.rs
swenson: It's not truly *needed* except on paper, as we don't have any 8- or 16-bit registers.

runtime/src/drivers.rs
swenson: Would we plan on building a separate active mode ROM then?

runtime/src/recovery_flow.rs
```diff
@@ -0,0 +1,64 @@
...
+        const SOC_MANIFEST_INDEX: u32 = 1;
+        const MCU_FIRMWARE_INDEX: u32 = 2;
+
+        let dma = &drivers.dma;
+        let dma_recovery =
+            DmaRecovery::new(drivers.soc_ifc.recovery_interface_base_addr().into(), dma);
+
+        // // download SoC manifest
+        let _soc_size_bytes = dma_recovery.download_image_to_mbox(SOC_MANIFEST_INDEX)?;
+        let Some(manifest) = AuthorizationManifest::read_from_prefix(drivers.mbox.raw_mailbox_contents()) else {
```
mhatrevi, line 46: This is not efficient. AuthorizationManifest is of 14284 bytes; most of the time there will only be a handful of image digests in it. I would simply add a TODO here to call the existing set_manifest_code.

ureg/src/lib.rs
swenson: I think we can drop this and replace it with `TryFrom<u32>`.

ureg/src/lib.rs
swenson: I added it back in.

swenson:
PR updated and rebased. PTAL

drivers/src/dma.rs
```diff
@@ -353,7 +404,218 @@ impl Dma {
...
+    pub fn with_regs<T, F>(&self, f: F) -> CaliptraResult<T>
+    where
+        F: FnOnce(I3CRegisterBlock<&DmaMmio>) -> T,
+    {
+        let mmio = DmaMmio::new(self.base, self.dma);
+        // SAFETY: we aren't referencing memory directly
+        let regs = unsafe {
+            I3CRegisterBlock::new_with_mmio(
+                // substract the recovery offset since all recovery registers are relative to it
+                core::ptr::null_mut::<u32>().sub(0x100 / core::mem::size_of::<u32>()),
```
mhatrevi, line 496: Suggest defining a constant for the offset.

mhatrevi (dismissed):

drivers/src/dma.rs
swenson: I missed this comment, and will address in a follow-up PR.

runtime/src/drivers.rs
swenson: I think we can leave this for now. If we want to split the build, we can change this to an `if cfg!`.

ureg/src/lib.rs
swenson: I cleaned this up.

mhatrevi (approved):

swenson:
Thanks!

PR #1907, opened 2025-01-24
ci: Build ROM release from versioned git ref (revive)

leongross:
Revive of PR https://github.com/chipsalliance/caliptra-sw/pull/1604.


.github/workflows/fpga.yml
```diff
@@ -109,6 +112,24 @@ jobs:
...
     steps:
+      - name: Checkout versioned ROM (${{ inputs.rom-ref }})
+        uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+          ref: ${{ inputs.rom-ref }}
+
+      - name: Build ROM
+        run: |
+          # TODO: Confirm if this is the desired test behaviour
```
jhand2, line 123: Ya, this lgtm

.github/workflows/fpga.yml
```diff
@@ -25,6 +25,9 @@ on:
       rom-version:
         default: "latest"
         type: string
+      rom-ref:
```
jhand2, line 28: I think ultimately our goal is for the nightly release workflow to be runnable with specified ROM & FW refs. Would it make sense to try to factor out some of the steps in this file so it can be reused in the nightly one as well?

Not sure if that would improve things or just make the CI more confusing.

.github/workflows/versioned-full-build-test.yml
```diff
@@ -13,47 +13,110 @@ on:
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          submodules: 'true'
+
+      - name: Generate release zip
+        run: |
+          ./ci-tools/release/build_release.sh "_versioned_build_test_" ${{ inputs.rom-ref }} ${{ inputs.firmware-version }}
```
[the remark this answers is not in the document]
jhand2: Ya, I think currently it's fine to build FMC and RT from the same revision.

.github/workflows/fpga.yml
leongross: One thing is still unclear to me: We want to build a release (probably using the `build_release.sh` script) for specific revisions of FMC, ROM, and RT.
You also said it should be integrated into the nightly release flow iirc.

So what we can do is to set the default values to HEAD and main and if desired we can change that. How about the testing? As of know, all the test suites are run as part of the nightly release, which makes sense. Is this also a requirement for the specifically 
versioned checkouts or should we speed things up by only packaging the release?

leongross: When you say you want to have this ultimately being run in the nightly workflow I think I don't fully understand what you mean by that.
The FPGA workflows are all pinned to versions, or do you mean we should add another test case that is run for these said specified versions that might be passed as inputs to the nightly workflow? Or should we just consider building the release using the `build_release.sh` script for these versions?

jhand2:
Hey @leongross, it looks like there's some failing CI in this PR. Any questions you have for getting this running?

leongross:
The underlying problem was introduced by my patches, which resulted in the builder tests failing due to a missing Rust toolchain installation:

```sh
--- STDERR:              caliptra-builder::bin/image test_binaries_are_identical ---
warning: unknown feature specified for `-Ctarget-feature`: `unaligned-scalar-mem`
  |
  = note: it is still passed through to the codegen backend
  = help: consider filing a feature request

info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2025-04-03, rust version 1.86.0 (05f9846f8 2025-03-31)
info: downloading component 'rust-src'
info: installing component 'rust-src'
error[E0463]: can't find crate for `core`
  |
  = note: the `riscv32imc-unknown-none-elf` target may not be installed
  = help: consider downloading the target with `rustup target add riscv32imc-unknown-none-elf`
```

When adding the suggested `riscv32imc-unknown-none-elf` toolchain, the runner runs out of hard drive space while pulling. I will try to investigate why the toolchain is not available in this run; otherwise, we could also increase the disk space to counter this as a quick fix.

.github/workflows/fpga.yml
```diff
@@ -4,7 +4,6 @@ name: FPGA Build
 on:
   push:
     branches: ["main"]
-  pull_request:
```
clundin25, line 7: I think we should keep the PR check for FPGAs, is this change intentional?

.github/workflows/fpga.yml
```diff
@@ -290,6 +304,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: 'true'
+
+      - name: Pull dpe submodule
+        run: |
+          git submodule update --init dpe
```
clundin25, line 310:
```suggestion
          submodules: 'recursive'
```

Looks like a few places could use this pattern to remove a step.

.github/workflows/fw-test-emu.yml
```diff
@@ -47,19 +62,45 @@ jobs:
...
         run: |
-          export CALIPTRA_PREBUILT_FW_DIR=/tmp/caliptra-test-firmware
+          export CALIPTRA_PREBUILT_FW_DIR=/tmp/caliptra-unified/
+          export CALIPTRA_PREBUILT_ROM_BIN=/tmp/caliptra-unified/caliptra-rom.bin
+          export RUST_BACKTRACE=full
+
+          rustup toolchain install nightly-2023-04-15-x86_64-unknown-linux-gnu
+          rustup default nightly-2023-04-15-x86_64-unknown-linux-gnu
+          rustup target add riscv32imc-unknown-none-elf --toolchain nightly-2023-04-15-x86_64-unknown-linux-gnu
+          rustup component add rust-src --toolchain nightly-2023-04-15-x86_64-unknown-linux-gnu
```
clundin25, line 100: I opened https://github.com/chipsalliance/caliptra-sw/pull/2138 to fix the compilation error that @MDr164 mentioned this morning.

I believe the toolchain was set by `rust-toolchain.toml` and then the tests change directories and use the global toolchain. 

My PR changes this snippet to make the toolchain match `rust-toolchain.toml`.

Can you clarify the usage of `nightly-2023-04-15-x86_64-unknown-linux-gnu` here?

clundin25:
> I am not entirely sure what the reason is for the error; in the `sw-emulator` test suites for the `builder` the rust toolchain (module `core`) is not picked up properly; will investigate. The FPGA test complete successfully though.

The FPGA suite passes because the build steps using an image that has `1.70` as the installed rust toolchain.

Ref: https://github.com/chipsalliance/caliptra-sw/blob/main/ci-tools/github-runner/scripts/tweak_runner_image.sh#L47

.github/workflows/fpga.yml
```diff
@@ -381,11 +399,11 @@ jobs:
           name: caliptra-test-binaries${{ inputs.artifact-suffix }}
           path: /tmp/caliptra-test-binaries.sqsh
 
-      - name: 'Download Test Firmware Artifact'
-        uses: actions/download-artifact@v4
-        with:
-          name: caliptra-test-firmware${{ inputs.artifact-suffix }}
-          path: /tmp/caliptra-test-firmware
+      # - name: 'Download Test Firmware Artifact'
```
clundin25, line 402: Please remove commented code from PR

.github/workflows/nightly-release.yml
```diff
@@ -385,6 +396,9 @@ jobs:
 
       - name: Generate release zip
         run: |
+          # argc == 1: <release_name>
+          # argc == 3: <release_name> <rom_ref> <firmware_ref> 
+
```
clundin25, line 401: nit: This comment is vague. Is it describing the args to `build_release.sh`?

.github/workflows/fpga.yml
```diff
@@ -106,9 +110,21 @@ jobs:
...
+  # this workflow uploads a file named caliptra-unified-<rom-ref>-<fw-ref>-<hw-version>-<artifact-suffix>
+  versioned-checkout-build:
+    uses: ./.github/workflows/versioned-checkout.yml
+    with:
+      rom-ref: "${{ inputs.rom-ref }}"
+      rom-logging: "${{ inputs.rom-logging }}"
+      fw-ref: "${{ inputs.fw-ref }}"
+      hw-version: "${{ inputs.hw-version }}"
+      workflow_call: "${{ inputs.workflow_call }}"
+      artifact-suffix: "${{ inputs.artifact-suffix }}"
```
clundin25, line 122: Is it possible to isolate the FPGA jobs to the nightly run? 

I think running all of these tests on each PR will effectively DDOS the FPGA infra.

.github/workflows/fpga.yml
leongross: No this was not intentional, I will revert this change.

PR #1914, opened 2025-01-27
[feat] Implement PQC Key type validation in FW Image Verifier

mhatrevi:
This change incorporates the newly added 'pqc_key_type' fuse check in the image verifier. The change ensures that the pqc key type value in the image is the same as the one in the pqc_key_type fuse.

### DIFF
--- a/api/types/src/lib.rs
+++ b/api/types/src/lib.rs
@@ -19,6 +19,8 @@ pub const DEFAULT_CPTRA_OBF_KEY: [u32; 8] = [
     0xa0a1a2a3, 0xb0b1b2b3, 0xc0c1c2c3, 0xd0d1d2d3, 0xe0e1e2e3, 0xf0f1f2f3, 0xa4a5a6a7, 0xb4b5b6b7,
 ];
 
+pub const DEFAULT_PQC_KEY_TYPE: u32 = 0x1; // MLDLSA = 1, LMS = 3.
+
 // Based on device_lifecycle_e from RTL
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub enum DeviceLifecycle {
@@ -168,6 +170,7 @@ pub struct Fuses {
     pub fuse_lms_revocation: u32,
     pub fuse_mldsa_revocation: u32,
     pub soc_stepping_id: u16,
+    pub fuse_pqc_key_type: u32,
 }
 impl Default for Fuses {
     fn default() -> Self {
@@ -186,6 +189,7 @@ impl Default for Fuses {
             fuse_lms_revocation: Default::default(),
             fuse_mldsa_revocation: Default::default(),
             soc_stepping_id: Default::default(),
+            fuse_pqc_key_type: DEFAULT_PQC_KEY_TYPE,
         }
     }
 }
--- a/api/types/src/lib.rs
+++ b/api/types/src/lib.rs
@@ -1,6 +1,8 @@
 // Licensed under the Apache-2.0 license
 #![cfg_attr(not(test), no_std)]
 
+use caliptra_image_types::FwVerificationPqcKeyType;
+
 // Rationale behind this choice
 //
 // * The constant should be easily recognizable in waveforms and debug logs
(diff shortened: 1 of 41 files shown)


api/types/src/lib.rs
```diff
@@ -19,6 +19,8 @@ pub const DEFAULT_CPTRA_OBF_KEY: [u32; 8] = [
     0xa0a1a2a3, 0xb0b1b2b3, 0xc0c1c2c3, 0xd0d1d2d3, 0xe0e1e2e3, 0xf0f1f2f3, 0xa4a5a6a7, 0xb4b5b6b7,
 ];
 
+pub const DEFAULT_PQC_KEY_TYPE: u32 = 0x1; // MLDLSA = 1, LMS = 3.
```
swenson, line 22: Use the enum type for this (or move the enum here)?

drivers/src/fuse_bank.rs
```diff
@@ -325,6 +325,18 @@ impl FuseBank<'_> {
         let soc_ifc_regs = self.soc_ifc.regs();
         soc_ifc_regs.fuse_lms_revocation().read()
     }
+
+    /// Get the PQC (MLDSA or LMS) key type.
+    ///
+    /// # Arguments
+    /// * None
+    ///
```
swenson, line 333: nit: I feel like if there are no arguments then we can just remove that part of the comment.

```suggestion
```

drivers/src/fuse_bank.rs
```diff
@@ -325,6 +325,18 @@ impl FuseBank<'_> {
...
+    /// # Returns
+    ///    PQC key type (1 for MLDSA, 3 for LMS)
+    ///
+    pub fn pqc_key_type(&self) -> u32 {
```
swenson, line 337: Ditto here about using the enum type so that we don't have to clarify in the comments what the values mean.

runtime/tests/runtime_integration_tests/test_fips.rs
```diff
@@ -7,17 +7,19 @@ use caliptra_common::mailbox_api::{
     CommandId, FipsVersionResp, MailboxReqHeader, MailboxRespHeader,
 };
 use caliptra_hw_model::HwModel;
+use caliptra_image_types::FwVerificationPqcKeyType;
 use caliptra_runtime::FipsVersionCmd;
 use zerocopy::{AsBytes, FromBytes};
 
-const HW_REV_ID: u32 = 0x11; // TODO 2.0
+const HW_REV_ID: u32 = 0x02;
```
swenson, line 14: Should this be `0x20` maybe?

test/tests/fips_test_suite/common.rs
```diff
@@ -33,57 +33,37 @@ pub struct HwExpVals {
     pub hw_revision: u32,
 }
 
-const HW_EXP_1_0_0: HwExpVals = HwExpVals { hw_revision: 0x1 };
+const HW_EXP_2_0_0: HwExpVals = HwExpVals { hw_revision: 0x2 };
```
swenson, line 36: Ditto here about `0x20`?

api/types/src/lib.rs
mhatrevi: Fixed.

drivers/src/fuse_bank.rs
mhatrevi: Fixed.

drivers/src/fuse_bank.rs
mhatrevi: This function simply returns the value in the fuse. Update the function description accordingly.

runtime/tests/runtime_integration_tests/test_fips.rs
mhatrevi: This is correct per: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.CPTRA_HW_REV_ID

Caliptra official release version. Bit field encoding is:
[15:8] Patch version
[ 7:4] Minor version
[ 3:0] Major version

test/tests/fips_test_suite/common.rs
mhatrevi: Please see the comment above.

swenson (approved):
LGTM

PR #1930, opened 2025-01-30
Reduce code size by enforcing ascii instead of utf-8

clundin25:
Checking that byte slice is utf-8 requires a significant amount of code space. By enforcing ascii this space can be saved at the cost of disallowing subject alt names that contain non-ascii utf-8 characters.

### DIFF
--- a/runtime/src/subject_alt_name.rs
+++ b/runtime/src/subject_alt_name.rs
@@ -12,8 +12,6 @@ Abstract:
 
 --*/
 
-use core::str::from_utf8;
-
 use arrayvec::ArrayVec;
 use caliptra_common::mailbox_api::{AddSubjectAltNameReq, MailboxResp};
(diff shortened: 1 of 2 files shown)


runtime/src/subject_alt_name.rs
```diff
@@ -53,13 +51,19 @@ impl AddSubjectAltNameCmd {
     }
 
     fn validate_dmtf_device_info(dmtf_device_info: &[u8]) -> CaliptraResult<()> {
```
rusty1968, line 53: NIT: Could you please add a function header ?

rusty1968 (approved):

runtime/src/subject_alt_name.rs
clundin25: Done!

jhand2 (approved):
Thanks!

PR #1937, opened 2025-02-03
Bump up FMC_SIZE - FMC overlfowing in 12 bytes during nightly release tests.

rusty1968:


common/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub use fuse::{FuseLogEntry, FuseLogEntryId};
 pub use pcr::{PcrLogEntry, PcrLogEntryId, RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR};
 
 pub const FMC_ORG: u32 = 0x40000000;
-pub const FMC_SIZE: u32 = 21 * 1024;
```
jhand2, line 41: This results in a somewhat odd memory layout where the max size of FMC + RT is 117 KiB + 256 bytes. I think I would prefer to keep that at a 1K boundary if possible.

I think the reason we don't use the full 128K of ICCM is that `FMC + RT + Manifest` must be <= 128KiB to fit in the mailbox. Is the manifest size > 10KiB?

jhand2: Also I think we could probably save 12 bytes in some other way fairly easily rather than changing the max size of FMC.

rusty1968: Hi, Jordan - I did trim strings last time and got feedback to leave them as is, I got away with trimming a few bytes by removing space. If we have the constraint of  combined mutable size <= 118k , I will express that in the code as a compile assert.

PR #1938, opened 2025-02-03
emulator: Add event bus between Caliptra, MCU, BMC, etc.

swenson:
And hook it up to the emulated AXI bus for DMA transfers.
    
The events are used to simulate some of the ways that Caliptra communicates with others in hardware, in particular, I3C, writing to MCU's memory aperture, and the recovery communications.

We wire these up in the emulated AXI DMA controller and recovery interface.

We also have to modify the emulated AXI DMA controller to support asynchronous reads and writes, which should more accurately emulates how the hardware works anyway.

This PR will be coordinated with a PR on the MCU repository, which will provide the emulated MCU and BMC cores, and will test the interactions.

I've run this with the MCU and BMC code through loading the recovery image from the recovery interface in the ROM's cold reset flow and gotten it to key verification.

Builds on some of the DMA changes in #1904.


ArthurHeymans (approved):

ArthurHeymans (dismissed):

drivers/src/dma.rs
```diff
@@ -551,64 +651,70 @@ impl<'a> DmaRecovery<'a> {
...
                 .modify(|val| val.agent_caps(Self::PROT_CAP2_FLASHLESS_BOOT_BIT));
 
-            // Set DEVICE_STATUS:Byte0 to 0x3 ('Recovery mode - ready to accept recovery image').
-            // Set DEVICE_STATUS:Byte[2:3] to 0x12 ('Recovery Reason Codes' 0x12 - Flashless/Streaming Boot (FSB)).
-            recovery.device_status_0().modify(|val| {
-                val.rec_reason_code(Self::FLASHLESS_STREAMING_BOOT_VALUE)
-                    .dev_status(Self::READY_TO_ACCEPT_RECOVERY_IMAGE_VALUE)
-            });
-
             // Set Byte0 Bit[3:0] to 0x1 ('Awaiting recovery image')
```
mhatrevi, line 654: Why is this sequence moved? This will cause the hardware test to fail.

mhatrevi (changes requested):
Please see the feedback

drivers/src/dma.rs
swenson: I will put rearrange it back.

mhatrevi (approved):

PR #1940, opened 2025-02-04
Adding hw-1.0 build test and FMC size fix

nquarton:
Adding check that runs during build and test for FMC and RT build for hw-1.0
Adjusting FMC logging to fix size issue on hw-1.0
Adding compile-time check to limit runtime (FMC+FW) size to 118kB (128-10kB for the manifest)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -107,6 +107,13 @@ jobs:
           (cd rom/dev && ./build.sh)
           sccache --show-stats
 
+      # Make sure FMC and runtime can build for hw-1.0
+      - name: hw-1.0 build test
+        run: |
+          mkdir hw-1.0_build_test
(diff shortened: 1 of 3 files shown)


nquarton:
> We should add a compile time assertion to constrain the combined size of FMC and RT to 118kb as per Jordan's comment.

That is probably a good idea. I'm not quite sure how to go about doing that though. Do you have an idea?

nquarton:
> > We should add a compile time assertion to constrain the combined size of FMC and RT to 118kb as per Jordan's comment.
> 
> That is probably a good idea. I'm not quite sure how to go about doing that though. Do you have an idea?

This has been addressed now

rusty1968 (approved):

clundin25 (approved):

jhand2 (approved):

PR #1942, opened 2025-02-05
image: Implement new methods for RustCrypto provider

swenson:
Some new methods were added to the ` ImageGeneratorCrypto` trait but the RustCrypto implementation was not updated, so the build was failing when using the `rustcrypto` feature flag.

Also, the RustCrypto provider was specifying compressed points for ECDSA public keys, even though for OpenSSL we always use uncompressed, which was causing keys not to be parsed correctly with the RustCrypto provider.

To ensure that won't happen again, we also add a test that the workspace members that use the `rustcrypto` feature flag are now built with both sets of flags.

We also add an option to print the vendor and owner PK hashes when building a firmware image bundle. This makes it easier to test the recovery flow.

### DIFF
--- a/image/app/src/create/mod.rs
+++ b/image/app/src/create/mod.rs
@@ -27,10 +27,6 @@ use caliptra_image_serde::ImageBundleWriter;
 use caliptra_image_types::*;
 use clap::ArgMatches;
 use hex::ToHex;
-#[cfg(feature = "openssl")]
-use openssl::sha::sha384;
-#[cfg(feature = "rustcrypto")]
-use sha2::{Digest, Sha384};
 use std::path::Path;
 use std::path::PathBuf;
 use zerocopy::AsBytes;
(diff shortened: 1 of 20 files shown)


image/crypto/src/rustcrypto.rs
```diff
@@ -118,6 +118,24 @@ impl ImageGeneratorCrypto for RustCrypto {
...
+        let key_bytes = std::fs::read(path)
+            .with_context(|| format!("Failed to read public key file {}", path.display()))?;
+        Ok(ImageMldsaPubKey(to_hw_format(&key_bytes)))
+    }
+
+    fn mldsa_priv_key_from_file(path: &Path) -> anyhow::Result<ImageMldsaPrivKey> {
+        let key_bytes = std::fs::read(path)
+            .with_context(|| format!("Failed to read private key file {}", path.display()))?;
+        Ok(ImageMldsaPrivKey(to_hw_format(&key_bytes)))
+    }
```
jhand2, line 138: nit: Since this doesn't actually depend on a crypto library I don't think we need to have separate implementations. Should this just be implemented on the trait directly instead of in each implementation?

swenson: Good call.

image/app/src/create/mod.rs
```diff
@@ -372,3 +386,20 @@ fn owner_config(
         Ok(None)
     }
 }
+
+#[cfg(feature = "rustcrypto")]
+fn sha384(data: &[u8]) -> [u8; 48] {
```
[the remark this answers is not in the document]
swenson: The biggest downside to this is that the `Crypto::sha384_digest()` returns things in HW format (big-endian u32s) instead of standard u8, but we can work around that.

ArthurHeymans (approved):

mhatrevi (approved):

swenson:
Thanks!

jhand2 (approved):

Issue #1922, opened 2025-01-29
KAT needed for MLDSA


nquarton:
MLDSA is being added in 2.0 and needs a KAT like the other crypto engines.

PR #1943, opened 2025-02-05
MLDSA KAT tests

mtimkovich:
Write tests for the 3 ways to set `MldsaSeed`.


mtimkovich:
#1922

kat/src/mldsa87_kat.rs
```diff
@@ -0,0 +1,480 @@
...
+    result
+}
+
+#[derive(Default, Debug)]
+pub struct Mldsa87Kat {}
+
+impl Mldsa87Kat {
+    /// This function executes the Known Answer Tests (aka KAT) for MLDSA87.
+    ///
+    /// Test vector source:
```
mhatrevi, line 432: Is this correct?

kat/src/mldsa87_kat.rs
```diff
@@ -0,0 +1,480 @@
...
+        &self,
+        mldsa87: &mut Mldsa87,
+        trng: &mut Trng,
+    ) -> CaliptraResult<()> {
+        let mut priv_key = Mldsa87PrivKey::default();
+        let pub_key = mldsa87
+            .key_pair(&Mldsa87Seed::Array4x8(&SEED), trng, Some(&mut priv_key))
+            .map_err(|_| CaliptraError::KAT_MLDSA87_KEY_PAIR_GENERATE_FAILURE)?;
+
+        priv_key.0.reverse();
```
mhatrevi, line 456: Why do we need to reverse this?

mhatrevi (changes requested):
Please see the feedback.

kat/src/mldsa87_kat.rs
mtimkovich: Fixed

kat/src/mldsa87_kat.rs
mtimkovich: Removed

mhatrevi (approved):

PR #1946, opened 2025-02-05
dma: Remove use of `Cell<Option<>>` in DMA driver

swenson:
Generally we only like to keep one copy of `AxiDmaReg` in existence, and use Rust's borrowing rules around `&mut` to ensure that it is safely accessed. However, this is not generally possible when implementing the `Mmio` trait, since the method receiver must be immutable.

`Cell<Option<>>` or `RefCell` can provide interior mutability so that we can safely share mutable references from immutable receivers, but it complicates the code and adds bloat to the binary.

However, since Caliptra is single-threaded, as long as we are careful about how we structure the code so that we don't ever have two `AxiDmaReg` instances in existence at the same time (which might cause conflicts in programming the underlying registers), then we will have fulfilled the contract with the `unsafe` code. This safety is ensured as long as there are no nested callers of `Dma::with_dma`.

Based on feedback from #1904.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -113,36 +113,25 @@ pub struct DmaWriteTransaction {
 }
 
 /// Dma Widget
-pub struct Dma {
-    dma: Cell<Option<AxiDmaReg>>,
-}
+/// Safety: only one instance of the DMA widget should be created.
+#[derive(Default)]
+pub struct Dma {}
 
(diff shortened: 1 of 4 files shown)


ArthurHeymans (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #1948, opened 2025-02-06
Update SoC Manifest and Mailbox Spec to support Streaming Boot / Hitless Update(Caliptra 2.0)

mlvisaya:
1. SoC Manifest
* Add staging address in the image metadata entry. This will be used by
MCU to temporarily store the image being downloaded through PLDM during
Hitless Update

* Add component id in the Image Metadata Entry. This will indicate which
image component in the PLDM package it is associated with.

2. Mailbox
Update SET_AUTH_MANIFEST command to reflect all the fields in the SoC
Manifest IME
Update AUTHORIZE_AND_STASH command to specify if the digest to be
computed will be coming from the load or staging area. This command will
need to DMA the contents of the load/staging area to the SHA accelerator
to compute the digest.
Add ACTIVATE_FIRMWARE - this command activates the particular firmware
image. This is sent by MCU when it receives an 'ActivateFirmware'
message from PLDM.

3. Remove GET_IMAGE_LOAD_ADDRESS, GET_IMAGE_SIZE and
GET_IMAGE_LOCATION_OFFSET instead just have one command to get all the
image metadata using GET_IMAGE_INFO

4.  Implement ACTIVATE_FIRMWARE command

  Activate Firmware command accepts a list of image_ids to activate.
  If the Image_ID is for MCU, Caliptra will initiate an MCU reset, copies
  the staged firmware to the load address and sets the corresponding EXEC
  Go bit.
  For SOC images, it will just set the corresponding Exec Go bit


auth-manifest/README.md
```diff
@@ -33,14 +33,17 @@ The Caliptra SOC manifest has two main components:
...
-| Image Identifier        | 4            | Unique value selected by the vendor to distinguish between component images.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+| Image Identifier        | 4            | Unique value selected by the vendor to distinguish between image locations.                                          |
```
mhatrevi: this should read "images"

mhatrevi (dismissed):

auth-manifest/README.md
mlvisaya: Changed to images

mhatrevi (approved):

PR #1956, opened 2025-02-10
Update to rust 1.84

ArthurHeymans:
The reason for doing this is to be able to generate templates using rustcrypto which already supports ML-DSA87 (needs 1.81).
This PR fixes all the warnings and errors that the new toolchain brings.
Stack usage is increase by 4K with this newer version


builder/src/lib.rs
```diff
@@ -193,9 +193,15 @@ pub fn build_firmware_elfs_uncached<'a>(
                 .arg("target.'cfg(all())'.rustflags = [\"-Dwarnings\"]");
         }
 
+        //        cmd.env("RUSTFLAGS", "--emit asm");
```
swenson, line 196: Can we delete these parts that are commented out?

ureg/lib/systemrdl/src/lib.rs
```diff
@@ -473,3 +456,20 @@ pub fn translate_addrmap(addrmap: systemrdl::ParentScope) -> Result<Vec<Register
     }
     Ok(blocks)
 }
+
+#[cfg(test)]
```
swenson, line 460: I think we can actually delete the usage of `next_multiple_of` now that we have the builtin method on integer types.

swenson (dismissed):
LGTM, a couple of nits.

This will also unblock us using the Caliptra image builder in the MCU so we can do full integration tests with both emulators.

ArthurHeymans:
It fails formatting on DPE. Main branch of DPE seems to have that fixed, however it would require https://github.com/chipsalliance/caliptra-dpe/pull/388 to be cherry-picked

ureg/lib/systemrdl/src/lib.rs
ArthurHeymans: This change just happens via cargo fmt. I think functional differences are best left to a separate PR?

ArthurHeymans:
So the ubuntu 20.04 linker seems to have problems with the libcaliptra.a and fails to find some symbols. This does not happen on my local machine. Trying out more recent images in CI (20.04 is deprecated soon anyway)

swenson (approved):
LGTM

swenson:
@ArthurHeymans The symbols dropping appear to be due to them only accidentally being included in the first place -- they weren't actually defined except in test modes. I'm not sure how the code ever compiled. So, I added those symbols into the C code properly.

I also fixed an additional problem where the Go code was passing in a nil slice with 0 length, which was previously allowed but now panics.

swenson (dismissed):

swenson (approved):

PR #1959, opened 2025-02-12
Add new feature hw-2.x

jacquelinewong:
- Dependency fips204 is included only if hw-2.x feature is enabled


sw-emulator/lib/periph/Cargo.toml
```diff
@@ -29,3 +29,4 @@ zerocopy.workspace = true
 
 [features]
 "hw-1.0" = []
+"hw-2.x" = ["fips204"]
```
jhand2, line 32: nit: I think this should be `"dep:fips204"`. As-written, this means setting the `hw-2.x` feature will also set the `fips204` feature (which doesn't exist, so it would probably fail to build)

jacquelinewong: Done.

jhand2 (approved):

PR #1961, opened 2025-02-12
FPGA Boss changes

clundin25:
* Retry's with a cached token to avoid creating multiple dead runners.
* Timeout on boot and jobs in case an FPGA gets stuck.
* Panic check while FPGA waits for a job.

### DIFF
--- a/ci-tools/fpga-boss/src/main.rs
+++ b/ci-tools/fpga-boss/src/main.rs
@@ -151,13 +151,58 @@ fn copy_file(dest: &mut File, src: &mut File) -> std::io::Result<()> {
     Ok(())
 }
 
-fn log_uart_until<R: BufRead>(lines: &mut Lines<R>, needle: &str) -> std::io::Result<()> {
+fn check_for_kernel_panic(input: &str) -> std::io::Result<()> {
+    if input.contains("Kernel panic") {
+        Err(Error::new(ErrorKind::BrokenPipe, "FPGA had a kernel panic"))?;
+    }
+    Ok(())
+}
+
+fn log_uart_until<R: BufRead>(
+    lines: &mut Lines<R>,
+    needle: &str,
+    error_parser: impl Fn(&str) -> std::io::Result<()>,
+) -> std::io::Result<()> {
+    log_uart_until_helper(lines, needle, None, Some(error_parser))
+}
+
+fn log_uart_until_with_timeout<R: BufRead>(
+    lines: &mut Lines<R>,
+    needle: &str,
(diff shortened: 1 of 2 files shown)


ci-tools/fpga-boss/src/main.rs
```diff
@@ -301,10 +346,14 @@ fn main_impl() -> anyhow::Result<()> {
             }
         }
         Some(("serve", sub_matches)) => {
+            // Reuse the previous token if we never connect to GitHub.
+            // This avoids creating a bunch of offline runners if the FPGA fails to establish a
+            // connection.
+            let mut cached_token: Option<Vec<u8>> = None;
```
clundin25, line 352: Once the VCK-190 is "stable" I would like to remove the caching logic and the loop label, in favor of failing and letting a supervisor process restart the serve command.

ci-tools/fpga-boss/src/main.rs
```diff
@@ -301,10 +346,14 @@ fn main_impl() -> anyhow::Result<()> {
...
             let mut fpga = get_fpga_ftdi()?;
             let mut sd_mux = get_sd_mux()?;
             let sd_dev_path = get_sd_dev_path()?;
-            loop {
+            'outer: loop {
```
jhand2, line 356: I wonder if this code would be clearer with a helper function like `run_single_job`. That way you can just have simple error handling in that function and just handle the timeout error condition once in this loop.

jhand2: On the other hand, maybe handling it in both places is better, since it allows for a more specific error message

clundin25: I agree on both counts, the error message isn't super important since it can be inferred from the FPGA logs.

I plan to put the FPGA state management behind a trait, which I think would clean this up

clundin25: The timeout does vary based on what the FPGAs current state is

jhand2: sgtm. Were you planning to do that cleanup in this PR or a follow-up?

clundin25: Whatever is easier to review works for me - this is the code I am using on the host runner, it doesn't have to be checked in.

jhand2: Ok, let's go ahead and check this in and then can clean up in a different PR.

jhand2 (approved):

PR #1965, opened 2025-02-14
Compare MLDSA signature digest to save on ROM size.

mtimkovich:


jhand2 (dismissed):

swenson (dismissed):
LGTM. It looks like you should be able to also do the public and private key as well.

kat/src/mldsa87_kat.rs
```diff
@@ -13,9 +13,12 @@ Abstract:
...
-    Array4x8, Array4xN, CaliptraError, CaliptraResult, Mldsa87, Mldsa87PrivKey, Mldsa87PubKey,
-    Mldsa87Seed, Mldsa87SignRnd, Trng,
+    Array4x16, Array4x8, Array4xN, CaliptraError, CaliptraResult, Mldsa87, Mldsa87PrivKey,
+    Mldsa87PubKey, Mldsa87Seed, Mldsa87SignRnd, Sha2_512_384, Trng,
 };
+use caliptra_registers::sha512::Sha512Reg;
+
+use zerocopy::IntoBytes;
 
 const KAT_PUB_KEY: Mldsa87PubKey = Array4xN::new([
```
mhatrevi, line 23: We need to use digests for public and private keys as well.

mtimkovich: I use a digest for the priv_key now, but pub key is necessary for signing.

swenson: I think you could just use `pub_key` in the signature piece, after checking that the hash matches.

mtimkovich: Oh yeah, of course

swenson (approved):
LGTM

mhatrevi (approved):

PR #1966, opened 2025-02-14
rom: csrng driver should not check ack flag

swenson:
The ack flag is used to indicate that a command has been processed, and is set for only [a single clock
cycle](https://opentitan.org/book/hw/ip/csrng/doc/theory_of_operation.html#command-response). The hardware register does not seem to persist that ack flag for any longer, so we are all but guaranteed to only ever see that flag as false.

This disables checking the ack flag for now.


swenson:
Related HW issue https://github.com/chipsalliance/caliptra-rtl/issues/738

mhatrevi (approved):

swenson:
Thanks!

PR #1972, opened 2025-02-21
Update to the latest zerocopy.

clundin25:
Zerocopy's derive macro for KnownLayout has changed. Compiling in caliptra-sw fails because public interfaces are leaking these private types.

The solution is to mark them as a public field. There may be another workaround, but since these types are being serialized to and from raw bytes, it makes sense that these private fields should be visible to callers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2972,18 +2972,18 @@ dependencies = [
 
 [[package]]
 name = "zerocopy"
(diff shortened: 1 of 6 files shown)


jhand2 (approved):

swenson (approved):
LGTM

PR #1976, opened 2025-02-24
Fix FPGA build; add to build and test

swenson:
We don't do the verilator build since that requires additional tools to be installed.


clundin25 (approved):

jhand2 (approved):

swenson:
Thanks!

PR #1977, opened 2025-02-25
[fix] Fixes for recovery interface fw download

mhatrevi:
This change contains various fixes for issues found during firmware download testing from recovery interface with the hardware.


drivers/src/dma.rs
```diff
@@ -543,70 +537,89 @@ impl<'a> DmaRecovery<'a> {
...
-            // Byte0 Bit[7:4] to 0 (Recovery image index).
-            // [TODO][CAP2] the spec says this register is read-only, but there is no other way to select an image index?
+            // Set Byte0 Bit[3:0] to 0x1 ('Awaiting recovery image')
+            // Set Byte0 Bit[7:4] to recovery image index
             recovery.recovery_status().modify(|recovery_status_val| {
-                // Set Byte0 Bit[3:0] to 0x1 ('Awaiting recovery image')
-                // Set Byte0 Bit[7:4] to recovery image index
-                (recovery_status_val & 0xFFFFFF00) | (fw_image_index << 4) | 0x1
+                (recovery_status_val & Self::RESET_BYTE_0_MASK)
+                    | (fw_image_index << Self::FW_IMG_INDEX_BIT_POS)
```
swenson, line 577: I feel like we're hard-coding a lot more offsets here, but the updated RDL has a lot of these defined: https://github.com/chipsalliance/i3c-core/blob/dc3afa3df7a6dfc992d9bcbc1db8a52fe6695011/src/rdl/secure_firmware_recovery_interface.rdl#L468

Could we update the RDL and regenerate the registers, and use the autogenerated offsets instead (where possible)?

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -214,9 +215,9 @@ impl FirmwareProcessor {
                 report_fw_error_non_fatal(0);
 
                 // Drop all commands for invalid PAUSER
-                if txn.id() == RESERVED_PAUSER {
-                    return Err(CaliptraError::FW_PROC_MAILBOX_RESERVED_PAUSER);
-                }
+                // if txn.id() == RESERVED_PAUSER {
```
swenson, line 218: Does this need to be kept in?

drivers/src/dma.rs
swenson: (and the `.set()`, `etc.` methods.

ArthurHeymans:
I had to manually drop the mailbox for the tests to succeed.

rom/dev/src/flow/cold_reset/fw_processor.rs
mhatrevi: Fixed.

mhatrevi:
> I had to manually drop the mailbox for the tests to succeed.

I fixed this differently by releasing the lock in the complete function of the fake mbox transaction.

ArthurHeymans (approved):

swenson (approved):
LGTM

PR #1980, opened 2025-02-26
Respect the no-cfi feature in the drivers crate.

clundin25:
caliptra-drivers fails to compile with the "no-cfi" feature. There are portions of CFI code that is not guarded by the "no-cfi" feature.

### DIFF
--- a/drivers/src/array.rs
+++ b/drivers/src/array.rs
@@ -13,6 +13,7 @@ Abstract:
 
 --*/
 
+#[cfg(not(feature = "no-cfi"))]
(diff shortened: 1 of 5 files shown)


swenson (approved):
LGTM; maybe we could add a no-cfi build to build & test to ensure this doesn't regress?

jhand2 (approved):

clundin25:
> LGTM; maybe we could add a no-cfi build to build & test to ensure this doesn't regress?

A good point!

I'll follow up on this.

PR #1984, opened 2025-02-27
Fuse controller reg

ArthurHeymans:
This adds the caliptra-ss as submodule in order to have access to the fuse controller RDL.
The fuse controller RDL is however written from the perspective of the SS and not caliptra, which probably means that the sw and hw fields in the RDL should be reversed. An alternative hypothesis is that the sw/hw field are not correct in RDL.

TODO: fix RDL?


ArthurHeymans:
https://github.com/chipsalliance/caliptra-ss/pull/109 is a pr with suggested fix in RDL

drivers/src/dma.rs
```diff
@@ -612,3 +613,55 @@ impl<'a> DmaRecovery<'a> {
...
+    #[inline(always)]
+    pub fn with_regs<T, F>(&self, f: F) -> CaliptraResult<T>
+    where
+        F: FnOnce(FuseCtrlRegisterBlock<&DmaMmio>) -> T,
+    {
+        let mmio = DmaMmio::new(self.base, self.dma);
+        // SAFETY: we aren't referencing memory directly
+        let regs = unsafe {
+            FuseCtrlRegisterBlock::new_with_mmio(
+                // substract the recovery offset since all recovery registers are relative to it
```
swenson, line 639: nit: comment refers to recovery registers, but should be OTP registers (or just change to "registers")

registers/bin/generator/src/main.rs
```diff
@@ -152,6 +154,14 @@ fn real_main() -> Result<(), Box<dyn Error>> {
         .collect();
     rdl_files.append(&mut i3c_core_rdl_files);
 
+    let otp_ctrl_rdl_dir = rtl_dir.join("hw").join("latest").join("caliptra-ss");
```
swenson, line 157: nit: there is a block further on where you can trim the `caliptra_` prefix from the generated code, which makes it a little cleaner looking.

(We do some more aggressive pruning of these prefixes and suffixes in MCU: https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/xtask/src/registers.rs#L269C35-L269C44)

swenson (approved):
LGTM

mhatrevi (approved):

registers/bin/generator/src/main.rs
mhatrevi: Let's tackle this as a separate task. Resolving this for now.

PR #1989, opened 2025-03-03
rom/runtime: Verify auth manifest in recovery flow

swenson:
This uses the same code as the SET_AUTH_MANIFEST command, which has been slightly refactored to accommodate reading directly from the entire mailbox as an option.

Due to a timing issue when writing the FW index, the Caliptra FW (index 0) was getting sent instead of the SoC manifest (index 1). To account for this, we made a small ROM change to pre-emptively set the recovery FW index to 1 when it is finishing up.

I did some cleanup on the constant names in the recovery interface, since it was not always clear which values were for the device status and which were for the recovery status.

This was verified as working with the MCU emulator.

Next step is validating the MCU firmware, but this will be more complex since we'll have to do some DMA gymnastics to ship the MCU SRAM to the SHA accelerator, so we will do that in a separate PR.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -440,15 +440,15 @@ impl<'a> DmaRecovery<'a> {
     const MCU_SRAM_OFFSET: u64 = 0x20_0000;
 
     const FLASHLESS_STREAMING_BOOT_VALUE: u32 = 0x12;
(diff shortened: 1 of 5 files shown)


mhatrevi (approved):

swenson:
Thanks!

PR #1990, opened 2025-03-03
feat: Use the same code path for mbox access in passive mode

ArthurHeymans:
To get the manifest, fmc and runtime from the mailbox use the same code as the active mode.


swenson (approved):
LGTM

mhatrevi (approved):

PR #1992, opened 2025-03-03
feat: Add macro to define error constants with uniqueness check

ArthurHeymans:
This makes it easier to check that numbers are not used more than once.
Also it's a bit less boilerplate to add new error codes.
This also adds docstrings for each error code.

TODO: Add this to main so that cherry-picking merge conflicts are a bit easier.


leongross (approved):
Lgtm, nice to see the boilerplate being reduced and duplication tests.

PR #1998, opened 2025-03-05
Add REVOKE_EXPORTED_CDI_HANDLE command

clundin25:
Also, check available CDI slots before derivation to avoid overwriting existing CDI.

### DIFF
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -464,6 +464,9 @@ impl CaliptraError {
     pub const RUNTIME_GET_FMC_CSR_UNSUPPORTED_FMC: CaliptraError =
         CaliptraError::new_const(0x000E0055);
 
+    pub const RUNTIME_REVOKE_EXPORTED_CDI_HANDLE_NOT_FOUND: CaliptraError =
+        CaliptraError::new_const(0x000E005A);
+
     /// FMC Errors
     pub const FMC_GLOBAL_NMI: CaliptraError = CaliptraError::new_const(0x000F0001);
     pub const FMC_GLOBAL_EXCEPTION: CaliptraError = CaliptraError::new_const(0x000F0002);
--- a/libcaliptra/examples/generic/test.c
+++ b/libcaliptra/examples/generic/test.c
@@ -64,7 +64,7 @@ static void caliptra_wait_for_csr_ready(void)
  *
  * @return bool True if the signature passes verification, false otherwise
  */
-static bool caliptra_verify_ecdsa_signature(struct dpe_derive_context_exported_cdi_response* dpe_resp, struct caliptra_sign_with_exported_ecdsa_resp* sign_resp, uint8_t* tbs, size_t tbs_size)
+static bool caliptra_verify_ecdsa_signature_helper(struct dpe_derive_context_exported_cdi_response* dpe_resp, struct caliptra_sign_with_exported_ecdsa_resp* sign_resp, uint8_t* tbs, size_t tbs_size)
 {
     bool status = true;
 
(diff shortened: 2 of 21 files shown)


runtime/README.md
```diff
@@ -946,6 +946,22 @@ Command Code: `0x5357_4545` ("SWEE")
...
+Command Code: `5256_4348` ("RVCH")
+
+*Table: `REVOKE_EXPORTED_CDI_HANDLE` input arguments*
+
+| **Name**             | **Type** | **Description**
+| --------             | -------- | ---------------
+| chksum               | u32      | Checksum over other input arguments, computed by the caller. Little endian.         |
+| exported_cdi_handle  | u8[32]   | The Exported CDI handle returned by the DPE `DeriveContext` command. Little endian. |
+
+The `exported_cdi` can be created by calling `DeriveContext` with the `export-cdi` and `create-certificate` flags.
```
ekoohest, line 960: exported_cdi vs exported_cdi_handle can be confusing. One is tied with DPE spec, which is purely a CDI vector, and the other is a Caliptra concept, which is a token.

I checked the DPE spec and I am not convinced that it allows using token. (I can read wrong).

Eventually, SoC Manager should not care which one is, as long as Caliptra does the favor of providing the proprietary commands. :)

Either way, I suggest that Caliptra doc be consistent. ;)

runtime/src/revoke_exported_cdi_handle.rs
```diff
@@ -0,0 +1,49 @@
...
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[inline(never)]
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
+        let cmd = RevokeExportedCdiHandleReq::ref_from_bytes(cmd_args)
+            .map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        match drivers.caller_privilege_level() {
+            // REVOKE_EXPORTED_CDI_HANDLE MUST only be called from PL0
+            PauserPrivileges::PL0 => (),
+            PauserPrivileges::PL1 => {
```
ekoohest, line 28: I must be too behind, but can you refer me to the doc where mentioned EXPORT_CDI is limited to PL0?

Only noticed that in the @jhand2 's proposal page.

runtime/tests/runtime_integration_tests/test_sign_with_export_ecdsa.rs
```diff
@@ -157,3 +171,211 @@ fn test_sign_with_exported_never_derived() {
...
+        &mut model,
+        &mut Command::DeriveContext(&export_cdi_cmd),
+        DpeResult::Success,
+    ) else { panic!("expected derive context resp!") };
+
+    let measurement_cmd = DeriveContextCmd {
+        handle: ContextHandle::default(),
+        data: [0xa; DPE_PROFILE.get_tci_size()],
+        flags: DeriveContextFlags::RECURSIVE,
+        tci_type: 1296193104, // TODO(clundin): Figure out why this is the tci_type.
```
ekoohest, line 262: "MBVP", rings the bells? :)

runtime/README.md
ekoohest, line 960: Thinking about expored_cdi vs exported_cdi_handle, that makes a lot of sense for derive_ctx(export-cdi=true) expose only the token and not the pure CDI. In case, pure CDI is leaked, anyone can pretend being SoC manager with Caliptra endorsement.

Curious, why DPE main spec is so vague about it.

_exported-cdi: If export-cdi is true, this argument is the exported CDI value._

libcaliptra/examples/generic/test.c
```diff
@@ -925,6 +925,161 @@ int sign_with_exported_ecdsa_cdi(const test_info* info)
...
+        printf("DPE Command: OK\n");
+    }
+
+    struct dpe_derive_context_response *derive_resp = &dpe_resp.derive_context_resp;
+
+    // Revoke existing exported CDI 
+    struct caliptra_revoke_exported_cdi_handle_req revoke_req = { 0 };
+    memcpy(&revoke_req.exported_cdi_handle, exported_resp.exported_cdi_handle, sizeof(uint8_t) * 32);
+
+    if (caliptra_revoke_exported_cdi_handle(&revoke_req, false)) {
```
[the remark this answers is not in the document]
clundin25: > I have been told before any derive_ctx(export_cdi=true), revoke_export_cdi should be called to release the KeyVault to get ready accepting the derived private key.

If there is an already exported CDI occupying the slot you will first need to revoke it before a new CDI can be exported.

> Apparently, revoke_export_cdi is only required in hitless update sequence. Because it is not called in sign_with_exported_ecdsa_cdi

The exported-cdi is still present in the key vault, so there is no need to export a new one. The exported-cdi-handle accepted by `sign_with_exported_ecdsa` is a handle to the real CDI, and that CDI will be used to sign the `tbs` buffer.

> Can you elaborate, why is that?

I believe that you only need to create a new CDI with `derive_ctx(export_cdi=true)` when you need a new CA certificate. An example I've been given is a hitless update where there has been a measurement change in the TCB and a new CA certificate is needed to reflect that change.

@jhand2 - PTAL

libcaliptra/inc/caliptra_enums.h
```diff
@@ -103,6 +103,7 @@ enum dpe_error_codes {
 };
 
 enum dpe_derive_context_cmd_flags {
+    DPE_DERIVE_CONTEXT_FLAG_RECURSIVE          = ( 1UL << 24),
```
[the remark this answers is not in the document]
clundin25: I've only added the flags I need to keep the diff small.

libcaliptra/examples/generic/test.c
```diff
@@ -925,6 +925,161 @@ int sign_with_exported_ecdsa_cdi(const test_info* info)
...
+        printf("Error invalid signature.\n");
+        failure = 1;
+    }
+
+    // Add new measurement.
+    memset(&derive_context_cmd, 0, sizeof(struct dpe_derive_context_cmd));
+    derive_context_cmd.cmd_hdr.magic = DPE_MAGIC;
+    derive_context_cmd.cmd_hdr.cmd_id = DPE_DERIVE_CONTEXT;
+    derive_context_cmd.cmd_hdr.profile = 0x4;
+    derive_context_cmd.flags = DPE_DERIVE_CONTEXT_FLAG_RECURSIVE;
```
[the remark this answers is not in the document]
clundin25: I am working on it and expect to open a PR later today.

libcaliptra/examples/generic/test.c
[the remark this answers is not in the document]
jhand2: This is to support SoC FW hitless update, not Caliptra hitless update.

If the SoC needs to re-derive parts of the tree (or call `DeriveContext(recursve=true)` to update measurements), then to get a new CA cert + exported CDI, they will need to call `DeriveContext(export=true)` again. But in order to call export again, the SoC will first need to clear the exported CDI slot so that there is an available slot.

runtime/src/revoke_exported_cdi_handle.rs
jhand2: Looks like we missed this in the docs. SIGN_WITH_EXPORTED and this new command should both be marked as PL0 only.

The reason is:

1. CA cert rooted in the vendor's PKI is powerful. Want to ensure that power is limited to trusted code.
2. Only known use-case for CDI export is confidential compute which is generally expected to be implemented in PL0.

runtime/tests/runtime_integration_tests/test_sign_with_export_ecdsa.rs
clundin25: Thanks!

libcaliptra/examples/generic/test.c
```diff
@@ -925,6 +925,161 @@ int sign_with_exported_ecdsa_cdi(const test_info* info)
...
+    derive_context_cmd.cmd_hdr.profile = 0x4;
+    derive_context_cmd.flags = DPE_DERIVE_CONTEXT_FLAG_EXPORT_CDI | DPE_DERIVE_CONTEXT_FLAG_CREATE_CERTIFICATE;
+    memcpy(derive_context_cmd.context_handle, derive_resp->new_context_handle, sizeof(uint8_t) * 16);
+
+    memset(&dpe_req, 0, sizeof(struct caliptra_invoke_dpe_req));
+    dpe_req.derive_context_cmd = derive_context_cmd;
+    dpe_req.data_size = sizeof(struct dpe_derive_context_cmd);
+
+    memset(&dpe_resp, 0, sizeof(struct caliptra_invoke_dpe_resp));
+    status = caliptra_invoke_dpe_command(&dpe_req, &dpe_resp, false);
```
jhand2, line 1049: nit: since this is fairly verbose and done twice in this test, would it make sense to put into a helper function?

clundin25: Wrapped most of this with macros, so it's somewhat less verbose

libcaliptra/examples/generic/test.c
```diff
@@ -37,6 +37,23 @@ void hwmod_init(struct caliptra_buffer rom, const test_info *info);
 static int dump_array_to_file(struct caliptra_buffer* buffer, const char *filename);
 #endif
 
+#define DPE_DERIVE_CONTEXT_INIT(x)                        \
```
jhand2, line 40: I think I have a minor preference for helper functions rather than macros. I think it makes the code a bit easier to follow. For example:

```c
int invoke_dummy_derive_context(uint32_t flags, caliptra_invoke_dpe_resp *resp) {
  ...
}
```

error/src/lib.rs
```diff
@@ -464,6 +464,9 @@ impl CaliptraError {
     pub const RUNTIME_GET_FMC_CSR_UNSUPPORTED_FMC: CaliptraError =
         CaliptraError::new_const(0x000E0055);
 
+    pub const RUNTIME_REVOKE_EXPORTED_CDI_HANDLE_FAILED: CaliptraError =
```
jhand2, line 467: nit: Maybe rename to `RUNTIME_REVOKE_EXPORTED_CDI_HANDLE_NOT_FOUND`

runtime/tests/runtime_integration_tests/test_revoke_exported_cdi_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+        exported_cdi_handle: original_cdi_resp.exported_cdi,
+    });
+    cmd.populate_chksum().unwrap();
+
+    model
+        .mailbox_execute(
+            CommandId::REVOKE_EXPORTED_CDI_HANDLE.into(),
+            cmd.as_bytes().unwrap(),
+        )
+        .expect("Expected REVOKE_EXPORTED_CDI_HANDLE to pass");
```
jhand2, line 49: nit: Might be worth a test that calls export a second time after the revoke. Although I think that's covered in the libcaliptra tests, so maybe it's redundant.

runtime/tests/runtime_integration_tests/test_sign_with_export_ecdsa.rs
```diff
@@ -157,3 +171,211 @@ fn test_sign_with_exported_never_derived() {
         )
     );
 }
+
+#[test]
+fn test_sign_with_exported_cdi_measurement_update_duplicate_cdi() {
```
jhand2, line 176: nit: can we have a test case that covers the prior bug? I'm thinking

1. export
2. Sign & check signature
3. export again (should fail because we didn't revoke)
4. Sign again and make sure the original key from 1 still works

libcaliptra/examples/generic/test.c
clundin25: No problem, I'll make that change.

runtime/tests/runtime_integration_tests/test_sign_with_export_ecdsa.rs
clundin25: This test has steps 1, 3 and 4. I'll add step 2 and it would match this.

`original_cdi_resp` carries the original export to the very end, and checks that the signature still matches the certificate and that the export-cdi handle is valid.

error/src/lib.rs
clundin25: Will do

runtime/tests/runtime_integration_tests/test_revoke_exported_cdi_handle.rs
clundin25: Added a test case for this.

runtime/src/revoke_exported_cdi_handle.rs
jhand2: @clundin25 could you update the README for both commands?

runtime/tests/runtime_integration_tests/test_sign_with_export_ecdsa.rs
```diff
@@ -157,3 +171,227 @@ fn test_sign_with_exported_never_derived() {
         )
     );
 }
+
+#[test]
```
jhand2, line 175: nit: the tests in this file are pretty long. Could you add a comment to the top of each of them explaining what scenarios are being tested

libcaliptra/examples/generic/test.c
```diff
@@ -854,75 +939,104 @@ int rom_test_devid_csr(const test_info* info)
...
-    if (status) {
-        printf("Sign with exported Ecdsa failed: 0x%x\n", status);
+    // Add new measurement.
+    if(derive_context(&derive_resp, DPE_DERIVE_CONTEXT_FLAG_RECURSIVE)) {
+        printf("Failed to add new measurement.\n");
+        return 1;
+    }
+
+    // Revoke existing exported CDI 
+    memcpy(&revoke_req.exported_cdi_handle, exported_resp.exported_cdi_handle, sizeof(uint8_t) * 32);
```
jhand2, line 1017: nit: Can use `sizeof(revoke_req.exported_cdi_handle)` instead of computing manually

runtime/src/revoke_exported_cdi_handle.rs
clundin25: Done  👍

fdamato (approved):

jhand2 (approved):

PR #2004, opened 2025-03-05
Re-enable CFI in ROM

swenson:
Also creates a `cfi_check` macro.

CFI was disabled to save space during development, but can now be re-enabled since we have expanded the ROM to 96 KB.

However, one challenge is that we updated the Rust version to 1.84, which seems to make a poor decision around copying the TRNG-generated `Array4x12` during initialization instead of allocating it to the right place on the stack to begin with. In turn, this causes the compiler to generate a `memcpy`, which we explicitly check for and disallow.

The way I would generally prefer to avoid this would be to remove `Copy` from the `Array4xN` type so that we would be more mindful about when we could potentially be `memcpy`ing the arrays around, but this would be a fairly big change.

The compiler seems to be particularly stubborn about the particular TRNG output one, and the only workaround I could fine was to add a new method, `generate4()`, that generates exactly the four `u32`s we need to initialize the CFI. This seems to avoid the extraneous `memcpy`.

I also enabled adding `debug` symbols to our ELF files, as it aids in debugging issues like this, and does not add any space to the final binaries.

### DIFF
--- a/common/src/macros.rs
+++ b/common/src/macros.rs
@@ -1,0 +1,12 @@
+// Licensed under the Apache-2.0 license.
+
+#[macro_export]
+macro_rules! cfi_check {
+    ($result:expr) => {
+        if cfi_launder($result.is_ok()) {
+            cfi_assert!($result.is_ok());
+        } else {
+            cfi_assert!($result.is_err());
+        }
+    };
+}
--- a/rom/dev/src/flow/cold_reset/fw_processor.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor.rs
@@ -19,6 +19,7 @@ use crate::key_ladder;
 use crate::pcr;
 use crate::rom_env::RomEnv;
 use crate::run_fips_tests;
+#[cfg(not(feature = "no-cfi"))]
 use caliptra_cfi_derive::cfi_impl_fn;
 use caliptra_cfi_lib::CfiCounter;
 use caliptra_common::capabilities::Capabilities;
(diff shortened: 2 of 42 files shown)


ArthurHeymans (dismissed):

jlmahowa-amd (approved):

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -19,7 +19,6 @@ use crate::key_ladder;
 use crate::pcr;
 use crate::rom_env::RomEnv;
 use crate::run_fips_tests;
-#[cfg(not(feature = "no-cfi"))]
```
jhand2, line 22: Shouldn't these features stay (here and all below files)? The no-cfi feature should still be supported

common/src/macros.rs
```diff
@@ -0,0 +1,12 @@
+// Licensed under the Apache-2.0 license.
+
+#[macro_export]
+macro_rules! cfi_check {
+    ($result:expr) => {
+        if cfi_launder($result.is_ok()) {
```
jhand2, line 6: I think we want to support the no-cfi flag for this macro

rom/dev/src/flow/cold_reset/fw_processor.rs
swenson: I can add them back.

common/src/macros.rs
swenson: I believe this is baked into the underlying `cfi_assert!` macro and `cfi_launder()` function, so we don't need to worry about it at this level.

mhatrevi (approved):

swenson:
Thanks!

PR #2005, opened 2025-03-06
This re-enables the libcaliptra tests

ArthurHeymans:
- main.mk changes reflect the image preamble changes
- test.c and idev_csr_array.h changes reflect changes in idevid and CSR request return value.


libcaliptra/examples/generic/test.c
```diff
@@ -847,7 +847,7 @@ int rom_test_devid_csr(const test_info *info)
     int failure = 0;
 
     struct caliptra_buffer caliptra_idevid_csr_buf = {0};
-    caliptra_idevid_csr_buf.len = IDEV_CSR_LEN;
+    caliptra_idevid_csr_buf.len = 9001; // This holds ecc and mldsa csr
```
jlmahowa-amd, line 850: Is there not a MLDSA_IDEV_CSR_LEN yet? ECC_IDEV_CSR_LEN + MLDSA_IDEV_CSR_LEN seems like it would be better.

libcaliptra/examples/generic/test.c
```diff
@@ -883,7 +883,7 @@ int rom_test_devid_csr(const test_info *info)
     }
 
     // Compare the retrieved IDEV CSR with the expected IDEV CSR
-    if (memcmp(caliptra_idevid_csr_buf.data, idev_csr_bytes, caliptra_idevid_csr_buf.len) != 0)
+    if (memcmp(&caliptra_idevid_csr_buf.data[12], ecc_idev_csr_bytes, ECC_IDEV_CSR_LEN) != 0)
```
jlmahowa-amd, line 886: Why 12? Is there another header wrapping both CSRs?

jlmahowa-amd (commented):
Would like to see named values instead of raw numbers.

libcaliptra/examples/generic/test.c
ArthurHeymans: yes

libcaliptra/examples/generic/test.c
ArthurHeymans: It needs to be large enough to contain InitDevIdCsrEnvelope.

ArthurHeymans:
I updated the comments to reflect better what is going on.

mhatrevi (approved):

MDr164 (approved):

PR #2012, opened 2025-03-07
runtime: Add CM_STATUS, CM_IMPORT commands

swenson:
The first two cryptographic mailbox commands, which show the status of the system (for now, usage counters for AES keys) and the ability to import keys and return them as wrapped output.

(For now, this is more of a no-op, since we don't have the AES driver yet to encrypt the CMKs.)

We add some of the additional structures we will need to track AES keys (in the form of usage counters), and start adding in the necessary mailbox requests, responses, and errors.


mhatrevi (approved):

api/src/mailbox.rs
```diff
@@ -1308,6 +1322,113 @@ impl Request for ProductionAuthDebugUnlockToken {
...
+
+impl Default for Cmk {
+    fn default() -> Self {
+        Self([0u8; CMK_SIZE_BYTES])
+    }
+}
+
+#[repr(u32)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum CmKeyUsage {
```
jhand2, line 1339: Just for my understanding, is import only intended for symmetric keys? And ECDH/ML-KEM keys should be generated inside Caliptra?

error/src/lib.rs
```diff
@@ -1402,6 +1392,26 @@ impl CaliptraError {
             0x000E005A,
```
jhand2, line 1392: Not exactly related to this PR, but it looks like we got a bit out of sync between the 1.x and 2.x error codes: https://github.com/chipsalliance/caliptra-sw/blob/main/error/src/lib.rs#L468. We can do it in a follow-up PR, but we should align those.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+#[derive(Default)]
+struct KeyUsageInfo {
+    key_id: u32,
+    #[allow(unused)]
+    counter: u64,
+}
+
+impl KeyUsage {}
```
jhand2, line 72: Remove? Looks like `impl KeyUsage` is already added above.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+use caliptra_error::CaliptraError;
+use zerocopy::{transmute, FromBytes, Immutable, IntoBytes, KnownLayout};
+
+pub const KEY_USAGE_MAX: usize = 256;
+
+// We have 24 bits for the key ID.
+const MAX_KEY_ID: u32 = 0xffffff;
+
+#[derive(Default)]
+pub struct KeyUsage {
```
jhand2, line 32: KeyUsage term seems a little overloaded. The CmKeyUsage in mailbox.rs matches how I would expect the term to use used.

Should this driver be something like `CmStorage` and `CmStorageInfo`?

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Default)]
+struct KeyUsageInfo {
+    key_id: u32,
+    #[allow(unused)]
+    counter: u64,
```
jhand2, line 69: Is this the gcm counter? If so, isn't it 128 bits? (96 bit iv + 32 bit counter)

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+        // keep track of counters for AES
+        if matches!(key_usage, CmKeyUsage::AES) {
+            unencrypted_cmk.id = drivers.cryptographic_usage_data.add_counter()?;
+        }
+
+        let encrypted_cmk = EncryptedCmk {
+            domain: 0,
+            domain_metadata: [0u8; 16],
+            iv: [0u8; 12],
+            ciphertext: transmute!(unencrypted_cmk),
```
jhand2, line 154: nit: I know there's no key material in here yes, but it makes me a bit nervous to return the plaintext cmk here. Seems a bit error prone when adding code later since we have to remember to fix it.

It might make sense to return an error here for now until we have the CMK encryption code set up.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+        let key_usage = CmKeyUsage::from(cmd.key_usage);
+
+        let valid = matches!((key_usage, cmd.input_size), (CmKeyUsage::AES, 32));
+        if !valid {
+            Err(CaliptraError::RUNTIME_CMB_INVALID_KEY_USAGE_AND_SIZE)?;
+        }
+
+        let _raw_key = &cmd.input[..cmd.input_size as usize];
+        // [TODO][CAP2]: we need to generate our internal key to encrypt the CMK
+        let mut unencrypted_cmk = UnencryptedCmk {
```
bluegate010, line 136: Seems like this could be non-mut if `add_counter()` were called inline as the value is being defined?

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+        let mut key_id =
+            (self.counters.last().map(|last| last.key_id).unwrap_or(0) + 1) % MAX_KEY_ID;
+
+        // normally we could just append to the end of the list, but if keys have been deleted,
+        // we could potentially wraparound the 24-bit key id space, so we have to check
+        loop {
+            match self.counters.binary_search_by_key(&key_id, |k| k.key_id) {
+                Ok(_) => {
+                    // key_id already exists, increment and try again
+                    // this only happens where there has been wraparound due to many keys imported and deleted
```
bluegate010, line 51: We probably wouldn't want to test this every time `cargo test` is invoked, but maybe a one-off manually-invoked test or something?

runtime/src/cryptographic_mailbox.rs
bluegate010: [SP 800-38D](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf#page=29) (section 8.3) requires that we track key usage and ensure no AES key is used more than 2^32 times.

jhand2: Ah, thanks for clarifying

swenson: This is a usage counter for how many times the key has been used.

api/src/mailbox.rs
swenson: I think for now, we should only import symmetric keys (AES, HMAC / HKDF secrets). Probaly MLDSA seeds would be fine too, once IETF standardizes the format (this is currently a hotly debated topic, from what I gather). Definitely we want to avoid importing raw ECDSA keys, since they are full of weird things we have to check for.

runtime/src/cryptographic_mailbox.rs
swenson: I'll just return 0s for now instead.

api/src/mailbox.rs
jhand2: lgtm. I think for how they get used in most protocols, generating asymmetric keys internally only makes sense.

runtime/src/cryptographic_mailbox.rs
swenson: I'll add a note to do this once we add the DELETE or CLEAR commands.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -0,0 +1,178 @@
...
+            id: if matches!(key_usage, CmKeyUsage::AES) {
+                drivers.cryptographic_usage_data.add_counter()?
+            } else {
+                [0u8; 3]
+            },
+            usage_counter: 0,
+            key_material: [0u8; CMK_MAX_KEY_SIZE_BITS / 8],
+        };
+
+        // keep track of counters for AES
```
bluegate010, line 148: Orphaned comment?

bluegate010 (approved):

bluegate010 (approved):

PR #2014, opened 2025-03-08
[feat] Lock the datavault in DCCM using PMP

mhatrevi:
This change locks the cold and warm reset entries in the DCCM via RISCV Physical Memory Protection support. These entries are unlocked on a hart reset (i.e either Cold or Warm reset) The Cold Reset entries are locked immediately on a Warm/Update reset since they need to remain locked during those reset types. Locking a locked configuration is an idempotent operation.


swenson (approved):
LGTM

PR #2020, opened 2025-03-11
Add aes_clp peripheral

ArthurHeymans:
This adds the registers from the RDL. 
TODO how should it behave and is a driver needed?


swenson:
Closing as this is incorporated in #2082

PR #2021, opened 2025-03-11
Update DPE to mark TCG DICE extensions as non-critical

clundin25:
By default these extensions are marked non-critical.

This PR does not expose a way to configure this outside of the firmware interacting directly with DPE.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
+++ b/runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
@@ -16,8 +16,8 @@ use cms::{
 };
 use dpe::{
     commands::{
-        CertifyKeyCmd, CertifyKeyFlags, Command, GetCertificateChainCmd, InitCtxCmd, RotateCtxCmd,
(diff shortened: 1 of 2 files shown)


runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
```diff
@@ -297,3 +298,69 @@ fn test_invoke_dpe_rotate_context() {
...
+        // Unknown extensions are DICE extensions, and they should match the
+        // criticality set by the DPE instance.
+        if extension.parsed_extension().unsupported() {
+            assert_eq!(extension.critical, expected_criticality);
+        }
+    }
+}
+
+#[test]
+fn test_invoke_dpe_certify_key_with_critical_dice_extensions() {
```
jhand2, line 318: nit: Shouldn't these test names be `with_non_critical_dice_extensions`?

clundin25: Whoops, Yes!

Thanks!

jhand2 (approved):

PR #2027, opened 2025-03-13
Updating RT & FMC version numbers for rt-1.2.1 release

nquarton:


builder/test_data/default_image_options.toml
```diff
@@ -5,9 +5,9 @@
 # NOTE: Only the first keys in the vendor config are populated. The keys are
 # copied from image/fake-keys/src/lib.rs.
 
-fmc_version = 0x880
+fmc_version = 0x881
```
nquarton, line 8: @jhand2 I saw you had updated these to 1.2.0 at one point. I'm not 100% sure what the impact is (since most builds get the version added here: https://github.com/chipsalliance/caliptra-sw/blob/c247608cc9a240a9b329281eec5b52c14483b6a3/builder/bin/image_gen.rs#L87) So just let me know if this should be left as-is.

jhand2: Keeping this up-to-date is helpful. If you use the image_options toml file, then we use the if branch on like 81 and skip the code you linked above. Thanks!

PR #2028, opened 2025-03-14
Add digest to ECDSA / LMS verify; remove external SHA acc

swenson:
The SHA accelerator external access was removed in the hardware for 2.0, so we should stop using it in our tests. We also remove it from the emulator so that it better matches the hardware.

This has the consequence that it can no longer be used with the `ECDSA384_VERIFY` and `LMS_VERIFY` mailbox commands in the runtime.

So, we modify the runtime verification commands to allow the SHA384 hash to be passed in directly, rather than going through the peripheral. If the caller wishes to use Caliptra to compute the hash, they can use the (forthcoming) cryptographic mailbox commands for computing hashes.

This also will fix several tests for the FPGA, since it cannot use the non-existent external register access for the accelerator.


runtime/tests/runtime_integration_tests/test_ecdsa.rs
```diff
@@ -78,9 +79,7 @@ fn ecdsa_cmd_run_wycheproof() {
                 id: test.tc_id,
                 comment: test.comment.to_string(),
             });
-            model
-                .compute_sha512_acc_digest(test.msg.as_slice(), ShaAccMode::Sha384Stream)
-                .unwrap();
+            let hash = sha384(test.msg.as_slice());
```
nquarton, line 82: Once the new SHA interface is added to Caliptra, should we use that here instead or continue to use openSSL? It should be identical either way. But we may want to cover the full flow somewhere. If this is the right place, a "TODO" here might be a good idea. (Same applies to LMS).

nquarton (dismissed):

libcaliptra/inc/caliptra_types.h
```diff
@@ -102,6 +102,7 @@ struct caliptra_ecdsa_verify_req {
     uint8_t pub_key_y[48];
     uint8_t signature_r[48];
     uint8_t signature_s[48];
+    uint8_t hash[48];
 };
 
 struct caliptra_lms_verify_req {
```
jhand2, line 108: nit: Could you change the names of these structures to `*_verify_v2_req`. Makes it a little more clear that the new command is being called.

jhand2 (commented):
Overall looks good. Just some minor comments. Thanks!

runtime/tests/runtime_integration_tests/test_ecdsa.rs
swenson: Good idea. I'll add a note to switch over.

sw-emulator/lib/periph/src/root_bus.rs
```diff
@@ -427,9 +426,6 @@ pub struct SocToCaliptraBus {
     #[peripheral(offset = 0x3002_0000, mask = 0x0000_0fff)]
     mailbox: MailboxExternal,
 
-    #[peripheral(offset = 0x3002_1000, mask = 0x0000_0fff)]
-    sha512_acc: Sha512Accelerator,
```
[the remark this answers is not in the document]
swenson: I think the problem is that emulator peripherals have diverged pretty significantly at this point, and it would be a lot of work to bring them together again.

I do think it is possible to keep the CPU, bus, derive, and types emulator crates in sync though between versions, and even with the MCU, and we're working now on syncing up the MCU changes with caliptra-sw.

I think a good approach would be to keep the `main.rs` and peripheral root buses separate for the emulators, and other crates the identical as much as possible.

(We can have a longer discussion on this elsewhere, even.)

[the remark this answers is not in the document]
jhand2: > and it would be a lot of work to bring them together again

Possibly true, I just think it may be worth doing.

> I think a good approach would be to keep the main.rs and peripheral root buses separate for the emulators, and other crates identical as much as possible.

Ok that makes sense. Especially since we have PMP in the 2.0 core, it's probably best to keep that part separate. 

Can leave how you have it here.

jhand2 (approved):

jlmahowa-amd:
Will this be permanent or temporary?
Like when there is a mailbox interface for doing the SHA, would these commands revert back and just have the expectation that the SHA mailbox command was done first?

swenson:
@jlmahowa-amd This is permanent. After the SHA mailbox command, I don't think we should necessarily leave the result in the accelerator.

libcaliptra/examples/generic/test.c
```diff
@@ -567,7 +567,7 @@ int rt_test_all_commands(const test_info *info)
     }
 
     // LMS_VERIFY
-    struct caliptra_lms_verify_req lms_req = {};
+    struct caliptra_lms_verify_v2_req lms_req = {};
```
mhatrevi, line 570: Given we are not maintaining the previous caliptra_lms_verify_req, does it make sense keep the name same?

mhatrevi (dismissed):

libcaliptra/examples/generic/test.c
swenson: I think it is so that people get a more obvious compilation error if they try to use the old request

mhatrevi: Ok

jlmahowa-amd (dismissed):

nquarton (approved):

jhand2 (approved):

swenson:
Thanks!

PR #2029, opened 2025-03-14
Update rust to 1.85 and edition to 2024

ArthurHeymans:
This might be necessary for using the latest rustcrypto for mldsa in x509. No reason to attend this yet however.


ArthurHeymans:
Updating rust to 1.85 can be done separate from edition2024

PR #2031, opened 2025-03-14
[feat] Update QUOTE_PCR command with MLDSA signature

mhatrevi:
This change updates the QUOTE_PCR command to add support for PCR quote signing with the MLDSA FMC Alias private key.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -2,6 +2,7 @@
 
 use bitflags::bitflags;
 use caliptra_error::{CaliptraError, CaliptraResult};
+use caliptra_image_types::MLDSA87_SIGNATURE_BYTE_SIZE;
 use core::mem::size_of;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Ref};
 
@@ -968,8 +969,9 @@ pub struct QuotePcrsResp {
     pub nonce: [u8; 32],
     pub digest: [u8; 64],
     pub reset_ctrs: [u32; 32],
-    pub signature_r: [u8; 48],
-    pub signature_s: [u8; 48],
+    pub ecc_signature_r: [u8; 48],
+    pub ecc_signature_s: [u8; 48],
+    pub mldsa_signature: [u8; MLDSA87_SIGNATURE_BYTE_SIZE],
 }
 
 impl Response for QuotePcrsResp {}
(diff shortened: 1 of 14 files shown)


libcaliptra/inc/caliptra_types.h
```diff
@@ -145,91 +161,107 @@ struct caliptra_fw_info_resp {
...
+struct caliptra_get_tagged_tci_resp
+{
     struct caliptra_resp_header hdr;
     uint8_t tci_cumulative[48];
     uint8_t tci_current[48];
 };
 
-struct caliptra_increment_pcr_reset_counter_req {
+struct caliptra_increment_pcr_reset_counter_req
+{
```
swenson, line 185: Do we have style settings for the C code?

runtime/tests/runtime_integration_tests/test_pcr.rs
```diff
@@ -59,15 +59,15 @@ fn test_pcr_quote() {
     h.update(&resp.nonce).unwrap();
     let res = h.finish().unwrap();
     let digest: [u8; 48] = res.as_bytes().try_into().unwrap();
-    assert_eq!(resp.digest, digest);
+    assert_eq!(resp.digest[0..48], digest);
```
swenson, line 62: I don't think this is correct. This should be SHA512?

libcaliptra/inc/caliptra_types.h
mhatrevi: Don't think so. I reverted the formatting changes.

runtime/tests/runtime_integration_tests/test_pcr.rs
mhatrevi: Correct. Rebased with main-2.x and pulled the fix.

api/src/mailbox.rs
```diff
@@ -968,8 +969,9 @@ pub struct QuotePcrsResp {
     pub nonce: [u8; 32],
     pub digest: [u8; 64],
     pub reset_ctrs: [u32; 32],
-    pub signature_r: [u8; 48],
-    pub signature_s: [u8; 48],
+    pub ecc_signature_r: [u8; 48],
+    pub ecc_signature_s: [u8; 48],
+    pub mldsa_signature: [u8; MLDSA87_SIGNATURE_BYTE_SIZE],
```
jhand2, line 974: nit: what do you think about adding a u32 `flags` field to the request? Then caller can pass flags to pick which signature(s) they want to generate.

jhand2 (approved):

api/src/mailbox.rs
mhatrevi: Sounds good. Will take this as a subsequent task.

mhatrevi: https://github.com/chipsalliance/caliptra-sw/issues/2039

swenson (approved):
LGTM

PR #2032, opened 2025-03-15
DMA read/write u32

mtimkovich:
* Refactor the DMA read/write code to use u32 instead of u8 buffers.
  * This also meant that we could remove some now unnecessary `?` calls.


swenson (approved):
LGTM; I feel like this is a fair bit cleaner

mhatrevi (approved):

PR #2033, opened 2025-03-15
CSRNG, PCR quote, and HMAC fixes to match hardware

swenson:
These were all found testing against the FPGA running the 2.0-rc1 RTL.

## Correct CSRNG entropy state machine hang alert enum to match latest hw

The CSRNG entropy state machine's enums were updated, and the ALERT_HANG enum value changed, so we need to update our emulator and driver to prevent the driver from hanging.

## QUOTE_PCRS now returns a SHA512

The PCR quote hardware functionality was changed to use SHA512 instead of SHA384.

This means we need to adapt the QUOTE_PCRS command to return a SHA512 digest as well.

**Open**: Should we change the QUOTE_PCRS requrest magic number?

## HMAC HW will fail if KV value is zero or wrong mode
    
The HMAC hardware gets an error if the KV value is all 0 or if an HMAC384 key is attempted to be used with HMAC512 mode.
    
This error condition presents as ready being true and valid being false.
    
So we update the emulator to reflect that behavior, protect against it in the driver, and fix the fake ROM flow to use the correct key size and not an all-zero key.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -950,7 +950,7 @@ pub struct QuotePcrsResp {
     /// The PCR values
     pub pcrs: [PcrValue; 32],
     pub nonce: [u8; 32],
-    pub digest: [u8; 48],
+    pub digest: [u8; 64],
     pub reset_ctrs: [u32; 32],
     pub signature_r: [u8; 48],
     pub signature_s: [u8; 48],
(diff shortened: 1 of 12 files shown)


mhatrevi (approved):

swenson:
Thanks!

PR #2042, opened 2025-03-20
[fix] MLDSA driver fix for zero signatures

mhatrevi:
This update modifies the MLDSA driver to handle signatures where the last 64 bytes (known as verify_result) are zero. Since the MLDSA hardware returns a verify_result of all zeros when a signature verification fails, it becomes impossible to differentiate between valid and invalid signatures that end with 64 zero bytes. With this change, any signature with the last 64 bytes as zeros will be treated as invalid.

### DIFF
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -49,6 +49,8 @@ pub type Mldsa87SignRnd = Array4x8;
 
 type Mldsa87VerifyRes = Array4x16;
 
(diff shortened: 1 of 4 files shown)


swenson (approved):
LGTM

The last 32 bytes of the signature are the hint, which there is a very real chance (~2^-75) it could be all 0s.

But, the second-to-last 32 bytes are almost completely random, so there is virtually no chance (~2^-256) that they are all 0.

So, there should be no danger of any legitimate signature ending in 64 bytes of 0.

PR #2048, opened 2025-03-21
FPGA fixes for 2.0

swenson:
A collection of misc fixes for tests to run on the FPGA.

* Update `uio` to `0.4.0` so that when it `drop()`s it closes its file descriptor. (Otherwise, tests that run multiple hardware models will block forever.)
* The FPGA is faster and will miss fast boot status changes. We update the boot status check loop to be faster, and modify the tests to be more lax on checking state transitions when running on the FPGA.
* The boot status is currently not observable on the the FPGA due to bugs in debug unlock. So, we turn back on boot status observability in the FPGA unconditionally.
* The ARM host that the FPGA uses is slow, and if we start the `HwModel` before starting the build process, usually the FPGA will have run millions of cycles. So, we move builds to before starting the `HwModel` when timing is important.
* Fix some `fpga_realtime` feature propagation in `Cargo.toml` so that the right versions are used.
* Make sure we drop an `HwModel` before creating another one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "433cfd6710c9986c576a25ca913c39d66a6474107b406f34f91d4a8923395241"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "cipher",
  "cpufeatures",
 ]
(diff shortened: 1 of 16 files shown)


jlmahowa-amd (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #2050, opened 2025-03-24
Add flags to QUOTE_PCRS.

mtimkovich:
Specify what signature to use and zero out the other.


runtime/src/pcr.rs
```diff
@@ -54,10 +55,10 @@ impl GetPcrQuoteCmd {
         let pcr_hash = drivers.sha2_512_384.gen_pcr_hash(args.nonce.into())?;
 
         // Ecc384 signature
-        let ecc_signature = drivers.ecc384.pcr_sign_flow(&mut drivers.trng)?;
+        let mut ecc_signature = drivers.ecc384.pcr_sign_flow(&mut drivers.trng)?;
```
mhatrevi, line 58: If the flag for a signature type is not set, the sig generation should be skipped.

runtime/src/pcr.rs
```diff
@@ -53,11 +53,14 @@ impl GetPcrQuoteCmd {
...
         let pcr_hash = drivers.sha2_512_384.gen_pcr_hash(args.nonce.into())?;
 
-        // Ecc384 signature
-        let ecc_signature = drivers.ecc384.pcr_sign_flow(&mut drivers.trng)?;
+        let mut ecc_signature = Default::default();
+        let mut mldsa_signature = Default::default();
 
-        // Mldsa signature
-        let mldsa_signature = drivers.mldsa87.pcr_sign_flow(&mut drivers.trng)?;
+        if args.flags == QuotePcrsFlags::MLDSA_SIGNATURE {
```
mhatrevi, line 59: The case where both signatures are requested is not handled.

mtimkovich: Done, also added a test case for it.

mhatrevi (approved):

PR #2055, opened 2025-03-26
runtime: Add cryptographic mailbox commands for SHA

swenson:
Adds `SHA_INIT`, `SHA_UPDATE`, and `SHA_FINAL` commands using the hardware "restore" functionality for SHA384 and SHA512.

The hardware does not support restore functionality for SHA256, so I deleted that from the spec.

This was tested on the FPGA as well. I also manually tested every size between 0 and 16383 but didn't check it in since it would be too slow.

### DIFF
--- a/drivers/src/sha2_512_384.rs
+++ b/drivers/src/sha2_512_384.rs
@@ -13,18 +13,18 @@ Abstract:
 --*/
 
 use crate::kv_access::{KvAccess, KvAccessErr};
-use crate::PcrId;
-use crate::{array::Array4x32, wait, Array4x12, Array4x16, Array4x8};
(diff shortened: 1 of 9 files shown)


drivers/src/sha2_512_384.rs
```diff
@@ -386,15 +388,26 @@ impl Sha2_512_384 {
     //
     /// * `first` - Flag indicating if this is the first block
     /// * `last` - Flag indicating if this is the last block
```
mhatrevi, line 390: NIT: Update the new param description.

mhatrevi (dismissed):

mhatrevi (approved):

PR #2060, opened 2025-03-31
feat: Add real mldas87 cert and csr to rom

ArthurHeymans:
The previous certificates were handcrafted and most certainly not correct.

For now this uses out of tree code to generate cert and csr based on existing code, but using rustcrypto instead of openssl. Later work will try to integrate that code into caliptra-x509 build.rs.

The repo for generating these tbs is
"https://github.com/ArthurHeymans/test-rustcrypto-mldsa"

A difference is that x509-cert from rustcrypto always uses UTC time rather than General time which has a 2049 date limit and uses 13 instead of 15 bytes.

This also updates the smoke test for idev_id. It uses rustcrypto to verify the signature.


jhand2 (changes requested):
Please check in the code used to generate these templates. It's ok if it's in a separate crate (even a separate workspace if necessary), but we shouldn't check in binaries without also reviewing how they were generated.

ArthurHeymans:
> Please check in the code used to generate these templates. It's ok if it's in a separate crate (even a separate workspace if necessary), but we shouldn't check in binaries without also reviewing how they were generated.

Alright that turned out to be easier than expected :-D

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -41,6 +41,21 @@ use zeroize::Zeroize;
 pub struct LocalDevIdLayer {}
 
 impl LocalDevIdLayer {
+    /// Convert GeneralTime (15 bytes) to UTCTime (13 bytes)
+    fn generaltime_to_utctime(general_time: &[u8; 15]) -> [u8; 13] {
```
jhand2, line 45: This is duplicated in a few places. can you move this to a helper?

x509/build/cert_rustcrypto.rs
```diff
@@ -0,0 +1,534 @@
...
+
+impl BuilderKeys for ml_dsa::KeyPair<MlDsa87> {
+    type Signature = ml_dsa::Signature<MlDsa87>;
+    fn key_gen() -> Self {
+        let mut rng = rand::thread_rng();
+        <MlDsa87 as KeyGen>::key_gen(&mut rng)
+    }
+}
+
+impl<'a, Key> CertTemplateBuilder<'a, Key>
```
jhand2, line 186: This is a lot of duplicated code. Can you just factor out the openssl/rustcrypto calls?

x509/build/csr_rustcrypto.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl AsExtension for TcgUeid<'_> {
+    fn critical(&self, _subject: &Name, _extensions: &[Extension]) -> bool {
+        true
+    }
+}
+
+/// CSR Tempate Builder
+pub struct CsrTemplateBuilder<'a, Key> {
```
jhand2, line 57: Same comment as cert builder

jhand2 (dismissed):
Approving to remove block since I will be out for a couple weeks. But I would appreciate if you can address the couple comments before merging.

x509/build/cert_rustcrypto.rs
ArthurHeymans: I'm not sure if there is that much to gain to factor things out? It's basically adding dummy entries either rustcrypto or openssl format and having needles to those.

jhand2: I think at least factoring out the struct definitions would make sense.

ArthurHeymans: Alright done.

mhatrevi (approved):

rom/dev/src/flow/cold_reset/ldev_id.rs
mhatrevi, line 45: Let's put this helper function in some common crate.

mhatrevi (approved):

PR #2078, opened 2025-04-08
Separate caliptra nightly FPGA runner label

clundin25:
This will allow FPGAs to be kept in reserve for PRs when the nightly tests are running.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -68,6 +68,7 @@ jobs:
       rom-version: "1.0"
       rom-logging: true
       fpga-itrng: false
+      fpga-runs-on: '["caliptra-fpga-nightly"]'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -81,6 +82,7 @@ jobs:
       rom-version: "1.0"
       rom-logging: false
       fpga-itrng: false
+      fpga-runs-on: '["caliptra-fpga-nightly"]'
```
jhand2, line 85: How are the runners tagged? I'm wondering if we submit this PR will we not have any FPGAs with this label applied?

clundin25: They are tagged offline. I've configured two runners with this label.

jhand2 (approved):

PR #2082, opened 2025-04-08
runtime: Implement basic AES GCM encryption and decryption driver

swenson:
This has a stub for the peripheral, which is not functional yet.

This only has an implementation of the one-shot encryption and decryption for AES-GCM-256 calls. We will also have a more sophisticated save / restore functionality soon.

We've started integration into the cryptographic mailbox as well.

Tests (including runtime KATs) and finishing the emulator peripheral will come in the next PR.


drivers/src/aes.rs
```diff
@@ -0,0 +1,397 @@
...
+    _Cfb = 1 << 2,
+    _Ofb = 1 << 4,
+    _Ctr = 1 << 5,
+    Gcm = 1 << 6,
+    _None = (1 << 7) - 1,
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum AesKeyLen {
+    _128 = 1,
```
mhatrevi, line 79: Why are these prepended with '_'?

mhatrevi (approved):

swenson:
Thanks!

PR #2083, opened 2025-04-08
runtime: Add cryptographic mailbox random number generate command

swenson:
And integration tests.


runtime/src/cryptographic_mailbox.rs
```diff
@@ -392,4 +393,38 @@ impl Commands {
...
+
+        let size = cmd.size as usize;
+        if size > MAX_CMB_DATA_SIZE {
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+
+        let mut data = [0u8; MAX_CMB_DATA_SIZE];
+
+        for i in (0..data.len()).step_by(48) {
+            let rand: [u8; 48] = drivers.trng.generate()?.into();
```
bluegate010, line 417: To implement [CM_RANDOM_STIR](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#cm_random_stir), seems like we can't just read directly from the TRNG, but instead would need to pull from an in-memory DRBG?

swenson: Yeah, we can't implement it is as-is. I'm going to take a look at that command as one of my next few. Whether it is supported or not might depend on what the hardware (CSRNG, iTRNG) allows us to do.

swenson:
Thanks!

PR #2086, opened 2025-04-09
runtime: Set device status to running recovery in to load additional images

swenson:
And add support in the emulator recovery interface to accept incoming recovery block writes so that the MCU emulator (or other code) can write to the recovery registers. This is necessary for the MCU recovery flow to complete.

This is tested and working with the MCU emulator as working.


mhatrevi (dismissed):

mhatrevi (approved):

swenson:
Thanks!

Issue #2019, opened 2025-03-11
Incorrect condition for calling  DISABLE_ATTESTATION


sunguk-bin-fad:
I'm using Caliptra1.1 RTL and FW

In libcaliptra, once caliptra_mailbox_send_complete() is called, mbox_status is always busy. (MBOX_CSR_MBOX_STATUS == 0)  If warm reset is triggered, Caliptra finally disables attestation, because mbox_status is busy.

For example, (check these from caliptra_check_status_get_response())
If mbox_status == CMD_COMPLETE, once clearing mbox_execute mbox_status goes to CMD_BUSY.
If mbox_status == DATA_READY, once clearing mbox_execute mbox_status goes to CMD_BUSY.

I don't think this is normal operation. Is there any history of mbox-related fix? or is there something that I need to check?

sunguk-bin-fad:
@jhand2 
Can you check this issue or invite the appropriate person to discuss it? This mailbox issue is reproducible from sw-emulator. After checking-out the main branch, I ran the libcaliptra example application while checking mbox_status both before and after calling caliptra_mbox_write_execute(false).
https://github.com/chipsalliance/caliptra-sw/blob/8b9133a35c3097d91fcebe5ef6d2dfaa9b77d736/libcaliptra/src/caliptra_api.c#L571-L585

before releasing lock : mbox_status (0x102 or 0x101) - CMD_COMPLETE or DATA_READY
after releasing lock : mbox_status (0x0) - which means **CMD_BUSY**

Once warm-reset is triggered, it finally calls DISABLE_ATTESTATION.
https://github.com/chipsalliance/caliptra-sw/blob/8b9133a35c3097d91fcebe5ef6d2dfaa9b77d736/runtime/src/lib.rs#L271-L276

I think either the mailbox status is incorrect(must not be busy), or the condition indicating that Caliptra was running(mbox status = cmd_busy) is wrong. I'd like to hear your opinion on this.

jhand2:
mbox_status is only valid while Caliptra is executing a command. Once the lock is released and `mbox_fsm_ps == MBOX_IDLE` then `mbox_status` will reset to 0 (CMD_BUSY).

[...]

PR #2090, opened 2025-04-10
Store runtime execution status in DCCM

clundin25:
This is because the `mailbox_flow_done` register is zeroed on warm resets.

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -51,6 +51,7 @@ pub const MEASUREMENT_MAX_COUNT: usize = 8;
 const DPE_DCCM_STORAGE: usize = size_of::<DpeInstance>()
     + size_of::<u32>() * MAX_HANDLES
     + size_of::<U8Bool>() * MAX_HANDLES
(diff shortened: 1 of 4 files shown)


drivers/src/persistent.rs
```diff
@@ -237,6 +238,8 @@ pub struct PersistentData {
     #[cfg(feature = "runtime")]
     pub attestation_disabled: U8Bool,
     #[cfg(feature = "runtime")]
+    pub runtime_cmd_active: U8Bool,
+    #[cfg(feature = "runtime")]
```
clundin25, line 242: This isn't necessarily the most correct place to store this bool. At the same time, I don't want to allocate a big new chunk for this.

Thoughts?

jhand2: Since it's related to disabled attestation, I think this is fine.

jhand2 (approved):

swenson (approved):
LGTM

sunguk-bin-fad:
This resolves #2019.

PR #2094, opened 2025-04-11
runtime: Cryptographic mailbox random stir implementation

swenson:
This calls the "update" function on the TRNG to add in additional entropy. The amount of data is padded to a multiple of 48 bytes, and then chunked into 48-byte chunks to feed into the update function to stir it in.

This is only possible if we are using the internal CSRNG, so we simply return an error if the external TRNG is used.

I also noticed that `svn` was missing from the `AUTHORIZE_AND_STASH` command, so I added that back in while I was modifying the README.


runtime/src/cryptographic_mailbox.rs
```diff
@@ -501,4 +502,37 @@ impl Commands {
...
+        if cmd_bytes.len() > core::mem::size_of::<CmRandomStirReq>() {
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+        let mut cmd = CmRandomStirReq::default();
+        cmd.as_mut_bytes()[..cmd_bytes.len()].copy_from_slice(cmd_bytes);
+
+        let size = (cmd.input_size as usize)
+            .next_multiple_of(MAX_SEED_WORDS * 4)
+            .min(MAX_CMB_DATA_SIZE);
+        if size > MAX_CMB_DATA_SIZE {
```
bluegate010, line 521: With the `.min(MAX_CMB_DATA_SIZE)` just above, is this case reachable?

runtime/src/cryptographic_mailbox.rs
```diff
@@ -501,4 +502,37 @@ impl Commands {
...
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+
+        let mut additional_data = [0u32; MAX_CMB_DATA_SIZE / 4];
+        for (i, chunk) in cmd.input[..size].chunks_exact(4).enumerate() {
+            // avoid panic even though this is impossible
+            if i >= additional_data.len() {
+                break;
+            }
+            additional_data[i] = u32::from_be_bytes(chunk.try_into().unwrap());
```
jhand2, line 531: Why swap endianness? We're stirring the same data either way, right?

runtime/src/cryptographic_mailbox.rs
swenson: Ah, I think I left these in when I was trying to remove a panic. I can delete the `.min()` above now.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -501,4 +502,37 @@ impl Commands {
...
+
+        let size = (cmd.input_size as usize)
+            .next_multiple_of(MAX_SEED_WORDS * 4)
+            .min(MAX_CMB_DATA_SIZE);
+        if size > MAX_CMB_DATA_SIZE {
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+
+        let mut additional_data = [0u32; MAX_CMB_DATA_SIZE / 4];
+        for (i, chunk) in cmd.input[..size].chunks_exact(4).enumerate() {
```
jhand2, line 526: Assuming the endianness swap isn't necessary, I think you should be able to use zerocopy to interpret this as a word buffer instead of copying.

runtime/src/cryptographic_mailbox.rs
swenson: In a cryptographic sense it doesn't matter at all.

But it's possible NIST might care that the bits that are brought in are loaded into the block cipher in the correct orientation.

Since we have to convert between `u8` and `u32` anyway, I thought we might as well swap endianess. :shrug:

runtime/src/cryptographic_mailbox.rs
swenson: I think it's best to swap endianness in case NIST cares about the order the bits are loaded.

bluegate010: Can we document this as the reason we're not using zerocopy? Personally I think it's doubtful NIST would care about bit order here. SP 800-90A 8.7.2 states that "the additional input used in DRBG requests is not considered to be a critical security parameter unless any secret information included in the additional input qualifies as a critical security parameter... The value of the additional input may be either secret or publicly known; its value is arbitrary".

swenson: Ah, good quote. I'll just use zerocopy in that case.

bluegate010 (approved):

swenson:
(FPGA test failures are unrelated and I will investigate separately.)

jhand2 (approved):

mhatrevi (approved):

PR #2098, opened 2025-04-14
Build debian image for VCK-190

clundin25:
Changes between the FPGA image for the VCK-190 and ZCU-104

* UART changes
* Bootloader change
* Kernel change


clundin25:
> Sorry for being slow to get to reviewing this.

No problem! I think it's finally in a good spot.

> Carl does there need to be updates to create_boot.bin.sh/update_boot.bin and the fpga README? Since you are adding a petalinux_project directory I'm concerned that create_boot.bin will make a mess of things since it deletes petalinux_project. 

Good point, I'll tweak the script to preserve the existing petalinux_project.

> Does the procedure change since you have the kernel modules under petalinux_project? Do we still need the kernel modules under hw/fpga/XXXX since you have them embedded?

Currently in the PR they are copied, which is not great. I think it would be easiest/clearest to leave the C code where it is and modify the script to copy it over. That avoids burying the kernel modules in the petalinux project folder.

hw/fpga/create_boot_bin.sh
```diff
@@ -22,39 +22,34 @@ trap '{
...
 sed -i '/versal-gpio/{n;s/disabled/okay/}' images/linux/system.dts
 dtc -I dts -O dtb -o images/linux/system.dtb images/linux/system.dts
 
 echo Packaging boot files
 petalinux-package --boot --format BIN --plm --psmfw --u-boot --dtb --force
+
+echo "Building io-module"
+petalinux-build -c io-module
+
+# Currently unused
```
jlmahowa-amd: Suggest you not have the rom-backdoor mentioned here.
We should remove it since it is unused and uses the Zynq addressing from 1.x

jlmahowa-amd (approved):
Nice, that resolves my concerns.

hw/fpga/create_boot_bin.sh
clundin25: Sounds good - I will delete it and rollup the commits.

jlmahowa-amd (approved):

PR #2104, opened 2025-04-17
fix: Don't have stack size blow up

ArthurHeymans:


mhatrevi:
This issue is fixed and this change is no longer needed.

PR #2109, opened 2025-04-18
drivers: AES driver fixes and KAT

swenson:
I wrote a KAT for the AES driver from the FIPS GCM test suite and validated that it passes on an FPGA.

I then wrote the emulator AES peripheral to match that behavior.

This doesn't yet support streaming mode, which is necessary for the mailbox commands, but that will be next.

To support the AES KAT, I had to add the driver to the ROM as well. It might be possible to avoid this, but it is a bit difficult even with feature flags due to feature flag unification (which tends to turn on the AES driver in the KAT when compiling even if only requested in the runtime).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,14 +3,38 @@
 version = 3
 
 [[package]]
-name = "aes"
-version = "0.8.2"
+name = "aead"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


bluegate010 (approved):

swenson:
FPGA failures appear unrelated. Debugging those separately.

mhatrevi (approved):

swenson:
Thanks!

PR #2112, opened 2025-04-20
Cherrypick from main

mhatrevi:


runtime/src/main.rs
```diff
@@ -28,14 +28,7 @@ use core::hint::black_box;
 #[cfg(feature = "std")]
 pub fn main() {}
 
-const BANNER: &str = r#"
```
swenson, line 31: Was this just to save space in 1.x?

I'll miss it. :) We aren't as space-constrained in 2.x so we could maybe keep it.

sw-emulator/lib/periph/src/mailbox.rs
```diff
@@ -194,7 +194,7 @@ pub enum MailboxRequester {
 impl From<u32> for MailboxRequester {
     fn from(value: u32) -> Self {
         match value {
-            0 => MailboxRequester::Caliptra,
+            0xFFFF_FFFF => MailboxRequester::Caliptra,
```
swenson, line 197: I thought this user was no longer valid?

swenson (dismissed):
LGTM, two questions

sw-emulator/lib/periph/src/mailbox.rs
jhand2: It's not valid for the SoC to use it because it's reserved for Caliptra's own use.

runtime/src/main.rs
jhand2: Ya, just space savings. I'm inclined to leave it off since I think we'll end up space constrained on 2.x as well.

jhand2 (dismissed):

swenson (approved):

PR #2113, opened 2025-04-20
Support streaming boot (Caliptra 2.0)

mlvisaya:
This patch include changes to support streaming boot and hitless update
of SOC images as outlined in this documentation:

https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/docs/src/image_loading.md

1. Changes in SOC Manifest
Add component_id field : This refers to the component identifier used in
PLDM
Add classification: This is the PLDM classification of the image
Add image_load_address: The AXI address where the image will be loaded
during streaming boot
Add image_staging_address: The AXI address where the image will be
temporarily stored during firmware update

2. Addition of GET_IMAGE_INFO Mailbox command
GET_IMAGE_INFO mailbox command returns the metadata of the image.
This is used by the MCU so that it can load the image to the correct
location during streaming boot.

3. Changes in AuthorizeAndStash Mailbox command
Added LoadAddress and StagingAddress as source for hash -> when
specified, AuthorizeAndStash command will stream the image from load
address or staging address to the SHA Accelerator, then the calculated
SHA will be compared with the one in the manifest.

Added image_size field -> This is to specify how many bytes does the
AuthorizeAndStash need to transfer to the SHA Accelerator.

4. Addition of Test SRAM peripheral

The Test SRAM peripheral emulates an external memory used by a custom
MCU where the streaming boot image will be consumed.
An arbitrary address of 0x50000 is used for this purpose.

5. Addition of DMAImage utility

This utility provides an interface to transfer images from a given AXI
address to the SHA accelerator

6. Increased the stack size from 101 to 107 kB

Due to the increased size of the SOC manifest, the stack usage has also
increased since the mailbox buffer is allocated from the stack. This
causes a stack overflow. The stack size needs to be increased by 6kB to
compensate for this increase.


api/src/mailbox.rs
```diff
@@ -1244,13 +1254,17 @@ pub enum ImageHashSource {
...
     ShaAcc,
+    LoadAddress,
+    StagingAddress,
 }
 
 impl From<u32> for ImageHashSource {
     fn from(val: u32) -> Self {
         match val {
             1_u32 => ImageHashSource::InRequest,
             2_u32 => ImageHashSource::ShaAcc,
```
mhatrevi, line 1265: Let's remove ShaAcc as image source.

drivers/src/dma.rs
```diff
@@ -725,6 +725,79 @@ impl<'a> DmaRecovery<'a> {
...
+        let mut acc_op = sha_acc
+            .try_start_operation(ShaAccLockState::NotAcquired)?
+            .ok_or(CaliptraError::RUNTIME_INTERNAL)?;
+
+        acc_op.stream_start_384(length, true)?;
+
+        let write_addr = Sha512AccCsr::PTR as u64 + SHA_ACC_DATA_IN_OFFSET;
+
+        cprintln!(
+            "[dma-image] SHA384 image digest calculation: source = {}, length = {}",
```
mhatrevi, line 758: Print the source addr in hex.

drivers/src/dma.rs
```diff
@@ -725,6 +725,79 @@ impl<'a> DmaRecovery<'a> {
...
+        Self { dma }
+    }
+
+    pub fn sha384_image(
+        &self,
+        sha_acc: &'a mut Sha2_512_384Acc,
+        source: AxiAddr,
+        length: u32,
+    ) -> CaliptraResult<Array4x12> {
+        const SHA_ACC_DATA_IN_OFFSET: u64 = 0x14;
```
mhatrevi, line 745: Add comment explaining what 0x14 is.

drivers/src/dma.rs
```diff
@@ -725,6 +725,79 @@ impl<'a> DmaRecovery<'a> {
...
+        self.dma.flush();
+
+        let read_transaction = DmaReadTransaction {
+            read_addr,
+            fixed_addr: read_fixed_addr,
+            length: payload_len_bytes,
+            target: DmaReadTarget::AxiWr(write_addr, write_fixed_addr),
+        };
+        self.dma.setup_dma_read(read_transaction, block_size);
+        cprintln!("Wait for DMA to complete");
```
mhatrevi, line 795: Trace is not needed.

runtime/src/get_image_info.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// * `auth_manifest_image_metadata_col` - A reference to the `AuthManifestImageMetadataCollection` containing the metadata entries.
+    /// * `cmd_fw_id` - The firmware ID from the command to search for.
+    ///
+    /// # Returns
+    ///
+    /// * `Option<&AuthManifestImageMetadata>` - Returns `Some(&AuthManifestImageMetadata)` if a matching entry is found,
+    ///   otherwise returns `None`.
+    ///
+    #[inline(never)]
+    fn find_metadata_entry(
```
mhatrevi, line 81: This function is duplicated. Can we move it to a common place?

mhatrevi (changes requested):
Please see the feedback.

mhatrevi (approved):

PR #2117, opened 2025-04-23
runtime: Add cryptographic mailbox commands for AES GCM

swenson:
Encrypt and decrypt.

The tests were also validated on the FPGA.

I tried to use the standard `ghash` crate to implement the AES GCM GHASH function, but it doesn't support save/restore, and adding that functionality using mathematics ended up being much more complicated than just implementing our own GHASH function.

The AES GCM emulator peripheral is not particularly efficient, but could be improved by moving to a proper AES-CTR that has save/restore support since we calculate GHASH ourselves.


swenson:
I see the 1-byte-at-a-time encryption is failing on the FPGA. It's likely an IV bookkeeping issue. I will test this more and get a fix later today and also fix the clippy lints.

swenson:
Added a commit to fix clippy lints and to avoid an extra AES GHASH save/restore when doing small updates, which was causing strange behavior on real hardware. Should be good to go now.

mhatrevi (approved):

vsonims (approved):

drivers/src/aes.rs
```diff
@@ -124,6 +150,379 @@ impl Aes {
...
+        let tag = self.compute_tag(context.aad_len as usize, len)?;
+        Ok((written, tag))
+    }
+
+    /// Saves and returns the current GHASH state.
+    fn save(&mut self) -> CaliptraResult<[u8; AES_BLOCK_SIZE_BYTES]> {
+        let mut ghash_state = [0u8; 16];
+        self.with_aes(|aes| {
+            wait_for_idle(&aes);
+            for _ in 0..2 {
```
bluegate010, line 413: Why is this repeated? (Here and elsewhere)

swenson: It's explained at the top of the file. Any register listed as "shadowed" is required to be programmed twice, presumably for glitch protection. It's an OpenTitan thing: https://opentitan.org/book/hw/ip/aes/doc/programmers_guide.html#block-operation

swenson:
Thanks!

PR #2120, opened 2025-04-24
runtime: Add ECDH generate and finish commands

swenson:
These are validated to work against OpenSSL's ECDH commands in the integration test.
    
I noticed that there was an unnecessary reference in enums used as arguments to the ECDSA driver's arguments, so I removed that while I was there.
    
Builds on top of https://github.com/chipsalliance/caliptra-sw/pull/2117 to encrypt and decrypt the context, as well as to use the result of ECDH as an AES key to prove that the result is correct.


mhatrevi (approved):

vsonims (approved):

swenson:
Thanks. Rebased.

runtime/README.md
```diff
@@ -1530,16 +1530,14 @@ Command Code: `0x434D_4547` ("CMEG")
...
-| chksum        | u32              |                                       |
-| fips_status   | u32              | FIPS approved or an error             |
-| context size  | u32              | size of context                       |
-| context       | u8[context size] | Used as the input to `CM_ECDH_FINISH` |
-| exchange data | u8[96]           | i.e., the public point                |
+| **Name**      | **Type** | **Description**                       |
+| ------------- | -------- | ------------------------------------- |
+| chksum        | u32      |                                       |
+| fips_status   | u32      | FIPS approved or an error             |
+| context       | u8[76]   | Used as the input to `CM_ECDH_FINISH` |
```
bluegate010, line 1539: In a separate file you explain the math behind this magic number (`pub const CMB_ECDH_ENCRYPTED_CONTEXT_SIZE: usize = 76; // = unencrypted size + 12 bytes IV + 16 bytes tag`). Could we add an explainer here as well? (And could refer to that in CM_ECDH_FINISH input args table as well.)

bluegate010 (dismissed):

runtime/README.md
swenson: Sure, I can add that.

vsonims (approved):

PR #2122, opened 2025-04-24
runtime: AES CBC commands

swenson:
Adds support for AES CBC commands, and adds them to the driver,
emulator, and docs.

Programming in the AES peripheral in "block" modes, like CBC, requires particular sequencing for reading and writing blocks, where you have to write data one block ahead of the block you are reading. Some of this is detailed in: https://opentitan.org/book/hw/ip/aes/doc/programmers_guide.html#block-operation

This PR builds on #2117 and #2120

## Stack size increase
tl;dr I had to temporarily increase the stack size, but I think I can
fix this in the next couple of weeks.

The longer version:

The `MailboxResp` `enum` is large enough to trigger some edge cases in
what I think is LLVM. I found that, if you take the state of the
`MailboxResp` before this commit and add a single enum option to
`MailboxResp`, even something as simple as `A(())`, then the stack size
blows up by +17 KB. Even if that enum option is never used anywhere in
the code.

I had some success in splitting the crytographic mailbox commands off
into a separate struct and handling them separate in the
`runtime/src/lib.rs` handler, and managed to get the stack bloat down to
about +8 KB from where it is now.

I have a WIP branch where I'm starting to rewrite the larger commands to
stage their responses directly into the mailbox SRAM using `zerocopy`'s
`mut_from_prefix()` function, so that we never spill these large
responses to the stack. I think this is a better way forward to reduce
stack usage, and it will also mean we don't have to copy through
the mailbox registers.

I'm open to suggestion on how else to split things up or reduce the size
of the enum, but for now, I think we should just keep the stack size
change in until we have time to implement one of these fixes.


api/src/mailbox.rs
```diff
@@ -1971,6 +2032,291 @@ pub struct GetImageInfoResp {
 }
 impl Response for GetImageInfoResp {}
 
+// CM_AES_CBC_ENCRYPT_INIT
+#[repr(C)]
+#[derive(Debug, IntoBytes, FromBytes, KnownLayout, Immutable, PartialEq, Eq)]
+pub struct CmAesCbcEncryptInitReq {
```
jhand2, line 2038: For GCM is makes sense to have separate commands.

But for all the unauthenticated cipher modes, should we just have a single set of commands that accepts a mode enum?

I think inevitibly someone will ask for another mode since the hw supports it. It would be good to not need to add a bunch of new commands for that.

bluegate010 (dismissed):

sw-emulator/lib/crypto/src/aes256cbc.rs
```diff
@@ -12,17 +12,54 @@ Abstract:
...
+        Self::new(&[0u8; AES_256_IV_SIZE], &[0u8; AES_256_KEY_SIZE])
+    }
+}
 
 const AES_256_IV_SIZE: usize = AES_256_BLOCK_SIZE;
 
+type Aes256Encryptor = cbc::Encryptor<aes::Aes256>;
 type Aes256Decryptor = cbc::Decryptor<aes::Aes256>;
 
 impl Aes256Cbc {
```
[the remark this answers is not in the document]
swenson: The emulator peripheral is generic. This is just the helper library to make implementing those modes more easily.

api/src/mailbox.rs
swenson: I thought about this and even did that at first, but realized that the encryption and decryption are different enough to warrant a separate command. To follow NIST recommendations, the encryption operation should return the initial random IV that it generates on the first command.

We could just always return that as a parameter and have it be 0 for decryption, but that felt a little weird to me for some reason.

vsonims (dismissed):

api/src/mailbox.rs
swenson: We discussed this offline and are going to go with four commands that have a mode so that we can add additional modes later.

swenson:
Stack issue has been filed as #2123

vsonims (approved):

mhatrevi (approved):

PR #2131, opened 2025-04-29
Add streaming support to sw-emulator

ArthurHeymans:
This fixes the driver to properly set the strobe, which also rewriting parts to make it a bit more clear using rust iterators.

### DIFF
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -340,32 +340,28 @@ impl Mldsa87 {
         Mldsa87::wait(mldsa, || mldsa.status().read().msg_stream_ready())?;
 
         // Reset the message strobe register.
-        mldsa.msg_strobe().write(|_| 0xF.into());
+        mldsa.msg_strobe().write(|s| s.strobe(0xF));
 
         // Stream the message to the hardware.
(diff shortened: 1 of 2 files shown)


drivers/src/mldsa87.rs
```diff
@@ -340,31 +340,32 @@ impl Mldsa87 {
...
+        let last_strobe = match remainder.len() {
+            0 => 0b0000,
+            1 => 0b0001,
+            2 => 0b0011,
+            3 => 0b0111,
+            _ => 0b0000, // should never happen
+        };
+        mldsa.msg_strobe().write(|s| s.strobe(last_strobe));
+        match remainder.len() {
+            0 => (), // do nothing
```
mhatrevi, line 362: This is not correct. We need to write a dummy value in the msg[0] register for the case when remainder.len() is 0 (i.e. when strobe is set to 0)

mhatrevi (changes requested):
Please see the feedback.

ArthurHeymans:
> Please see the feedback.

thanks. I addressed it

mhatrevi:
This PR is not required as the change is added to https://github.com/chipsalliance/caliptra-sw/pull/2133

PR #2138, opened 2025-05-05
Use Rust stable version 1.70 to match 'rust-toolchain.toml'

clundin25:
This fixes the compiler errors in #1907


leongross:
Thanks for pointing this out!

PR #2143, opened 2025-05-08
ci/fpga-boss: add SDMux implementation for usbsdmux

leongross:
Add support for [usbsdmux](https://github.com/linux-automation/usbsdmux/).
This introduces breaking changes and renders the old syntax used in scripts and workflows unusable.

### DIFF
--- a/ci-tools/fpga-boss/src/ftdi.rs
+++ b/ci-tools/fpga-boss/src/ftdi.rs
@@ -29,7 +29,6 @@ pub struct FtdiCtx {
     ctx: *mut libftdi1_sys::ftdi_context,
     port_path: UsbPortPath,
 }
-
 impl FtdiCtx {
     pub fn open(port_path: UsbPortPath, iface: FtdiInterface) -> anyhow::Result<Self> {
         let dev = device_from_path(&port_path)?;
--- a/ci-tools/fpga-boss/src/main.rs
+++ b/ci-tools/fpga-boss/src/main.rs
@@ -33,7 +33,7 @@ fn cli() -> clap::Command<'static> {
             arg!(--"sdwire" [PORT_PATH] "USB port path to the hub chip on the SDWire (ex: 3-1.2)")
                 .value_parser(value_parser!(OsString)))
         .arg(
-            arg!(--"usbsdmux" [USBID] "USB uniuqe ID for the usbsdmux device (ex: 00048.00643)")
+            arg!(--"usbsdmux" [USBID] "USB unique ID for the usbsdmux device (ex: 00048.00643)")
                 .value_parser(value_parser!(OsString))
         )
         .arg(
--- a/ci-tools/fpga-boss/src/sd_mux.rs
+++ b/ci-tools/fpga-boss/src/sd_mux.rs
@@ -120,36 +120,22 @@ impl SdMux for UsbsdMux {
     /// The ID used to construct an UsbdsdMux instance is the symlink name without the leading "id-".
     /// Resolving these symlinks by ID resolves them to the according /dev/sgX device.
     fn open(sdmux_id: String) -> anyhow::Result<Self> {
-        for entry in fs::read_dir("/dev/usb-sd-mux/")? {
-            let entry = entry?;
-            let entry_id = format!("id-{}", sdmux_id);
-            if entry.file_name() == std::ffi::OsStr::new(&entry_id) {
-                // this will be of format ../sgx
-                let had_dev = entry
-                    .path()
-                    .read_link()
-                    .unwrap()
-                    .to_string_lossy()
-                    .to_string();
+        let symlink_path = format!("/dev/usb-sd-mux/id-{}", sdmux_id);
 
-                return Ok(Self {
-                    scsi_generic_name: had_dev.strip_prefix("../").unwrap().to_string(),
-                });
-            }
-        }
-        Err(anyhow::anyhow!(
-            "Failed to find SCSI generic device for USBSDMUX with ID {}",
-            sdmux_id
-        ))
+        // this will be of format ../sgx
+        let had_dev = std::fs::read_link(symlink_path)?
+            .to_string_lossy()
+            .to_string();
+
+        return Ok(Self {
+            scsi_generic_name: had_dev.strip_prefix("../").unwrap().to_string(),
+        });
     }
 
     // For now we use the python cli tool implementation provided by the vendor.
     // Further, we assume there is only one usbsdmux device connected, exposed
     // via /dev/sg0.
     fn set_target(&mut self, target: SdMuxTarget) -> anyhow::Result<()> {
-        // UsbSDMux has three targets: DUT, HOST and CLIENT.
-        // TODO: handle client
-
         let target = match target {
             SdMuxTarget::Dut => "dut",
             SdMuxTarget::Host => "host",


ci-tools/fpga-boss/README.md
```diff
@@ -68,7 +89,12 @@ $ sudo bash build.sh
 
 ```
 $ cd ci-tools/fpga-boss
-$ cargo run -- --zcu104 1-14.3 --sdwire 1-14.6 flash ../fpga-image/out/image.img
+# SDwire
+$ cargo run -- --zcu104 1-14.3 --sdmux_model sdwire --sdmux_id 1-14.6 flash ../fpga-image/out/image.img
```
clundin25, line 93: I think it's is enough to add a separate flag for `usbsdmux` to avoid a breaking change.

E.g. keep `--sdwire` as is, and add a `--usbsdmux` that takes a  `UsbSDMux` ID.

leongross: I see your point, which is fair, about not breaking backward compatibility. The reason I added the `sdmux_model` flag was to make it ready for possible further models to be added in the future. Then one would have to add a new flag for each new SDMux, which, in my opinion, can be quite confusing.

clundin25: I agree this approach scales better but I think we can punt on it for now, until a third mux tool is introduced

leongross: I talked to @MDr164 and he made a proposal I think could work out for us:
We could keep the `--sdwire` flag and also add the generic `--sdmux_model` and `--sdmux_id` flags, such that `--sdiwire <foo>` will internally be translated to `--sdmux_model sdwire --sdmux_id <foo>`. This way we keep the --sdwire flag for compatibility and can still introduce the new changes. What is your opinion on this?

ci-tools/fpga-boss/Cargo.toml
```diff
@@ -7,6 +7,11 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+# used for unit testing on systems with guaranteed attached usbsdmux.
+sdmux_usbsdmux = []
+
+
```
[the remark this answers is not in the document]
leongross: The idea was to prevent the regular CI from running this since it does not have a `usbsdmux` connected to it. Developers who have one can run it locally to ensure functionality using `cargo test --features sdmux_usbsdmux`.
Alternatively, we could add a check in the test that skips it if no `usbsdmux` is detected.

ci-tools/fpga-boss/README.md
clundin25: I think it's not too heavy of a lift to update the scripts, so if the preference is to add the generic flags let's move forward with it.

Another alternative:

Identify the correct backend by sequentially attempting to open the path in the `--sdwire` with each mux backend. That avoids a breaking change and keeps the interface generic.

ci-tools/fpga-boss/Cargo.toml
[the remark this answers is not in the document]
clundin25: Ah okay, so this crate will eventually be tied into the CI test suite?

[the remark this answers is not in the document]
leongross: I wrongly assumed this was the case, but you are right, it's not atm. I will remove it.

ci-tools/fpga-boss/README.md
clundin25: I think it should be okay but I don't want to bikeshed on this. I am okay with moving forward with the proposal.

> My personal favorite is the 'overshadowing' option.

Which option is this?

leongross: As per our last conversation, I implemented both the `--sdwire` and `--usbsdmux` flags; closing up this thread.

leongross:
The CI seems to be flaky again, will investigate:
```
Error: Unable to download artifact(s): Artifact not found for name: caliptra-fpga-kmod
        Please ensure that your artifact is not expired and the artifact was uploaded using a compatible version of toolkit/upload-artifact.
```

ci-tools/fpga-boss/README.md
```diff
@@ -80,8 +99,9 @@ fpga-boss as root.
 
 ### To observe the UART output of the now booting FPGA:
 
+Example for `sdwire`
 ```
-$ cargo run -- --zcu104 1-14.3 --sdwire 1-14.2  console
+$ cargo run -- --zcu104 1-14.3 --sdwire 1-14.6  console
```
clundin25, line 104: nit: This change seems inadvertent (And conflicts with the earlier example command).

ci-tools/fpga-boss/src/ftdi.rs
```diff
@@ -24,10 +24,12 @@ pub enum BitMode {
     BitBang = 0x01,
     CBus = 0x20,
 }
+
 pub struct FtdiCtx {
     ctx: *mut libftdi1_sys::ftdi_context,
     port_path: UsbPortPath,
 }
+
```
clundin25: nit: Prefer to omit whitespace changes

ci-tools/fpga-boss/src/main.rs
```diff
@@ -24,15 +23,18 @@ use std::{
...
 
 fn cli() -> clap::Command<'static> {
     clap::Command::new("fpga-boss")
         .about("FPGA boss tool")
         .arg(
             arg!(--"sdwire" [PORT_PATH] "USB port path to the hub chip on the SDWire (ex: 3-1.2)")
-                .value_parser(value_parser!(UsbPortPath))
+                .value_parser(value_parser!(OsString)))
+        .arg(
+            arg!(--"usbsdmux" [USBID] "USB uniuqe ID for the usbsdmux device (ex: 00048.00643)")
```
clundin25: nit: Typo
```suggestion
            arg!(--"usbsdmux" [USBID] "USB unique ID for the usbsdmux device (ex: 00048.00643)")
```

ci-tools/fpga-boss/src/ftdi.rs
clundin25: Or was this added by cargo fmt?

ci-tools/fpga-boss/src/sd_mux.rs
```diff
@@ -38,24 +42,174 @@ impl FromStr for SdMuxTarget {
...
+                    .read_link()
+                    .unwrap()
+                    .to_string_lossy()
+                    .to_string();
+
+                return Ok(Self {
+                    scsi_generic_name: had_dev.strip_prefix("../").unwrap().to_string(),
+                });
+            }
+        }
```
clundin25: nit: It may be clearer to just check if `sdmux_id` is in the directory, instead of iterating each entry.

ci-tools/fpga-boss/src/sd_mux.rs
```diff
@@ -38,24 +42,174 @@ impl FromStr for SdMuxTarget {
...
+            sdmux_id
+        ))
+    }
+
+    // For now we use the python cli tool implementation provided by the vendor.
+    // Further, we assume there is only one usbsdmux device connected, exposed
+    // via /dev/sg0.
+    fn set_target(&mut self, target: SdMuxTarget) -> anyhow::Result<()> {
+        // UsbSDMux has three targets: DUT, HOST and CLIENT.
+        // TODO: handle client
```
clundin25: Do we need this mode?

clundin25 (dismissed):

ci-tools/fpga-boss/src/ftdi.rs
leongross: No this was no cargo fmt change, I will revert it.

ci-tools/fpga-boss/src/sd_mux.rs
leongross: No, this was a note I took during development. I will remove it since we do not need it.

ci-tools/fpga-boss/README.md
leongross: Maybe I am confused, but to me it looks like sdwire has been `1-14.6` in the above file, and that was the reason I corrected the occurrence of `1-14.2` I found to `1-14.6` to be aligned to the others (cf [L34](https://github.com/chipsalliance/caliptra-sw/blob/main/ci-tools/fpga-boss/README.md?plain=1#L43) and [L71](https://github.com/chipsalliance/caliptra-sw/blob/main/ci-tools/fpga-boss/README.md?plain=1#L71)).

clundin25: Ah you are right! Sorry for the noise!

PR #2144, opened 2025-05-08
runtime: Don't use the giant MailboxResp enum of doom

### LINKED ISSUE #2123: Refactor big mailbox response enum to use traits
The large `MailboxResp` enum in `api/src/mailbox.rs` seems to be causing strange behavior with the Rust or LLVM code emitter, which causes the enum to be unnecessarily copied to the stack. Since the enum is ~9 KB, an extra copy or two of the enum eats stack SRAM quickly.

There are a couple of ways to do work around this, but the easiest one would be to replace the enum with a trait. The only methods we need are the `as_bytes()`, `as_mut_bytes()`, and `populate_checksum()`, and those are mostly delegated to traits already (`Response`, `IntoBytes`), so it should not be a huge amount of work to remove this enum entirely, and just refer to the traits underneath.

If possible, it would be nice to do something similar with mailbox requests.

swenson:
Rust or LLVM don't seem to handle using enums with a lot of variants (some of which are very large) very well: extra copies end up on the stack unnecessarily.

So instead, we allocate a fixed block of memory and pass a mutable reference to that to each mailbox command for them to stage their responses directly into using `zerocopy`.

As a further optimization, since in the runtime we are only using a fraction of the mailbox SRAM, we can use the unused portion of the mailbox SRAM as a staging area. Once the response is finalized, we copy it to the beginning of the mailbox SRAM.

In total, this reduces the stack memory consumption by 18 KB and also reduces code size by about 1400 bytes in the runtime.

Fixes #2123


ArthurHeymans (approved):

runtime/src/lib.rs
```diff
@@ -193,134 +204,161 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
-    // Handle the request and generate the response
-    let mut resp = match CommandId::from(req_packet.cmd) {
+    // The mailbox is much larger than the request and response buffers.
+    // We can use the second half of the mailbox to stage the response, which we
+    // can copy to the first half of the mailbox before sending it. This saves us
+    // ~10 KB of stack space.
+    let mbox = Mailbox::raw_mailbox_contents_mut();
+    let mbox_len = mbox.len();
+    let (_, resp) = mbox.split_at_mut(mbox_len / 2);
+    let resp = &mut resp[..MAX_RESP_SIZE];
```
ArthurHeymans, line 214: Maybe it's worth some kind of validation that requests don't overlap this value? One one way would be to statically compare sizes, another is to dynamically compare to payload().len().

mhatrevi (approved):

runtime/src/lib.rs
swenson: Will resolve in separate PR

swenson:
Thanks!

PR #2146, opened 2025-05-09
runtime: Don't return computed tag in GCM decrypt

swenson:
Originally I added returning the computed GCM tag on decryption as a flexibility measure.

However, I've given this some further thought, and I think it is a bad idea for a general API to return a GCM tag for arbitrary inputs: a bad actor could request tags for any AAD or ciphertext. If they included no AAD or ciphertext, then they could discover the ECB encryption of the 0 block. Between these, it opens the door to attacks like Invisible Salamanders, which a bad actor could then craft valid tags for arbitrary inputs for any key.

So, I've removed returning the computed GCM tag for decryption and return only whether the input matched or not.

This also means that we need to limit the size of tags we check to be at least 64 bits (8 bytes) long when checking, as otherwise someone could simply brute force compute the GCM tag a byte-at-a-time.

Also, I added a memory check as a follow on to #2144 to ensure that the mailbox requests and responses don't overlap.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -450,6 +450,8 @@ pub enum MailboxReq {
     ProductionAuthDebugUnlockToken(ProductionAuthDebugUnlockToken),
 }
 
+pub const MAX_REQ_SIZE: usize = size_of::<MailboxReq>();
+
 impl MailboxReq {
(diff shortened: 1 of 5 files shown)


mhatrevi (approved):

swenson:
Thanks!

PR #2153, opened 2025-05-13
runtime: Add HMAC mailbox command

swenson:
The hardware does not support save/restore. We could potentially offer this functionality in firmware by implementing HMAC on top of the SHA engine, but that means we would have to do KATs for both HW and FW, FIPS validation for HW and FW, and so on. Plus, we would not be able to use the mailbox commands to validate the HMAC hardware engine.

So, instead we do a one-shot HMAC command that is limited to the cryptographic mailbox max size (currently 4 KB).

I did some minor refactoring of the HMAC driver to eliminate a few unnecessary `&` references in the code on `enum`s that are either a single word or slices themselves.


bluegate010 (approved):

jhand2 (approved):

mhatrevi (approved):

swenson:
Thanks!

swenson:
Closing in favor of #2164

PR #2160, opened 2025-05-14
runtime: MLDSA public key, sign, and verify mailbox commands

swenson:
And update docs, including clarifying their usage with respect to DPE.
    
I took the opportunity to clean up an unnecessary reference in the MLDSA driver (since the enum type only contains references or small elements).

Builds on #2153


swenson:
Thanks!

swenson:
Closing in favor of #2164

PR #2161, opened 2025-05-14
Repurpose fips_status field in FW_INFO

bluegate010:
By making this field indicate whether Caliptra has been operating in FIPS mode since boot, other mailbox commands shouldn't need to include a fips_status field.


jhand2 (approved):
No objections. cc @JohnTraverAmd or @nquarton since they added the fips_status fields originally.

swenson:
I think with the changes in #2494 this may no longer be possible?

bluegate010:
> I think with the changes in #2494 this may no longer be possible?

Yeah this isn't needed anymore anyways.

PR #2162, opened 2025-05-14
runtime: Add ECDSA public key, sign, verify commands

swenson:
For the cryptographic mailbox.

Builds on #2160 and #2153


api/src/mailbox.rs
```diff
@@ -3148,6 +3173,151 @@ impl Request for CmMldsaVerifyReq {
...
+        }
+    }
+}
+
+impl Response for CmEcdsaSignResp {}
+
+// CM_ECDSA_VERIFY
+#[repr(C)]
+#[derive(Debug, IntoBytes, FromBytes, KnownLayout, Immutable, PartialEq, Eq)]
+pub struct CmEcdsaVerifyReq {
```
jhand2, line 3276: Should we deprecate the other ECDSA verify command? They're very similar, but it probably makes sense to keep the CM one since it follows the semantics of the other CM commands.

swenson: Hmm. We could, though the ECDSA384_SIGNATURE_VERIFY takes a public key hash, which people might be useful if the private key has not been loaded from the cryptographic mailbox or if the message is larger than the CM mailbox message limit.

mhatrevi (approved):

swenson:
Thanks!

swenson:
Closing in favor of #2164

PR #2164, opened 2025-05-15
runtime: Add HMAC KDF in Counter Mode mbox command

swenson:
This is similar to, but not quite, the same as HKDF-Expand. I'll add HKDF-Expand and HKDF-Extract in a separate PR.

Builds on #2162, #2160, and #2153

PR #2168, opened 2025-05-16
runtime: Add HKDF extract and expand

swenson:
Follows RFC 5869 and NIST SP 800-56Cr2 Section 5 (Two-Step Derivation)
with a few limitations:

- The salt input for HKDF-Extract is limited up to 0..48 (SHA384) or 0..64
  (SHA512) bytes. Technically we could support 129 bytes or greater and follow
  the spec by hashing the salt, but for sizes 48..128 (or 64..128), we
  cannot match the spec using the HMAC hardware, since it limits the key
  to 48 or 64 bytes.
- The IKM is fixed to 48 or 64 bytes (for SHA384 and SHA512) even though
  the spec is more lax.
- The label is limited to the standard cryptographic mailbox limit
  (currently 4 KB).
- The Expand output size must be less than or equal to the SHA size it
  uses, e.g., you can't output a 64-byte key from HKDF-SHA384, even
  though the spec allows this by iterating the HMAC function.

We couldn't use the official NIST test vectors for HKDF-Extract as their
HMAC test vectors all have sizes that don't match our keys.

We could potentially allow arbitrary sizes for the IKM to HKDF-Extract
by having an option in the mailbox message to either use the CMK input
or to have a buffer that is passed in.

Users can also bypass HKDF-Extract entirely and use HKDF-Expand on any
valid HMAC CMK, which is also allowed by the spec. In that case, it
becomes a special case of HMAC KDF Counter Mode (with an 8-bit counter
appended after the info instead of 32 bits before).

Builds on #2164, #2162, #2160, and #2153


mhatrevi (approved):

bluegate010 (approved):

swenson:
Thanks!

Issue #2159, opened 2025-05-14
Zero Exported CDI on Disable Attestation


clundin25:
We do this for the DICE alias key and CDI, let's also do it for the exported CDI.

clundin25:
FYI @nquarton

PR #2170, opened 2025-05-16
Zero exported CDI when attestation is disabled.

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/2159


jhand2 (dismissed):
Looks like it needs a cargo fmt. Otherwise, lgtm.

PR #2176, opened 2025-05-17
FPGA build: Don't disable networking during boot.bin build

swenson:
Disabling networking fails due to a permission issue for me on Ubuntu 24.04 at least. It isn't necessary to do anyway.


mhatrevi (approved):

swenson:
Thanks!

PR #2178, opened 2025-05-21
[update] Add authorization manifest digest to FW_INFO command

### LINKED ISSUE #2179: Add measurement manifest measurement to FW info command
A digest of the measurement manifest should be included in the FW info command. Ideally, this will also be added in some way within DPE, but this solution should be sufficient for 1.x

mhatrevi:


nquarton (approved):

runtime/src/fips.rs
```diff
@@ -195,7 +195,6 @@ pub mod fips_self_test_cmd {
             env.persistent_data.get().fht.rom_info_addr.get()? as *const RomInfo as usize / 64;
 
         let mut digest = unsafe { env.sha256.digest_blocks_raw(rom_start, n_blocks)? };
-        cprintln!("ROM Digest: {}", HexBytes(&<[u8; 32]>::from(digest)));
```
swenson, line 198: Is this just for space reasons? This is useful generally.

swenson (approved):

runtime/src/fips.rs
mhatrevi: Correct, for space reasons. Wanted to avoid changing binary sizes at this point.

PR #2180, opened 2025-05-21
Enable fake_boot_test

ArthurHeymans:
- Set the pqc_key
- Set enable fake_rom bit
- change check for boot status (we're in production so that's not available)
- For now comment out the Dice info check


ArthurHeymans:
TODO: add MLDSA equivalent. https://github.com/chipsalliance/caliptra-sw/pull/2169 the derive functions in here for MLDSA will be needed

PR #2181, opened 2025-05-21
Add empty verify res cfi checks

mtimkovich:


swenson (commented):
The assembly looks good.

Not something for this PR, but we should really write some ROM tests for these to check that they really do panic, since we will never trigger the panics in normal code.

swenson (dismissed):

swenson (approved):
LGTM

mhatrevi (approved):

PR #2196, opened 2025-05-30
Switch to OpenSSL for MLDSA image signing.

mtimkovich:


image/crypto/src/openssl.rs
```diff
@@ -116,45 +118,37 @@ impl ImageGeneratorCrypto for OsslCrypto {
...
     }
 
-    // [TODO][CAP2]: Update to use OpenSSL API when available.
     fn mldsa_sign(
         &self,
         msg: &[u8],
         priv_key: &ImageMldsaPrivKey,
         pub_key: &ImageMldsaPubKey,
     ) -> anyhow::Result<ImageMldsaSignature> {
-        // Private key is received in hw format which is also the library format.
```
mhatrevi, line 126: Please keep this comment. It's important to understand why the DWORDs are not being reversed (as we do it for ECC)

mhatrevi (approved):

PR #2197, opened 2025-05-31
[update] Recovery Interface flow update

mhatrevi:


runtime/src/recovery_flow.rs
```diff
@@ -49,7 +49,7 @@ impl RecoveryFlow {
             dma_recovery.set_device_status(DmaRecovery::DEVICE_STATUS_RUNNING_RECOVERY_IMAGE)?;
```
swenson, line 49: I think this needs to be changed to be 3 (awaiting recovery image) now.

mhatrevi: It's ok to have an intermediate device status of 'DEVICE_STATUS_RUNNING_RECOVERY_IMAGE'. request_image function will set the device status to 'awaiting recovery image' before downloading the authorization manifest and MCU RT.

swenson (approved):
LGTM

PR #2200, opened 2025-06-02
[update] Runtime POPULATE_IDEV_MLDSA87_CERT command + README updates

mhatrevi:


runtime/src/populate_idev.rs
```diff
@@ -38,14 +38,49 @@ impl PopulateIDevIdCertCmd {
...
+                return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+            }
+
+            let flags = drivers.persistent_data.get().manifest1.header.flags;
+            // PL1 cannot call this mailbox command
+            if flags & PL0_PAUSER_FLAG == 0 {
+                return Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL);
+            }
+
+            let mut tmp_chain = ArrayVec::<u8, MAX_MLDSA_CERT_CHAIN_SIZE>::new();
```
swenson, line 76: I suspect this is why the stack size blew up. Could we instead check the sizes here, and if they are okay, clear `drivers.mldsa_cert_chain` and populate it directly?

mhatrevi: Issue is with the create_mldsa_cert_chain function. I'll address this as a separate change.

swenson (approved):
LGTM

PR #2205, opened 2025-06-03
drivers: Add AES CMAC driver, KAT, emulator

swenson:
Adds AES support for the CMAC mode, including emulator support, KAT, and a firmwre test.

In addition, we also add support for loading AES keys from the key vault.

This was validated on the FPGA.


mhatrevi (approved):

drivers/src/aes.rs
```diff
@@ -661,9 +675,11 @@ impl Aes {
             wait_for_idle(&aes);
         });
 
-        self.load_key(key);
+        if !key.sideload() {
+            self.load_key(key)?;
+        }
```
bluegate010, line 680: There's a couple places here where we load the key either before or after we program the control register, based on whether the key is in KV. Just curious, why is that?

drivers/src/aes.rs
```diff
@@ -1096,6 +1128,113 @@ impl Aes {
...
+                        .manual_operation(false)
+                        .sideload(key.sideload())
+                });
+            }
+            wait_for_idle(&aes);
+        });
+
+        if !key.sideload() {
+            self.load_key(key)?;
+        }
```
bluegate010, line 1157: Optional: this pattern repeats four times in this file. Could be worth a helper.

bluegate010 (approved):

drivers/src/aes.rs
swenson: If the key is being sideloaded, then the KV has to be primed before the control register is programmed. Once the control register is set to sideload, then it immediately loads from the KV.

drivers/src/aes.rs
swenson: It's a bit tricky since the pattern is repeated, but the control register is programmed differently each time. I'm not sure it would be economical to define a pair of functions to remove an if statement?

swenson:
Thanks!

mhatrevi (approved):

PR #2209, opened 2025-06-03
[feat] Implementing FMC Alias MLDSA CSR and corresponding GET_FMC_ALIAS_MLDSA87_CSR mailbox command

mhatrevi:


runtime/src/get_fmc_alias_csr.rs
```diff
@@ -20,14 +20,48 @@ impl GetFmcAliasCsrCmd {
...
+    }
+}
+
+pub struct GetFmcAliasMldsaCsrCmd;
+impl GetFmcAliasMldsaCsrCmd {
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[inline(never)]
+    pub(crate) fn execute(
+        drivers: &mut Drivers,
+        _cmd_args: &[u8],
```
swenson: nit: you can just not pass this in

swenson (dismissed):
LGTM, one minor nit

runtime/src/get_fmc_alias_csr.rs
mhatrevi: Updated.

swenson (approved):
LGTM

PR #2218, opened 2025-06-05
[update] Add zeroization for hmac finalize

mhatrevi:


swenson (approved):
LGTM but zeroize appears to be missing from several other functions in `impl Hmac`.

mhatrevi:
> LGTM but zeroize appears to be missing from several other functions in `impl Hmac`.

Ack, we'll tackle this in a separate item for all peripherals.

PR #2220, opened 2025-06-05
Add go.sum

clundin25:
This makes `rtool` easier to build with nix.


leongross (approved):
Good to see this coming in.

PR #2222, opened 2025-06-05
Add runtime support for 1.0 and 1.1 RTL with single build

nquarton:
- added simultaneous support for HW 1.0 and 1.1 in FMC and RT FW
- removed hw-1.0 feature from crates where possible
- added test to ensure FMC and FW binaries do not change with hw-1.0 feature


builder/src/firmware.rs
```diff
@@ -45,6 +45,12 @@ pub const FMC_WITH_UART: FwId = FwId {
     features: &["emu"],
 };
 
+pub const FMC_WITH_UART_HW_1_0: FwId = FwId {
```
nquarton, line 48: These were just added for a test to confirm that adding the hw-1.0 feature does not affect the resultant binary. They are inherently useless otherwise given that requirement. If there is a better way to do this, let me know.

common/src/lib.rs
```diff
@@ -40,11 +40,11 @@ pub use pcr::{PcrLogEntry, PcrLogEntryId, RT_FW_CURRENT_PCR, RT_FW_JOURNEY_PCR};
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 21 * 1024;
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 96 * 1024;
+pub const RUNTIME_SIZE: u32 = 97 * 1024;
 
-// Max size of runtime code should be 118K to allow room for the manifest
+// Max size of runtime code should be 120K to allow 8k for the manifest
```
nquarton, line 45: I believe we can relax this a bit given the header is under 6k.

nquarton:
The size changes from adding this support as built with "cargo run --release -p caliptra-builder --bin image  -- --fw <name>"

**Original Image (latest HW):**
FMC: 18800 B
RT FW: 89660 B

**Original Image HW-1.0:**
FMC: 18820 B
RT FW: 89316 B

**Combined RTL support:**
FMC: 18848 B
RT FW: 89962 B

drivers/src/soc_ifc.rs
```diff
@@ -52,6 +52,12 @@ pub fn reset_reason() -> ResetReason {
     }
 }
 
+#[allow(dead_code)]
+pub fn is_hw_gen_1_0() -> bool {
+    let soc_ifc = unsafe { SocIfcReg::new() };
+    soc_ifc.regs().cptra_hw_rev_id().read().cptra_generation() == 0x1
```
zhalvorsen, line 58: We don't have any patched versions of 1.0 HW right? I'm wondering if we should mask out the patch portion of this register. If I remember right, we jumped straight to 1.1.

nquarton: There are a couple patches for the 1.0 RTL but I don't believe they are reflected in the generation register. My understanding was the 1.0.x RTL versions show 0x1 here and the 1.1.x versions show a 0x11. In my opinion, there probably should be a place for a patch though so it should have been 0x100 and 0x110 or something.

But as it is, this function just checks for 0x1 as being 1.0 HW. I'll dig through the patch_v1.0 branch to confirm this is correct for all the 1.0.x patches. The RTL tagging is a little confusing though unfortunately.

nquarton: I confirmed it was 0x1 for all of 1.0.x. From the most recent 1.0 patch, [this line](https://github.com/chipsalliance/caliptra-rtl/blob/7c953512251707d6f77d7d9a2f6be186ed013d50/src/soc_ifc/rtl/soc_ifc_reg.sv#L5487) dates back to July 2023. It was [updated to 0x11](https://github.com/chipsalliance/caliptra-rtl/blob/5f85fb4bc95b753a2f7d042db7dc2644ca1e8c49/src/soc_ifc/rtl/soc_ifc_reg.sv#L5839) prior to 1.1.0 so there should be no overlap.

zhalvorsen: Great, thanks for double checking this!

swenson (approved):
LGTM

PR #2233, opened 2025-06-06
emulator: SHA accelerator streaming mode is not limited to 256 KB

swenson:
I checked the RTL and with the hardware team and they confirmed that we can use the full 4 GB (minus 1 byte) for streaming mode.

This is going to matter as the MCU firmware could potentially be larger than 256 KB.


swenson:
Thanks!

PR #2237, opened 2025-06-06
rom: I3C recovery PROT_CAP needs to set mandatory bits and version

swenson:
We thought initially that the hardware would have set the version and the mandatory bits of PROT_CAP, but in hardware, these values are 0 by default.


mhatrevi (approved):

swenson:
Thanks!

PR #2244, opened 2025-06-11
Add CMAC KDF counter mode

swenson:
And KAT.

This will be used by the Device Ownership Transfer flows in another PR soon.


swenson:
Thanks!

PR #2250, opened 2025-06-12
[update] Split PCR_QUOTE command into PCR_QUOTE_ECC384 and PCR_QUOTE_MLDSA87 commands.

### LINKED ISSUE #2210: Split QUOTE_PCRS into two commands: one with an ECC signature and another with an MLDSA signature.
Unless there's a specific use case requiring verification of both ECC and MLDSA signatures, could we separate the QUOTE_PCR command for ECC and MLDSA? 
For instance, in the SPDM protocol, the user will verify the corresponding signature based on the negotiated asymmetric algorithm. Implementing two commands would streamline this use case and ensure compatibility

mhatrevi:


runtime/src/pcr.rs
```diff
@@ -49,54 +49,98 @@ impl GetPcrQuoteCmd {
...
+                resp.digest.copy_from_slice(pcr_hash.0.as_bytes());
+                resp.signature = signature.into();
+
+                Ok(core::mem::size_of::<QuotePcrsMldsa87Resp>())
+            }
+        }
     }
 }
 
+// pub struct GetPcrQuoteMldsa87Cmd;
```
swenson: Does this need to be here?

mhatrevi: Removed.

swenson (approved):
LGTM

PR #2258, opened 2025-06-18
[fix] Check MLDSA status.valid signal before programming variable message

mhatrevi:
This fix addresses a condition where the MLDSA peripheral may complete a signature verification operation before the variable-sized message has been programmed. Previously, the MLDSA driver waited for the msg_stream_ready signal to be asserted, and then proceeded to program the message. This update ensures that the driver exits the wait early if the hardware aborts the verification operation, allowing it to handle premature termination scenarios gracefully and avoid waiting on an invalid condition.


swenson (approved):
LGTM

PR #2259, opened 2025-06-18
DMA: recovery flow fixes based on FPGA testing

swenson:
This fixes multiple issues with the DMA driver and recovery flow found when testing on the FPGA:

* The MCU SRAM offset has changed in the MCI registers
* Caliptra notifies the MCU to boot through bit 2 in the exec MCI register
* The SHA accelerator stores the AXI user who locks it. So if Caliptra is going to feed the SHA accelerator from DMA, it has to lock the SHA accelerator through DMA (ditto for dlen and execute). So, we added a DMA wrapper to the SHA accelerator registers.
* The recovery device status need to be 3 for the recovery FIFO to respond.
* The FPGA does not support fixed burst AXI transfers, so we add a temporary FPGA-specific DMA driver to translate fixed burst transfers to a series of single dword transfers.

This is tested on a full subsystem FPGA through validating and running the MCU runtime firmware. (This requires a patch on the SHA accelerator to honor the endian toggle for streaming mode.)


mhatrevi (approved):

swenson:
Thanks!

PR #2262, opened 2025-06-18
Derive stable identity root secrets as part of LDevID

swenson:
These will be used by Device Ownership Transfer and OCP LOCK.

We also generate the AES CMB keys in ROM and store them in persistent data, as this will be needed for the upcoming ROM mailbox commands.


bluegate010 (approved):

bluegate010:
It would be a pain but may be worth adding some smoke tests to ensure the key doesn't change across firmware updates or lifecycle states. I did something similar in https://github.com/chipsalliance/caliptra-sw/pull/1874 by extending mbox_responder.rs and test_update_reset.rs

swenson:
@bluegate010 my plan was to add such tests when we add the mailbox commands in the next PR

swenson:
Thanks!

PR #2263, opened 2025-06-18
Skip FPGA ROM tests in nightly when testing with older ROMs

nquarton:


nquarton:
All the nightly tests passed with this branch (the action failed at the end for the publish step as expected).

https://github.com/chipsalliance/caliptra-sw/actions/runs/15936937163

drivers/src/hmac384.rs
```diff
@@ -176,7 +176,8 @@ impl Hmac384 {
 
         let rand_data = trng.generate()?;
 
-        #[cfg(any(feature = "fmc", feature = "runtime"))]
+        // Support HW 1.0 RTL (except for ROM)
+        #[cfg(not(feature = "rom"))]
```
zhalvorsen, line 180: Should these return an error if they are called from 1.0 ROM? It'd pretty bad if we accidentally add code that calls these

nquarton: Actually this function can be called in ROM, it will just use the flow below starting on line 188. Rather than being a normal if/else I set it up like this so that 1.0 check can be completely removed by that feature switch.

zhalvorsen: Even for 1.0 ROM?

nquarton: I saw you marked this as resolved, but the 1.0 ROM is only buildable from the caliptra-1.0 branch. So we can still backport patches to those old ROM branches, but otherwise the ROM moves on and does not maintain backward compatibility with older ROMs or older HW.

zhalvorsen (approved):

drivers/Cargo.toml
```diff
@@ -29,6 +29,7 @@ caliptra-cfi-derive-git = { workspace = true, optional = true }
 
 [features]
 emu = []
+rom = []
```
swenson, line 32: In general, mutually exclusive tags are not good without compile-time assertions to check that multiple aren't getting included, since it is easy to accidentally include several because of feature unification.

However, in this PR, it looks like it is relatively safe, but it would still be best to add a compile-time assert.

nquarton: This is a good point, I have filed an issue for this and will look for some time to look into this soon (though I am going to be out for a bit starting Thursday). 

Test evidence would indicate this is working okay in the case of the builder creating each component separately as it does. But I am a little confused what will happen when something like the runtime test suite goes and pulls in a whole bunch of stuff. I feel like you may get a compilation of drivers that is not actually executed that has a whole mess of features it in.

PR #2269, opened 2025-06-25
Add option in emulator to use MCU Recovery Interface

mlvisaya:
Added an option in emulator to use MCU I3C Recovery Interface instead of
the local recovery interface in Caliptra emulator.

This allows the MCU to have better synchronization with the Caliptra
Recovery flow. MCU has synchronous access to the I3C Recovery Interface,
Caliptra uses DMA to asynchronously access Recovery Interface. If MCU is
using the Caliptra Recovery Interface, this synchronous access is not
possible.

For AXI streaming boot, MCU also needs to have better control on the
FIFO data going to the Recovery Interface.

If Caliptra is using the MCU Recovery Interface, register reads/writes
to the recovery registers will send memory reads/writes events to the
MCU. MCU will respond back with the corresponding events.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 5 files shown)


swenson (commented):
Ideally we would have tests for this in the emulator DMA file.

Also, have you checked this out locally and done any tests with `caliptra-mcu-sw`?

mlvisaya:
> Ideally we would have tests for this in the emulator DMA file.
> 
> Also, have you checked this out locally and done any tests with `caliptra-mcu-sw`?

Hey Chris, yes I did test it with the MCU changes: https://github.com/chipsalliance/caliptra-mcu-sw/pull/243
I can't create a new unit test for this in the Caliptra core, because it would need MCU to respond to the events.
By default, it would still use recovery interface in the Caliptra core. MCU will need to supply the use_mcu_recovery_interface to true  in the CaliptraRootBusArgs to allow Caliptra to use MCU's recovery interface.
I have unit tests in MCU to verify this flow.

swenson (approved):
LGTM

swenson (approved):

PR #2276, opened 2025-06-26
Adding most recent non-fatal FW error to FW_INFO cmd

nquarton:
FIPS level 3 requires the most recent error to always be available. Because we clear non-fatal errors, this change stores the most recent cleared error in persistent data and exposes it in the FW_INFO command

### DIFF
--- a/drivers/test-fw/src/bin/error_reporter_tests.rs
+++ b/drivers/test-fw/src/bin/error_reporter_tests.rs
@@ -15,16 +15,51 @@ Abstract:
 #![no_std]
 #![no_main]
 
-use caliptra_drivers::{report_fw_error_fatal, report_fw_error_non_fatal};
+use caliptra_drivers::{
+    clear_fw_error_non_fatal, get_fw_error_non_fatal, report_fw_error_fatal,
+    report_fw_error_non_fatal, PersistentDataAccessor,
+};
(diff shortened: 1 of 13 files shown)


drivers/test-fw/src/bin/error_reporter_tests.rs
```diff
@@ -15,16 +15,51 @@ Abstract:
...
+    assert_eq!(err2, persistent_data.cleared_non_fatal_fw_error);
+
+    // Repeatedly clearing should not overwrite the stored previous error when no error is present
+    clear_fw_error_non_fatal(persistent_data);
+    clear_fw_error_non_fatal(persistent_data);
+    assert_eq!(0, retrieve_fw_error_non_fatal());
+    assert_eq!(err2, persistent_data.cleared_non_fatal_fw_error);
 }
 
 fn retrieve_fw_error_non_fatal() -> u32 {
```
zhalvorsen: We might as well remove this function now that there is a public version

zhalvorsen (dismissed):

drivers/test-fw/src/bin/error_reporter_tests.rs
nquarton: Good point, this is addressed in the update I just pushed

swenson (approved):
LGTM

Issue #1423, opened 2024-03-21
Add basic compatibility check to ROM


FerralCoder:
ROM should add a simple check to verify it is running on an RTL version that it is compatible with. If not, a clear error should be reported to aid in debug.

FerralCoder:
Similar issues filed for ROM, FMC, RT:

| Component | Issue Link |
|-|-|
| ROM | https://github.com/chipsalliance/caliptra-sw/issues/1423 |
| FMC | https://github.com/chipsalliance/caliptra-sw/issues/1424 |
| RT | https://github.com/chipsalliance/caliptra-sw/issues/1425 |

zhalvorsen:
This is done in 2.0 ROM. It is not needed for the 1.x and 2.1 generations

PR #2279, opened 2025-06-30
ROM/RTL compat. check

embediver:
#1423


embediver:
Is checking only when not fake-rom the wanted behaviour?
I just assumed that.

Also, when #1503 is merged, that implementation to get the ROM version can be used.
Leaving this a draft until then.

PR #2280, opened 2025-06-30
[feat] DOT_DERIVE_KEY and HMAC mailbox commands

mhatrevi:
This change implements the DOT_DERIVE_KEY operation in the Caliptra ROM, enabling key derivation functionality to support Device Ownership Transfer (DOT) services. It also enhances the ROM's cryptographic capabilities by adding support for the HMAC mailbox command, allowing the generation of message authentication codes using HMAC operations.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -133,6 +133,9 @@ impl CommandId {
     // Image metadata commands
     pub const GET_IMAGE_INFO: Self = Self(0x494D_4530); // "IME0"
 
+    // Device Ownership Transfer (DOT) commands
+    pub const DERIVE_DOT_KEY: Self = Self(0x444F544B); // "DOTK"
+
     // Cryptographic mailbox commands
     pub const CM_IMPORT: Self = Self(0x434D_494D); // "CMIM"
     pub const CM_DELETE: Self = Self(0x434D_444C); // "CMDL"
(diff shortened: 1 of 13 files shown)


common/src/hmac_cm.rs
```diff
@@ -0,0 +1,129 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    crypto.rs
```
mhatrevi: Fix the file name

api/src/mailbox.rs
```diff
@@ -3696,6 +3699,65 @@ impl Request for CmEcdsaVerifyReq {
     type Resp = MailboxRespHeader;
 }
 
+// DERIVE_DOT_KEY
```
swenson: We should add some documentation for these as well.

api/src/mailbox.rs
```diff
@@ -133,6 +133,9 @@ impl CommandId {
     // Image metadata commands
     pub const GET_IMAGE_INFO: Self = Self(0x494D_4530); // "IME0"
 
+    // Device Ownership Transfer (DOT) commands
+    pub const DERIVE_DOT_KEY: Self = Self(0x444F544B); // "DOTK"
```
swenson: I wonder if we should name these `DERIVE_STABLE_KEY` so that it is clear they can be used in non-DOT flows as well?

swenson (dismissed):
LGTM, a couple of suggestions for this or a follow-up PR

api/src/mailbox.rs
mhatrevi: Added

api/src/mailbox.rs
mhatrevi: Makes sense. Updated.

swenson (approved):
LGTM

PR #2281, opened 2025-06-30
Add a mapping between Firmware Image ID to EXEC/GO Bit

### LINKED ISSUE #2207: Firmware Image IDs are constrained to values from 0 to 127
Currently Caliptra Runtime code rejects values of firmware Image IDs beyond 127. 
This is done to simplify the mapping between a Firmware Image ID and the Exec/Go Bit (which has 128 bits).

This means that in the Authentication Manifest, even though the Image ID is defined as 4 bytes, the values it can take is only 0-127 (with 0 and 1 as reserved, and 2 allocated for MCU FW).

mlvisaya:
Implement mapping between a specific image to its corresponding EXEC/Go
Bit in the SOC Manifest.

Added an 8-bit field within the 'flags' field in the Image Metadata
entry that will map to the exec bit for the image.

This will be used when determining which EXEC/GO bit will be set during
firmware activation when executing the ACTIVATE firmware command.


auth-manifest/README.md
```diff
@@ -35,7 +35,7 @@ The Caliptra SOC manifest has two main components:
...
+| Flags                   | 4            | A set of 32-bit flags specific to the image.<br />**Bit0**: If set, the image hash will not be verified otherwise the metadata image hash will be compared against the calculated hash of the image.<br />**Bit1**: If set, it indicates that the image is a MCU Runtime image, otherwise it indicates a SOC Image.<br />**Bit 15..7** The firmware execution control bit mapped to this image                                                                                                                                                                                                                                                  |
```
mhatrevi: These are 9 bits. We need 7 bits to specify 128 bit locations. Can we keep this to Bits 14..8?

mlvisaya: you're right. yes this should be just 7 bits. i'll change to 14..8

runtime/src/activate_firmware.rs
```diff
@@ -90,6 +89,12 @@ impl ActivateFirmwareCmd {
             if fw_id == ActivateFirmwareReq::MCU_IMAGE_ID && mcu_image_size == 0 {
                 return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
             }
+            let exec_bit = Self::get_exec_bit(drivers, fw_id)
+                .map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+            // Check if exec_bit is valid
+            if exec_bit == 0 || exec_bit == 1 || exec_bit > MAX_EXEC_GO_BIT_INDEX {
```
mhatrevi, line 97: Why is the exec_bit == 1 an invalid condition?

mlvisaya: Added a comment in the code on why exec_bit 0 and 1 are reserved. It's in the register description: https://chipsalliance.github.io/caliptra-rtl/main/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.SS_GENERIC_FW_EXEC_CTRL%5B0%5D

mhatrevi (approved):

PR #2285, opened 2025-07-01
feat(rom): Add ECDSA and MLDSA verify commands

ArthurHeymans:
This adds ECDSA verify to ROM, MLDSA verify to runtime and ROM.
To achieve this the verify commands are placed in caliptra-common.
Since the API for mailbox access between ROM and runtime differ a little introduce a trait to access raw mailbox sram.


ArthurHeymans:
Updated with methods on the Req to decode a portions of struct from a byteslice. This avoids making copies of the data structure and does not require the struct to be backed by the max size of the data which zerocopy requires.

mhatrevi (approved):

PR #2287, opened 2025-07-02
rom: Use slots 0 and 1 for stable identity root secret

swenson:
These previously were used for UDS and FE, but will have been erased by the time we use them for stable identity root secrets.

This is to prepare for potential hardware changes in 2.1 where keys 10-15 might have special treatment for OCP LOCK.


mhatrevi (approved):

swenson:
Thanks!

PR #2288, opened 2025-07-02
 [feat] Device Ownership Transfer: INSTALL_OWNER_PK_HASH Mailbox command implementation

mhatrevi:
This update introduces the INSTALL_OWNER_PK_HASH mailbox command, enabling the addition of the firmware ownership public key hash to the Caliptra ROM. It also updates the firmware verification flow to utilize this hash when the CPTRA_OWNER_PK_HASH register is set to 0.


swenson (commented):
Code LGTM. I think we should also update the ROM docs to document this command how the precedence between the two owner PK hashes works, since it is part of the core boot flow.

swenson (approved):

PR #2289, opened 2025-07-02
[fix] Restrict HMAC key usage

mhatrevi:


rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -179,6 +184,7 @@ impl LocalDevIdLayer {
             None,
             stable_idev,
             HmacMode::Hmac512,
+            KeyUsage::default().set_aes_key_en(),
```
zhalvorsen, line 187: Should this enable `set_hmac_key_en` as well for deriving DOT and LOCK stable identity keys? Similar question for `derive_stable_identity_root_ldev` below

zhalvorsen (approved):

rom/dev/src/flow/cold_reset/ldev_id.rs
swenson: No, it's not necessary.

The goal for these keys is to derive firmware-visible keys through using AES CMAC -> HMAC. So they will always go through AES CMAC first so that the result is visible to firmware.

We won't be deriving any other key vault keys off of them.

swenson (approved):
LGTM

rom/dev/src/flow/cold_reset/ldev_id.rs
mhatrevi: For DOT, we want to limit it strictly to AES key derivation. If there's a need to relax this restriction for LOCK in the future, we can revisit it then.

PR #2290, opened 2025-07-02
Update DPE submodule

zhalvorsen:
Changes are being made for DPE to support ML-DSA. This updates the DPE submodule to make sure they don't drift too far apart.

It is broken into separate commits to hopefully make the changes easier to review.

### DIFF
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -24,9 +24,17 @@ use caliptra_drivers::{
     ExportedCdiEntry, ExportedCdiHandles, Hmac384, Hmac384Data, Hmac384Key, Hmac384Tag, KeyId,
     KeyReadArgs, KeyUsage, KeyVault, KeyWriteArgs, Sha384, Sha384DigestOp, Trng,
 };
-use crypto::{AlgLen, Crypto, CryptoBuf, CryptoError, Digest, EcdsaPub, EcdsaSig, Hasher};
+use crypto::{
+    ecdsa::{
+        curve_384::{Curve384, EcdsaPub384, EcdsaSignature384},
+        EcdsaPubKey, EcdsaSignature,
+    },
+    Crypto, CryptoError, CryptoSuite, Digest, DigestAlgorithm, DigestType, Hasher, PubKey,
+    Signature, SignatureAlgorithm, SignatureType,
+};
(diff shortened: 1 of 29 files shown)


runtime/src/drivers.rs
```diff
@@ -424,10 +459,11 @@ impl Drivers {
                 .map_err(|_| CaliptraError::RUNTIME_ADD_VALID_PAUSER_MEASUREMENT_TO_DPE_FAILED)?,
             flags: DeriveContextFlags::MAKE_DEFAULT
                 | DeriveContextFlags::CHANGE_LOCALITY
-                | DeriveContextFlags::INPUT_ALLOW_CA
+                | DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT
```
zhalvorsen: @clundin25 can you confirm whether we want this attribute here and below?

clundin25: Yes 👍

zhalvorsen: Thanks!

runtime/src/dpe_crypto.rs
```diff
@@ -324,88 +312,56 @@ impl<'a> Crypto for DpeCrypto<'a> {
...
+            y: Ecc384Scalar::from(y),
+        };
+
+        let mut digest_arr = [0u8; SIZE];
+        digest_arr.copy_from_slice(
+            digest
+                .as_slice()
+                .get(..SIZE)
+                .ok_or(CryptoError::CryptoLibError(0))?,
+        );
```
clundin25: In a follow up we should double check if this copy is still necessary

runtime/src/drivers.rs
```diff
@@ -477,7 +511,8 @@ impl Drivers {
         }
 
         // Write DPE to persistent data.
-        pdata.dpe = dpe;
+        drop(env);
+        pdata.dpe_state = state;
```
clundin25: I think a comment that explains the drop could be helpful

clundin25 (dismissed):

runtime/src/dpe_crypto.rs
zhalvorsen: Good call, this was significantly cleaned up as well

clundin25 (approved):

PR #2292, opened 2025-07-03
Add option to trim unwanted padding when updating SHA input data

### LINKED ISSUE #2291: sw-emulator: SHA ACC panics when (block_size-4)<dlen<block_size
The SHA ECC emulator panics when hashing data that is less than one word shorter than the next block-size border.

The implementation already predicts this case in a [todo](https://github.com/chipsalliance/caliptra-sw/blob/main/sw-emulator/lib/crypto/src/sha512.rs#L157).

Since the register only passes words to the `update_bytes` function, a whole block of 128 bytes has arrived after 32 calls to `update_bytes`, which triggers the `update` function which processes the block.
After this `blocks_processed` * `BLOCK_SIZE` is up to 3 byte larger than `dlen`.

As to my understanding of SHA2, `update` should not be called for the last block(s), until the padding bits (0b1 + 0b0 * n + dlen_BE) have been appended.

Probably the fix is to include the `dlen` in every call to `update_bytes` to then check if a couple of bytes have to be removed from the received data.

embediver:
Fixes #2291.

I tested this with the libcaliptra SHA ACC implementation from #1637, let me know if this should have its own unit test.


embediver:
Is there a possibility that the rtl has the same bug?
If the rtl has the bug, should the sw-emulator replicate the bug?

zhalvorsen:
> Is there a possibility that the rtl has the same bug?

It would be good to verify the behavior in the hardware. I can run it on the FPGA and see what happens. All I have to do is hash 127 bytes of data right? What is the expected behavior, incorrect hash or a panic?

> If the rtl has the bug, should the sw-emulator replicate the bug?

I think the sw-emulator should alway model the hardware as closely as possible. So if it is a bug, lets keep it the same until the hardware has a change.

embediver:
> All I have to do is hash 127 bytes of data right? What is the expected behavior, incorrect hash or a panic?

Yes.
SHA384/512 has a block size of 1024bit/128Byte, so when (in stream mode) 125-127 Bytes have been send to the sw-emulator, it starts processing a whole 128 Byte block without applying the correct padding at that stage.
This results in a panic of the sw-emulator as soon as the final execution of hashing is triggered (because more data than possible has already been processed).

If the RTL has the same or a similar bug, I would expect a wrong hash and/or some other UB.

zhalvorsen:
It looks like there isn't a problem with the FPGA in https://github.com/chipsalliance/caliptra-sw/pull/2295

sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -1171,4 +1172,116 @@ mod tests {
         assert_eq!(sha_accl.read(RvSize::Word, OFFSET_STATUS).unwrap(), 0);
         assert_eq!(sha_accl.read(RvSize::Word, OFFSET_EXECUTE).unwrap(), 0);
     }
+
+    #[test]
+    fn test_accelerator_sha512_stream_mode() {
+        // In stream mode, every write is a 32bit word.
+        // When 127 bytes are to be hashed, 128 bytes have thus to be written.
+        // Since the SHA284/512 block size is exactly 128 bytes, 127 bytes are choosen.
```
zhalvorsen: typo:
```suggestion
        // Since the SHA384/512 block size is exactly 128 bytes, 127 bytes are choosen.
```

sw-emulator/lib/crypto/src/sha512.rs
```diff
@@ -322,6 +322,48 @@ mod tests {
         assert_eq!(&hash, &expected);
     }
 
+    #[test]
+    fn test_finalize_sha384() {
+        let mut sha = Sha512::new(Sha512Mode::Sha384);
+        sha.update_bytes(&SHA_512_TEST_BLOCK, None);
+        sha.finalize(128);
+
+        // 3e1650ca726e447ecadf8fa9e5d2b981029d5a76fc2c68a9db341ea13d01c7ff7e827f0f6a62de5a6fa32fcda23eb83b
```
zhalvorsen: IMO we can leave these comments out because the same value and the code

sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -1171,4 +1172,116 @@ mod tests {
...
+            assert_eq!(sha_accl.write(RvSize::Word, OFFSET_DATAIN, dword), Ok(()));
+        }
+        let mut remaining: [u8; 4] = [0; 4];
+        remaining[0..3].copy_from_slice(&DATA[dword_bytes..]);
+        assert_eq!(
+            sha_accl.write(RvSize::Word, OFFSET_DATAIN, u32::from_be_bytes(remaining)),
+            Ok(())
+        );
+
+        // Trigger thea accelerator by writing to the execute register.
```
zhalvorsen: typo:

```suggestion
        // Trigger the accelerator by writing to the execute register.
```

sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -1171,4 +1172,116 @@ mod tests {
...
+                .write(RvSize::Word, OFFSET_EXECUTE, execute.get())
+                .ok(),
+            Some(())
+        );
+
+        // Wait for operation to complete.
+        loop {
+            let status = InMemoryRegister::<u32, Status::Register>::new(
+                sha_accl.read(RvSize::Word, OFFSET_STATUS).unwrap(),
+            );
```
zhalvorsen, line 1268: I'm not super familiar with this code, but can this register be created outside the loop so it is only done once?

embediver: <s>I believe this is in no loop, is it?</s>
Looked at the wrong line...

But either way, checking the status register this way in the test is fine imo.
`InMemoryRegister` is just wrapping a `UnsafeCell<T>`, <u32> in this case.
Due to transparent memory layout this should also be a no-op.

Except the data streaming this is mostly just copied from another test.

zhalvorsen (approved):

zhalvorsen (approved):

PR #2295, opened 2025-07-07
Check Sha384 behavior.

zhalvorsen:


embediver:
Thanks, the test looks correct to me.

Should probably be moved to `sha2_512_384acc_tests.rs` if we want to merge and keep this, right?

PR #2300, opened 2025-07-11
Increase disk size when using VCK190 tools

clundin25:


ci-tools/github-runner/launch.go
```diff
@@ -151,14 +151,18 @@ func Launch(ctx context.Context, client *github.Client, labels []string) error {
 	if err != nil {
 		return err
 	}
-	disks := singleDisk("global/images/family/github-runner", 16)
+
+    var disks []*computepb.AttachedDisk
+    disks = singleDisk("global/images/family/github-runner", 16)
+
```
clundin25, line 157: I will run the formatter on my next PR, should be soon.

PR #2301, opened 2025-07-12
Add CM_DERIVE_STABLE_KEY to runtime mailbox as well

swenson:
This does a few other fixups as well:

* Allow the ROM mailbox processor to understand variable-length requests and responses for CM_HMAC (bug fix)
* Make the CM_DERIVE_STABLE_KEY command have the same `CM` prefix as other cryptographic mailbox commands
* Test that the CMK generated by the ROM is usable by the runtime
* Test that HMACs generated by the ROM are consistent with those generated by the runtime
* Test that if we derive the same stable key in runtime that it generates the same HMACs.
* nit: use the same `Cmk` type in all mailbox API commands
* documentation: Update docs for CM_DERIVE_STABLE_KEY and add missing docs for MLDSA87_SIGNATURE_VERIFY. Make the names consistent with the code for ECDSA384_SIGNATURE_VERIFY and MLDSA87_SIGNATURE_VERIFY.

I ran the the new test on the FPGA as well.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
+++ b/runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
@@ -2,7 +2,7 @@
 
 use std::collections::VecDeque;
 
-use crate::common::{assert_error, run_rt_test, RuntimeTestArgs};
+use crate::common::{assert_error, run_rt_test, start_rt_test_pqc_model, RuntimeTestArgs};
 use aes::Aes256;
 use aes_gcm::{aead::AeadMutInPlace, Key};
 use caliptra_api::mailbox::{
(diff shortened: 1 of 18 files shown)


bluegate010 (dismissed):

clundin25 (approved):

runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
```diff
@@ -2323,4 +2325,125 @@ fn test_ecdsa_sign_verify() {
     }
 }
 
-// #[test]
+#[test]
+fn test_derive_stable_key() {
```
bluegate010, line 2329: Could be done later, but might be useful to also have a test that the stable key remains stable across RT / FMC updates, if such a test doesn't already exist.

bluegate010 (approved):

runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
swenson: In the interest of time, we can do that later I think. But good idea.

mhatrevi (approved):

swenson:
Thanks!

PR #2302, opened 2025-07-13
drivers: Don't restore AES state if no work has been done previously

swenson:
If we do an AES GCM init with no AAD, then the GHASH that is output by the hardware when we do a SAVE state does not behave correctly when we subsequently RESTORE it as part of an update or final operation.

This caused mailbox AES GCM operations with no AAD to output incorrect tags (even though the ciphertext was correct).

Luckily there is an easy workaround: we simply do not save or restore the incorrect GHASH intermediate state if there was no AAD during the encrypt or decrypt init. The GHASH will still be correctly initialized to the correct value.

We add some extra logic to zeroize the GHASH state as best we can, since it is not zeroized when the other registers are.

With this fix, all cryptographic mailbox tests pass on the FPGA again.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -47,7 +47,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 32 * 1024;
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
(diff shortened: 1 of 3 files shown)


bluegate010 (approved):

mhatrevi (approved):

mhatrevi (dismissed):

swenson:
Thanks!

mhatrevi (approved):

PR #2303, opened 2025-07-14
runtime: SPDM runtime cryptographic mailbox fixes

swenson:
* HKDF extract salt should use CMK; if needed, users can import non-secret salts using CM_IMPORT
* Add AES GCM SPDM flags. This tells the GCM init operation to derive a key and IV from the passed in CMK using the SPDM key schedule. The alternative would be to allow arbitrary IVs to be passed in, which is generally not allowed by FIPS.
* Check the GCM key usage counter when encrypting.


runtime/README.md
```diff
@@ -1741,16 +1743,39 @@ Command Code: `0x434D_4155` ("CMAU")
...
 Command Code: `0x434D_4749` ("CMGI")
 
 *Table: `CM_AES_GCM_ENCRYPT_INIT` input arguments*
 | **Name**       | **Type**           | **Description**                  |
 | -------------- | ------------------ | -------------------------------- |
 | chksum         | u32                |                                  |
+| spdm flags     | u8                 | If non-zero, signals that SPDM   |
+|                |                    | mode will be used to derive the  |
+|                |                    | key anv IV (see above).          |
+|                |                    | The value should be equalt to the|
```
bluegate010:
```suggestion
|                |                    | The value should be equal to the |
```

runtime/src/cryptographic_mailbox.rs
```diff
@@ -1068,17 +1069,48 @@ impl Commands {
...
+        } else {
+            // increment and check usage
+            if !matches!(key_usage, CmKeyUsage::Aes) {
+                Err(CaliptraError::RUNTIME_CMB_INVALID_KEY_USAGE_AND_SIZE)?;
+            }
+            let counter = drivers.cryptographic_mailbox.increment_counter(&cmk)?;
+            if counter > GCM_MAX_KEY_USES {
+                Err(CaliptraError::RUNTIME_GCM_KEY_USAGE_LIMIT_REACHED)?;
+            }
+            let key: [u8; 32] = cmk.key_material[..32].try_into().unwrap();
```
bluegate010, line 1109: Should we inspect cmk.length here? Ditto in aes_256_gcm_decrypt_init(). (Should the new logic added in this commit be a helper called from both?)

runtime/README.md
```diff
@@ -1838,22 +1863,44 @@ Starts an AES-256-GCM decryption computation.
...
-| iv              | u8[12]              |                               |
-| aad size        | u32                 |                               |
-| aad             | u8[aad size]        | Additional authenticated data |
+| **Name**        | **Type**            | **Description**                  |
+| --------------- | ------------------- | -------------------------------- |
+| chksum          | u32                 |                                  |
+| spdm flags      | u8                  | If non-zero, signals that SPDM   |
+|                 |                     | mode will be used to derive the  |
+|                 |                     | key anv IV (see above).          |
+|                 |                     | The value should be equalt to the|
```
bluegate010:
```suggestion
|                 |                     | The value should be equal to the |
```

runtime/README.md
```diff
@@ -1741,16 +1743,39 @@ Command Code: `0x434D_4155` ("CMAU")
...
+when encrypting or decrypting in SPDM mode as doing could compromise the
+plaintext of the messages.
+
 Command Code: `0x434D_4749` ("CMGI")
 
 *Table: `CM_AES_GCM_ENCRYPT_INIT` input arguments*
 | **Name**       | **Type**           | **Description**                  |
 | -------------- | ------------------ | -------------------------------- |
 | chksum         | u32                |                                  |
+| spdm flags     | u8                 | If non-zero, signals that SPDM   |
```
bluegate010, line 1800: > If non-zero

Optional, could future-proof a bit by reserving all values except 0 and 1.

bluegate010: Oh, disregard this.

runtime/src/cryptographic_mailbox.rs
swenson: I don't think we have any mechanisms that could create a key of size less than 48, so this should always be safe.

bluegate010 (approved):

mhatrevi (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #2306, opened 2025-07-15
Pre-conditioned keys 

clundin25:
Used throughout the OCP LOCK spec to mix keys together.

### DIFF
--- a/drivers/src/preconditioned_key.rs
+++ b/drivers/src/preconditioned_key.rs
@@ -1,0 +1,70 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    preconditioned_key.rs
+
+Abstract:
+
+    A FIPS-approved method for combining keys together leveraging an HMAC-based key-extraction process.
+
+--*/
+
+use crate::{
+    hmac_kdf, Aes, AesKey, AesOperation, Hmac, HmacData, HmacKey, HmacMode, HmacTag, KeyReadArgs,
+    KeyUsage, KeyWriteArgs, Trng,
+};
+
+use caliptra_error::{CaliptraError, CaliptraResult};
+
+pub fn preconditioned_key_extract(
+    input_key: HmacKey,
+    output_key: HmacTag,
+    kdf_label: &[u8],
+    salt: HmacKey,
+    trng: &mut Trng,
+    hmac: &mut Hmac,
+    aes: &mut Aes,
+) -> CaliptraResult<()> {
+    let HmacTag::Key(output_kv) = output_key else {
+        Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?
+    };
+
+    let mut aes_key = [0; 32];
+    let aes_key = match salt {
+        HmacKey::Array4x16(arr) => {
+            let arr: [u8; 64] = arr.into();
+            aes_key.clone_from_slice(&arr[0..32]);
+            AesKey::Array(&aes_key)
+        }
+        HmacKey::Key(kv) => AesKey::KV(kv),
+        _ => Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?,
+    };
+
+    let mut checksum = [0; 16];
+    aes.aes_256_ecb(aes_key, AesOperation::Encrypt, &[0; 16], &mut checksum)?;
+
+    hmac_kdf(
+        hmac,
+        input_key,
+        kdf_label,
+        Some(&checksum),
+        trng,
+        HmacTag::Key(KeyWriteArgs::new(
+            output_kv.id,
+            KeyUsage::default().set_hmac_data_en(),
+        )),
+        HmacMode::Hmac512,
+    )?;
+
+    hmac.hmac(
+        salt,
+        HmacData::Key(KeyReadArgs::new(output_kv.id)),
+        trng,
+        HmacTag::Key(output_kv),
+        HmacMode::Hmac512,
+    )?;
+    Ok(())
+}
--- a/drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
+++ b/drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
@@ -1,0 +1,86 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    preconditioned_key_extract_gen.rs
+
+Abstract:
+
+    A vector generator for the preconditioned_key_extract function.
+    https://chipsalliance.github.io/Caliptra/ocp-lock/specification/HEAD/#fig:preconditioned-key-extract
+
+--*/
+
+use crate::utils::*;
+
+use openssl::symm::{encrypt, Cipher};
+
+pub struct PreconditionedKeyExtractVector {
+    pub input_key: Vec<u8>,
+    pub input_salt: Vec<u8>,
+    pub input_kdf_label: Vec<u8>,
+    pub checksum: [u8; 16],    // AES-ECB-Encrypt(input_salt[0..32], 0x0000...)
+    pub kdf_output: [u8; 64],  // KDF(input_key, input_kdf_label, checksum)
+    pub output_key: [u8; 64],  // HMAC(input_salt, kdf_output)
+    pub fingerprint: [u8; 16], // AES-ECB-Encrypt(output_key[0..32], 0x0000...)
+}
+
+impl Default for PreconditionedKeyExtractVector {
+    fn default() -> PreconditionedKeyExtractVector {
+        PreconditionedKeyExtractVector {
+            input_key: Vec::new(),
+            input_salt: Vec::new(),
+            input_kdf_label: Vec::new(),
+            checksum: [0; 16],
+            kdf_output: [0; 64],
+            output_key: [0; 64],
+            fingerprint: [0; 16],
+        }
+    }
+}
+
+pub fn gen_vector(
+    input_key_length: usize,
+    salt_length: usize,
+    kdf_length: usize,
+) -> PreconditionedKeyExtractVector {
+    let cipher = Cipher::aes_256_ecb();
+    let mut vec = PreconditionedKeyExtractVector::default();
+
+    vec.input_key.resize(input_key_length, 0);
+    vec.input_salt.resize(salt_length, 0);
+    vec.input_kdf_label.resize(kdf_length, 0);
+
+    rand_bytes(&mut vec.input_key);
+    rand_bytes(&mut vec.input_salt);
+    rand_bytes(&mut vec.input_kdf_label);
+
+    let mut input_key = vec![0; 64];
+    let mut salt = vec![0; 64];
+
+    hmac(&vec.input_key, &[0], &mut input_key, Digest::SHA512);
+    hmac(&vec.input_salt, &[0], &mut salt, Digest::SHA512);
+
+    let aes_key = &salt[0..32];
+
+    let checksum = encrypt(cipher, aes_key, None, &[0u8; 16]).unwrap();
+    vec.checksum.copy_from_slice(&checksum[0..16]);
+
+    kdf(
+        &input_key,
+        &vec.input_kdf_label,
+        Some(&vec.checksum),
+        &mut vec.kdf_output,
+        Digest::SHA512,
+    );
+
+    hmac(&salt, &vec.kdf_output, &mut vec.output_key, Digest::SHA512);
+
+    let aes_key = &vec.output_key[0..32];
+    let fingerprint = encrypt(cipher, aes_key, None, &[0u8; 16]).unwrap();
+    vec.fingerprint.copy_from_slice(&fingerprint[0..16]);
+
+    vec
+}
--- a/drivers/test-fw/src/bin/preconditioned_keys_tests.rs
+++ b/drivers/test-fw/src/bin/preconditioned_keys_tests.rs
@@ -1,0 +1,244 @@
+// Licensed under the Apache-2.0 license
+
+#![no_std]
+#![no_main]
+
+use caliptra_cfi_lib::CfiCounter;
+use caliptra_drivers::preconditioned_key::preconditioned_key_extract;
+use caliptra_drivers::{
+    Aes, AesKey, AesOperation, Array4x12, Array4x16, Hmac, HmacData, HmacKey, HmacMode, HmacTag,
+    KeyId, KeyReadArgs, KeyUsage, KeyWriteArgs, Trng,
+};
+use caliptra_kat::CaliptraError;
+use caliptra_registers::aes::AesReg;
+use caliptra_registers::aes_clp::AesClpReg;
+use caliptra_registers::csrng::CsrngReg;
+use caliptra_registers::entropy_src::EntropySrcReg;
+use caliptra_registers::hmac::HmacReg;
+use caliptra_registers::soc_ifc::SocIfcReg;
+use caliptra_registers::soc_ifc_trng::SocIfcTrngReg;
+use caliptra_test_harness::test_suite;
+
+const OUTPUT_KEY_ID: KeyId = KeyId::KeyId15;
+const SALT_KEY_ID: KeyId = KeyId::KeyId9;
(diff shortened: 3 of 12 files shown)


drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,82 @@
...
+                HmacTag::Key(KeyWriteArgs::new(
+                    output_kv.id,
+                    KeyUsage::default().set_hmac_key_en(),
+                )), // TODO(clundin): This re-uses the output key vault slot as a scratch space for the KDF. Verify this is okay and that it works.
+                HmacMode::Hmac512,
+            )?;
+
+            hmac.hmac(
+                HmacKey::Key(KeyReadArgs::new(output_kv.id)),
+                HmacData::Key(salt_kv),
```
bluegate010: I believe these should be swapped. Ditto in the other switch case.

```suggestion
                HmacData::Key(salt_kv),
                HmacKey::Key(KeyReadArgs::new(output_kv.id)),
```

drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,82 @@
...
+
+            hmac.hmac(
+                HmacKey::Key(KeyReadArgs::new(output_kv.id)),
+                HmacData::Slice(salt_arr),
+                trng,
+                HmacTag::Key(output_kv),
+                HmacMode::Hmac512,
+            )?;
+        }
+        _ => Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?,
```
bluegate010, line 43: Since all of the allowed cases require `output_kv` to be an `HmacTag::Key`, we could just enforce that in the function signature.

drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,82 @@
...
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    preconditioned_key.rs
+
+Abstract:
+
+    A FIPS-approved method for combining keys together leveraging a HMAC-based key-extraction process.
```
bluegate010: Nit.

```suggestion
    A FIPS-approved method for combining keys together leveraging an HMAC-based key-extraction process.
```

drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,82 @@
...
+                HmacTag::Key(KeyWriteArgs::new(
+                    output_kv.id,
+                    KeyUsage::default().set_hmac_key_en(),
+                )), // TODO(clundin): This re-uses the output key vault slot as a scratch space for the KDF. Verify this is okay and that it works.
+                HmacMode::Hmac512,
+            )?;
+
+            hmac.hmac(
+                HmacKey::Key(KeyReadArgs::new(output_kv.id)),
+                HmacData::Slice(salt_arr),
```
bluegate010: It looks like between the two match cases, this is the only line that differs. Could hoist this out and make the switch case much smaller.

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    preconditioned_key_extract(
+        input_key, output_key, kdf_label, salt, &mut trng, &mut hmac, &mut aes,
+    )
+    .unwrap();
+}
+
+test_suite! {
+    test_preconditioned_key_kv_salt,
+    test_preconditioned_key_firmware_salt,
```
bluegate010: I know LOCK doesn't use these combos, but could we also add the following?

- `test_preconditioned_key_kv_salt_fw_key`
- `test_preconditioned_key_fw_salt_fw_key`

drivers/src/preconditioned_key.rs
clundin25: In the other case I think this could be an issue:

`AesKey::Array` == 32 bytes
`HmacKey::Array` == 48 || 64 bytes

In that case, do you want to pad the AesKey so it can be used as an `HmacKey`?

drivers/src/preconditioned_key.rs
clundin25: Ah yes, cleaned this up!

drivers/src/preconditioned_key.rs
clundin25: Hmm I don't think Rust supports that: https://users.rust-lang.org/t/passing-enum-variant-as-argument/94422/3

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
clundin25: Will do!

drivers/src/preconditioned_key.rs
bluegate010: Ah. I see two options:

1. We do what you suggest and pad the AesKey when creating an HmacKey.
2. We change the type of `salt` to HmacKey, and then create an AesKey from that, truncating the bytes if needed. If a caller wants to pass a shorter salt, they can pad it themselves.

Option (2) would be needed if we ever want to pass a firmware-visible key larger than 256 bits as the salt. I don't think the LOCK spec currently has that. The only place that might qualify would be the initial MPK secret. The actual MPK secrets are all 512 bits, but the initial MPK secret, which comes from fw, would need to be 256 bits if `salt` remains an AesKey.

clundin25: After discussion we will add option 2 - I will modify the implementation to match this.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+use openssl::symm::{encrypt, Cipher};
+
+pub struct PreconditionedKeyExtractVector {
+    pub input_key: [u8; 48],
+    pub salt: [u8; 48],
+    pub kdf_label: Vec<u8>,
+    pub checksum: [u8; 16],
+    pub kdf_output: [u8; 64],
+    pub output_key: [u8; 64],
+    pub fingerprint: [u8; 16],
```
bluegate010:
```suggestion
    pub input_key: [u8; 48],
    pub input_salt: [u8; 48],
    pub input_kdf_label: Vec<u8>,
    pub checksum: [u8; 16],     // AES-ECB-Encrypt(input_salt[0..32], 0x0000...)
    pub kdf_output: [u8; 64],   // KDF(input_key, input_kdf_label, checksum)
    pub output_key: [u8; 64],   // HMAC(input_salt, kdf_output)
    pub fingerprint: [u8; 16],  // AES-ECB-Encrypt(output_key[0..32], 0x0000...)
```
This would help readers understand the shape of the vectors.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+
+--*/
+
+use crate::utils::*;
+
+use openssl::symm::{encrypt, Cipher};
+
+pub struct PreconditionedKeyExtractVector {
+    pub input_key: [u8; 48],
+    pub salt: [u8; 48],
```
bluegate010: Can we also make these variable-length vectors? The gen_vector() function can pass the desired lengths for each of the key, salt, and KDF label, and we can fill them each with randomness.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+            kdf_label: Vec::<u8>::default(),
+            checksum: [0; 16],
+            kdf_output: [0; 64],
+            output_key: [0; 64],
+            fingerprint: [0; 16],
+        }
+    }
+}
+
+fn populate_kv_slot(key: &[u8], tag: &mut [u8]) {
```
bluegate010: This function isn't really doing anything with KV. Is this wrapper needed?

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+    hmac(key, &[0], tag, Digest::SHA512);
+}
+
+pub fn gen_vector() -> PreconditionedKeyExtractVector {
+    // Generates Test Vector for OCP LOCK Preconditioned Key Extract.
+    // https://chipsalliance.github.io/Caliptra/ocp-lock/specification/HEAD/#fig:preconditioned-key-extract
```
bluegate010: Nit: this reference probably belongs at the top of the file.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+    let cipher = Cipher::aes_256_ecb();
```
bluegate010, line 49: Nit: this is used in just one place, could be inlined.

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
```diff
@@ -0,0 +1,304 @@
...
+const SALT_SEED: [u8; 48] = [
+    0x33, 0x98, 0x12, 0x45, 0x35, 0x1c, 0xf3, 0x16, 0x08, 0xa4, 0xb0, 0x36, 0x10, 0xcf, 0x17, 0x74,
+    0xff, 0xa1, 0xc9, 0xbc, 0xb9, 0x4a, 0x21, 0x63, 0x40, 0x7e, 0xc9, 0xde, 0x23, 0x47, 0xf1, 0xa5,
+    0x66, 0xaf, 0x9e, 0xc7, 0x3c, 0x9d, 0x0d, 0x55, 0xae, 0xcf, 0x24, 0x12, 0x64, 0xef, 0x29, 0x72,
+];
+const EXPECTED_FINGERPRINT: [u8; 16] = [
+    0x5e, 0x39, 0x4d, 0xfb, 0x29, 0x11, 0x55, 0x6d, 0x76, 0x5b, 0x69, 0xb8, 0x40, 0xb9, 0xbf, 0xcd,
+];
+
+fn test_preconditioned_key_kv_salt_kv_key() {
```
bluegate010: Test organization suggestion:

```
fn test_preconditioned_key_combo(key: &[u8], salt: &[u8], label: &[u8], result: &[u8], key_in_kv: bool, salt_in_kv: bool) {
  ...
}

fn test_preconditioned_key(key: &[u8], salt: &[u8], label: &[u8], result: &[u8]) {
  test_preconditioned_key_combo(key, salt, label, result, true, true);
  test_preconditioned_key_combo(key, salt, label, result, true, false);
  test_preconditioned_key_combo(key, salt, label, result, false, true);
  test_preconditioned_key_combo(key, salt, label, result, false, false);
}

test_preconditioned_key(TEST_VEC_1_KEY, TEST_VEC_1_SALT, TEST_VEC_1_LABEL, TEST_VEC_1_RESULT);
test_preconditioned_key(TEST_VEC_2_KEY, TEST_VEC_2_SALT, TEST_VEC_2_LABEL, TEST_VEC_2_RESULT);
...
```

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
```diff
@@ -0,0 +1,304 @@
...
+        &mut hmac,
+        &mut aes,
+    )
+    .unwrap();
+
+    let fingerprint = get_kv_slot_fingerprint(&mut aes, OUTPUT_KEY_ID);
+    assert_eq!(fingerprint, EXPECTED_FINGERPRINT);
+}
+
+fn test_preconditioned_key_cross_check() {
```
bluegate010: I don't think this test is needed. If each of the combos match the test vectors, then we don't need to also check two of the combos against each other.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
bluegate010: (Where the key-as-KV test would only be done if the key is 512-bits in size, and ditto for salt-in-KV.)

drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,77 @@
...
+            aes_key.clone_from_slice(&arr[0..32]);
+            AesKey::Array(&aes_key)
+        }
+        HmacKey::Array4x16(arr) => {
+            let arr: [u8; 64] = arr.into();
+            aes_key.clone_from_slice(&arr[0..32]);
+            AesKey::Array(&aes_key)
+        }
+        HmacKey::Key(kv) => AesKey::KV(kv),
+        HmacKey::CsrMode() => Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?,
```
bluegate010: I think we can actually just only support fw-provided salts that are provided as an Array4x16. If the salt is shorter, the caller can pad it out. Which the caller would need to do so anyways because fw-provided salts will be 32 bytes in many cases.

```suggestion
        HmacKey::Array4x16(arr) => {
            let arr: [u8; 64] = arr.into();
            aes_key.clone_from_slice(&arr[0..32]);
            AesKey::Array(&aes_key)
        }
        HmacKey::Key(kv) => AesKey::KV(kv),
        _ => Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?,
```

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
clundin25: It's in two places but I'll go ahead and inline it.

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
clundin25: Removed and inlined this function.

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
clundin25: Removed!

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
bluegate010: Oh sorry! I missed that. Disregard.

drivers/src/preconditioned_key.rs
clundin25: Sounds good!

drivers/test-fw/src/bin/preconditioned_keys_tests.rs
clundin25: Much nicer 👍

drivers/test-fw/scripts/vector_gen/src/preconditioned_key_extract_gen.rs
clundin25: @bluegate010 IIUC I think we can still use a 512-bits key size for the key-as-KV test here, since we are just using the key as a seed.

bluegate010 (approved):

drivers/src/preconditioned_key.rs
```diff
@@ -0,0 +1,70 @@
...
+    let HmacTag::Key(output_kv) = output_key else {
+        Err(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_KEY_INVALID_INPUT)?
+    };
+
+    let mut aes_key = [0; 32];
+    let aes_key = match salt {
+        HmacKey::Array4x16(arr) => {
+            let arr: [u8; 64] = arr.into();
+            aes_key.clone_from_slice(&arr[0..32]);
+            AesKey::Array(&aes_key)
```
swenson, line 40: I suppose it doesn't matter as long as we're consistent, but I think the the HMAC keys come out with the words byte-reversed from what AES expects.

swenson (approved):
LGTM

drivers/src/preconditioned_key.rs
clundin25: Good to know - I think this will almost always be a KV vault, but I will add a comment here in a separate PR

PR #2309, opened 2025-07-17
feat(mbox): Make mbox sram size dependent on HW revision

ArthurHeymans:


ArthurHeymans:
Does this need to go to a different branch?

PR #2313, opened 2025-07-18
feat(hw-model): Add hardware revision support

ArthurHeymans:
- Added hw_rev field to InitParams
- Updated mailbox RAM sizing based on HW revision
- Exposed HW revision in root bus args
- Updated test cases to use MailboxRam::default()
- Changed mailbox capacity constants for different HW revisions

### DIFF
--- a/sw-emulator/lib/periph/src/sha512_acc.rs
+++ b/sw-emulator/lib/periph/src/sha512_acc.rs
@@ -657,7 +657,7 @@ mod tests {
 
     fn test_sha_accelerator(data: &[u8], expected: &[u8], start_address: usize, sha_mode: u32) {
         // Write to the mailbox.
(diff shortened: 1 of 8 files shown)


sw-emulator/lib/periph/src/sha512_acc.rs
```diff
@@ -659,7 +657,7 @@ mod tests {
 
     fn test_sha_accelerator(data: &[u8], expected: &[u8], start_address: usize, sha_mode: u32) {
         // Write to the mailbox.
-        let mut mb_ram = MailboxRam::new();
+        let mut mb_ram = MailboxRam::new((2, 0));
```
zhalvorsen, line 660: Should we just use `MailboxRam::default()` for these?

zhalvorsen (approved):

PR #2316, opened 2025-07-22
Remove kernel module build

clundin25:
The kernel modules are currently part of the image, so this step is unnecessary right now.


swenson (approved):
LGTM

PR #2317, opened 2025-07-22
Filter Failing FPGA tests

clundin25:
Skip these tests so we can enable the FPGA tests as a required check.


rom/dev/tests/rom_integration_tests/test_image_validation.rs
```diff
@@ -1948,7 +2026,9 @@ fn test_fmc_rt_load_address_range_overlap() {
     }
 }
 
+//TODO: https://github.com/chipsalliance/caliptra-sw/issues/2070
 #[test]
+#[cfg(not(feature = "fpga_realtime"))]
```
swenson, line 2031: These FMC ones seem to be legitimate failures due to the ROM setting a fatal error instead of returning a mailbox command error.

Issue #1891, opened 2025-01-14
Test case needed to retrieve certs with max measurements submitted


nquarton:
Issue https://github.com/chipsalliance/caliptra-sw/issues/1832 was about the DPE max cert size being too small for the number of measurements supported. A test case needs to be added that submits all the possible measurements in both contexts and retrieves the cert. Parsing the cert for the measurements is a good idea, but technically DPE returning the cert without error should indicate it was not too large when being generated.

nquarton:
This change Anthony is making should enable us to test this easier in the SW emulator: https://github.com/chipsalliance/caliptra-sw/pull/1681

leongross:
@nquarton Just to clarify the terminology here: with "max measurements" you mean that we should measure and validate the maximum possible **size** for a x509 certificate that we can create? What are you referring to when you say 

> the number of measurements supported

nquarton:
Stash measurement (and derive context) support 16 measurements for each PL0 and PL1. Those all add to the TCB info in the DPE leaf cert and increase its size. The suggestion here was to add a test that stores all 32 possible measurements and then retrieves the DPE leaf cert to ensure there is no error when Caliptra generates and provides it.

zhalvorsen:
Fixed in 2.1 by #3417

PR #2318, opened 2025-07-22
Correctly set the max number of handles in DPE.

zhalvorsen:
#1891 

Previously the environment variable was not cascading to the DPE crate. The crate was building prior to RT setting the environment variable.

This sets the environment variable in all cargo invocations so DPE detects it. This also adds a line in the GetProfile test to make sure it is set correctly.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
+++ b/runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
@@ -57,6 +57,7 @@ fn test_invoke_dpe_get_profile_cmd() {
     assert_eq!(profile.vendor_id, VENDOR_ID);
     assert_eq!(profile.vendor_sku, VENDOR_SKU);
     assert_eq!(profile.flags, DPE_SUPPORT.bits());
+    assert_eq!(profile.max_tci_nodes, 32);
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
```diff
@@ -57,6 +57,7 @@ fn test_invoke_dpe_get_profile_cmd() {
     assert_eq!(profile.vendor_id, VENDOR_ID);
     assert_eq!(profile.vendor_sku, VENDOR_SKU);
     assert_eq!(profile.flags, DPE_SUPPORT.bits());
+    assert_eq!(profile.max_tci_nodes, 32);
```
jhand2, line 60: Can we also add a test that tries to exhause the contexts and makes sure DPE actually lets use all the contexts?

jhand2 (commented):
Thanks!

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
zhalvorsen: That's a good idea. I'll add that

zhalvorsen: This uncovered some problems with `is_dpe_context_threshold_exceeded_helper`. I'll fix them in later PRs.

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
```diff
@@ -464,3 +465,60 @@ fn test_export_cdi_destroyed_root_context() {
...
+    for _ in 0..14 {
+        let _ = execute_dpe_cmd(
+            &mut model,
+            &mut Command::DeriveContext(&BASE_CMD),
+            DpeResult::Success,
+        );
+    }
+
+    // Trigger failure by trying to derive one more context to PL1
+    let _ = execute_dpe_cmd(
```
nquarton, line 519: It may make sense to also trigger this for PL0. But maybe there is some reason that would mess the test up.

nquarton (approved):

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
zhalvorsen: Yeah, that is one of the bugs this uncovered. That function isn't smart enough to know "all of PL0 slots are full, but I can still call DeriveContext because it will be for the other locality"

zhalvorsen: Created #2340 to track this

PR #2319, opened 2025-07-22
emulator: MCU can write to the SoC UDS address registers

swenson:
These registers are shared with MCU on its bus and it needs to be able to write that base address.


swenson:
Thanks!

PR #2321, opened 2025-07-23
feat(field entropy): Move extend uds programming for field entropy

ArthurHeymans:
Move the code to program the UDS seed to a common place and extend it to be able to program the field entropy.


common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl UdsFeProgrammingFlow {
+    /// Returns true if this is UDS programming mode
+    pub fn is_uds(self) -> bool {
+        matches!(self, UdsFeProgrammingFlow::Uds)
+    }
+
+    /// Returns the seed length in 32-bit words for this mode
+    pub fn seed_length(self) -> usize {
```
zhalvorsen: nit: Can we name this function something to imply it is the number of words instead of bytes. This might be hard to read from the call site? Maybe something like `seed_num_words`.

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    /// Programs either UDS (64 bytes) or FE (32 bytes) based on the enum variant
+    ///
+    /// # Arguments
+    /// * `soc_ifc` - SoC interface for accessing fuse controller and related registers
+    /// * `trng` - TRNG for generating random seeds
+    /// * `dma` - DMA engine for OTP control
+    pub fn program_uds_fe(
```
zhalvorsen: nit: I think we could just name this `program` to make it scalable in case we decide to add more fuse programming to this enum.

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+            soc_ifc.set_uds_programming_flow_state(true);
+        } else {
+            // [CAP2][TODO]: What needs to be done here?
+        }
+
+        let result = (|| {
+            // Generate a 512-bit random value.
+            let seed: [u32; 16] = trng.generate16()?.into();
+
+            let uds_fuse_row_granularity_64 = soc_ifc.uds_fuse_row_granularity_64();
```
zhalvorsen, line 136: Sorry, I'm still getting familiar with 2.x stuff. Does this granularity apply to both both UDS and FE?

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+            let fuse_controller_base_addr = soc_ifc.fuse_controller_base_addr();
+            let otp_ctrl = DmaOtpCtrl::new(AxiAddr::from(fuse_controller_base_addr), dma);
+            let uds_seed_dest_address = self.get_dest_address(soc_ifc);
```
zhalvorsen: nit: This could now be named `seed_dest_address` because it applies to both UDS and FE

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+                        let dest = uds_seed_dest_address + (index * seed.len() * 4) as u32;
+
+                        // Poll the STATUS register until the DAI state returns to idle
+                        while !regs.status().read().dai_idle() {}
+
+                        let wdata_0 = seed[0];
+                        cprintln!(
+                            "[{}] Writing the seed to the DIRECT_ACCESS_WDATA_0 register, wdata_0: {:#x}",
+                            self.prefix(),
+                            wdata_0
```
swenson: I'm not sure it is wise to print these values. @mhatrevi what do you think?

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+
+            let uds_fuse_row_granularity_64 = soc_ifc.uds_fuse_row_granularity_64();
+            let fuse_controller_base_addr = soc_ifc.fuse_controller_base_addr();
+            let otp_ctrl = DmaOtpCtrl::new(AxiAddr::from(fuse_controller_base_addr), dma);
+            let uds_seed_dest_address = self.get_dest_address(soc_ifc);
+
+            let _ = otp_ctrl.with_regs_mut(|regs| {
+                seed[..self.seed_length()].chunks(if uds_fuse_row_granularity_64 { 2 } else {1}).enumerate().filter(|(i, _)| {
+                    match self {
+                        Self::Fe { bitmask } => (*bitmask >> *i) == 1,
```
swenson: I think this needs to be `&`'d?

```suggestion
                        Self::Fe { bitmask } => (*bitmask >> *i) & 1 == 1,
```

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,187 @@
...
+
+            Ok(())
+        })();
+
+        // Set the programming result.
+        cprintln!("[{}] Setting the programming result", self.prefix());
+        if self.is_uds() {
+            soc_ifc.set_uds_programming_flow_status(result.is_ok());
+        } else {
+            // [CAP2][TODO]: What needs to be done here?
```
swenson: I don't think we need to set any of these programming flow states, since that only affect the debug service register.

swenson (commented):
Adding tests for these two commands would also be nice.

common/src/uds_fe_programming.rs
swenson: Generally they are both 64-bit granularity. We don't currently have a way to differentiate, but we could potentially use another generic input wire for this (and default to 64, like we do for UDS). cc @mhatrevi

ArthurHeymans:
> Adding tests for these two commands would also be nice.

Done. The UDS programming already had tests.

swenson (approved):
LGTM. I'll see if I can get this tested today on the subsystem FPGA.

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,165 @@
...
+                seed[..self.seed_length_words()].chunks(if uds_fuse_row_granularity_64 { 2 } else {1}).enumerate().filter(|(i, _)| {
+                    match self {
+                        Self::Fe { bitmask } => (*bitmask >> *i) & 1 == 1,
+                        Self::Uds => true,
+                    }
+                }).
+                    for_each(
+                    |(index, seed)| {
+
+                        let dest = seed_dest_address + (index * seed.len() * 4) as u32;
```
swenson: When testing this, I realized these field entropy fields are not adjacent: there are four separate partitions that need to be written and finalized separately: https://github.com/chipsalliance/caliptra-ss/blob/main/src/fuse_ctrl/doc/otp_ctrl_mmap.md

swenson (changes requested):
See my comment about the separate partitions

common/src/uds_fe_programming.rs
```diff
@@ -83,14 +83,17 @@ impl UdsFeProgrammingFlow {
             let _ = otp_ctrl.with_regs_mut(|regs| {
                 seed[..self.seed_length_words()].chunks(if uds_fuse_row_granularity_64 { 2 } else {1}).enumerate().filter(|(i, _)| {
                     match self {
-                        Self::Fe { bitmask } => (*bitmask >> *i) & 1 == 1,
+                        Self::Fe { partition } => *partition as usize == *i,
```
mhatrevi: Need a validity check for partition values.

ArthurHeymans:
> See my comment about the separate partitions

I accounted for it.

common/src/uds_fe_programming.rs
```diff
@@ -0,0 +1,217 @@
...
+            let fuse_controller_base_addr = soc_ifc.fuse_controller_base_addr();
+            let otp_ctrl = DmaOtpCtrl::new(AxiAddr::from(fuse_controller_base_addr), dma);
+            let seed_dest_address = self.get_dest_address(soc_ifc);
+
+            let _ = otp_ctrl.with_regs_mut(|regs| {
+                let seed = &seed[..self.seed_length_words()]; // Get random bytes of desired size
+                let chunk_size = if uds_fuse_row_granularity_64 { 2 } else { 1 };
+                let chunked_seed = seed.chunks(chunk_size);
+                for (index, seed_part) in chunked_seed.enumerate() {
+                    let dest = seed_dest_address + (index * seed.len() * size_of::<u32>()) as u32;
```
swenson: this is no longer correct for UDS, as it writes addresses 0x48, 0x88, 0xc8, etc., instead of 0x48, 0x50, etc.

runtime/src/fe_programming.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[inline(never)]
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_bytes: &[u8]) -> CaliptraResult<usize> {
+        let cmd = FeProgReq::ref_from_bytes(cmd_bytes)
+            .map_err(|_| CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
+
+        // [CAP2][TODO] Check if we're in manufacturing mode. Is this to be done for FE?
+        if drivers.soc_ifc.lifecycle() != Lifecycle::Manufacturing {
+            Err(CaliptraError::RUNTIME_FE_PROG_ILLEGAL_LIFECYCLE_STATE)?;
+        }
```
swenson: I believe we do not want this check for field entropy, as it is field-programmable in production mode.

swenson: I checked, this is only supported in production: https://github.com/chipsalliance/caliptra-ss/blob/main/src/fuse_ctrl/rtl/otp_ctrl_part_pkg.sv#L195

swenson (dismissed):
Tested on the FPGA and this now works for both UDS and FE.

I'd really like a follow-up PR that checks for both flows that the addresses emitted are correct, data written looks random, etc. (Perhaps by emitting memory write events, capturing them in the test, and examining them?)

swenson (approved):

PR #2323, opened 2025-07-24
Add bitflags to fe prog

mtimkovich:


ArthurHeymans:
Thanks, I added it to https://github.com/chipsalliance/caliptra-sw/pull/2321 but I placed the bitmask for FE in the enum. That made the function signature a bit cleaner.

mhatrevi:
This PR is no longer applicable.

PR #2324, opened 2025-07-24
Update NixOS config for hostrunners

clundin25:
* Add zhalvorsen ssh key
* Add systemd timer to fetch latest fpga image
* Get rtool from flake


ci-tools/host-runner/configuration.nix
```diff
@@ -1,6 +1,20 @@
...
+{ pkgs, identifier, user, lib, rtool, ... }:
+let 
+    update-fpga-script = pkgs.writeShellScriptBin "update-fpga-image" ''
+      export GCP_ZONE="us-central1"
+      export GITHUB_ORG="chipsalliance"
+      export GCP_PROJECT="caliptra-github-ci"
+
+      set -eux
+      cd /home/${user}
+      ${rtool}/bin/rtool download_artifact 379559 40993215 fpga-image.yml caliptra-fpga-image > caliptra-fpga-image.zip
```
zhalvorsen: Does this download a specific version of the FPGA image? If so, would it be worth while to check a hash of the artifact to make sure it is what expect it to be?

ci-tools/host-runner/hostrunners/kir-1.nix
```diff
@@ -65,6 +65,13 @@
...
         export ZCU_SDWIRE="1-1.1"
 
+        caliptra-fpga-boss --zcu104 $ZCU_FTDI --sdwire $ZCU_SDWIRE "$@"
+     ''))
+      ((pkgs.writeShellScriptBin "vck-0-debug" ''
+        #!${pkgs.bash}/bin/bash
+        export ZCU_FTDI="1-1.3.2"
+        export ZCU_SDWIRE="1-1.3.1"
+
         caliptra-fpga-boss --zcu104 $ZCU_FTDI --sdwire $ZCU_SDWIRE "$@"
```
zhalvorsen, line 75: Do the VCK boards also use the `--zcu104` flag?

zhalvorsen (dismissed):

ci-tools/host-runner/configuration.nix
clundin25: It will download the latest FPGA-Image built on the main branch on GitHub. I expect the hash to change each week.

Let's chat some more offline.

zhalvorsen (approved):

ci-tools/host-runner/hostrunners/kir-1.nix
clundin25: Yes, it could use a rename or alias flag. It is the usb path to the FTDI UART

PR #2329, opened 2025-07-26
emulator: Add mcycle CSR

swenson:
And reset internal timer counter to 0 when it fires to match the hardware.

Since the internal timer counter resets to 0, it cannot be used to track the current clock cycle counter, so we need to implement the `mcycle` and `mcycleh` CSRs to track this.

I validated that this matches the behavior on the FPGA more closely, as with this change, MCU applications run on both the emulator and the FPGA.


ArthurHeymans (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #2331, opened 2025-07-29
emulator: Support 255 PIC interrupts

swenson:
Caliptra core's VeeR supports only 31 but MCU's supports up to 255.

Caliptra core doesn't use the extra interrupts so there is no harm in allocating the extra registers.


mhatrevi (approved):

swenson:
Thanks!

PR #2334, opened 2025-07-30
MCI support for MCU Hitless Update

mlvisaya:
Changes to support MCU Hitless update
(https://github.com/chipsalliance/caliptra-ss/blob/main/docs/CaliptraSSIntegrationSpecification.md#mcu-hitless-fw-update)

1. MCI interrupt When Caliptra clears FW_EXEC_CTRL[2], MCI should be able to generate an interrupt to MCU. Added an outgoing event that will be sent to the MCU to process the MCI interrupt.

2. MCI interrupt registers Added several MCI interrupt registers in the MCI peripheral. Not all are needed for MCU hitless update, but it's added for completeness.

3. External SRAM size increase For non-flash based FW Update, the external test sram is used temporarily for staging area (this can be changed in the future). The firmware is ~512KB in size, so the capacity is bumped up from 4K to 1 MB.

4. Fixes on the ACTIVATE_FIRMWARE mailbox command. a. Since clearing the FW_EXEC_CTRL[2] triggers a reset on MCU, the reset reason should be set first by Caliptra before clearing this field. b. The load address for MCU should always be the MCU SRAM address and need not be taken from the manifest.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -26,7 +26,7 @@ use core::{cell::Cell, mem::size_of, ops::Add};
 use ureg::{Mmio, MmioMut, RealMmioMut};
 
 const BLOCK_SIZE: u32 = 256; // Block size for DMA transfers
-const MCU_SRAM_OFFSET: u64 = 0xc0_0000;
+pub const MCU_SRAM_OFFSET: u64 = 0xc0_0000;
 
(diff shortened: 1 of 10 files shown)


swenson (approved):
LGTM

mlvisaya:
Thanks!

mhatrevi (approved):

PR #2343, opened 2025-08-04
Add more Caliptra Subsystem register files.

zhalvorsen:
These register files will be used by the Caliptra Subsystem test infrastructure to allow reading and writing to specific registers as if it were the external SoC.

The mailbox registers will be added in a future change.


swenson (approved):
LGTM

PR #2346, opened 2025-08-05
runtime: Separate SPDM AES enc/dec

swenson:
Secure SPDM requires a counter be XOR'd into IV. The standard requires that this is XOR'd to the beginning of the IV ("little endian") but libspdm also supports XORing to the end ("big endian"), so we support both.
    
This can potentially introduce collision problems if used, so I think we should separate out the AES GCM SPDM init commands, which also keeps the standard (random IV) GCM commands simpler.


runtime/README.md
```diff
@@ -1833,6 +1811,62 @@ The size of the (encrypted) context is always exactly 128 bytes,
...
+
+Derives the AES key and IV following the [SPDM 1.4](https://www.dmtf.org/dsp/dsp0274) and [Secured Messages using SPDM 1.1](https://www.dmtf.org/dsp/dsp0277) specifications.
+
+Currently only supports AES-256-GCM.
+
+Additional authenticated data (AAD) can only be passed during the `INIT` command, so is limited to the maximum cryptographic mailbox data size (4096 bytes).
+
+The CMK must have been created for **HMAC** usage.
+
+The CMK passed in should be the SPDM major secret CMK created for HMAC usage. The key and IV used for encryption shall follow the [SPDM 1.4](https://www.dmtf.org/dsp/dsp0274) section 12.7 derivation with `key_length` 256 and `iv_length` 96.
```
steven-bellock:
```suggestion
The CMK passed in should be the SPDM major secret CMK created for HMAC usage. The key and IV used for encryption shall follow the [SPDM 1.4](https://www.dmtf.org/dsp/dsp0274) section 12.7 derivation with `key_length` equal to 32 bytes and `iv_length` equal to 12 bytes.
```
`HKDF-Expand` uses octets (bytes).

runtime/README.md
```diff
@@ -1833,6 +1811,62 @@ The size of the (encrypted) context is always exactly 128 bytes,
...
+```
+EncryptionKey = HKDF-Expand(major-secret, bin_str5, key_length);
+IV = HKDF-Expand(major-secret, bin_str6, iv_length);
+bin_str5 = BinConcat(key_length, Version, "key", null);
+bin_str6 = BinConcat(iv_length, Version, "iv", null);
+```
+
+The provided 64-bit message counter will be XOR'd with the IV to produce the message IV.
+
+Note that it is **critical** that the same CMK and counter never be used more than once when encrypting or decrypting in SPDM mode as doing could compromise the plaintext of the messages.
```
steven-bellock:
```suggestion
Note that it is **critical** that the same CMK and counter never be used more than once when encrypting or decrypting in SPDM mode as doing so could compromise the plaintext of the messages.
```

runtime/README.md
```diff
@@ -1909,28 +1973,30 @@ bin_str5 = BinConcat(key_length, Version, "key", null);
...
-|                 |                     | version, e.g., 0x13 = SPDM 1.3.  |
-| CMK             | CMK                 | CMK to use for decryption        |
-| iv              | u8[12]              | Ignored in SPDM mode.            |
-| aad size        | u32                 |                                  |
-| aad             | u8[aad size]        | Additional authenticated data    |
-
-*Table: `CM_AES_GCM_DECRYPT_INIT` output arguments*
+The provided 64-bit message counter will be XOR'd with the IV to produce the message IV.
+
+Note that it is **critical** that the same CMK and counter never be used more than once when encrypting or decrypting in SPDM mode as doing could compromise the plaintext of the messages.
```
steven-bellock:
```suggestion
Note that it is **critical** that the same CMK and counter never be used more than once when encrypting or decrypting in SPDM mode as doing so could compromise the plaintext of the messages.
```

runtime/README.md
```diff
@@ -1833,6 +1811,62 @@ The size of the (encrypted) context is always exactly 128 bytes,
...
+Command Code: `0x434D_5345` ("CMSE")
+
+*Table: `CM_AES_GCM_SPDM_ENCRYPT_INIT` input arguments*
+| **Name**           | **Type**           | **Description**                   |
+| ------------------ | ------------------ | --------------------------------- |
+| chksum             | u32                |                                   |
+| spdm version       | u8                 | The value should be equal to the  |
+|                    |                    | byte representation of the SPDM   |
+|                    |                    | version, e.g., 0x13 = SPDM 1.3    |
+| counter endianness | u8                 | XOR the counter with the IV using |
```
steven-bellock, line 1848: 277 1.2 requires little endianness, so might be good to state that.

swenson: I'll add a comment to that effect. The standard requires "little endian", but it fails (AFAICT) to define how that related to the layout of IV (is the natural byte order of the IV considered big or little endian) and libspdm supports both little and big.

steven-bellock: 277 1.2 defines this explicitly.
image

>libspdm supports both little and big.

libspdm only supports both for 277 1.0/1.1. For 1.2 it is little endian only.

swenson: Thanks. Let me know if what I wrote clarifies this.

steven-bellock: Yes, that should work.

runtime/README.md
```diff
@@ -1909,28 +1973,30 @@ bin_str5 = BinConcat(key_length, Version, "key", null);
...
+|                    |                    | the given endianness.             |
+|                    |                    | 0 = Little endian (standard)      |
+|                    |                    | 1 = Big endian                    |
```
zhalvorsen, line 1992: Looks like the reserved bytes are missing from this table:

```
| reserved           | u8[2]              | Reserved                          |
```

runtime/src/cryptographic_mailbox.rs
```diff
@@ -1126,6 +1128,79 @@ impl Commands {
...
+            &mut drivers.aes,
+            &mut drivers.trng,
+            encrypted_cmk,
+        )?;
+
+        if cmk.length as usize > cmk.key_material.len() {
+            Err(CaliptraError::RUNTIME_CMB_INVALID_KEY_USAGE_AND_SIZE)?;
+        }
+
+        let key_usage: CmKeyUsage = CmKeyUsage::from(cmk.key_usage as u32);
```
zhalvorsen: uber nit: optional: 
```suggestion
        let key_usage = CmKeyUsage::from(cmk.key_usage as u32);
```

zhalvorsen (dismissed):
LGTM with minor comments

runtime/src/cryptographic_mailbox.rs
```diff
@@ -1126,6 +1128,79 @@ impl Commands {
...
+
+        if !matches!(key_usage, CmKeyUsage::Hmac) {
+            Err(CaliptraError::RUNTIME_CMB_INVALID_KEY_USAGE_AND_SIZE)?;
+        }
+        let spdm_version: SpdmVersion = (cmd.spdm_flags as u8).try_into()?;
```
clundin25: I think `as` can overflow here

swenson: That should be fine -- it will just automatically cut off the other bits, so it's equivalent to `& 0xff`.

runtime/README.md
```diff
@@ -1833,6 +1811,62 @@ The size of the (encrypted) context is always exactly 128 bytes,
...
+Command Code: `0x434D_5345` ("CMSE")
+
+*Table: `CM_AES_GCM_SPDM_ENCRYPT_INIT` input arguments*
+| **Name**           | **Type**           | **Description**                   |
+| ------------------ | ------------------ | --------------------------------- |
+| chksum             | u32                |                                   |
+| spdm version       | u8                 | The value should be equal to the  |
+|                    |                    | byte representation of the SPDM   |
+|                    |                    | version, e.g., 0x13 = SPDM 1.3    |
+| counter endianness | u8                 | XOR the counter with the IV using |
```
clundin25, line 1848: I think the code is using these flags in a flipped order.

swenson: I think the code is correct, as we load the flags as a little-endian u32. However, I'll refactor to use a bit field so that this doesn't use hard-coded shifts.

clundin25: Mainly the issue is the doc not reflecting the code, but I don't think it necessarily matters.

swenson: I think the docs and code are consistent? We load the SPDM flags as a u32 in little endian order. The first byte (lowest 8 bits) are the version, the second byte (next 8 bits) have the counter endianness, and the high 2 bytes are reserved.

That matches the order of the bytes listed here.

swenson: I specified them as `u8`s in sequence in the README so that there would not be any ambiguity about their endianness.

clundin25: Ah you are right, sorry!

swenson:
Thanks!

zhalvorsen (approved):

PR #2348, opened 2025-08-05
remove mldsa zeroize

mtimkovich:


ArthurHeymans:
// Wait for hardware ready. Ignore errors
        let _ = Mldsa87::wait(mldsa, || mldsa.status().read().ready()); 
        
Maybe it makes sense to add that before writing the zeroize bit like the rest of the code does? It's a patern that seems to be common in other places in the driver code.

PR #2351, opened 2025-08-05
Add VERIFY_AUTH_MANIFEST mailbox command

mlvisaya:
Add a mailbox command to verify an auth manifest without setting it in
DCCM (in contrast with the SET_AUTH_MANIFEST command).

This will be used during firmware update to verify the downloaded
Auth Manifest without setting it.


api/src/mailbox.rs
```diff
@@ -1626,6 +1633,9 @@ impl Default for SetAuthManifestReq {
     }
 }
 
+// VERIFY_AUTH_MANIFEST
+pub type VerifyAuthManifestReq = SetAuthManifestReq;
```
swenson: Technically the types are slightly different since when they `impl Request` they have to provide the correct `CommandId`.

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -1179,3 +1180,180 @@ fn test_authorize_from_staging_address_incorrect_digest() {
...
+    );
+
+    match result {
+        Ok(Some(resp)) => panic!(
+            "Expected an error response but got a valid response: {:?}",
+            resp
+        ),
+        Ok(None) => panic!("Expected a response but got None"),
+        Err(_) => {
+            // Expected error due to invalid manifest
```
swenson, line 1387: After this fails, could you verify that the old manifest still works?

api/src/mailbox.rs
mlvisaya: got it. i'll separate the struct for VerifyAuthManifestReq

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
mlvisaya: yes, will do.

swenson (approved):
LGTM

PR #2352, opened 2025-08-06
drivers: Use LEArray types instead of byte slices when possible in AES driver to save code space

swenson:
This saves around ~300 bytes of ROM space.

I don't know if it is worth backporting to 2.0 to save space there, but that is always an option.


swenson:
Thanks! I'll merge this for now so we get the code size benefits.

PR #2359, opened 2025-08-07
Print human-readable name of mailbox commands in log.

mtimkovich:


runtime/src/lib.rs
```diff
@@ -204,11 +204,30 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
 
-    cprintln!(
-        "[rt] Received command=0x{:x}, len={}",
-        req_packet.cmd,
-        req_packet.payload().len()
-    );
+    // Create human-readable name of command.
+    let bytes = req_packet.cmd.to_be_bytes();
+    let ascii = {
+        if bytes.len() != 4 || bytes.iter().any(|&c| !c.is_ascii_uppercase()) {
```
swenson: There are one or two commands that have numbers in them. So maybe any ASCII letter or number?

runtime/src/lib.rs
```diff
@@ -204,11 +204,30 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
 
-    cprintln!(
-        "[rt] Received command=0x{:x}, len={}",
-        req_packet.cmd,
-        req_packet.payload().len()
-    );
+    // Create human-readable name of command.
+    let bytes = req_packet.cmd.to_be_bytes();
+    let ascii = {
+        if bytes.len() != 4 || bytes.iter().any(|c| !c.is_ascii()) {
```
swenson:
```suggestion
        if bytes.len() != 4 || bytes.iter().any(|c| !c.is_ascii_alphanumeric()) {
```

swenson (approved):

PR #2360, opened 2025-08-08
feat(MCU staging): Emulate MCU staging area

ArthurHeymans:
I want to see how the FPGA fares against this test.

### DIFF
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -260,6 +260,7 @@ jobs:
               "${COMMON_ARGS[@]}" \
               --test-threads=1 \
               --no-fail-fast \
+              --no-capture \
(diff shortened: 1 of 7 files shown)


swenson (approved):
LGTM

The DMA SHA384 tests will probably fail on FPGA since the current bitstream has subsystem disabled.

ArthurHeymans:
ow this has changes in github ci tests to see what is going on...

clundin25:
@ArthurHeymans Looks like this fails with:

```
error: the argument '--test-threads <N>' cannot be used with '--no-capture'
```

A method with less noise, is to specify the test filter `-E` parameter in cargo nextest, to only include the tests you are interested. With a PR you can then quickly validate against the FPGA.

PR #2361, opened 2025-08-11
Making vendor key configs optional in auth-manifest tool

nquarton:
This allows users to create an auth manifest without populating vendor signatures or keys. This may be useful for vendors that will be signing using a different tool or process.


auth-manifest/gen/src/generator.rs
```diff
@@ -113,33 +124,35 @@ impl<Crypto: ImageGeneratorCrypto> AuthManifestGenerator<Crypto> {
             .crypto
             .sha384_digest(auth_manifest.image_metadata_col.as_bytes())?;
 
-        // Sign the IMC with the vendor manifest public keys if indicated in the flags.
+        // Sign the IMC with the vendor manifest public keys if indicated in the flags and provided
```
mhatrevi: "if provided"

nquarton: Addressed this in latest push

PR #2368, opened 2025-08-12
Add subsystem FPGA testing

zhalvorsen:
This currently only enables one test with more to be enabled later. This also adds feature flags so compatibility can be shared with the core FPGA tests.

.github/workflows/fpga.yml was copied to create .github/workflows/fpga-subsystem.yml and the following changes were made:

* change build feature to `fpga_subsystem`
* limit testing to a single test
* change board type the tests are run on to `vck190-mcu`


clundin25:
Looks like you need the Alignment fix from https://github.com/chipsalliance/caliptra-mcu-sw/pull/305 for the Caliptra ROM

zhalvorsen:
> Looks like you need the Alignment fix from [chipsalliance/caliptra-mcu-sw#305](https://github.com/chipsalliance/caliptra-mcu-sw/pull/305) for the Caliptra ROM

Good catch! Thanks!

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -45,6 +45,9 @@ const FUSE_PQC_OFFSET: usize = FUSE_VENDOR_PKHASH_OFFSET + 48;
 const FUSE_LIFECYCLE_TOKENS_OFFSET: usize = 1184;
 const FUSE_LIFECYCLE_STATE_OFFSET: usize = 4008;
 
+const I3C_ADDR: u32 = 0x2000_4000;
+const MCI_ADDR: u32 = 0x2100_0000;
```
jhand2: nit: Do you know why these differ from the memory map in this file? Should we just update the memory map?

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1315,6 +1318,14 @@ impl HwModel for ModelFpgaSubsystem {
         // self.setup_mailbox_users(boot_params.valid_axi_user.as_slice())
         //     .map_err(ModelError::from)?;
 
+        // TODO: This isn't needed in the mcu-sw-model. It should be done by MCU ROM. There must be
```
jhand2, line 1329: Random guess, I wonder if the bootfsm_brkpoint integration parameter isn't set in the mcu-sw mode but is set in this one.

jhand2 (dismissed):
Overall lgtm, just a couple minor comments

hw-model/src/model_fpga_subsystem.rs
zhalvorsen: These are the default addresses given in the RTL, but the FPGA has them moved to where it makes sense in their memory map. Because we are dealing with the virtual address it doesn't really matter which physical address we use here. I like the idea of having the same values between the emulator and the FPGA. The `caliptra-mcu-sw` emulator model is using these addresses.

The `caliptra-sw` models are doing the same thing. The FPGA and emulator both use `0x3003_0000` as the Caliptra physical address.

hw-model/src/model_fpga_subsystem.rs
jlmahowa-amd: It's set the same for SS and core.
https://github.com/chipsalliance/caliptra-mcu-sw/blob/jlmahowa/fpga/hw/fpga/src/caliptra_wrapper_top.sv#L2063
I assumed that it was in MCU ROM as well.

zhalvorsen: Yeah, ROM should be setting the register [here](https://github.com/chipsalliance/caliptra-mcu-sw/blob/4d150bf2df9b1b321bb975e543158f0e9338f1f5/rom/src/cold_boot.rs#L119). But one possibility we are considering is the pre-built ROM checked in doesn't have that call. Carl said he would create a CI job to build the ROM from `caliptra-mcu-sw` to be confident in what we are running.

clundin25: @zhalvorsen I have the change ready. Will open a PR when this lands.

zhalvorsen: Ack thanks Carl!

drivers/src/soc_ifc.rs
```diff
@@ -38,7 +38,7 @@ pub fn report_boot_status(val: u32) {
         soc_ifc.regs_mut().cptra_boot_status().write(|_| val);
     }
     // [TODO][CAP2]: remove this when debug unlock is fixed
-    if cfg!(feature = "fpga_realtime") {
+    if cfg!(any(feature = "fpga_realtime", feature = "fpga_subsystem")) {
         soc_ifc.regs_mut().cptra_boot_status().write(|_| val);
```
jlmahowa-amd: Not really feedback on this PR. Where can I find more info on this TODO? It was added in #2048 back in March, would this be fixed by the newer RTL?

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1332,6 +1343,19 @@ impl HwModel for ModelFpgaSubsystem {
         );
         self.bmc.push_recovery_image(mcu_firmware);
 
+        let mut xi3c_configured = false;
+        for _ in 0..1_000_000 {
```
jlmahowa-amd, line 1357: Why 1M cycles instead of a while loop? Do we have a way to know that the recovery flow is done?

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1385,12 +1409,13 @@ pub struct FpgaRealtimeBus<'a> {
...
     fn ptr_for_addr(&mut self, addr: RvAddr) -> Option<*mut u32> {
         let addr = addr as usize;
-        unsafe {
-            match addr {
-                0x3002_0000..=0x3003_ffff => Some(self.mmio.add((addr - 0x3000_0000) / 4)),
-                _ => None,
-            }
-        }
+        let offset = match addr {
+            0x2000_4000..0x2000_5000 => 0x2000_4000,
```
jlmahowa-amd: If we are defining I3C_ADDR and MCI_ADDR can we use them here? Using the raw address for Caliptra registers was always just doing what was easy but you do have definitions for I3C_ADDR and MCI_ADDR.

hw-model/src/model_fpga_subsystem.rs
jlmahowa-amd: I agree that these addresses don't really matter, but I do find this confusing. My understanding is that to access i3c registers we call i3c_core which gets a register offset within address space specified by I3C_ADDR, and then we map that into the mapped virtual address space mapping in ptr_for_addr.
Are there uses of these address spaces that _don't_ go through i3c_core() that we need to be consistent with? I don't see any uses outside of this file.
On 1.x FPGA was playing catch up so I think the register file being accessed was not FPGA specific like it is here.
I would appreciate if this were named more clearly, like EMULATOR_I3C_ADDR, but at the end of the day this works.

jhand2: Ok makes sense. Could you just leave a comment for why it's being overridden?

drivers/src/soc_ifc.rs
zhalvorsen: I would assume so, we should ask Christopher when he gets back. I'm not sure what the [CAP2] tag means.

hw-model/src/model_fpga_subsystem.rs
zhalvorsen: Yeah, there were a lot of magic numbers here. I changed them to constants.

hw-model/src/model_fpga_subsystem.rs
zhalvorsen: I copy pasted this from the xtask binary in `caliptra-mcu-sw`. There should be a way to check this. I'll add the capability in a follow on PR. I added a todo. I'll use 1 million steps as a timeout instead of a fixed number

jlmahowa-amd (dismissed):
Thanks!

jhand2 (dismissed):

jhand2 (approved):

Issue #2369, opened 2025-08-13
Fix 2.x FPGA CI


zhalvorsen:
200 of the 692 tests are failing the FPGA tests. These need to be fixed to give confidence the firmware is working with the hardware.

The first step will be to remove the failing tests to create a passing baseline. Then tests need to be added back, probably piecemeal, until we can consistently pass all of the tests.

The following packages have failures:

Package | Total Tests | Failures
-|-|-
caliptra-drivers::drivers_integration_tests |32 |1
caliptra-runtime::runtime_integration_tests |208 |18
caliptra-test::fips_test_suite |115 |99
caliptra-rom::rom_integration_tests |150 |82

clundin25:
Just a few outstanding tests to fix on 2.1. Closing this as they are tracked as a release milestone.

PR #2370, opened 2025-08-13
Remove tests from FPGA CI to create baseline.

zhalvorsen:
#2369

This is the first step in fixing the CI. This will get the tests passing then we can add them back slowly.

### DIFF
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -240,6 +240,7 @@ jobs:
 
           echo VARS=${VARS}
 
+          # TODO(#2369): Remove the extra filters after fixing the broken tests
           COMMON_ARGS=(
               --cargo-metadata="${TEST_BIN}/target/nextest/cargo-metadata.json"
               --binaries-metadata="${TEST_BIN}/target/nextest/binaries-metadata.json"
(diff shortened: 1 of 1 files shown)


.github/workflows/fpga.yml
```diff
@@ -249,7 +249,11 @@ jobs:
                        package(caliptra-builder) |
                        package(caliptra-cfi-derive) |
                        package(caliptra-file-header-fix) |
-                       package(compliance-test))'
+                       package(compliance-test) |
```
clundin25, line 253: I think we should add a comment that includes a link to the tracking issue to remove this filter change

.github/workflows/fpga.yml
```diff
@@ -249,7 +249,11 @@ jobs:
...
+                       package(caliptra-drivers) |
```
clundin25: I think this could be reduced to `test(test_sign_validation_failure)`

.github/workflows/fpga.yml
```diff
@@ -249,7 +249,11 @@ jobs:
...
+                       package(caliptra-runtime) |
```
clundin25: I think it would be good to filter out the 18 failing tests, so we can keep most of the runtime test suite

.github/workflows/fpga.yml
```diff
@@ -249,7 +249,11 @@ jobs:
...
+                       package(caliptra-test) |
```
clundin25: I think this could be reduced to `fips_test_suite`.

E.g. ` binary_id(caliptra-test::fips_test_suite)'`

PR #2374, opened 2025-08-14
OCP LOCK FPGA Validation Tests

clundin25:


drivers/src/aes.rs
```diff
@@ -942,6 +944,72 @@ impl Aes {
...
+        }
+
+        // Key is always in KV, always load before starting OP.
+        self.load_key(key)?;
+
+        // Only KV 23 is allowed to destination KV.
+        let mek_slot = KeyWriteArgs::new(
+            KeyId::KeyId23, // MEK KV.
+            KeyUsage::default().set_dma_data_en(),
+        );
```
bluegate010: Nit: if we're hard-coding the output key, we could also hard-code the input key.

clundin25: Done. Tweaked the API so I can test the hardware but still enforce the correct KV usage by splitting into two functions.

Since I have no way to use a private method in an integration test, it is unfortunately still public.

drivers/src/fuse_bank.rs
```diff
@@ -307,6 +307,18 @@ impl FuseBank<'_> {
...
     }
+
+    /// Get the OCP HEK Seed
+    ///
+    /// # Arguments
+    /// * None
+    ///
+    /// # Returns
+    ///     OCP HEK Seed
+    pub fn ocp_heck_seed(&self) -> Array4x8 {
```
zhalvorsen: Ah heck :)
```suggestion
    pub fn ocp_hek_seed(&self) -> Array4x8 {
```

drivers/src/aes.rs
```diff
@@ -942,6 +944,75 @@ impl Aes {
         Ok(())
     }
 
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    /// AES ECB Decrypt to KV. Used by OCP LOCK for MEK release.
+    pub fn aes_256_ecb_decrypt_kv(&mut self, input: &[u8; 64]) -> CaliptraResult<()> {
+        // Only KV 23 is allowed to be key KV.
```
zhalvorsen:
```suggestion
        // Only KV 16 is allowed to be key KV.
```

zhalvorsen (approved):

swenson (approved):
LGTM

drivers/test-fw/src/bin/ocp_lock_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+// AES Decrypt to KV only work for KV 16 -> KV 23 when OCP in progress is set.
+fn test_aes_kv_release_locked() {
+    let mut test_regs = TestRegisters::default();
+    for (input_kv, output_kv) in OCP_LOCK_KV_RANGE.cartesian_product(OCP_LOCK_KV_RANGE) {
+        let input = KeyId::try_from(input_kv).unwrap();
+        let output = KeyId::try_from(output_kv).unwrap();
+
+        // Already exercised on the happy path.
+        // Can't HMAC into KV 23, so we skip that as input.
+        if (input == KeyId::KeyId16 || input == KeyId::KeyId23) || output == KeyId::KeyId23 {
```
bluegate010, line 114: > || input == KeyId::KeyId23

Do we need this bit?

drivers/test-fw/src/bin/ocp_lock_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+        &mut test_regs.hmac,
+        &mut test_regs.trng,
+        KeyId::KeyId3,
+        KeyUsage::default().set_hmac_key_en(),
+    )
+    .unwrap();
+    hmac_kdf(
+        &mut test_regs.hmac,
+        cdi_slot,
+        b"OCP_LOCK_MDK", // TODO: Use real label from spec.
```
bluegate010, line 155: > TODO: Use real label from spec.

We can also go the other direction and update the spec with the labels used during implementation.

drivers/test-fw/scripts/vector_gen/src/main.rs
```diff
@@ -72,4 +74,11 @@ fn main() {
     println!("  let out_pub_x = {};", hex_arr(&vec_kdf_c0.out_pub_x));
     println!("  let out_pub_y = {};", hex_arr(&vec_kdf_c0.out_pub_y));
     println!();
+
+    let mdk = Mdk::default();
+    println!("mdk");
+    println!("  let input_key = {};", hex_arr(&mdk.input_key));
+    println!("  let mdk = {};", hex_arr(&mdk.mdk));
```
bluegate010, line 81: I don't see this vector in any of the tests.

drivers/src/aes.rs
```diff
@@ -942,6 +944,75 @@ impl Aes {
...
 
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    /// AES ECB Decrypt to KV. Used by OCP LOCK for MEK release.
+    pub fn aes_256_ecb_decrypt_kv(&mut self, input: &[u8; 64]) -> CaliptraResult<()> {
+        // Only KV 16 is allowed to be key KV.
+        let mdk_slot = KeyReadArgs::new(KeyId::KeyId16);
+        // Only KV 23 is allowed to be destination KV.
+        let mek_slot = KeyWriteArgs::new(KeyId::KeyId23, KeyUsage::default().set_dma_data_en());
+
+        self.aes_256_ecb_decrypt_kv_internal(AesKey::KV(mdk_slot), mek_slot, input)
```
bluegate010, line 955: Just a comment, I'm not sure having this as a separate helper is all that necessary.

drivers/test-fw/src/bin/ocp_lock_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+    hmac_kdf(
+        &mut test_regs.hmac,
+        cdi_slot,
+        b"OCP_LOCK_HEK", // TODO: Use real label from spec.
+        Some(hek_seed.as_bytes()),
+        &mut test_regs.trng,
+        hek_slot,
+        HmacMode::Hmac512,
+    )
+    .unwrap();
```
bluegate010, line 192: It would be useful to test that the HEK is expected, based on known UDS and HEK seed values.

drivers/test-fw/src/bin/ocp_lock_tests.rs
clundin25: Yes to avoid overwriting KV23. In a follow up I'll work on removing the order of tests dependency

drivers/test-fw/src/bin/ocp_lock_tests.rs
clundin25: ACK - will do

drivers/test-fw/scripts/vector_gen/src/main.rs
clundin25: It's not, it could probably be omitted from the print statement. I just included it since the encrypted MEK is dependent on it.

drivers/src/aes.rs
clundin25: This way I can test the AES KV rules without duplicating the driver code, although there may be a cleaner way

PR #2376, opened 2025-08-15
[wip] ci: add automatic ipv4 configuration to fpga image

leongross:
It seems as if github can't be reliably reached via IPv6, which is currently the only way for the FPGA to connect. Let's try to use ipv4 instead.


jlmahowa-amd (dismissed):
I'd like IPV4. Manually running these steps on a CI image I downloaded worked to give IPV4 access. @clundin25 does this cause any issues for CI?

ci-tools/fpga-image/startup-script.sh
```diff
@@ -7,6 +7,7 @@
 
 # Stop spewing kernel noise to the UART
 echo 3 > /proc/sys/kernel/printk
+dhclient
```
clundin25: I think this can be removed before merging, correct?

clundin25:
@jlmahowa-amd should be OK. Once @leongross is ready for review, I will test this change.

ci-tools/fpga-image/startup-script.sh
leongross: Yes, ideally, this should be redundant due to the automatic configuration. For some reason, the autoconf seems to be flaky in my setup, such that the device does not get an IP address assigned automatically; running dhcleint fixes that. Until I know what the issue is, I will need it for debugging, but when getting this into upstream, it will definitely be removed.

ci-tools/fpga-image/build.sh
```diff
@@ -39,7 +39,9 @@ if [[ -z "${SKIP_DEBOOTSTRAP}" ]]; then
   chroot out/rootfs bash -c 'echo caliptra-fpga > /etc/hostname'
   chroot out/rootfs bash -c 'echo auto end0 > /etc/network/interfaces'
   chroot out/rootfs bash -c 'echo allow-hotplug end0 >> /etc/network/interfaces'
+  chroot out/rootfs bash -c 'echo iface end0 inet dhcp >> /etc/network/interfaces'
   chroot out/rootfs bash -c 'echo iface end0 inet6 auto >> /etc/network/interfaces'
```
leongross, line 43: `auto` refers to `SLSAC` [1], and I am not sure if our IPv6 networking setup supports address assignment via SLAAC, DHCP seems to work though. 

[1]: https://wiki.debian.org/NetworkConfiguration#Using_DHCP_to_automatically_configure_the_interface

clundin25:
@leongross Tested this change and I don't see any issues. Let's wait for tonight's nightly to check the results.

Can you port this to `main-2.x` as well?

ci-tools/fpga-image/build.sh
leongross: I am not too familiar with the `/etc/network/interfaces` configuration, but maybe it's possible to create a sort of fall-back solution, such that if `auto` fails, DHCP will be tried next.

clundin25:
LGTM, nightly looks OK

PR #2388, opened 2025-08-19
Update RTL for ML-KEM and SHA3

ArthurHeymans:
This updates the RTL on the Adams Bridge fix branch

### DIFF
--- a/libcaliptra/examples/generic/main.mk
+++ b/libcaliptra/examples/generic/main.mk
@@ -11,7 +11,7 @@ LIBCALIPTRA_INC  = $(LIBCALIPTRA_ROOT)/inc
 OBJS := $(patsubst %.c,%.o, $(filter %.c,$(SOURCE)))
 
 # SOC REFERENCE
-RTL_SOC_IFC_INCLUDE_PATH = ../../../hw/latest/rtl/src/soc_ifc/rtl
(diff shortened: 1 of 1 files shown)


hw/latest/registers/src/sha3.rs
```diff
@@ -0,0 +1,1491 @@
...
+//
+#![allow(clippy::doc_lazy_continuation)]
+#![allow(clippy::erasing_op)]
+#![allow(clippy::identity_op)]
+/// A zero-sized type that represents ownership of this
+/// peripheral, used to get access to a Register lock. Most
+/// programs create one of these in unsafe code near the top of
+/// main(), and pass it to the driver responsible for managing
+/// all access to the hardware.
+pub struct Sha3 {
```
nquarton, line 13: I know these are auto-generated, but this ends up as "Sha3" instead of following the convention of "Sha3Reg" as all the others seem to.

swenson: It would be quite annoying to fix this I think (we'd have to basically hardcode a fix for this one struct), since the reason this happens is because the name is slightly inconsistent in the integration `addrmap`: https://github.com/chipsalliance/caliptra-rtl/blob/main/src/integration/rtl/caliptra_reg.rdl#L44

PR #2394, opened 2025-08-20
Make the Subsystem hardware model more accessible.

zhalvorsen:
The subsystem hardware model in `caliptra-mcu-sw` isn't working as consistently. This change is made to reduce code duplication and unnecessary debugging in both repositories. It makes some functionality more public so it can be used by the `caliptra-mcu-sw` hardware model.

This also adds an optional parameter for MCU ROM.

https://github.com/chipsalliance/caliptra-mcu-sw/pull/307 shows what the MCU hardware model is reduced to.


clundin25 (approved):

swenson (approved):
LGTM

PR #2398, opened 2025-08-22
feat(driver): Move enum generation to RDL register parser

ArthurHeymans:
The RDL register parser used to generate this enum for mldsa. Now it just needed a rename for abr.

This was part of #2388 but probably a force push removed it.


swenson (approved):
LGTM

PR #2401, opened 2025-08-22
Add AES DMA basic test

swenson:
Tests DMA with AES GCM for all sizes between 4 and 2080 bytes.

This only runs on the subystem 2.1 FPGA at the moment as we have not updated the emulator for this feature yet.


.github/workflows/fpga-subsystem.yml
```diff
@@ -286,6 +286,7 @@ jobs:
               --target-dir-remap="${TEST_BIN}/target"
               --workspace-remap=.
               -E 'package(caliptra-hw-model) and test(tests::test_execution)'
+              -E 'package(caliptra-drivers) and test(test_dma_aes)'
```
clundin25, line 289: I think line 292 might need `--features fpga_subsystem` for this test to appear in the list. 

I don't think this will impact the actual tests being run, since the test binaries were built earlier

drivers/test-fw/src/bin/dma_aes_tests.rs
```diff
@@ -0,0 +1,279 @@
...
+use caliptra_registers::aes::AesReg;
+use caliptra_registers::aes_clp::AesClpReg;
+use caliptra_registers::csrng::CsrngReg;
+use caliptra_registers::entropy_src::EntropySrcReg;
+use caliptra_registers::soc_ifc::SocIfcReg;
+use caliptra_registers::soc_ifc_trng::SocIfcTrngReg;
+use caliptra_test_harness::test_suite;
+use zerocopy::IntoBytes;
+
+const MCU_SRAM_OFFSET: u64 = 0xc0_0000;
```
clundin25: Out of curiosity, why this address? Is it just a random unused address?

swenson: This is the beginning of the MCU SRAM within the MCI peripheral: https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.mci_top.mcu_sram

(I don't think there's a great way to get this from the currently generated registers.)

zhalvorsen: The offset within MCI is here https://github.com/chipsalliance/caliptra-sw/blob/1cc9e5db651feb496f554e9e51759ccc41d1506e/hw/latest/registers/src/mcu_sram.rs#L16. I think it would be better if it was the full physical address though.

zhalvorsen: As in I would prefer `PTR` in mcu_sram.rs to be the full address

drivers/test-fw/src/bin/dma_aes_tests.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    // Read and decrypt 0s from MCU SRAM to MCU SRAM
+    let mcu_base_addr = soc.mci_base_addr() + MCU_SRAM_OFFSET;
+
+    let zeroize_mcu_sram = |dma: &mut Dma| {
+        zeroize_axi(dma, mcu_base_addr, MCU_SRAM_SIZE);
+    };
+
+    let src = soc.mci_base_addr() + 0xc0_0000;
+    let dst = soc.mci_base_addr() + 0xc0_0000;
```
zhalvorsen: nit:

```suggestion
    let src = soc.mci_base_addr() + MCU_SRAM_OFFSET;
    let dst = soc.mci_base_addr() + MCU_SRAM_OFFSET;
```

zhalvorsen (dismissed):

drivers/test-fw/src/bin/dma_aes_tests.rs
swenson: The full address is SoC-dependent though, since it depends on the strap. But, I'll switch to using this `PTR`. I forgot that was there!

.github/workflows/fpga-subsystem.yml
swenson: It looks like `--features` cannot be specified if `--binaries-metadata` is also:

```
error: the argument '--binaries-metadata <PATH>' cannot be used with:
...
--features
```

zhalvorsen (approved):

drivers/test-fw/src/bin/dma_aes_tests.rs
```diff
@@ -0,0 +1,281 @@
...
+            aes_gcm: false,
+        },
+        0,
+    );
+    dma.dma_read_fifo(&mut output[..max_len.div_ceil(4)]);
+    dma.wait_for_dma_complete();
+
+    let output: &[u8] = output.as_bytes();
+    assert_eq!(&output[..max_len], &EXPECTED_CIPHERTEXT[..max_len]);
+    println!("AES DMA len={} passed", max_len);
```
ArthurHeymans, line 234: nit: do we want these print statements in driver tests?

mhatrevi (approved):

clundin25:
Subsystem FPGA test timed out and got killed. Retrying.

clundin25 (approved):

swenson:
Thanks!

drivers/test-fw/src/bin/dma_aes_tests.rs
swenson: It was helping to check some failures I was seeing for sure. It may not be necessary though moving forward.

PR #2408, opened 2025-08-22
Implement SOC Manifest SVN Verification

mlvisaya:
* Added SVN in the Auth Manifest
* Added APIs to initialize SOC Manifest SVN Fuse
* Add SoC Manifest verification in SET_AUTH_MANIFEST mailbox command
* Added integration tests

### DIFF
--- a/api/types/src/lib.rs
+++ b/api/types/src/lib.rs
@@ -254,7 +254,7 @@ impl Default for Fuses {
             fuse_pqc_key_type: DEFAULT_PQC_KEY_TYPE,
             manuf_dbg_unlock_token: DEFAULT_MANUF_DEBUG_UNLOCK_TOKEN_HASH,
             debug_locked: false,
-            soc_manifest_max_svn: 127,
+            soc_manifest_max_svn: 128,
             soc_manifest_svn: Default::default(),
         }
     }
--- a/api/types/src/lib.rs
+++ b/api/types/src/lib.rs
@@ -232,6 +232,8 @@ pub struct Fuses {
     pub fuse_pqc_key_type: u32,
     pub manuf_dbg_unlock_token: [u32; 16],
     pub debug_locked: bool,
+    pub soc_manifest_svn: [u32; 4],
+    pub soc_manifest_max_svn: u8,
 }
 impl Default for Fuses {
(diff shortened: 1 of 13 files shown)


api/types/src/lib.rs
```diff
@@ -252,6 +254,8 @@ impl Default for Fuses {
             fuse_pqc_key_type: DEFAULT_PQC_KEY_TYPE,
             manuf_dbg_unlock_token: DEFAULT_MANUF_DEBUG_UNLOCK_TOKEN_HASH,
             debug_locked: false,
+            soc_manifest_max_svn: 127,
```
mhatrevi: This should be 128.

auth-manifest/app/src/main.rs
```diff
@@ -82,6 +82,10 @@ pub(crate) fn run_auth_man_cmd(args: &ArgMatches) -> anyhow::Result<()> {
         .get_one::<u32>("version")
         .with_context(|| "version arg not specified")?;
 
+    let svn: &u32 = args
```
mhatrevi, line 85: We need to validate that the SVN cannot be > 128.

sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -980,6 +987,8 @@ impl SocRegistersImpl {
             fuse_mldsa_revocation: Default::default(),
             fuse_soc_stepping_id: ReadWriteRegister::new(0),
             fuse_manuf_dbg_unlock_token: [0; 16],
+            fuse_soc_manifest_svn: [0; 4],
+            fuse_soc_manifest_max_svn: 127,
```
mhatrevi: This should be 128.

api/types/src/lib.rs
mlvisaya: nice catch! i'll update.

auth-manifest/app/src/main.rs
mlvisaya: will do.

sw-emulator/lib/periph/src/soc_reg.rs
mlvisaya: i'll update

PR #2409, opened 2025-08-23
fpga: Reset AXI before and after test

swenson:
This makes the FPGA significantly more stable and should remove the need to rebooting the FPGA between runs occasionally.


zhalvorsen (approved):

swenson:
Thanks!

PR #2411, opened 2025-08-24
Increase emulated MCU SRAM size.

zhalvorsen:
The FPGA has 512 KB of SRAM. It would be good to try to keep it below 384 KB though.


swenson (approved):
LGTM

PR #2412, opened 2025-08-26
drivers: AES DMA additional tests

swenson:
For very large DMA as well as out-of-place.

These were manually confirmed as working on an FPGA with the 2.1 subsystem bitstream (latest). This adds them to CI as well.


swenson:
Thanks!

PR #2422, opened 2025-08-28
Add sha3 periph and crypto to emulator.

mtimkovich:


swenson (dismissed):
LGTM

I think this looks good. We might need to update the offsets shortly as there is an RDL change from the HW team coming in soon.

sw-emulator/lib/periph/src/hash_sha3.rs
```diff
@@ -0,0 +1,793 @@
...
+const SHA3_MSG_FIFO_SIZE: usize = 2048 / 32;
+// TODO: I think this value is wrong.
+const SHA3_MSG_FIFO_MAX_DEPTH: usize = 32;
+
+/// SHA3 Peripheral
+#[derive(Bus)]
+#[poll_fn(poll)]
+#[warm_reset_fn(warm_reset)]
+#[update_reset_fn(update_reset)]
+pub struct HashSha3 {
```
swenson, line 139: This also needs to be added to the `root_bus.rs` so it will be available to the RISC-V core.

nquarton:
Hi, is this at a point where I can try testing the sha3 driver in PR https://github.com/chipsalliance/caliptra-sw/pull/2427 against it? This has already been enabled on the FPGA.

swenson:
Includes #2427 now as a cherry pick.

mhatrevi (dismissed):

nquarton (dismissed):
I don't have the emulator experience to have my review carry any weight for that portion. The driver looked the same overall to me so if all the testing is passing on both FPGA and emulator, I think we are good.

mhatrevi (dismissed):

mhatrevi (approved):

PR #2423, opened 2025-08-28
Add definition of CM_SEAL and CM_LOAD.

bluegate010:
This allows long-lived keys to be escrowed into Caliptra and used in subsequent boots, without needing to expose them to MCU.


swenson (approved):
Design LGTM, but I don't want to merge this until we have code to back it up.

PR #2424, opened 2025-08-28
fpga: Clean up of subsystem I3C controller logic

swenson:
* Add methods to make it easier to send and receive I3C packets.
* Removes unnecessary mutability from Command.
* Makes Controller so it can be used from const references (and won't need `RwLock` or `RefCell` to share).
* Adds proper error types to Controller


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1132,7 +1210,19 @@ impl HwModel for ModelFpgaSubsystem {
             )
         }));
 
-        let i3c_controller = xi3c::Controller::new(i3c_controller_mmio);
+        let xi3c_config = xi3c::Config {
+            device_id: 0,
+            base_address: i3c_controller_mmio,
+            input_clock_hz: 199_999_000,
```
zhalvorsen:
```suggestion
            input_clock_hz: AXI_CLK_HZ,
```

zhalvorsen (approved):

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -423,6 +563,10 @@ impl ModelFpgaSubsystem {
         }
     }
 
+    pub fn i3c_controller(&self) -> XI3CWrapper {
+        self.i3c_controller.clone()
```
clundin25, line 584: Could this be wrapped with a Mutex or similar, instead of cloning the raw pointers?

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -180,6 +183,143 @@ impl Mci {
...
+        self.controller.set_s_clk(AXI_CLK_HZ, I3C_CLK_HZ, I3C_MODE);
+        self.controller.cfg_initialize().unwrap();
+        println!("I3C controller finished initializing");
+    }
+
+    /// Start receiving data (non-blocking).
+    pub fn read_start(&self, len: u16) -> Result<(), XI3cError> {
+        let target_addr = self.get_primary_addr();
+        let mut cmd = xi3c::Command {
+            cmd_type: 1,
```
clundin25: nit: It would be good to (eventually) use an enum for `cmd_type`.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -526,6 +644,18 @@ impl ModelFpgaSubsystem {
             }
         }
 
+        let status = self
+            .i3c_core()
+            .sec_fw_recovery_if()
+            .recovery_status()
+            .read()
+            .dev_rec_status();
+        if status == 3 {
```
clundin25: nit: Similar enum comment here, not sure what `3` means at a glance, but can be refactored later

hw-model/src/model_fpga_subsystem.rs
swenson: Hmm. I'm realizing now that although `&Controller` can be used as a reference, really, we should be using `RwLock` since we don't want two threads to access the registers at the same time. I'll fix this.

hw-model/src/model_fpga_subsystem.rs
swenson: Will do. This was originally copied over from the auto-translated C code so these were lost.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -180,6 +183,156 @@ impl Mci {
...
 }
 
+#[derive(Clone)]
+pub struct XI3CWrapper {
+    pub controller: Arc<Mutex<xi3c::Controller>>,
+    pub i3c_mmio: *mut u32,
+    pub i3c_controller_mmio: *mut u32,
+}
+
+// needed to copy pointers
```
clundin25, line 195: nit: Add safety comment

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -180,6 +183,156 @@ impl Mci {
     }
 }
 
+#[derive(Clone)]
+pub struct XI3CWrapper {
+    pub controller: Arc<Mutex<xi3c::Controller>>,
+    pub i3c_mmio: *mut u32,
+    pub i3c_controller_mmio: *mut u32,
```
clundin25, line 192: Since these fields are pub, and not gated behind a mutable reference, it seems we can't guarantee exclusive access to these points

clundin25 (dismissed):

zhalvorsen (approved):

hw-model/src/model_fpga_subsystem.rs
swenson: Correct. I think I would like to leave them open as `pub` a little while longer while we move the I3C socket over to the FPGA, and then once we know we have all of the methods we'll need to build off, we can lock it back down and remove `pub`.

wmaroneAMD (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #2427, opened 2025-08-29
Sha3 shake driver

nquarton:


drivers/src/sha3.rs
```diff
@@ -0,0 +1,366 @@
+/*++
+
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    shake_sha3.rs
```
zhalvorsen: nit: 
```suggestion
    sha3.rs
```

drivers/src/sha3.rs
```diff
@@ -0,0 +1,366 @@
...
+        let dst_u32 = reg.msg_fifo().at(0).ptr;
+        let dst_u8 = dst_u32 as *mut u8;
+
+        // Write 4-byte chunks as u32 dwords
+        for chunk in src_dwords.chunks_exact(4) {
+            let src_dword = u32::from_ne_bytes(chunk.try_into().unwrap());
+            unsafe {
+                dst_u32.write_volatile(src_dword);
+            }
+        }
```
zhalvorsen: I'm not sure of a better way to handle the u8 writes, but we might as well avoid the unsafe block in the u32 writes:
```suggestion
        let dst_u32 = reg.msg_fifo().at(0);
        let dst_u8 = dst_u32.ptr as *mut u8;

        // Write 4-byte chunks as u32 dwords
        for chunk in src_dwords.chunks_exact(4) {
            let src_dword = u32::from_ne_bytes(chunk.try_into().unwrap());
            dst_u32.write(|_| src_dword);
        }
```

hw/latest/registers/src/sha3.rs
```diff
@@ -1,6 +1,9 @@
 // Licensed under the Apache-2.0 license.
 //
-// generated by caliptra_registers_generator with caliptra-rtl repo at 525f8291c8766b6281eb3f5c5b5720e38cab3ce5
+// generated by caliptra_registers_generator with caliptra-rtl repo at 6f63855a49300fe0248c9ab0c6d190c25f95f68e
+//
+//
+// Warning: rtl-caliptra was dirty: M submodules/adams-bridge
```
zhalvorsen: Is this warning something we should be concerned about?

Or was it required as some of the changes to get it working? Except I wouldn't expect adam's bridge to be related to the SHA3 engine.

drivers/src/sha3.rs
nquarton: Oh good catch. Fixed.

hw/latest/registers/src/sha3.rs
nquarton: I must have messed up a bit generating this. Looks like Luke updated the RTL and these files in a recent commit so this is out of this PR now.

drivers/src/sha3.rs
nquarton: This is a good point. I think we do want to use write_volatile over just a normal write for register writes (though I'm not sure if this core actually even does out-of-order execution to make this matter).

However. there isn't any reason I can't use the register directly for the 32 bit writes. It boils down to the same unsafe write_volatile in the back-end. But better to let ureg handle it than re-implement when not needed.

This is updated in the latest push.

nquarton:
> LGTM; mind rebasing?

Sure, CI is still going to be blocked until the emulator support is in. But I can still update this now.

nquarton:
Absorbed into https://github.com/chipsalliance/caliptra-sw/pull/2422

PR #2428, opened 2025-08-30
[fix] Don't wait for zeroization to complete for MLDSA in error handler

mhatrevi:
This is a temporary fix to unblock numerous test timeout failures due to MLDSA zeroization hang in fatal error handler. Once the root cause is investigated, this change should be reverted.


ArthurHeymans (approved):

swenson (approved):
LGTM

PR #2430, opened 2025-09-03
Feat(sw-emulator): Align with RTL to disallow unaligned access to mbo…

ArthurHeymans:
…x sram

The RTL does not allow unaligned access to the mailbox sram. This makes the LMS verify runtime command fail like the RTL.

TODO make sure unaligned stores don't get generated on the mailbox sram.


swenson:
I think this needs to be re-targeted at `main-2.x`?

ArthurHeymans:
> I think this needs to be re-targeted at `main-2.x`?

Done

PR #2434, opened 2025-09-03
Fix HW Rev ID

clundin25:


bluegate010 (approved):

zhalvorsen (approved):
I feel like this value should be something like `0x12`, but that's probably something we need to discuss with the hardware folks.

clundin25:
> I feel like this value should be something like `0x12`, but that's probably something we need to discuss with the hardware folks.

@zhalvorsen it's currently `0x12` in RTL but there was a window where RTL was using `0x2`. Unfortunately thats what is in the current CI image. Next bitstream update should have the correct revision.

zhalvorsen:
Good to know, thanks @clundin25!

PR #2435, opened 2025-09-03
drivers: DMA accelerator has to release unlock itself

swenson:
The SHA accelerator hardware does not support calculating the digest of an empty stream, so we workaround that by hardcoding the SHA384("") (still being confirmed by hardware team).

Also fixed a bug where we were feeding the SHA accelerator too much data if the data is not a multiple of 256 bytes.

### DIFF
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -222,7 +222,7 @@ pub mod driver_tests {
 
     pub const KEYVAULT_FPGA: FwId = FwId {
         bin_name: "keyvault",
-        features: &["fpga_realtime"],
+        features: &["emu", "fpga_realtime"],
         ..BASE_FWID
     };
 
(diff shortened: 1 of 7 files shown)


builder/src/firmware.rs
```diff
@@ -222,7 +222,7 @@ pub mod driver_tests {
 
     pub const KEYVAULT_FPGA: FwId = FwId {
         bin_name: "keyvault",
-        features: &["fpga_realtime"],
+        features: &["emu", "fpga_realtime"],
```
clundin25, line 225: Should this change be checked in? I think the above `KEYVAULT` binary has the emu feature.

clundin25 (dismissed):

builder/src/firmware.rs
swenson: I'm not sure why this works without it, tbh. If you don't include the `emu` feature, then the test harness won't generate the correct code to exit the test, so it will hang forever.

It is required for the DMA test to work, so I thought I would make the keyvault test match.

clundin25 (approved):

mhatrevi (approved):

zhalvorsen (approved):

swenson:
Thanks!

PR #2442, opened 2025-09-04
[fix] Update fatal error handler cleanup sequence

mhatrevi:
This change moves the zeroization of the peripherals before setting the fatal/no-fatal error registers.


swenson (approved):
LGTM

mhatrevi:
Abandoning this change as it causes a smoke test failure.

PR #2452, opened 2025-09-05
[jtag,openocd] add initial OpenOCD JTAG TAP infra

timothytrippel:
This updates the Subsystem HW model to add the initial OpenOCD/JTAG TAP infrastructure that is required to actuate variouse JTAG TAP interfaces in E2E tests. This simply adds functions to spawn and terminate an OpenOCD server subprocess during a test. In a follow up, additional functionality will be added to implement various TAP operations such as read/write memory/registers.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -6,6 +6,7 @@
 use crate::api_types::{DeviceLifecycle, Fuses};
 use crate::bmc::Bmc;
 use crate::fpga_regs::{Control, FifoData, FifoRegs, FifoStatus, ItrngFifoStatus, WrapperRegs};
+use crate::openocd::openocd_jtag_tap::{JtagParams, JtagTap, OpenOcdJtagTap};
 use crate::otp_provision::{
     lc_generate_memory, otp_generate_lifecycle_tokens_mem, LifecycleControllerState,
     LifecycleRawTokens, LifecycleToken,
@@ -13,6 +14,7 @@ use crate::otp_provision::{
 use crate::output::ExitStatus;
 use crate::xi3c::XI3cError;
 use crate::{xi3c, BootParams, Error, HwModel, InitParams, ModelError, Output, TrngMode};
+use anyhow::Result;
 use caliptra_api::SocManager;
 use caliptra_emu_bus::{Bus, BusError, BusMmio, Device, Event, EventData, RecoveryCommandCode};
 use caliptra_emu_types::{RvAddr, RvData, RvSize};
@@ -350,6 +352,7 @@ pub struct ModelFpgaSubsystem {
     pub i3c_mmio: *mut u32,
     pub i3c_controller_mmio: *mut u32,
     pub i3c_controller: XI3CWrapper,
+    pub jtag_tap: Option<Box<OpenOcdJtagTap>>,
     pub otp_mmio: *mut u32,
     pub lc_mmio: *mut u32,
 
@@ -1117,6 +1120,11 @@ impl ModelFpgaSubsystem {
     fn cycle_count(&mut self) -> u64 {
         self.wrapper.regs().cycle_count.get() as u64
     }
+
+    pub fn jtag_tap_connect(&mut self, params: &JtagParams, tap: JtagTap) -> Result<()> {
+        self.jtag_tap = Some(OpenOcdJtagTap::new(params, tap)?);
+        Ok(())
+    }
 }
 
 impl HwModel for ModelFpgaSubsystem {
@@ -1275,6 +1283,7 @@ impl HwModel for ModelFpgaSubsystem {
                 i3c_mmio,
                 i3c_controller_mmio,
             },
+            jtag_tap: None,
             otp_mmio,
             lc_mmio,
 
--- a/hw-model/src/openocd/openocd_jtag_tap.rs
+++ b/hw-model/src/openocd/openocd_jtag_tap.rs
@@ -1,0 +1,100 @@
+// Licensed under the Apache-2.0 license
+//
+// Derived from OpenTitan's opentitanlib with original copyright:
+//
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::PathBuf;
+
+use anyhow::{bail, Result};
+use serde::{Deserialize, Serialize};
+
+use crate::openocd::openocd_server::{OpenOcdError, OpenOcdServer};
+
+/// Available JTAG TAPs in Calitpra Subsystem.
+#[derive(Clone, Copy, Debug, Deserialize, Serialize, PartialEq)]
+pub enum JtagTap {
+    /// RISC-V Veer core's TAP for Caliptra Core.
+    CaliptraCoreTap,
+    /// RISC-V Veer core's TAP for Caliptra MCU.
+    CaliptraMcuTap,
+    /// Lifecycle Controller's TAP.
+    LccTap,
+    /// No TAP selected.
+    NoTap,
+}
+
+/// JTAG parameters to pass to OpenOCD server on startup.
+#[derive(Debug, Clone)]
+pub struct JtagParams {
+    /// OpenOCD binary path.
+    pub openocd: PathBuf,
+
+    /// JTAG adapter speed in kHz.
+    pub adapter_speed_khz: u64,
+
+    /// Whether or not to log OpenOCD server messages to stdio.
+    pub log_stdio: bool,
+}
+
+/// A JTAG TAP accessible through an OpenOCD server.
+pub struct OpenOcdJtagTap {
+    /// OpenOCD server instance.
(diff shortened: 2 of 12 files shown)


hw-model/src/openocd/openocd_jtag_tap.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub openocd: PathBuf,
+
+    /// JTAG adapter speed in kHz.
+    pub adapter_speed_khz: u64,
+
+    /// Whether or not to log OpenOCD server messages to stdio.
+    pub log_stdio: bool,
+}
+
+/// An JTAG TAP accessable through an OpenOCD server.
```
steven-bellock:
```suggestion
/// A JTAG TAP accessible through an OpenOCD server.
```

hw-model/src/openocd/openocd_jtag_tap.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct OpenOcdJtagTap {
+    /// OpenOCD server instance.
+    openocd: OpenOcdServer,
+    /// JTAG TAP OpenOCD server is connected to.
+    jtag_tap: JtagTap,
+}
+
+impl OpenOcdJtagTap {
+    /// Start OpenOCD server and connect to a given JTAG TAP yet.
```
steven-bellock:
```suggestion
    /// Start OpenOCD server and connect to a given JTAG TAP.
```

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -350,6 +355,7 @@ pub struct ModelFpgaSubsystem {
     pub i3c_mmio: *mut u32,
     pub i3c_controller_mmio: *mut u32,
     pub i3c_controller: XI3CWrapper,
+    pub jtag_tap: Option<Box<OpenOcdJtagTap>>,
```
clundin25, line 355: Note sure how much this matters, but may want to use `Arc` instead of `Box` since wrapper is using an `Arc`.

hw-model/src/openocd/openocd_server.rs
```diff
@@ -0,0 +1,228 @@
...
+
+use std::io::{BufRead, BufReader, Write};
+use std::net::TcpStream;
+use std::os::unix::process::CommandExt;
+use std::path::Path;
+use std::process::{Child, Command, Stdio};
+use std::time::{Duration, Instant};
+
+use anyhow::{bail, Context, Result};
+use once_cell::sync::Lazy;
```
clundin25, line 17: I think the standard library could be used here: https://doc.rust-lang.org/std/cell/struct.OnceCell.html

One less dependency :)

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -3,30 +3,35 @@
 #![allow(clippy::mut_from_ref)]
 #![allow(dead_code)]
 
+use std::marker::PhantomData;
+use std::process::exit;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{mpsc, Arc, Mutex};
+use std::thread;
+use std::time::{Duration, Instant};
+
```
zhalvorsen: nit: I'd recommend keeping all of the use statements together and just let rustfmt take care of them. Is there a specific reason you broke them up?

hw-model/src/openocd/openocd_server.rs
```diff
@@ -0,0 +1,228 @@
...
+use thiserror::Error;
+
+use crate::openocd::printer;
+
+/// Errors related to the OpenOCD server.
+#[derive(Error, Debug, Deserialize, Serialize)]
+pub enum OpenOcdError {
+    #[error("OpenOCD initialization failed: {0}")]
+    InitializeFailure(String),
+    #[error("OpenOCD server exists prematurely")]
```
zhalvorsen:
```suggestion
    #[error("OpenOCD server exited prematurely")]
```

hw-model/src/openocd/openocd_server.rs
```diff
@@ -0,0 +1,228 @@
...
+    server_process: Child,
+    /// Receiving side of the stream to the telnet interface of OpenOCD.
+    reader: BufReader<TcpStream>,
+    /// Sending side of the stream to the telnet interface of OpenOCD.
+    writer: TcpStream,
+}
+
+impl Drop for OpenOcdServer {
+    fn drop(&mut self) {
+        let _ = self.server_process.kill();
```
zhalvorsen: Should this use `self.shutdown()`, or does it not need to end gracefully?

hw-model/src/openocd/openocd_server.rs
```diff
@@ -0,0 +1,228 @@
...
+        {
+            if let Ok(version) = openocd.execute("version") {
+                println!("OpenOCD version: {version}");
+            } else {
+                panic!("Failed to execute an openocd command.");
+            }
+            assert!(openocd.shutdown().is_ok());
+        } else {
+            panic!("Failed to spawn an openocd subprocess.");
+        }
```
zhalvorsen: nit: this might be more readable using let else syntax:
```suggestion
        let Ok(mut openocd) =
            OpenOcdServer::spawn(Path::new("openocd"), /*log_stdio=*/ true) else {
            panic!("Failed to spawn an openocd subprocess.");
        }
        let Ok(version) = openocd.execute("version") else {
            panic!("Failed to execute an openocd command.");
        }
        println!("OpenOCD version: {version}");
        assert!(openocd.shutdown().is_ok());
```

hw-model/src/openocd/openocd_server.rs
timothytrippel: oncecell is already brought into the workspace and used by another package: https://github.com/chipsalliance/caliptra-sw/blob/74ebff299062deac9d3bf152d5cb2997f6480e88/Cargo.toml#L161

I can explore cleaning up this and the other package up in a follow up.

hw-model/src/model_fpga_subsystem.rs
timothytrippel: Using Arc implies we will have multiple threads accessing the openocd server process. This is exactly what we want to avoid since a series of JTAG transactions executed across several threads could impact on another and cause unintended test failures. I think we want to enforce single ownership of this resource.

hw-model/src/model_fpga_subsystem.rs
timothytrippel: There is a grouping feature of rustfmt: https://github.com/rust-lang/rustfmt/issues/2979 that follows a specific order. That is what is implemented here and is common across some rust codebases. I can remove this if this is not desired in this project.

jlmahowa-amd (commented):
The cfg files you added imply to me that the file I created, openocd_caliptra.txt, is no longer needed. Can you update the FPGA README with the current steps?

hw-model/src/openocd/openocd_server.rs
timothytrippel: Ah nice, thanks for the pro tip! Fixed.

hw-model/src/openocd/openocd_server.rs
timothytrippel: fixed

hw-model/src/openocd/openocd_jtag_tap.rs
timothytrippel: fixed

hw-model/src/openocd/openocd_jtag_tap.rs
timothytrippel: fixed

hw-model/src/openocd/openocd_server.rs
timothytrippel: good point; fixed

hw-model/src/model_fpga_subsystem.rs
zhalvorsen: Its currently not being done elsewhere in this project. I wouldn't mind migrating to this at some point, but it would probably be best to do it using rustfmt.toml so it can be uniform

timothytrippel:
> The cfg files you added imply to me that the file I created, openocd_caliptra.txt, is no longer needed. Can you update the FPGA README with the current steps?

The files I added split the (sysfsgpio) adapter config and TAP config into two, but are really only to be used with the automatic OpenOCD server spawning / management infra I added to the SS HW model in this PR (as it is useful to split them up for configurability purposes). The openocd_caliptra.txt file you have is still useful for doing both in a single shot, which I have found useful for debugging when needing to launch openocd manually, outside of the HW model instantiation typically only done in a rust test.

That said, I am working on a follow up PR that can include my two files into a new file that can replace your openocd_caliptra.txt file to have a single source of truth, I just didn't want to cram that into this PR as it was already getting fairly large (500+ lines). I will include both said cleanup and doc change in a follow up shortly after if that is OK?

zhalvorsen (dismissed):

hw-model/src/model_fpga_subsystem.rs
timothytrippel: sg. fixed.

zhalvorsen (dismissed):

PR #2453, opened 2025-09-05
Mix HEK into MDK to test contents

clundin25:


drivers/test-fw/scripts/vector_gen/src/mdk.rs
```diff
@@ -19,26 +19,36 @@ use openssl::{
...
 }
 
-impl Default for Mdk {
+impl Default for Mek {
     fn default() -> Self {
         let mut input_key = [0; 64];
         hmac(&[0; 64], &[0], &mut input_key);
 
+        // This doesn't exactly match the spec, but mix in HEK so we can verify that it's contents
+        // look good in tests.
```
bluegate010, line 41: I have a few high-level points of feedback:

- There should be random string generation happening, similar to https://github.com/chipsalliance/caliptra-sw/blob/a4f6f0161273722b2f455224985239d0a1a2d771/drivers/test-fw/scripts/vector_gen/src/hmac384_kdf_vector_gen.rs#L75-L77

- Eventually we should add routines that extend https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/test/src/derive.rs to include LOCK functionality. Would let us write an end-to-end smoke test with real cryptographic test values, starting from UDS.

- For now, light-weight vector generation is ok, but the routines here should more closely follow the spec. Reduces cognitive overhead for future maintenance.

I feel like these light-weight test vectors can just start with the HEK and MDK as random values, and proceed from there. Something like:

```
rand_bytes(rand_seed);  // Have to start with a seed in order to get known values into KV.
kdf(rand_seed, "hek", None, hek);
kdf(rand_seed, "mdk", None, mdk);

// Encrypted MEK test vector
rand_bytes(random_mek);
encrypt(mdk, random_mek, encrypted_mek);
// (will need to expand this with preconditioned AES for the outer layer of encryption)

// Derived MEK test vector
rand_bytes(sek);
rand_bytes(dpk);
kdf(hek, sek, "epk", epk);
kdf(epk, dpk, "mek_secret_seed", mek_secret_seed);
kdf(mek_secret_seed, "derived_mek", derived_mek_secret);
cmac_kdf(derived_mek_secret, "mek_seed", mek_seed);
decrypt(mdk, mek_seed, derived_mek);
```

bluegate010: > Eventually we should add routines that extend https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/test/src/derive.rs to include LOCK functionality. Would let us write an end-to-end smoke test with real cryptographic test values, starting from UDS.

On reflection I'm actually not sure about this one. It could work for derived MEKs, but that smoke test suite is fundamentally deterministic in nature. If we were to use it to test randomly-generated MEKs then we'd need to predict what values would come out of the DRBG, which is a taller order.

clundin25:
Closing this PR in favor of validating everything based on `DOEInput` in the smoke tests (or equivalent).

PR #2454, opened 2025-09-05
[subsystem] Continue booting after loading fuses.

zhalvorsen:
This will let the MCU test ROM know to go to the next pause point after loading fuse values into the OTP peripheral.

Because this can now release the FPGA from reset during `new_unbooted`, the custom `new` function can also be removed.

This also changes the default AXI user to 0x1 similar to the Core hardware model.

These changes allow us to enable more tests in CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,6 @@ hw/fpga/vivado*.log
 libcaliptra/libcaliptra.a
 
 hw-latest/verilated/out/*
-/cross-target/
+cross-target/
 caliptra-test-binaries.tar.zst
 junit.xml
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ hw/fpga/vivado*.log
 libcaliptra/libcaliptra.a
 
 hw-latest/verilated/out/*
+/cross-target/
+caliptra-test-binaries.tar.zst
+junit.xml
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1443,11 +1443,14 @@ impl HwModel for ModelFpgaSubsystem {
 
         self.otp_slice().copy_from_slice(&otp_mem);
 
+        // TODO(zhalvorsen): this should be referencing the other MCI GPIO word.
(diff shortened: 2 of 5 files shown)


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1404,6 +1387,18 @@ impl HwModel for ModelFpgaSubsystem {
             .regs()
             .mcu_reset_vector
             .set(FPGA_MEMORY_MAP.rom_offset);
+
+        println!("Taking subsystem out of reset");
+        m.set_subsystem_reset(false);
+
+        // Wait for McuRomBootStatus::CaliptraBootGoAsserted
+        while m.mci_flow_status() != 0xC1 {}
```
jhand2: nit: Probably preferable to reference the defined constant instead of hard-coding.

zhalvorsen: I didn't add it originally because the definition is in caliptra-mcu-sw, but I copied the file here and referenced it directly

.gitignore
```diff
@@ -30,3 +30,6 @@ hw/fpga/vivado*.log
 libcaliptra/libcaliptra.a
 
 hw-latest/verilated/out/*
+/cross-target/
```
clundin25: nit: 
```suggestion
cross-target/
```

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1447,6 +1442,12 @@ impl HwModel for ModelFpgaSubsystem {
         otp_mem[FUSE_PQC_OFFSET] = val;
 
         self.otp_slice().copy_from_slice(&otp_mem);
+
+        // Notify MCU ROM it can start loading the fuse registers
+        let gpio1 = self.wrapper.regs().mci_generic_input_wires[1]
+            .extract()
+            .get();
+        self.wrapper.regs().mci_generic_input_wires[0].set(gpio1 | 1 << 30);
```
zhalvorsen: Just realized, I'm reading from one word and writing to the other. I'll fix this.

jhand2 (approved):

PR #2456, opened 2025-09-05
Add OCP LOCK to ROM capabilities

clundin25:
This let's a caller check if ROM and HW support OCP LOCK.


rom/dev/tests/rom_integration_tests/test_ocp_lock.rs
```diff
@@ -0,0 +1,37 @@
...
+    };
+
+    let response = hw
+        .mailbox_execute(CommandId::CAPABILITIES.into(), payload.as_bytes())
+        .unwrap()
+        .unwrap();
+
+    let capabilities_resp = CapabilitiesResp::ref_from_bytes(response.as_bytes()).unwrap();
+    let caps = Capabilities::try_from(capabilities_resp.capabilities.as_bytes()).unwrap();
+    assert!(caps.contains(Capabilities::ROM_OCP_LOCK));
```
bluegate010: Should we also have a negative test?

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -416,6 +416,13 @@ impl FirmwareProcessor {
                         let mut capabilities = Capabilities::default();
                         capabilities |= Capabilities::ROM_BASE;
 
+                        #[cfg(feature = "ocp-lock")]
+                        {
+                            if soc_ifc.ocp_lock_enabled() {
```
bluegate010: I didn't see code in the test that tries to twiddle this bit. Is this default-enabled?

rom/dev/tests/rom_integration_tests/test_ocp_lock.rs
clundin25: Added a check to `test_capabilities`. That should be testing against "core", at least for now.

rom/dev/src/flow/cold_reset/fw_processor.rs
clundin25: This is checking the HW config register to see if the hardware supports OCP LOCK: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.CPTRA_HW_CONFIG#OCP_LOCK_MODE_en.desc

bluegate010: So there isn't ever a case where `#[cfg(feature = "ocp-lock")]` is true, but `soc_ifc.ocp_lock_enabled()` is false?

clundin25: > So there isn't ever a case where #[cfg(feature = "ocp-lock")] is true, but soc_ifc.ocp_lock_enabled() is false?

I think it's possible, if there is a reason for using the same ROM on OCP LOCK / non-OCP LOCK hardware

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -416,6 +416,13 @@ impl FirmwareProcessor {
                         let mut capabilities = Capabilities::default();
                         capabilities |= Capabilities::ROM_BASE;
 
+                        #[cfg(feature = "ocp-lock")]
```
mhatrevi: Why do we have a feature flag for this? We can simply rely on the hardware capability.

rom/dev/src/main.rs
```diff
@@ -131,6 +131,14 @@ pub extern "C" fn rom_entry() -> ! {
         }
     );
 
+    if cfg!(feature = "ocp-lock") {
```
mhatrevi: See previous comment on this. I would recommend avoiding feature flags.

clundin25:
@mhatrevi I have removed the feature flag. My only concern is that now the tests will "fail open" on non-OCP LOCK hardware. 

That could probably be remedied outside of the ROM.

zhalvorsen (dismissed):

clundin25:
Blocked on: https://github.com/chipsalliance/caliptra-sw/pull/2454

rom/dev/src/flow/cold_reset/fw_processor.rs
clundin25: Removed 👍

rom/dev/src/main.rs
clundin25: Removed 👍

rom/dev/tests/rom_integration_tests/test_capabilities.rs
```diff
@@ -42,4 +42,12 @@ fn test_capabilities() {
     // Verify Capabilities
     let caps = Capabilities::try_from(capabilities_resp.capabilities.as_bytes()).unwrap();
     assert!(caps.contains(Capabilities::ROM_BASE));
+
+    // `Capabilities::ROM_OCP_LOCK` should only be set if using a subsystem fpga with a ROM
+    // compiled with the `ocp-lock` feature.
+    if hw.subsystem_mode() && hw.supports_ocp_lock() {
```
bluegate010, line 48: It's difficult to look at this test and be confident that both branches are tested. Do we know whether they are?

clundin25: They are but on different jobs. `hw.subsystem_mode() && hw.supports_ocp_lock()` will happen on the fpga-subsystem job and the other case will happen on the fpga job.

bluegate010: Thanks. Is there a pointer to where that configuration lives? Could we put that as a comment here?

clundin25: Will follow up with a comment.

The configuration is here: https://github.com/chipsalliance/caliptra-sw/blob/vmhatre/staging-2.1/.github/workflows/fpga-subsystem.yml#L217

The bitstream is managed out of band, so it's a "global" config for this branch

PR #2463, opened 2025-09-08
[openocd] add functions to interface with LCC

timothytrippel:
This adds functions to read/write LCC registers over the JTAG/OpenOCD interface. This will be used to write LC transition tests on the FPGA.

### DIFF
--- a/hw-model/src/lcc.rs
+++ b/hw-model/src/lcc.rs
@@ -1,0 +1,54 @@
+// Licensed under the Apache-2.0 license
+
+#[derive(Clone, Copy, Debug)]
+#[repr(u32)]
+pub enum LcCtrlReg {
+    AlertTest = 0x0,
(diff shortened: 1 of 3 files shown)


hw-model/src/lcc.rs
zhalvorsen, line 1: FYI these registers and offsets are already accounted for in https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/hw/latest/registers/src/lc_ctrl.rs

zhalvorsen: Nevermind, those definitions won't help you much here

timothytrippel: You make a great point though: we should find a way to autogen these offsets using ureg (or a similar mechanism); I will file an issue to track said work.

PR #2466, opened 2025-09-08
Disable FPGA tests that uses test SRAM

mlvisaya:
Some integration tests uses test SRAM that does not exist in actual HW
which causes FPGA tests to fail.

Disable these tests for now until we find a way to use an actual SRAM
that can be accessed from the integration test.


clundin25:
Can we filter these tests with a nextest filter instead of a conditional build macro? I have a feeling it's going to be hard to track tests that are only temporarily disabled.

mlvisaya:
actually these tests are already disabled: https://github.com/chipsalliance/caliptra-sw/blob/a4f6f0161273722b2f455224985239d0a1a2d771/.github/workflows/fpga.yml#L255 so no need for this PR

PR #2467, opened 2025-09-08
Add Nextest default filter for FPGA profiles

clundin25:
This should ease local development by filtering out known failing tests by default.

To run tests, use `nextest` and specify the profile.

E.g. `cargo nextest -P fpga-core run`.


clundin25:
Blocked by: https://github.com/chipsalliance/caliptra-sw/issues/2472 to use the profile in CI.

Can still be used locally, but CI needs a newer nextest.

PR #2473, opened 2025-09-09
Fixing Runtime LMS command

mhatrevi:
This fix resolves the issue where the compiler incorrectly optimizes the read access to cmd.pub_key_tree_type from the LmsVerifyReq struct.

### DIFF
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -265,11 +265,6 @@ jobs:
                        test(test_debug_unlock::test_dbg_unlock_prod_wrong_public_keys) |
                        test(test_fe_programming::test_fe_programming_cmd) |
                        test(test_fips::test_fips_shutdown) |
-                       test(test_lms::test_lms_verify_cmd) |
-                       test(test_lms::test_lms_verify_failure) |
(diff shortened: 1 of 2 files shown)


nquarton (dismissed):
I'll continue to look into this issue and understand it more. But agree this seems like a good fix for now and we should get it merged.

swenson (dismissed):
LGTM

swenson (approved):

clundin25 (approved):

zhalvorsen (approved):

swenson:
What bothers me about this though is that this **doesn't** fail with the emulator.

nquarton:
> What bothers me about this though is that this **doesn't** fail with the emulator.

I agree. I am still looking into this. If it really were because of compiler optimizations not understanding this is volatile SRAM, it should be seen in both cases (assuming the RT binary is exactly the same which it better be).

Looking at the assembly, I also do think I see where it is reading from the mailbox SRAM direct access when it gets 0. (Then again in my modified test from a forced, read_volatile when it gets the correct value of 0xc). I am going to work with Luke to see if we can look at the specific reads happening to that address (0x3004_0004) and see if the FPGA is actually getting that read and returning a 0.

PR #2475, opened 2025-09-10
Add note about AES byte reversal

clundin25:


clundin25:
In 2.1 the AES Key API has changed, so we have to handle this explicitly: https://github.com/chipsalliance/caliptra-sw/pull/2477/commits/d01d9d36080b1d707f6d61bfe7dd68923d32361b

swenson (approved):
LGTM

PR #2480, opened 2025-09-10
Allow DMA access to MCU MBOX SRAMs

mlvisaya:
Add MCU MBOX SRAM DMA address ranges in AXI root bus. This allows Caliptra to access MCU MBOX SRAMs through DMA.

Allow MCI_OFFSET base address to be set by an external compiler. Currently the MCI_OFFSET is randomly assigned at compilation. However, in some cases, the MCI_OFFSET should be known before compilation. For example, when creating the authentication manifest, the load_address or staging_address may refer to MCU MBOX SRAM which is some offsets away from the MCI_OFFSET, therefore the MCI offset should be already known before creating the authentication manifest. To allow this, allow the external compiler to pass the desired MCI_OFFSET through an  environment variable. If the environment variable is not set, then the SS_MCI_OFFSET will be assigned randomly.

### DIFF
--- a/sw-emulator/lib/bus/src/event.rs
+++ b/sw-emulator/lib/bus/src/event.rs
@@ -21,6 +21,8 @@ pub enum Device {
     TestSram,
     ExternalTestSram,
     RecoveryIntf,
+    McuMbox0Sram,
+    McuMbox1Sram,
     External(&'static str),
 }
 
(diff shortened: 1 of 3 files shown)


zhalvorsen (approved):

swenson (approved):
LGTM

PR #2483, opened 2025-09-11
Add REPORT_HEK_METADATA command

clundin25:


drivers/src/ocp_lock.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl HekSeedState {
+    /// Checks if HEK is available based on the HEK seed state and the Caliptra lifecycle state.
+    ///
+    /// Section 4.6.4.1 of the OCP LOCK v1.0 spec.
+    pub fn hek_is_available(&self, lifecycle_state: Lifecycle) -> bool {
+        match (lifecycle_state, self) {
+            (Lifecycle::Unprovisioned | Lifecycle::Manufacturing, Self::Unerasable)
```
bluegate010: Does this line here mean that if the lifecycle state is Unprovisioned or Manufacturing, and the HEK state is _not_ UNERASABLE, the HEK is unavailable? If lifecycle state is not production, we'll want the HEK to be available regardless of what the HEK seed state is.

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,55 @@
...
+    Ok(hek_available)
+}
+
+/// Check HEK seed state. Must always be called before moving to next boot stage.
+///
+/// HEK seed can only be marked available if MCU has called `REPORT_HEK_METADATA`.
+///
+/// If HEK seed is available, NOP
+/// If HEK seed is unavailable, erase HEK.
+fn check_hek_seed(pdata: &mut PersistentData, kv: &mut KeyVault) -> CaliptraResult<()> {
```
clundin25: note: This and other methods should have CFI

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,55 @@
...
+    if pdata.ocp_lock_metadata.hek_available {
+        return Ok(());
+    }
+    // HEK seed is not available, erase the HEK.
+    kv.erase_key(KEY_ID_HEK)?;
+    Ok(())
```
bluegate010:
```suggestion
    kv.erase_key(KEY_ID_HEK)
```

drivers/src/ocp_lock.rs
clundin25: Correct, that's a misunderstanding on my part. I have updated this to reflect "If lifecycle state is not production, we'll want the HEK to be available regardless of what the HEK seed state is"

api/src/mailbox.rs
```diff
@@ -4061,6 +4068,42 @@ pub struct CmDeriveStableKeyResp {
...
+bitflags! {
+    impl ReportHekMetadataRespFlags: u32 {
+        const HEK_AVAILABLE = 1u32 << 31;
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Default, IntoBytes, FromBytes, Immutable, KnownLayout, PartialEq, Eq)]
+pub struct ReportHekMetadataResp {
+    pub hdr: MailboxRespHeader,
```
zhalvorsen, line 4100: `MailboxRespHeader` has the `fips_status` in it so we can drop it from `ReportHekMetadataResp`

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,57 @@
...
+use caliptra_api::mailbox::ReportHekMetadataReq;
+use caliptra_cfi_lib::{cfi_assert, cfi_assert_bool, cfi_launder};
+use caliptra_common::keyids::ocp_lock::KEY_ID_HEK;
+use caliptra_drivers::{CaliptraResult, HekSeedState, KeyVault, Lifecycle, PersistentData, SocIfc};
+
+/// Handle `ReportHekMetadataReq` command from MCU.
+///
+/// Returns:
+///  True if the HEK is available. False otherwise.
+pub fn handle_report_hek_metadata(
```
zhalvorsen, line 14: It looks like some of the other functions that handle requests return the response structure, then populate the checksum externally. What do you think of doing this here? It could be nice to have everything self contained in case the response structure changes or flags are added.

api/src/mailbox.rs
clundin25: Good catch, thanks!

rom/dev/src/flow/cold_reset/ocp_lock.rs
clundin25: Good idea, that's cleaner. Added!

drivers/src/persistent.rs
```diff
@@ -259,6 +259,16 @@ pub struct DOT_OWNER_PK_HASH {
     reserved: [u8; 3],
 }
 
+#[derive(TryFromBytes, IntoBytes, KnownLayout, Zeroize)]
+#[repr(C)]
+pub struct OcpLockMetadata {
+    pub total_slots: u16,
+    pub active_slots: u16,
+    pub seed_state: u16,
```
bluegate010:
```suggestion
    pub total_hek_seed_slots: u16,
    pub active_hek_seed_slot: u16,
    pub hek_seed_state: u16,
```

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// Checks the HEK seed state. Sets LOCK in Progress.
+pub fn complete_ocp_lock_flow(
+    soc_ifc: &mut SocIfc,
+    pdata: &mut PersistentData,
+    kv: &mut KeyVault,
+) -> CaliptraResult<()> {
+    // Check that ROM has reported the HEK seed's state.
+    // We must always do this on a cold reset after processing FW commands.
+    check_hek_seed(pdata, kv)?;
```
bluegate010: Nit: not sure having this as its own helper function is worth it, could just inline the logic.

(I feel somewhat similarly about `hek_is_available()` but up to you.)

rom/dev/tests/rom_integration_tests/test_ocp_lock.rs
```diff
@@ -0,0 +1,116 @@
...
+            .unwrap()
+            .unwrap();
+        let response = ReportHekMetadataResp::ref_from_bytes(response.as_bytes()).unwrap();
+        assert_eq!(
+            response
+                .flags
+                .contains(ReportHekMetadataRespFlags::HEK_AVAILABLE),
+            expects_hek_available
+        );
+    }
```
bluegate010, line 129: Eventually when runtime firmware implements REPORT_EPOCH_KEY_STATE, these test routines should be updated to invoke it. May be worth leaving as a TODO or something.

Also, we'll need to use that to test what happens when REPORT_HEK_METADATA is not invoked.

clundin25:
Note: Need model changes from https://github.com/chipsalliance/caliptra-sw/pull/2456 to filter tests, otherwise the core FPGA tests fail.

rom/dev/tests/rom_integration_tests/test_ocp_lock.rs
clundin25: Added a note and tracking ticket

rom/dev/src/flow/cold_reset/ocp_lock.rs
clundin25: Leaving this as is since it's a nit, feel free to push back and I am happy to make the change! I like small functions, but I think you make a valid point.

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,64 @@
...
+    Ok(resp)
+}
+
+/// Check HEK seed state. Must always be called before moving to next boot stage.
+///
+/// HEK seed can only be marked available if MCU has called `REPORT_HEK_METADATA`.
+///
+/// If HEK seed is available, NOP
+/// If HEK seed is unavailable, erase HEK.
+fn check_hek_seed(pdata: &mut PersistentData, kv: &mut KeyVault) -> CaliptraResult<()> {
```
bluegate010: `check_hek_seed()` could stand to be a bit more expressive. Also it's not really checking the _seed_. Maybe `zeroize_hek_if_needed()`

clundin25: Updated!

drivers/src/persistent.rs
```diff
@@ -259,6 +260,16 @@ pub struct DOT_OWNER_PK_HASH {
     reserved: [u8; 3],
 }
 
+#[derive(TryFromBytes, IntoBytes, KnownLayout, Zeroize)]
+#[repr(C)]
+pub struct OcpLockMetadata {
```
zhalvorsen, line 265: Should we have a build assert making sure this const above is satisfied? Mostly so we make sure the const above always stays accurate.

```
const _: () = assert!(size_of::<OcpLockMetadata>() <= OCP_LOCK_METADATA_SIZE);
```

zhalvorsen: Actually this should be caught by the test below. Resolving.

bluegate010 (approved):

zhalvorsen (approved):

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,65 @@
...
+    Ok(resp)
+}
+
+/// Check HEK availability. Must always be called before moving to next boot stage.
+///
+/// HEK can only be marked available if MCU has called `REPORT_HEK_METADATA`.
+///
+/// If HEK is available, NOP
+/// If HEK is unavailable, erase HEK.
+fn zeroize_hek_if_needed(pdata: &mut PersistentData, kv: &mut KeyVault) -> CaliptraResult<()> {
```
mhatrevi, line 39: This function cannot fail. Suggest removing the return type.

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -0,0 +1,65 @@
...
+    resp.flags
+        .set(ReportHekMetadataRespFlags::HEK_AVAILABLE, hek_available);
+    Ok(resp)
+}
+
+/// Check HEK availability. Must always be called before moving to next boot stage.
+///
+/// HEK can only be marked available if MCU has called `REPORT_HEK_METADATA`.
+///
+/// If HEK is available, NOP
```
mhatrevi, line 37: NIT: If HEK is available, lock the KV entry.

rom/dev/src/flow/cold_reset/ocp_lock.rs
clundin25: ACK - will follow up

rom/dev/src/flow/cold_reset/ocp_lock.rs
clundin25: ACK - will follow up

PR #2490, opened 2025-09-12
Simplify running all tests for core/subsystem

ArthurHeymans:
With FPGA we cannot just pass whether we want subsystem or caliptra core
via InitParams as the CI runs on different hardware.

So align the sw emulator to this too: use generic to make A
ModelEmulated and ModelEmulatedSS. Then use a feature flag to determine
which should be used.

For convenience add a cfg flag "has_subsystem" which can be used to
indicate whether a test needs to run on the current feature flag or not.


hw-model/Cargo.toml
```diff
@@ -46,6 +47,7 @@ libc.workspace = true
 caliptra-coverage = { workspace = true, optional = true }
 caliptra-image-types.workspace = true
 tock-registers.workspace = true
+build-print = "1.0.1"
```
ArthurHeymans: remove

zhalvorsen:
> With FPGA we cannot just pass whether we want subsystem or caliptra core via InitParams as the CI runs on different hardware.

This isn't quite true anymore. Now that we have segmented FPGA builds, we could instead change the FPGA to behave more like the emulator. We should load the bitstream according to the `subsystem_mode` flag given in the InitParams for hardware model.

I would prefer to not have to split the emulator based on feature flags.

timothytrippel:
I agree with @zhalvorsen; segmented FPGA builds for Caliptra Core bitstream have been added: https://github.com/chipsalliance/caliptra-sw/pull/2550

Support for segmented subsystem image is coming soon @jlmahowa-amd @clundin25

timothytrippel (commented):
Thus I would prefer this approach not be taken.

PR #2494, opened 2025-09-14
Mark stable keys as FIPS invalid for 2.0 subsystem mode

swenson:
In 2.0 subsystem mode, the UDS and FE secrets are not guaranteed to be zeroizable. For non-identity keys, such as stable keys and derivatives, the keys and any uses of them must be marked as FIPS invalid.

For 2.0 passive mode, the integrator is expected to guarantee that the UDS and FE are zeroizable. For 2.1 in both modes, UDS and FE must be zeroizable.

This reserves one of the bits in the CMKs to be used as a FIPS bit and adds the logic to propagate that to derived CMKs and to the mailbox commands that use them.

This adds tests for the behavior in subsystem and passive mode for all cryptographic mailbox commands as well.


drivers/src/soc_ifc.rs
```diff
@@ -600,6 +601,43 @@ impl SocIfc {
...
+                .read()
+                .cptra_generation(),
+        );
+        gen.major_version() == 2 && gen.minor_version() == 0
+    }
+
+    /// Returns true if the stable keys are zeroizable according to FIPS.
+    /// In Caliptra 2.0 subsystem mode, the fuse controller does not have the logic
+    /// to zeroize UDS and FE, so the stable keys are not valid for FIPS.
+    pub fn stable_key_zeroizable(&self) -> bool {
```
mhatrevi, line 620: This check will always return false on 2.1+

hw-model/src/lib.rs
```diff
@@ -1130,6 +1132,29 @@ pub trait HwModel: SocManager {
...
+    fn version_2_0(&mut self) -> bool {
+        let gen = CptraGeneration(self.soc_ifc().cptra_hw_rev_id().read().cptra_generation());
+        gen.major_version() == 2 && gen.minor_version() == 0
+    }
+
+    /// Returns true if the stable keys are zeroizable according to FIPS.
+    /// In Caliptra 2.0 subsystem mode, the fuse controller does not have the logic
+    /// to zeroize UDS and FE, so the stable keys are not valid for FIPS.
+    fn stable_key_zeroizable(&mut self) -> bool {
+        self.version_2_0() && !self.subsystem_mode()
```
mhatrevi: See comment above.

drivers/src/soc_ifc.rs
swenson: Ah, I think I have this wrong.

mhatrevi (approved):

swenson:
Thanks! Going to go ahead and merge for ROM stabilization.

PR #2497, opened 2025-09-15
[fix] MLDSA driver fix

mhatrevi:
An update to the MLDSA peripheral requires it to be zeroized after each cryptographic operation—key generation, signing, and verification. To comply with this change, the MLDSA driver must perform a zeroize operation immediately after signing, before proceeding to signature verification.

### DIFF
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -152,12 +152,6 @@ impl Mldsa87 {
         // Wait for hardware ready
         Mldsa87::wait(mldsa, || mldsa.status().read().ready())?;
 
-        // Clear the hardware before start
-        mldsa.ctrl().write(|w| w.zeroize(true));
-
-        // Wait for hardware ready
-        Mldsa87::wait(mldsa, || mldsa.status().read().ready())?;
-
         // Copy seed to the hardware
         match seed {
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

drivers/src/mldsa87.rs
```diff
@@ -222,12 +216,6 @@ impl Mldsa87 {
         // Wait for hardware ready
         Mldsa87::wait(mldsa, || mldsa.status().read().ready())?;
```
zhalvorsen, line 218: Do we not need to wait until hardware is ready anymore? Is it fast enough that between zeroizing at the end of the signature and starting verify it will always be ready?

zhalvorsen (approved):

drivers/src/mldsa87.rs
mhatrevi: verify_common_setup() ensures that the peripheral is ready. With this change, every function (keygen/sign/verify/pcr_quote) will wait for ready at the start and zeroize at the end and return without waiting for zeroization to complete.

zhalvorsen: Oh sorry, I should have looked higher in the `key_pair` and `sign` functions. I forgot it was doing `wait-zeroize-wait` and just saw that `zeroize-wait` were removed.

LGTM

PR #2499, opened 2025-09-15
hw-model: Release MCU test ROM from second pause point.

zhalvorsen:
This will notify the MCU test ROM that it is ready to start loading mutable firmware for both Caliptra and MCU.


ArthurHeymans:
> This will notify the MCU test ROM that it is ready to start loading mutable firmware for both Caliptra and MCU.

I suppose this fixes booting with core_test feature set on the MCU ROM?

zhalvorsen:
> > This will notify the MCU test ROM that it is ready to start loading mutable firmware for both Caliptra and MCU.
> 
> I suppose this fixes booting with core_test feature set on the MCU ROM?

Yes, this allows both the MCU and Caliptra to boot to mutable firmware

Issue #2470, opened 2025-09-09
Re-name ROM/FMC CDI


clundin25:
[KEY_ID_ROM_FMC_CDI](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/common/src/keyids.rs#L29) should be split into two names, to avoid confusion.

KEY_ID_IDEVID_CDI = KeyId::KeyId6;
KEY_ID_FMC_CDI = KeyId::KeyId6;

clundin25:
@clundin-bot please fix.

PR #2500, opened 2025-09-15
Split KEY_ID_ROM_FMC_CDI into two names

mtimkovich:
#2470


clundin25 (dismissed):

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -54,7 +54,7 @@ impl LocalDevIdLayer {
     #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
     pub fn derive(env: &mut RomEnv, input: &DiceInput) -> CaliptraResult<DiceOutput> {
         cprintln!("[ldev] ++");
-        cprintln!("[ldev] CDI.KEYID = {}", KEY_ID_ROM_FMC_CDI as u8);
+        cprintln!("[ldev] CDI.KEYID = {}", KEY_ID_IDEVID_CDI as u8);
```
bluegate010: Ah looks like we need to also represent KEY_ID_LDEVID_CDI using the same KeyID value.

clundin25 (dismissed):

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -74,14 +74,14 @@ impl FmcAliasLayer {
         let mut measurement = env.pcr_bank.read_pcr(PCR_ID_FMC_CURRENT);
 
         // Derive the DICE CDI from the measurement
-        let result = Self::derive_cdi(env, &measurement, KEY_ID_ROM_FMC_CDI);
+        let result = Self::derive_cdi(env, &measurement, KEY_ID_FMC_CDI);
```
bluegate010, line 77:
```suggestion
        let result = Self::derive_cdi(env, &measurement, KEY_ID_LDEVID_CDI);
```

As this is the parent CDI from which we derive the child.

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -68,21 +68,21 @@ impl LocalDevIdLayer {
         cprintln!("[ldev] FE.KEYID = {}", KEY_ID_FE as u8);
 
         // Derive the stable identity root IDevID key
-        Self::derive_stable_identity_root_idev(env, KEY_ID_ROM_FMC_CDI, KEY_ID_STABLE_IDEV)?;
+        Self::derive_stable_identity_root_idev(env, KEY_ID_LDEVID_CDI, KEY_ID_STABLE_IDEV)?;
```
bluegate010, line 71:
```suggestion
        Self::derive_stable_identity_root_idev(env, KEY_ID_IDEVID_CDI, KEY_ID_STABLE_IDEV)?;
```

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -68,21 +68,21 @@ impl LocalDevIdLayer {
...
 
         // The measurement for this layer is generated by previous layer
         // (Initial Device ID DICE Layer).
         //
         // This is the decrypted Field Entropy
```
bluegate010, line 76: This is a nit, but:

```suggestion
        // This is the decrypted Field Entropy
        //
        // derive_cdi() takes one slot ID and uses it as both the source and destination.
        // Because the IDevID and LDevID CDIs are stored in the same slot, we just pass
        // the LDevID CDI slot here.
```

... Or we change the implementation of derive_cdi() here to be consistent with the [version](https://github.com/chipsalliance/caliptra-sw/blob/c87a3f6086877857d2c83b33f9752ebb46021193/rom/dev/src/flow/cold_reset/fmc_alias.rs#L146) in fmc_alias.rs, where the function is explicit about targeting a particular output slot ID.

clundin25:
@mtimkovich friendly ping

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -240,7 +240,7 @@ impl InitDevIdLayer {
         Crypto::hmac_kdf(
             &mut env.hmac,
             &mut env.trng,
-            KEY_ID_ROM_FMC_CDI,
+            KEY_ID_FMC_CDI,
```
clundin25, line 243: I think this should be `KEY_ID_IDEVID_CDI`

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -260,7 +260,7 @@ impl InitDevIdLayer {
...
```
clundin25, line 263: I think this should be `KEY_ID_IDEVID_CDI`

PR #2504, opened 2025-09-16
fpga: Better support for I3C IBIs

swenson:
* Support resetting AXI I3C controller (this fixes some flakiness after recovery is finished)
* Add helpers to get I3C controller FIFO levels
* Simplify boot sequence for subsystem FPGA

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1530,9 +1530,6 @@ impl HwModel for ModelFpgaSubsystem {
         );
         self.bmc.push_recovery_image(mcu_fw_image);
 
-        println!("Taking subsystem out of reset");
-        self.set_subsystem_reset(false);
-
         while !self.i3c_target_configured() {
             self.step();
         }


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1478,7 +1507,35 @@ impl HwModel for ModelFpgaSubsystem {
...
+        self.bmc.push_recovery_image(
+            boot_params
+                .soc_manifest
+                .map(|s| s.to_vec())
+                .unwrap_or_default(),
+        );
+        self.bmc.push_recovery_image(mcu_fw_image);
+
+        println!("Taking subsystem out of reset");
+        self.set_subsystem_reset(false);
```
zhalvorsen: This is already done in `new_unbooted`, do we need to do it again?

swenson: Ah, I missed that. (This was originally branched from an old commit.) We don't need this extra one.

clundin25 (approved):

swenson:
Thanks!

PR #2509, opened 2025-09-17
ss: Add boot milestones

zhalvorsen:
This is preparation for https://github.com/chipsalliance/caliptra-mcu-sw/pull/424. The CI for that PR can't pass without these changes.

### DIFF
--- a/hw-model/src/mcu_boot_status.rs
+++ b/hw-model/src/mcu_boot_status.rs
@@ -12,16 +12,18 @@ Abstract:
 
 --*/
 
-const ROM_INITIALIZATION_BASE: u32 = 1;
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM, but can we merge this without updating the FPGA ROM in CI?

zhalvorsen:
> LGTM, but can we merge this without updating the FPGA ROM in CI?

The current ROM only modifies the bottom 16 bits and this change makes it so the model only looks at the bottom 16 bits. When https://github.com/chipsalliance/caliptra-mcu-sw/pull/424 merges we can update the hardware model to look at milestones instead which should be more stable

Issue #2353, opened 2025-08-06
1.2.3 RT FW Release Checklist


nquarton:
- [x] https://github.com/chipsalliance/caliptra-sw/pull/2318
- [x] https://github.com/chipsalliance/caliptra-sw/pull/1561
- [x] https://github.com/chipsalliance/caliptra-sw/issues/2320
- [x] https://github.com/chipsalliance/caliptra-sw/issues/2340
- [x] https://github.com/chipsalliance/caliptra-sw/issues/2530
- [x] (Cherry pick fix from 2.x) https://github.com/chipsalliance/caliptra-sw/issues/2543
- [x] https://github.com/chipsalliance/caliptra-sw/pull/2657

Issue #2320, opened 2025-07-22
Caliptra Dynamic Allocation for PL Slots


fdamato:
Description:

A new command will be introduced for runtime repartitioning using PL0 and PL1, with the default setting remaining at 16.

A capability command will be created to activate this feature.
Action Items:

-  Introduce a new command for runtime repartitioning using PL0 and PL1.
-  Create a capability command to activate dynamic allocation.

PR #2510, opened 2025-09-17
Allow PL0 to use context slots reserved for PL1

kaienl-nv:
The original limitation of maximum 16 contexts in each PL already cannot satisfy some complicated SoCs' requirements, which have of a large number of different TCBs in its identity. As a temporary mitigation, this patch allows the PL0 user to take those memory (context array) reserved for PL1 contexts. After this, the PL0 user can have up to (32 - 1) contexts (minus 1 because RT journey always occupies one contexts in PL1). Although this opens the possibility that the PL0 user can deny the DPE service for PL1 users, it is acceptable because PL0 user should be the SoC manager.

This also fixes the inaccurate description in `runtime/README.md`, where the context limit is actually applied to the summation of both active and retired contexts. Even if the user only retains one active context, there could be a lot of retired parent contexts occupying the context array.

Testing: `cargo test` passed.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -1090,11 +1090,11 @@ InvokeDpe mailbox command calling the DeriveContext or InitializeContext DPE
 commands. However, a caller could easily exhaust space in DPE's context array
 by repeatedly calling the aforementioned DPE commands with certain flags set.
 
-To prevent this, we establish active context limits for each PAUSER
-privilege level:
+To prevent this, we establish non-inactive (including active and retired) context
+limits based on PAUSER privilege level:
 
-* PL0 - 16 active contexts
-* PL1 - 16 active contexts
+* Total (PL0 + PL1) - 32 non-inactive contexts
+* PL1 - 16 non-inactive contexts
 
 If a DPE command were to activate a new context such that the total number of
 active contexts in a privilege level is above its active context limit, the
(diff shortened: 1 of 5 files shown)


nquarton:
Hi, there is a planned change to reallocate the contexts between PL0 and PL1 with a new command (issue https://github.com/chipsalliance/caliptra-sw/issues/2320). Would this satisfy your requirements? The concern voiced with an approach that allows PL0 and PL1 to share any portion of the contexts is that a rogue PL1 entity could use up contexts effectively creating a Denial of Service attack for PL0.

kaienl-nv:
Hi @nquarton , I believe the capability to repartition context limits should satisfy our requirements. Could you share the estimated release timeline for this feature? Time is also important for us so we want to check if it is okay to wait for it.

For addressing the concern, this change allows only PL0 to take portion from PL1 but not in reverse, so PL1 cannot DoS PL0.

nquarton:
I think a PR on this can be expected by end of next week. This is planned in the next patch release of 1.2 RT FW (1.2.3 FW outlined in issue https://github.com/chipsalliance/caliptra-sw/issues/2353) which should follow shortly after if no other needed changes arise.

Definitely agree PL0 is guaranteed its original 16 contexts with this implementation. But the concern was that if PL0 requires say 20 contexts in its boot flow, PL1 could still end up taking contexts that PL0 needs. I don't think it was a _major_ concern, but that's why the reallocation approach was preferred.

nquarton:
Here is the PR I was describing before: https://github.com/chipsalliance/caliptra-sw/pull/2525

kaienl-nv:
Closing this as rt-1.2.3 has been released. Thank you.

PR #2518, opened 2025-09-18
ml-dsa: Add external mu emulator support

zhalvorsen:
This adds support for using external mu to sign and verify in the software emulator. It currently uses a release candidate version of the `ml-dsa` crate because it now supports it. We should use an official release once it is available.


swenson (approved):
LGTM; I'm glad the ml-dsa dependency worked out because we usually have a very hard time with RustCrypto dependencies.

zhalvorsen:
Thanks! Me too, it could have been a nightmare

PR #2523, opened 2025-09-20
[fix] AXI mbox user warm reset fix

mhatrevi:
This update modifies the emulator to reset the MBOX AXI users and their locks during a warm reset. It also enhances the warm reset integration tests to reprogram the AXI users and their locks afterward.


swenson (approved):
LGTM

PR #2525, opened 2025-09-22
Adding REALLOCATE_DPE_CONTEXT_LIMIT command

nquarton:
- Adding command to reallocate contexts between PL0 and PL1
- Fixed off-by-one error for calculating PL0 context limit
- Now consider Caliptra's locality to be PL0 for context limit purposes


drivers/src/persistent.rs
```diff
@@ -292,6 +293,9 @@ pub struct PersistentData {
 
     reserved11: [u8; FMC_ALIAS_CSR_SIZE as usize - size_of::<FmcAliasCsr>()],
 
+    pub dpe_pl0_context_limit: u8,
+    pub dpe_pl1_context_limit: u8,
```
zhalvorsen, line 297: Is there a reason these are tracked separately instead of `dpe_pl1_context_limit` always being `dpe::MAX_HANDLES - dpe_pl0_context_limit`?

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
```diff
@@ -479,10 +479,9 @@ fn test_fill_max_contexts() {
...
     // 32 contexts = 1 root node +
     //               1 rt_journey auto init) +
-    //               1 (valid pauser at init) +
     //               14 PL0 contexts in loop +
     //               1 PL1 context as transition +
-    //               14 PL1 contexts in loop
+    //               15 PL1 contexts in loop
 
     // Fill PL0 contexts
     for _ in 0..14 {
```
zhalvorsen: Can you add a check after filling the PL0 contexts that adding another one will return an error:

```Rust
    // Trigger failure by trying to derive one more context to PL0
    let _ = execute_dpe_cmd(
        &mut model,
        &mut Command::DeriveContext(&BASE_CMD),
        DpeResult::MboxCmdFailure(CaliptraError::RUNTIME_PL0_USED_DPE_CONTEXT_THRESHOLD_REACHED),
    );
```

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
```diff
@@ -64,10 +65,10 @@ fn test_pl0_derive_context_dpe_context_thresholds() {
     let mut handle = rotate_ctx_resp.handle;
 
     // Call DeriveContext with PL0 enough times to breach the threshold on the last iteration.
-    // Note that this loop runs exactly PL0_DPE_ACTIVE_CONTEXT_THRESHOLD times. When we initialize
-    // DPE, we measure mailbox valid pausers in pl0_pauser's locality. Thus, we can call derive child
-    // from PL0 exactly 7 times, and the last iteration of this loop, is expected to throw a threshold breached error.
-    let num_iterations = PL0_DPE_ACTIVE_CONTEXT_THRESHOLD;
+    // 2 PL0 contexts are used by default by Caliptra. When we initialize DPE, we measure mailbox valid pausers in pl0_pauser's locality.
+    // The RT Journey measurement also is cointed against PL0's limit. Thus, we can call derive child
```
zhalvorsen:
```suggestion
    // The RT Journey measurement also is counted against PL0's limit. Thus, we can call derive child
```

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    // Trigger failure by trying to derive one more context to PL0
+    let _ = execute_dpe_cmd(
+        model,
+        &mut Command::DeriveContext(&BASE_DERIVE_CONTEXT_CMD),
+        DpeResult::MboxCmdFailure(CaliptraError::RUNTIME_PL0_USED_DPE_CONTEXT_THRESHOLD_REACHED),
+    );
+
+    // Fill PL1 contexts
+    if pl1_limit > 0 {
```
zhalvorsen: When pl1_limit is 0, the for loop will be skipped. Should we remove this if statement?

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    model.mailbox_execute(
+        u32::from(CommandId::REALLOCATE_DPE_CONTEXT_LIMITS),
+        cmd.as_bytes().unwrap(),
+    )
+}
+
+#[test]
+fn test_pl0_pl1_reallocation_range() {
+    for pl0_limit in 2..32 {
```
zhalvorsen: You could use the DPE definition for the maximum number of contexts here. But on the other hand, we are planning to bump that value to 64 for 2.x and it would be nice to see that fail here.

```suggestion
    for pl0_limit in 2..MAX_HANDLES {
```

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
```diff
@@ -0,0 +1,296 @@
...
+    )
+    .unwrap()
+    .to_bytes()
+    .unwrap();
+    model
+        .mailbox_execute(u32::from(CommandId::FIRMWARE_LOAD), &updated_fw_image)
+        .unwrap();
+
+    // Use the rest of the PL0 contexts
+    // (2 contexts are used by Ccaliptra)
```
zhalvorsen:
```suggestion
    // (2 contexts are used by Caliptra)
```

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
zhalvorsen: TBH, I wouldn't mind dropping this test and relying on your reallocation tests

drivers/src/persistent.rs
nquarton: Good question. It could be done that way. The only advantage here is this limits where you have to implement calculating the PL1 limit. It's retrieved in more than one place and I thought it was cleaner to just calculate it once when reallocating to maintain a single point of truth.

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
nquarton: Good point, I removed this test

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
nquarton: Fixed

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
nquarton: Good catch. Must have been leftover as I updated this during development.

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
nquarton: Good call, done

runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs
nquarton: Thanks, fixed

drivers/src/persistent.rs
zhalvorsen: Sounds good to me

PR #2526, opened 2025-09-23
hw-model: Add utf8 support

swenson:
Tock uses Unicode characters in utf8 occasionally and our demos might use them to make everything pretty.

This should have no performance impact when using plain ASCII, as the new behavior is only triggered on non-ASCII characters.

Miraculously, the characters used for tests (like `0xff`) are not valid in UTF-8 sequences.


swenson:
Thanks!

PR #2527, opened 2025-09-23
AXI bypass custom test ROM

clundin25:


drivers/test-fw/src/bin/axi_bypass.rs
```diff
@@ -0,0 +1,62 @@
...
+    let soc_ifc = SocIfc::new(unsafe { SocIfcReg::new() });
+    let dma = Dma::default();
+    let mut mbox = unsafe { Mailbox::new(MboxCsr::new()) };
+
+    let caliptra_base = AxiAddr::from(soc_ifc.caliptra_base_axi_addr());
+    let recovery_base = AxiAddr::from(soc_ifc.recovery_interface_base_addr());
+    let mci_base = soc_ifc.mci_base_addr();
+
+    let dma_recovery =
+        DmaRecovery::new(recovery_base, caliptra_base, AxiAddr::from(mci_base), &dma);
```
zhalvorsen: uber nit: Can you make this an AxiAddr from the beginning like the other addresses?
```suggestion
    let mci_base = AxiAddr::from(soc_ifc.mci_base_addr());

    let dma_recovery = DmaRecovery::new(recovery_base, caliptra_base, mci_base, &dma);
```

PR #2528, opened 2025-09-23
[Fix] Restore FMC & RT Firmware Versions on Warm Reset

### LINKED ISSUE #2502: Bug: FMC/RT version is missing after warm reset
caliptra 1.x/2.x

After a warm-reset, the VERSION command does not show the FMC/FW version.

get_version() reads the version from the register which will be cleared by warm-reset. However, the version information in that register is not updated during a warm-reset.

https://github.com/chipsalliance/caliptra-sw/blob/b7195a64c3683076917708c9169db42ee70ec12b/drivers/src/soc_ifc.rs#L357-L363

https://github.com/chipsalliance/caliptra-sw/blob/b7195a64c3683076917708c9169db42ee70ec12b/common/src/fips.rs#L17-L21

mhatrevi:

This is a temporary workaround until the hardware update planned for version 2.2 addresses the issue more robustly.

### DIFF
--- a/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
+++ b/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
@@ -1,15 +1,20 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_api::SocManager;
-use caliptra_builder::firmware::FMC_WITH_UART;
-use caliptra_builder::firmware::{APP_WITH_UART, ROM_WITH_UART};
-use caliptra_builder::ImageOptions;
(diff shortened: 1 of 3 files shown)


swenson (approved):
LGTM

rusty1968 (approved):

zhalvorsen (approved):

sunguk-bin-fad:
This needs to be applied to the 'main' branch as well.

zhalvorsen:
> This needs to be applied to the 'main' branch as well.

Thanks for the reminder, @nquarton is making this change in https://github.com/chipsalliance/caliptra-sw/pull/2615

PR #2545, opened 2025-09-26
runtime: Avoid immediate save after restore in CM_SHA_UPDATE

swenson:
If CM_SHA_INIT contains more than 128 bytes, then an intermediate hash is generated and restored when starting CM_SHA_UPDATE.

If the CM_SHA_UPDATE does not contain enough data to trigger another block write, then we would try to immediately save the SHA state. In that case, the hardware status would show ready but not valid, so the driver would get into an infinite loop waiting for the save state to be ready.

This addresses this condition by not doing triggering a save if we did not write enough data and instead using the old intermediate hash.

Tested on the FPGA.

### DIFF
--- a/runtime/src/cryptographic_mailbox.rs
+++ b/runtime/src/cryptographic_mailbox.rs
@@ -649,8 +649,8 @@ impl Commands {
 
         context.length = data_len as u32;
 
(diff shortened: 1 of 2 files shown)


swenson:
Thanks!

PR #2546, opened 2025-09-26
fpga: Support FE programming command test

swenson:
The test is disabled in the FPGA for now due to flakiness of booting all the way to runtime, but it has been tested as working and is ready for CI whenever that flakiness is resolved.

To support this command (and probably others) we need to pass in SoC manifest and MCU FW when operating in subsystem mode, otherwise the FPGA subsystem boot flow will fail.

We also need to set the SoC Manifest Max SVN in fuses so that these are routed back to the Caliptra fuse registers.

### DIFF
--- a/runtime/tests/runtime_integration_tests/common.rs
+++ b/runtime/tests/runtime_integration_tests/common.rs
@@ -1,11 +1,13 @@
 // Licensed under the Apache-2.0 license
 
+use crate::test_set_auth_manifest::create_auth_manifest_with_metadata;
 use caliptra_api::{
     mailbox::{GetFmcAliasMlDsa87CertResp, Request},
     SocManager,
 };
-use caliptra_auth_man_types::ImageMetadataFlags;
-use caliptra_auth_man_types::{AuthManifestImageMetadata, AuthorizationManifest};
+use caliptra_auth_man_types::{
+    AuthManifestImageMetadata, AuthorizationManifest, ImageMetadataFlags,
+};
 use caliptra_builder::{
     firmware::{APP_WITH_UART, APP_WITH_UART_FPGA, FMC_WITH_UART},
     FwId, ImageOptions,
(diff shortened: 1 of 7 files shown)


mhatrevi (dismissed):

runtime/tests/runtime_integration_tests/test_fe_programming.rs
```diff
@@ -12,10 +12,17 @@ use caliptra_runtime::RtBootStatus;
 
 #[test]
 fn test_fe_programming_cmd() {
-    let rom = caliptra_builder::rom_for_fw_integration_tests().unwrap();
+    if cfg!(feature = "fpga_realtime") {
```
zhalvorsen: Would it be better to use ignore here? I'm worried someone will run this thinking it is doing something when it isn't

runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -172,6 +170,37 @@ pub fn start_rt_test_pqc_model(
...
+    let init_params = match args.init_params {
+        Some(init_params) => init_params,
+        None => InitParams {
+            rom: &rom,
+            stack_info: Some(StackInfo::new(image_info)),
+            test_sram: args.test_sram,
+            security_state: args.security_state.unwrap_or_default(),
+            ..Default::default()
+        },
+    };
```
zhalvorsen: nit: optional: you could use the same pattern as above and use `unwrap_or` here

```suggestion
    let init_params = args.init_params.unwrap_or(|| InitParams {
        rom: &rom,
        stack_info: Some(StackInfo::new(image_info)),
        test_sram: args.test_sram,
        security_state: args.security_state.unwrap_or_default(),
        ..Default::default()
    });
```

runtime/tests/runtime_integration_tests/test_fe_programming.rs
swenson: I wanted to avoid ignore so that nextest would still be fine.

zhalvorsen: That's a good point, but if we are filtering it with nextest maybe it would be good to let this fail?

zhalvorsen: Or return an error indicating it should never work

swenson: I can move this to an ignore though. I realize nextest should honor the ignore.

swenson: If this test can never work on the core FPGA, I think I'm okay just marking it as ignore in the code.

PR #2547, opened 2025-09-26
Enable disabled ActivateFw and AuthAndStash integ tests

mlvisaya:
Previously all ActivateFw and AuthAndStash integ tests were disabled when testing FPGA.

This was because these integ tests needs an external test SRAM to exchange data with Caliptra. SW emulator is using a SW SRAM which is not available previously in HW.

To work around this, we will be using the MCU MBOX SRAM as a test SRAM to enable these tests.


runtime/tests/runtime_integration_tests/test_activate_firmware.rs
```diff
@@ -90,7 +100,38 @@ fn load_and_authorize_fw(images: &[Image]) -> DefaultHwModel {
...
+
+    #[cfg(feature = "fpga_subsystem")]
+    {
+        println!("locking  MCU mailbox SRAMs");
+        unsafe {
+            let mcu_mbox_exec_ptr = model.mci.ptr.add(0x600018 / 4) as *mut u32;
+            mcu_mbox_exec_ptr.write_volatile(0x1);
+
+            let mcu_mbox_lock_ptr = model.mci.ptr.add(0x600000 / 4) as *mut u32;
+            let _ = mcu_mbox_lock_ptr.read_volatile();
```
zhalvorsen: The order of these should be swapped. In general, it should lock and then write to the execute bit.


```suggestion
            let mcu_mbox_lock_ptr = model.mci.ptr.add(0x600000 / 4) as *mut u32;
            let _ = mcu_mbox_lock_ptr.read_volatile();

            let mcu_mbox_exec_ptr = model.mci.ptr.add(0x600018 / 4) as *mut u32;
            mcu_mbox_exec_ptr.write_volatile(0x1);
```

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
```diff
@@ -90,7 +100,38 @@ fn load_and_authorize_fw(images: &[Image]) -> DefaultHwModel {
...
+            let mcu_mbox_sram_ptr = model.mci.ptr.add(0x400000 / 4) as *mut u32;
+
+            for (count, chunk) in test_sram_contents.chunks(4).enumerate() {
+                let mut word = 0u32;
+                for (i, byte) in chunk.iter().enumerate() {
+                    word |= (*byte as u32) << (i * 8);
+                }
+                mcu_mbox_sram_ptr
+                    .offset(count as isize)
+                    .write_volatile(word);
```
zhalvorsen: This could be simplified with `u32::from_be_bytes`:


```suggestion
                mcu_mbox_sram_ptr
                    .offset(count as isize)
                    .write_volatile(u32::from_be_bytes(chunk.try_into().unwrap());
```

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -996,6 +1022,71 @@ fn test_authorize_and_stash_after_update_reset_multiple_set_manifest() {
...
+    #[cfg(feature = "fpga_subsystem")]
+    {
+        let staging_address = address.lo as usize - TEST_SRAM_BASE.lo as usize;
+        let mut test_sram_contents = vec![0u8; TEST_SRAM_SIZE];
+        let image_size = data.len();
+        test_sram_contents[staging_address..staging_address + image_size].copy_from_slice(data);
+
+        println!("locking  MCU mailbox SRAMs");
+        unsafe {
+            let mcu_mbox_exec_ptr = model.mci.ptr.add(0x600018 / 4) as *mut u32;
```
zhalvorsen: Same comments as above here. Actually, would you mind combining these into a separate function that can be used in both places?

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
```diff
@@ -143,16 +221,12 @@ fn test_activate_mcu_fw_success() {
...
-        .mailbox_execute(
-            u32::from(CommandId::ACTIVATE_FIRMWARE),
-            activate_cmd.as_bytes().unwrap(),
-        )
+    send_activate_firmware_cmd(&mut model, activate_cmd)
         .unwrap()
         .expect("We should have received a response");
 }
 
+#[cfg(not(feature = "fpga_realtime"))]
```
clundin25: Is this a long term filter? Is this to only run the test for subsystem?

.github/workflows/fpga-subsystem.yml
```diff
@@ -297,6 +297,7 @@ jobs:
               -E 'package(caliptra-image-elf)'
               -E 'package(caliptra-image-types)'
               -E 'package(caliptra-lms-types)'
+              -E 'package(caliptra-runtime) & test(test_activate_firmware::test_activate_fw_id_not_in_manifest) - test(test_activate_firmware::test_activate_invalid_fw_id) - test(test_activate_firmware::test_activate_mcu_fw_success) - test(test_activate_firmware::test_activate_mcu_soc_fw_success) - test(test_activate_firmware::test_activate_soc_fw_success) - test(test_activate_firmware::test_invalid_exec_bit_in_manifest) - test(test_authorize_and_stash::test_authorize_from_load_address) - test(test_authorize_and_stash::test_authorize_from_load_address_incorrect_digest) - test(test_authorize_and_stash::test_authorize_from_staging_address) - test(test_authorize_and_stash::test_authorize_from_staging_address_incorrect_digest)'
```
swenson: The `-` is subtracting the tests out I think?

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -1008,19 +1099,28 @@ fn test_authorize_from_load_address() {
...
+            lo: TEST_SRAM_BASE.lo,
+            hi: TEST_SRAM_BASE.hi,
         },
         ..Default::default()
-    }];
-    let auth_manifest = create_auth_manifest_with_metadata(image_metadata);
-    let mut model = set_auth_manifest_with_test_sram(Some(auth_manifest), &load_memory_contents);
-
+    };
+    let mcu_image = [0xAAu8; 256];
```
swenson: Could we change these to not all be the same byte? (We've had problems before where got the endianness wrong when feeding to the SHA accelerator, for instance.)

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
mlvisaya: yes, the test can only run on subsystem because fpga_realtime doesn't contain MCU MBOX SRAM HW which this test needs

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
zhalvorsen: @swenson what do you think of pulling the MCU manager into `caliptra-sw` so we can use all of the peripherals? I doubt this is the only place we would use them in this repo. I can do it in a follow on

mlvisaya: Actually the reason why the execute was done first, is to make sure that the lock is released, in case previous tests have not released it (for example due to some failure). For this use case there's actually no use for the execute command other than to release the lock, since we just want to use the SRAM as data storage.

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
clundin25: Hmm I think @swenson has been using the model to check for subsystem to exit early. I think that might be more future proof

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
mlvisaya: good idea

.github/workflows/fpga-subsystem.yml
mlvisaya: yes you're right. i thought it was a separator. i'll correct it.

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
mlvisaya: yes i can modify it to an increasing integer values.

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
mlvisaya: yes i'll put it one function

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
mlvisaya: i think the actual MCU RT FW will exit immediately, but in these tests we just supplied a random mcu image that will be placed in the MCU SRAM, so when it boots up, it will not call the fpga_exit

runtime/tests/runtime_integration_tests/test_activate_firmware.rs
zhalvorsen: Wouldn't you want to write `0` to execute in that scenario so it clears any previously started command?

mlvisaya: yes you're right it should be 0. i'll update it.

swenson (approved):
LGTM

zhalvorsen (approved):

mhatrevi (approved):

PR #2549, opened 2025-09-26
[ci-tools]: Remove Bitstream bucket delete lifecycle

clundin25:
Let's keep object indefinitely until we have a plan for generating new
bitstreams.


swenson (approved):
LGTM

clundin25:
Thanks!

PR #2563, opened 2025-10-01
Enabling Cryptographic Mailbox tests for Subsystem FPGA

mhatrevi:


runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
```diff
@@ -50,7 +50,21 @@ use zerocopy::{FromBytes, IntoBytes};
 
 #[test]
 fn test_status() {
-    let mut model = run_rt_test(RuntimeTestArgs::default());
+    let subsystem_mode = cfg!(feature = "fpga_subsystem");
```
clundin25, line 53: Could this be part of `InitParams::default`? 

Alternatively, could we create a new builder function for `InitParams` that handles this logic? E.g. `InitParams::default_with_subsystem`

mhatrevi:
Abandoing as this is fixed by other changes.

PR #2579, opened 2025-10-02
emulator: Add OCP LOCK fuse registers

swenson:
These are needed for MCU to boot in 2.1 since it writes to these fuses during its boot up.


swenson:
Thanks!

zhalvorsen (approved):

PR #2582, opened 2025-10-02
ss-fpga: Add recently removed tests back

zhalvorsen:
Extending this wait period makes it so it no longer panics running these tests.


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1561,7 +1561,7 @@ impl HwModel for ModelFpgaSubsystem {
         {
             // Give the FPGA some time to start. If this returns too quickly some of the tests fail
             // with a kernel panic.
-            for _ in 0..5_000 {
+            for _ in 0..10_000 {
```
clundin25: Why loop w/ a time out here?

Would looping until `flow_status.ready_for_mb_processing() == true` work?

If the FPGA get's stuck could emit a log every 30s or so.

swenson: suggestion: instead of stepping a certain amount of times, wait for the cycle count to have a certain delta? That way it will be more deterministic at least.

zhalvorsen: > Would looping until flow_status.ready_for_mb_processing() == true work?

The problem is not a lot of tests set that FW status

Changed it to use cycle count

clundin25 (approved):

swenson (approved):
LGTM

PR #2587, opened 2025-10-06
Fix test for FPGA subsystem

ArthurHeymans:


runtime/tests/runtime_integration_tests/test_boot.rs
```diff
@@ -161,11 +164,19 @@ fn test_stress_update() {
 
 #[test]
 fn test_boot_tci_data() {
+    let fw_id = if cfg!(all(feature = "fpga_realtime", feature = "fpga_subsystem")) {
```
zhalvorsen, line 167: Is this supposed to be `any`? I don't think we ever have both of them defined. If this should be `any`, it looks like there are a few other places in this PR that should also be changed. I can make the changes.
```suggestion
    let fw_id = if cfg!(any(feature = "fpga_realtime", feature = "fpga_subsystem")) {
```

ArthurHeymans: Oh you're right indeed. I'll fix it.

ArthurHeymans: https://github.com/chipsalliance/caliptra-sw/pull/2637

zhalvorsen: Thank you!

PR #2592, opened 2025-10-06
Add a test ROM for MCU hitless update

zhalvorsen:
This will be tested in the MCU repo to ensure th Hitless Update flow works correctly in hardware


hw-model/test-fw/mcu_hitless_update_flow.rs
```diff
@@ -0,0 +1,78 @@
+// Licensed under the Apache-2.0 license
+
+//! A very simple program that responds to the mailbox.
```
clundin25: Is this comment out of date?

Also, it might be worth including a comment that this test is meant to be run from the MCU sw repo

clundin25 (dismissed):

hw-model/test-fw/mcu_hitless_update_flow.rs
zhalvorsen: Good catch, I updated the comment.

PR #2606, opened 2025-10-08
Fix Hitless Update Activate Firmware Mailbox command sequence

mlvisaya:
During HW testing, when MCU sends Activate Firmware mailbox command to perform MCU hitless update, clearing FW_EXEC_CTRL[2] didn't trigger MCI notif_cptra_mcu_reset_req_sts interrupt to MCU. Caliptra needs to trigger the interrupt via notif0_intr_trig_r register.

After the interrupt occurs, MCU will reset itself and clears the interrupt, this will be an indication to Caliptra to clear FW_EXEC_CTRL[2].

Removed the unnecesary check RESET_STATUS.MCU_RESET_STS. This indicates that the MCU is on reset, but the transition happens so fast that the SW will not catch the transition. It is enough for Caliptra to wait on notif0_intr_trig_r getting cleared to know that MCU already reset.


rom/dev/Cargo.toml
```diff
@@ -25,6 +25,7 @@ caliptra-lms-types.workspace = true
 caliptra-registers.workspace = true
 caliptra-x509 = { workspace = true, default-features = false }
 ufmt.workspace = true
+ureg.workspace = true
```
mhatrevi: Why is this change required?

mhatrevi (changes requested):

rom/dev/Cargo.toml
mlvisaya: you're right this is not needed. I'll remove.

mhatrevi (approved):

PR #2615, opened 2025-10-10
[Fix] Restore FMC & RT Firmware Versions on Warm Reset (#2528)

nquarton:

### DIFF
--- a/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
+++ b/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
@@ -1,17 +1,22 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_api::SocManager;
-use caliptra_builder::firmware::FMC_WITH_UART;
-use caliptra_builder::firmware::{APP_WITH_UART, ROM_WITH_UART};
-use caliptra_builder::ImageOptions;
-use caliptra_common::mailbox_api::CommandId;
-use caliptra_common::RomBootStatus::*;
+use caliptra_api::{
+    mailbox::{FipsVersionResp, MailboxReqHeader, MailboxRespHeader},
+    SocManager,
+};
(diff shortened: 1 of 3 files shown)


nquarton:
Cherry pick of https://github.com/chipsalliance/caliptra-sw/pull/2528

rom/dev/tests/rom_integration_tests/test_warm_reset.rs
```diff
@@ -216,3 +221,135 @@ fn test_warm_reset_during_update_reset() {
         u32::from(CaliptraError::ROM_WARM_RESET_UNSUCCESSFUL_PREVIOUS_UPDATE_RESET)
     );
 }
+
+const HW_REV_ID: u32 = 0x11;
```
zhalvorsen: A few other tests in 1.x set this based on a feature flag:
```suggestion
const HW_REV_ID: u32 = if cfg!(feature = "hw-1.0") { 0x1 } else { 0x11 };
```

nquarton: Oh good catch. I would have been fixing this in another PR when the nightly failed otherwise.

runtime/src/drivers.rs
```diff
@@ -284,6 +285,20 @@ impl Drivers {
         Ok(())
     }
 
+    fn update_fw_version(drivers: &mut Drivers, update_fmc_ver: bool, update_rt_ver: bool) {
+        // This is a temp workaround since cptra_fw_rev_id registers are not sticky on a warm reset.
```
mhatrevi: Suggest removing this comment since the hw won't be updated for 1.x.

nquarton: Thanks. I caught that in the commit message but missed it in the code. Fixed now.

mhatrevi (approved):

PR #2616, opened 2025-10-10
[Fix] Restore FMC & RT Firmware Versions on Warm Reset

zhalvorsen:

This is a temporary workaround until the hardware update planned for version 2.2 addresses the issue more robustly.


zhalvorsen:
Dropped in favor of https://github.com/chipsalliance/caliptra-sw/pull/2615

PR #2624, opened 2025-10-14
Add additional warm reset tests.

mtimkovich:


runtime/tests/runtime_integration_tests/test_ecdsa.rs
```diff
@@ -280,3 +280,131 @@ fn test_ecdsa_hw_failure() {
...
+        0x25, 0x3d, 0x95, 0x51, 0x26, 0x68, 0x23, 0x77, 0x1b, 0x25, 0x17, 0x05, 0xb4, 0xa8, 0x51,
+        0x66, 0x02, 0x2a, 0x37, 0xac, 0x28, 0xf1, 0xbd,
+    ];
+
+    let hash = sha384(msg);
+
+    // ECDSAVS NIST test vector
+    let mut cmd = MailboxReq::EcdsaVerify(EcdsaVerifyReq {
+        hdr: MailboxReqHeader { chksum: 0 },
+        pub_key_x: [
```
zhalvorsen: Can we pull these values into separate variables so they can be used in both calls? It would be nice for quick readability to see it is the same request done over a warm reset.

runtime/tests/runtime_integration_tests/test_fe_programming.rs
```diff
@@ -79,3 +79,58 @@ fn test_fe_programming_invalid_partition() {
...
+    });
+    cmd.populate_chksum().unwrap();
+
+    let resp = model
+        .mailbox_execute(u32::from(CommandId::FE_PROG), cmd.as_bytes().unwrap())
+        .unwrap()
+        .expect("We should have received a response");
+
+    // Verify we got a successful response (should be at least header size)
+    assert!(resp.len() >= core::mem::size_of::<MailboxReqHeader>());
```
zhalvorsen: Should this be referring the response header? Same in the command below.

```Rust
    assert!(resp.len() >= core::mem::size_of::<MailboxRespHeader>());
```

runtime/tests/runtime_integration_tests/test_ecdsa.rs
```diff
@@ -280,3 +280,131 @@ fn test_ecdsa_hw_failure() {
...
+    assert_eq!(model.soc_ifc().cptra_fw_error_non_fatal().read(), 0);
+
+    // Perform warm reset
+    model.warm_reset_flow(&Fuses::default());
+
+    model.step_until(|m| {
+        m.soc_ifc().cptra_boot_status().read() == u32::from(RtBootStatus::RtReadyForCommands)
+    });
+
+    let mut cmd = MailboxReq::EcdsaVerify(EcdsaVerifyReq {
```
zhalvorsen, line 363: Can we remove this `cmd` definition and use the previous one similar to the tests below. It would be nice for quick readability to see it is the same request done over a warm reset.

zhalvorsen (dismissed):

runtime/tests/runtime_integration_tests/test_fe_programming.rs
mtimkovich: Yeah, I think you're right

PR #2634, opened 2025-10-16
Resize FPGA dev image partition on first boot

clundin25:


ci-tools/fpga-image/startup-script.sh
```diff
@@ -8,10 +8,14 @@
 # Stop spewing kernel noise to the UART
 echo 3 > /proc/sys/kernel/printk
 
+# TODO(clundin): There are a lot of hacks that get cleaned up if using initrd.
```
timothytrippel, line 11: Could you provide more details here?

clundin25: Having some connectivity issues this morning. Will update the comment in a later PR.

The issue is that the filesystem is read only when systemd's init script runs, which causes a lot of things to fail, for example NTP, partition resizing, etc.

clundin25: This would also resolve the overlayfs issue on the CI image variant

PR #2635, opened 2025-10-16
[FPGA-CI]: Cache kernel build job

clundin25:


clundin25:
Landed in https://github.com/chipsalliance/caliptra-sw/pull/2634

PR #2636, opened 2025-10-17
[hw_model] move OTP init into separate function

timothytrippel:
Caliptra Core fuses have always been initialized on boot, since Caliptra Core's "fuses" are CSRs that are populated by the host RoT complex. However, since Caliptra Subsystem's MCU ROM forwards these fuses values from the values that are backed in actual fuses, they need to be initialized before boot of the subsystem.

This consolidates all OTP memory (fuse) initialization to a single function in the subsystem HW model that is called *before* subsystem is booted (instead of after; which was the prior convention). Additionally, this updates the "init_fuses()" implementation for the subsystem model to do nothing, since for other Caliptra Core HW models this was forwarding fuse values to Caliptra Core CSRs. This makes it more clear that MCU ROM is responsible for this.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -44,14 +44,23 @@ const I3C_TARGET_MAPPING: (usize, usize) = (1, 2);
 const MCI_MAPPING: (usize, usize) = (1, 3);
 const OTP_MAPPING: (usize, usize) = (1, 4);
 
-// Offsets in the OTP for fuses.
-const FUSE_MANUF_DEBUG_UNLOCK_TOKEN_OFFSET: usize = 0x0;
-const FUSE_VENDOR_PKHASH_OFFSET: usize = 0x3f8;
-const FUSE_PQC_OFFSET: usize = FUSE_VENDOR_PKHASH_OFFSET + 48;
-const FUSE_SVN_OFFSET: usize = 0x390;
-const FUSE_SOC_MAX_SVN_OFFSET: usize = FUSE_SVN_OFFSET + 36;
-const FUSE_LIFECYCLE_TOKENS_OFFSET: usize = 0x2d8;
-const FUSE_LIFECYCLE_STATE_OFFSET: usize = 0xc80;
+// TODO(timothytrippel): autogenerate these from the OTP memory map definition
+// Offsets in the OTP for all partitions.
+// SW_TEST_UNLOCK_PARTITION
+const OTP_SW_TEST_UNLOCK_PARTITION_OFFSET: usize = 0x0;
+// SW_MANUF_PARTITION
+const OTP_SW_MANUF_PARTITION_OFFSET: usize = 0x0D0;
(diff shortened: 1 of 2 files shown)


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -44,14 +44,23 @@ const I3C_TARGET_MAPPING: (usize, usize) = (1, 2);
...
 
-// Offsets in the OTP for fuses.
-const FUSE_MANUF_DEBUG_UNLOCK_TOKEN_OFFSET: usize = 0x0;
-const FUSE_VENDOR_PKHASH_OFFSET: usize = 0x3f8;
-const FUSE_PQC_OFFSET: usize = FUSE_VENDOR_PKHASH_OFFSET + 48;
-const FUSE_SVN_OFFSET: usize = 0x390;
-const FUSE_SOC_MAX_SVN_OFFSET: usize = FUSE_SVN_OFFSET + 36;
-const FUSE_LIFECYCLE_TOKENS_OFFSET: usize = 0x2d8;
-const FUSE_LIFECYCLE_STATE_OFFSET: usize = 0xc80;
+//// TODO(timothytrippel): autogenerate these from the OTP memory map definition
```
zhalvorsen: fyi: It looks like clippy is mad these are 4 slashes instead of 3

zhalvorsen (dismissed):

hw-model/src/model_fpga_subsystem.rs
timothytrippel: fixed.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1397,45 +1488,8 @@ impl HwModel for ModelFpgaSubsystem {
...
-
-        // Compute and provision default manuf debug unlock token.
-        let mem = otp_generate_manuf_debug_unlock_token_mem(&DEFAULT_MANUF_DEBUG_UNLOCK_RAW_TOKEN)?;
-        let offset = FUSE_MANUF_DEBUG_UNLOCK_TOKEN_OFFSET;
-        otp_data[offset..offset + mem.len()].copy_from_slice(&mem);
-
-        let otp_mem = m.otp_slice();
-        otp_mem.copy_from_slice(&otp_data);
+        println!("Initializing subsystem OTP memory.");
+        m.init_otp(params.security_state.device_lifecycle(), &Fuses::default())?;
```
ArthurHeymans, line 1492: This is wrong. It should not always set the fuses to default value...

ArthurHeymans:
Can we revert this? The Bootparams.fuses need to be passed to the OTP setup which is only passed via boot().

hw-model/src/model_fpga_subsystem.rs
timothytrippel: We can make this be overrideable by an InitParam then.

timothytrippel: https://github.com/chipsalliance/caliptra-sw/pull/2660 should fix in the interim IIUC

PR #2647, opened 2025-10-21
[hw_model] expose prod debug hash number and offset straps

timothytrippel:
This exposes the prod debug hash number and offset straps as params so they can be configured by subsystem test cases that actuate the prod debug unlock mechanism.

### DIFF
--- a/.github/workflows/fpga-image.yml
+++ b/.github/workflows/fpga-image.yml
@@ -53,19 +53,31 @@ jobs:
         with: 
           ref: ${{ inputs.branch }}
 
+      - name: Cache Kernel output
+        id: cache-kernel
+        uses: actions/cache@v4
+        with:
+          path: |
+            /tmp/vck190-kernel-${{ matrix.image_variant }}.tar.gz
+            /tmp/io-module-${{ matrix.image_variant }}.ko
+          key: kernel-${{ matrix.image_variant }}-${{ matrix.bitstream_hash }}
+
       - name: Install pre-requisites
+        if: steps.cache-kernel.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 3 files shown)


hw-model/src/lib.rs
```diff
@@ -283,8 +291,12 @@ impl Default for InitParams<'_> {
             stack_info: None,
             soc_user: MailboxRequester::SocUser(1u32),
             test_sram: None,
+
+            // Only relevant for SS Model.
```
clundin25, line 295: nit: Could wrap this in `SubsystemInitParams` to make this clearer

timothytrippel: Good idea! This will be a more invasive change (will need to wrap the sub parameters in a struct and then propagate the change through the caliptra-mcu-repo), so if it is OK, let me do this in a follow up. I filed an issue to track this in the interim: https://github.com/chipsalliance/caliptra-sw/issues/2650.

zhalvorsen: Most of this is copied straight from the `InitParams` in `caliptra-sw`. Should make just wrap that structure directly and then add whatever we need on top of it?

timothytrippel: Yes, it is fairly straight forward: I need to implement Default (simple) for the new struct, and then also propagate the new InitParams structure through each implementation of the HwModel and then in mcu-sw repo where I currently have a bunch of changes stacked up to implement the prod debug unlock flow. Since it is orthogonal to the intent of this PR, I would prefer to have a specific refactor cleanup commit for a clean commit history than to tack the refactor onto an existing PR.

timothytrippel: Fixed in #2653

PR #2649, opened 2025-10-21
feat(caliptra-rom): Add more rom tests to the CI

ArthurHeymans:
Current status of ROM tests: 121 passed (1 slow), 0 failed, 40 skipped

These ones are currently failing and skipped in CI:
- test(test_debug_unlock::)
- test(test_fips_hooks::test_fips_hook_exit) needs debugging/fixing
- test(test_fmcalias_derivation::test_pcr_log) need a way to write owner_pk
- test(test_fmcalias_derivation::test_pcr_log_across_update_reset) Reset not working...
- test(test_fmcalias_derivation::test_zero_firmware_size) zero size is a problem for FPGA...
- test(test_fake_rom::test_skip_kats) flow is likely different
- test(test_fake_rom::test_fake_rom_production_enabled) No way to set register
- test(test_image_validation::test_preamble_owner_pubkey_digest_mismatch) No way to set owner key?
- test(test_image_validation::test_preamble_vendor_ecc_pubkey_revocation) Some part of fuse not set?
- test(test_image_validation::test_preamble_vendor_lms_pubkey_revocation)
- test(test_image_validation::test_preamble_vendor_mldsa_pubkey_revocation)
- test(test_image_validation::test_runtime_svn_less_than_fuse_svn) no fuse set?
- test(test_pmp::test_pmp_enforced) unsupported instruction mseccfg
- test(test_uds_programming::) No way to set request from mcu rom
- test(test_update_reset::test_update_reset_vendor_lms_pub_key_idx_dv_mismatch) No way to set key idx fuse?
- test(test_version::test_version)
- test(test_warm_reset::)
- test(test_wdt_activation_and_stoppage::) No way to set WDT


rom/dev/tests/rom_integration_tests/test_derive_stable_key.rs
```diff
@@ -94,6 +94,7 @@ fn test_derive_stable_key() {
                 InitParams {
                     rom: &rom,
                     subsystem_mode,
+                    enable_mcu_uart_log: cfg!(feature = "fpga_subsystem"),
```
zhalvorsen: We could just set this to true, because the only model that uses MCU UART log is `fpga_subsystem`

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1721,10 +1672,63 @@ impl HwModel for ModelFpgaSubsystem {
...
+        //     self.soc_ifc()
+        //         .cptra_i_trng_entropy_config_0()
+        //         .write(|_| reg);
+        // }
+
+        // TODO: support passing these into MCU ROM
+
+        // Set up the PAUSER as valid for the mailbox (using index 0)
+        // self.setup_mailbox_users(boot_params.valid_axi_user.as_slice())
+        //     .map_err(ModelError::from)?;
```
zhalvorsen: These can be left in the other function. They should be done separately as part of initialization and don't need to be involved in uploading firmware

zhalvorsen: Moved back

PR #2651, opened 2025-10-21
Update README.md

jihome-org:
Thank you for considering my change. Thanks.

### DIFF
--- a/rom/dev/README.md
+++ b/rom/dev/README.md
@@ -263,7 +263,7 @@ Initial Device ID Layer is used to generate Manufacturer CDI & Private Key.  Thi
 
     `kv_clear(KvSlot0)`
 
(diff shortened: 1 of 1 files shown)


rom/dev/README.md
```diff
@@ -263,7 +263,7 @@ Initial Device ID Layer is used to generate Manufacturer CDI & Private Key.  Thi
 
     `kv_clear(KvSlot0)`
 
-3. Derive ECC Key Pair using CDI in Key Vault Slot6 and store the generated private key in KeySlot7
+3. Derive ECC Key Pair using CDI in Key Vault Slot6 and store the generated public key in KeySlot7
```
mhatrevi, line 266: The update is incorrect. Only Private keys are stored in KV.

mhatrevi:
The update is incorrect. Only Private keys are stored in KV.

PR #2657, opened 2025-10-22
Fixing warm reset flow in test_warm_reset_version to call boot()

nquarton:
Fixing nightly release failure for test_warm_reset_version on FPGA.

Warm reset flow was not re-setting all config registers set by boot. This was not re-configuring the mailbox pausers and therefore was causing mailbox access to error out.

### DIFF
--- a/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
+++ b/rom/dev/tests/rom_integration_tests/test_warm_reset.rs
@@ -72,12 +72,7 @@ fn test_warm_reset_success() {
     }
 
     // Perform warm reset
-    hw.warm_reset_flow(&Fuses {
-        key_manifest_pk_hash: vendor_pk_hash,
-        owner_pk_hash,
-        fmc_key_manifest_svn: 0b1111111,
-        ..Default::default()
-    });
+    hw.warm_reset_flow().unwrap();
 
     // Wait for boot
     while !hw.soc_ifc().cptra_flow_status().read().ready_for_runtime() {
@@ -99,7 +94,7 @@ fn test_warm_reset_during_cold_boot_before_image_validation() {
     hw.step_until_boot_status(IDevIdDecryptUdsComplete.into(), true);
 
     // Perform a warm reset
-    hw.warm_reset_flow(&Fuses::default());
+    hw.warm_reset_flow().unwrap();
 
     // Wait for error
     while hw.soc_ifc().cptra_fw_error_fatal().read() == 0 {
@@ -135,7 +130,7 @@ fn test_warm_reset_during_cold_boot_during_image_validation() {
     }
 
     // Perform a warm reset
-    hw.warm_reset_flow(&Fuses::default());
+    hw.warm_reset_flow().unwrap();
 
     // Wait for error
     while hw.soc_ifc().cptra_fw_error_fatal().read() == 0 {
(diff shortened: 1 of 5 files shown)


mhatrevi (approved):

rom/dev/tests/rom_integration_tests/test_warm_reset.rs
```diff
@@ -338,7 +338,13 @@ fn test_warm_reset_version() {
     );
 
     // Perform warm reset
-    hw.warm_reset_flow(&fuses);
+    // Explicitly call boot() to re-set all config regs (otherwise mailbox access will not work)
+    hw.warm_reset();
+    let boot_params = BootParams {
```
zhalvorsen: I'm worried this is this a symptom of a different problem. Putting it here indicates this will either need to be done in hardware by integrators or in the firmware of the management core attached to it. Should some of this be done in firmware like the version number was?

Part of the reason this worries me is `boot` also tries to upload firmware which we shouldn't need to do during a warm reset.

nquarton: Yeah I am with you on that. I was a little surprised the [MBOX_VALID_PAUSER](https://chipsalliance.github.io/caliptra-rtl/v1_1/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_MBOX_VALID_PAUSER%5B0%5D) regs and [iTRNG config regs](https://chipsalliance.github.io/caliptra-rtl/v1_1/external-regs/?p=caliptra_top_reg.generic_and_fuse_reg.CPTRA_iTRNG_ENTROPY_CONFIG_0) both reset on rst_b instead of pwrgood.

My understanding is during a hitless update, we just reset the core without toggling rst_b so we should be okay there. But I wasn't sure if this behavior was right for other warm resets.

For the test flows, this was addressed in 2.x where the warm reset flow takes boot_params instead of just fuses. But I agree if we do that we would, at minimum, need to error out if a fw image was provided. It also requires a little refactoring in places but I can do that if we want.

nquarton: I also had to add this to the warm reset in the SW emulator because the lock was not getting cleared. It seems like we should have a lot more registers in these warm reset functions. I would think everything with a rst_b reset signal should be reset there.

zhalvorsen: Yeah, that makes sense. Do you mind, in a follow up is fine, changing the param to boot_params similar to 2.x?

nquarton: Yeah I can definitely do that. @mhatrevi are we concerned about the warm reset flows not actually clearing most registers on the SW emulator? In theory, we have the FPGA that should catch anything else. But it still seems like it's not ideal.

nquarton: I just added this as a separate commit in this PR. I can separate if desired though.

zhalvorsen: Thanks for adding this Nick!

Another concern I have, which I won't gate this PR with, is the fuses should be cached by hardware on cold boot and can't be changed over warm reset. It would be good to also add a panic if the fuses given to `warm_reset_flow` are not the defaults. So we can verify that the hardware is caching them correctly and not taking the new fuse values on warm reset.

zhalvorsen (approved):

nquarton:
After talking with Zach, I changed this implementation. Now, we just save and restore the boot regs that are not persistent across a warm reset. The warm_reset_flow function doesn't require any input anymore.

PR #2642, opened 2025-10-17
[hw-model] provision SW_MANUF OTP partition with defaults

timothytrippel:
This updates the subsystem HW model to provision the SW_MANUF OTP partition with default values, including a default prod debug unlock public key hash, to enable testing the prod debug unlock lifecycle flow on FPGA.

PR #2658, opened 2025-10-23
Re-enable subsystem FPGA runtime tests

ArthurHeymans:
An invalid newline in the CI made it never run.

This re-enables the runtime FPGA tests without the invalid newline. `test_lms_verify_invalid_lmots_type`, `test_lms_verify_cmd`, and `test_lms_verify_failure` were added to the filtered tests because they are flaky. I can get `test_lms_verify_invalid_lmots_type` to pass consistently by just adding a print statement in the command handler. We need to investigate this further.

### DIFF
--- a/.github/workflows/fpga-image.yml
+++ b/.github/workflows/fpga-image.yml
@@ -39,9 +39,9 @@ jobs:
             image_variant: subsystem-2.0
           - bitstream_name: xsa_ss2p0_141b54c5.xsa
             image_variant: subsystem-2.0
-          - bitstream_hash: 04612d7cc65cb8274c68d6fb39807f0da8ce4477cc8d38b6bba77b62cb1a4f09
+          - bitstream_hash: b2bf2ebd7a4ed829b8a08e249b0eee2716410995bbb8bfafb88053d25612c977
             image_variant: subsystem-2.1
-          - bitstream_name: xsa_ss2p1_f84723cd.xsa
+          - bitstream_name: xsa_ss2p1_4686dd0d.xsa
             image_variant: subsystem-2.1
     env:
       IMAGE_VARIANT: ${{ matrix.image_variant }}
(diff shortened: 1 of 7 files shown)


runtime/tests/runtime_integration_tests/test_fe_programming.rs
```diff
@@ -20,10 +18,7 @@ fn test_fe_programming_cmd() {
     let init_params = InitParams {
         rom: &rom,
         security_state: *SecurityState::default().set_device_lifecycle(DeviceLifecycle::Production),
-        ss_init_params: SubsystemInitParams {
-            enable_mcu_uart_log: true,
-            ..Default::default()
-        },
+        enable_mcu_uart_log: true,
```
timothytrippel: why are you also reverting #2653?

hw-model/src/keys.rs
```diff
@@ -1,114 +0,0 @@
-// Licensed under the Apache-2.0 license
```
timothytrippel, line 1: Why are you removing all of the refactoring of the OTP layout? This is needed for the debug unlock flow.

timothytrippel:
Just a question: why do you need to revert #2653, #2647 and #2642, when only #2636 is the issue?

timothytrippel:
Instead reverting a bunch of changes that are unrelated to the failure, why don't you call `m.init_otp(...)` from the `boot()` function like was originally done, i.e. right before `init_fuses()`?

ArthurHeymans:
> Just a question: why do you need to revert #2653, #2647 and #2642, when only #2636 is the issue?

It's not possible to cleanly revert just that one.

mhatrevi:
> If you would like, I can prepare the suggested change. If you revert all of these unrelated changes, it is going to delay the ROM release; FYI @mhatrevi

Let's work this out on Slack and close today.

timothytrippel:
Thanks for fixing @ArthurHeymans

PR #2660, opened 2025-10-23
[subsystem] call `init_otp()` in `boot()`

timothytrippel:
As pointed out in:
https://github.com/chipsalliance/caliptra-sw/pull/2636#discussion_r2454218959, some tests need to pass non-default values to `init_otp()`, so we move the invocation back to the `boot()` function.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1135,41 +1135,9 @@ impl ModelFpgaSubsystem {
         );
     }
 
-    fn init_otp(
-        &mut self,
-        device_lifecycle: DeviceLifecycle,
-        fuses: &Fuses,
-    ) -> Result<(), Box<dyn Error>> {
-        let mut otp_data = vec![0; OTP_SIZE];
-
-        // Users can provide data to initialize OTP a specific way. If an OTP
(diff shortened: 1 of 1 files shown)


zhalvorsen (approved):

clundin25 (dismissed):

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1580,6 +1576,11 @@ impl HwModel for ModelFpgaSubsystem {
     where
         Self: Sized,
     {
+        // TODO(timothytrippel): this should ideally be called in `new_unbooted()`, but we need a
+        // way to pass the fuses parameter to it from that function to allow tests to pass
+        // different (non-default) fuse values.
+        println!("Initializing subsystem OTP memory.");
+        self.init_otp(&boot_params.fuses)?;
```
zhalvorsen, line 1583: At this point you might as well call `init_otp` from `init_fuses`

zhalvorsen: Talked offline, we will keep it as is for now

clundin25 (approved):

mhatrevi (approved):

PR #2663, opened 2025-10-24
[keys] add ECC P384 and ML-DSA-87 key pairs for testing

timothytrippel:
This generates both ECC and ML-DSA fake key pairs for testing purposes, specifically for testing prod debug unlock flows. The instructions for re-generating the key pairs and formatting them as u32 Rust arrays is also included.

Additionally, this updates the default prod debug unlock public key hash provisioning configuration by provisioning the same public key (hashes) in all eight key slots. This will enable testing all 8 debug access levels.


hw-model/src/keys.rs
```diff
@@ -21,94 +21,128 @@ pub const DEFAULT_MANUF_DEBUG_UNLOCK_RAW_TOKEN: ManufDebugUnlockToken = ManufDeb
...
+// Default prod debug unlock ECDSA P384 private key.
+//
+// Array below was generated from the private key PEM file in `hw-model/src/keys/` using:
+// openssl ec -in hw-model/src/keys/ecc_p384_private_key.pem -text -noout | \
+//     sed '/priv:/,/pub:/!d' | \
+//     sed '1d;$d' | \
+//     tr -d '[:space:]:' | \
+//     xxd -r -p | \
+//     xxd -e -c 4 | \
+//     awk '{print "0x" $2 ","}'
```
clundin25, line 33: IMO a script written in Rust would be easier to maintain.

Here is an existing example: https://github.com/chipsalliance/caliptra-sw/tree/main-2.x/drivers/test-fw/scripts/vector_gen

hw-model/src/keys/README.md
```diff
@@ -0,0 +1,16 @@
+# Lifecycle Keys
```
clundin25, line 1: Similar comment here, I think a script written in Rust will be easier to use in the future.

hw-model/src/keys.rs
timothytrippel: Do you mean to avoid using openssl for keygen?

I think it is imperative that test vectors be generated with standard tools instead of home-grown scripts to verify the correctness of our home grown HW and SW. Think of it as a sanity check done by a third party.

If openssl is really what is not desired we should review that independently.

Or am I misunderstanding what you mean here?

clundin25: I mean porting this bash script to a small Rust binary. In the example I linked it uses OpenSSL to construct various pieces of test data. 

It's easier to maintain since it's just Rust code

timothytrippel: Ah got it; agreed; followed up offline and there are actually already fake keys in the repo we can use:  https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/image/fake-keys/src/lib.rs
(Thanks @zhalvorsen  and @clundin25 .) Let me get the downstream test working and then refactor it to use the keys already available.

timothytrippel:
Closing this as other fake keys already exist in the repo that are being used.

PR #2668, opened 2025-10-24
Modify buffers

clundin25:


drivers/src/sha1.rs
```diff
@@ -216,6 +217,12 @@ pub struct Sha1DigestOp<'a> {
     data_size: usize,
 }
 
+impl<'a> Drop for Sha1DigestOp<'a> {
```
zhalvorsen: Can we use `#[derive(Zeroize)]` on these structs directly?

zhalvorsen (dismissed):

PR #2669, opened 2025-10-27
[model,util] fix bugs in the HW model to enable prod debug unlock test cases

timothytrippel:
This contains several bug fixes / adjustments to enable the prod debug unlock FPGA test cases, including:
1. publically expose `init_otp` method in the Subsystem model,
2. fixing a bug in the OTP SW_MANUF partition's HSM ID field size,
3. adding a util to compute mailbox checksums over a u32 array (instead of just a u8 array), and
4. provisioning the default prod debug unlock key hash in all slots, thus enabling testing all 8 debug access levels by default.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -2,7 +2,7 @@
 #![cfg_attr(not(test), no_std)]
 
 mod capabilities;
-mod checksum;
+pub mod checksum;
 pub mod mailbox;
 pub mod soc_mgr;
 


api/src/mailbox.rs
```diff
@@ -2010,8 +2010,8 @@ impl Request for ProductionAuthDebugUnlockReq {
 }
 
 // PRODUCTION_AUTH_DEBUG_UNLOCK_CHALLENGE
+#[derive(Debug, FromBytes, Immutable, IntoBytes, KnownLayout, PartialEq, Eq, Clone, Copy)]
```
zhalvorsen: Any reason, you can't use `.clone()` in your scenario? I'm pretty sure to main reason `Copy` wasn't added originally was to avoid unintended copies from getting scattered around the firmware

api/src/checksum.rs
```diff
@@ -18,6 +18,22 @@ pub fn calc_checksum(cmd: u32, data: &[u8]) -> u32 {
...
+    for c in cmd.to_le_bytes().iter() {
+        checksum = checksum.wrapping_add(*c as u32);
+    }
+    for dw in data {
+        for d in dw.to_le_bytes().iter() {
+            checksum = checksum.wrapping_add(*d as u32);
+        }
+    }
+    0u32.wrapping_sub(checksum)
+}
```
clundin25: I think Zerocopy could be used to re-use code here:


```suggestion
pub fn calc_words_checksum(cmd: u32, data: &[u32]) -> u32 {
    calc_checksum(cmd, data.as_bytes())
}
```

api/src/mailbox.rs
timothytrippel: fixed.

api/src/checksum.rs
timothytrippel: fixed.

Issue #2583, opened 2025-10-03
Add ability to only verify the integrity and authenticity of a provided Caliptra run-time image without triggering its execution


christherene:
The current FW_LOAD (FWLD) Caliptra MB command support atomically the authentication and loading of a new Caliptra RT image. Requesting a new Caliptra MB command to only verify the integrity and authenticity of a provided Caliptra run-time image without triggering its execution, to support hitless firmware update (HFU) flows using PLDM Type 5 stages (transfer, verify, apply, activate). 

Essentially, this new MB command would be used during the Verify stage when a PLDM full package (Caliptra RT, SoC Manifest, MCU RT, SoC FW images) is being verified, before being applied and later activated. FW_LOAD would still be used during the Activation phase.

Caliptra Version: 2.1
Caliptra Mode: Subsystem Mode/Core Mode

RaunakGu:
Thanks for raising it, 

I do see need for the same for meeting NVMe FW Commit requirements as mentioned in below issue, 

adding for ref: https://github.com/chipsalliance/caliptra-mcu-sw/issues/490

zhalvorsen:
Fixed in #2671

PR #2671, opened 2025-10-27
Add FIRMWARE_VERIFY mailbox command

mlvisaya:
Add the FIRMWARE_VERIFY mailbox command that allows user to verify a Caliptra FW image bundle without activating it. This command is used during hitless update of the subsystem.

Integration test added in the caliptra-mcu-sw repo https://github.com/chipsalliance/caliptra-mcu-sw/pull/518


runtime/src/firmware_verify.rs
```diff
@@ -0,0 +1,88 @@
...
+                resp.verify_result = FirmwareVerifyResult::Failure as u32;
+            }
+        }
+
+        populate_checksum(resp.as_mut_bytes());
+        // Send the payload
+        drivers.mbox.write_response(resp.as_bytes())?;
+        // zero the original resp buffer so as not to leak sensitive data
+        resp.as_mut_bytes().fill(0);
+
```
zhalvorsen, line 76: Can we also zeroize the `manifest2` from persistent data because it is used to verify hitless updates?

zhalvorsen (dismissed):

zhalvorsen:
Fixes #2583

runtime/src/firmware_verify.rs
mlvisaya: nice catch! yes manifest2 should be zeroize too. i'll make the change

zhalvorsen (dismissed):

api/src/mailbox.rs
```diff
@@ -49,6 +49,7 @@ pub enum AlgorithmType {
 impl CommandId {
     pub const ACTIVATE_FIRMWARE: Self = Self(0x41435446); // "ACTF"
     pub const FIRMWARE_LOAD: Self = Self(0x46574C44); // "FWLD"
+    pub const FIRMWARE_VERIFY: Self = Self(0x46575652); // "FWVR"
```
mhatrevi, line 52: Please add integration test(s) to validate the change.

mhatrevi (changes requested):

api/src/mailbox.rs
mlvisaya: Added integration test in the latest commit.

zhalvorsen (approved):

mhatrevi (approved):

PR #2673, opened 2025-10-28
Add x509 template for P-384 ECDH OCP LOCK HPKE Endorsement

clundin25:
This will be used to endorse HPKE public keys used to wrap access keys in OCP LOCK.

### DIFF
--- a/x509/build/build.rs
+++ b/x509/build/build.rs
@@ -251,4 +251,32 @@ fn gen_ocp_lock_endorsement_cert(out_dir: &str) {
         "Caliptra 2.0 MlDsa87 Rt Alias",
     );
     CodeGen::gen_code("OcpLockMlKemCertTbsMlDsa87", template, out_dir);
+
+    let bldr = cert::CertTemplateBuilder::<EcdsaSha384Algo, EcdsaSha384Algo>::new()
+        .add_basic_constraints_ext(false, 0)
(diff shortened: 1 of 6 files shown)


x509/build/build.rs
```diff
@@ -251,4 +251,32 @@ fn gen_ocp_lock_endorsement_cert(out_dir: &str) {
...
+        .add_basic_constraints_ext(false, 0)
+        .add_key_usage_ext(usage)
+        .add_hpke_identifiers_ext(&HPKEIdentifiers::new(
+            HPKEIdentifiers::ECDH_P384_IANA_CODE_POINT,
+            HPKEIdentifiers::HKDF_SHA384_IANA_CODE_POINT,
+            HPKEIdentifiers::AES_256_GCM_IANA_CODE_POINT,
+        ));
+    let template = bldr.tbs_template(
+        "OCP LOCK HPKE Endorsement ECDH P-384",
+        "Caliptra 2.0 Ecc384 Rt Alias",
```
zhalvorsen, line 265: I realize this is the same with all of the previous templatized certs, but should these say `2.x` so it isn't confusing? Especially for these certs that will never exist in 2.0

clundin25: Good call, opened https://github.com/chipsalliance/caliptra-sw/issues/2680 to track

PR #2678, opened 2025-10-29
Add Warm reset test for OCP LOCK HW flow

clundin25:
TODO: Build full OCP LOCK key ladder in these tests, to exercise all KV slots.

### DIFF
--- a/drivers/tests/drivers_integration_tests/main.rs
+++ b/drivers/tests/drivers_integration_tests/main.rs
@@ -2,11 +2,12 @@
 
 use caliptra_api::SocManager;
 use std::error::Error;
-use std::iter;
+use std::time::Duration;
+use std::{iter, thread};
 
 use caliptra_builder::{firmware, FwId};
 use caliptra_drivers::{Array4x12, Array4xN, Ecc384PubKey};
-use caliptra_drivers_test_bin::DoeTestResults;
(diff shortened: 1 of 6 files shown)


drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -39,15 +40,27 @@ fn start_driver_test(test_rom: &'static FwId) -> Result<DefaultHwModel, Box<dyn
...
 }
 
 fn run_driver_test(test_rom: &'static FwId) {
+    run_driver_test_ext(test_rom, false);
+}
+
+fn run_driver_test_ext(test_rom: &'static FwId, perform_warm_reset: bool) {
     let mut model = start_driver_test(test_rom).unwrap();
     // Wrap in a line-writer so output from different test threads doesn't multiplex within a line.
+    if perform_warm_reset {
```
zhalvorsen: Will other tests also use this flag? It feels too specific to this test case. If not, could we just put this logic directly in the `test_ocp_lock_warm_reset` test?

clundin25: That's fair, I walked back the common code changes, and wrote a larger harness instead.

clundin25:
Do you mind taking a look @swenson?

PR #2679, opened 2025-10-29
Add firwmare flows for reduced mailbox

swenson:
* ROM loads firmware bundle into MCU SRAM and validates from there.
* Runtime loads auth manifest into MCU SRAM and validates from there.
* Add ability to run mailbox commands that are in an external AXI address (via DMA).
* Disable some failing tests for now
* Update reset tests need to be updated to use external mailbox commands
* testfmc still needs to account for external mailbox
* FIPS self test commands are disabled for now
* Disable some KATs for space (since ROM is full)

Replaces #2338.

Issue #2575, opened 2025-10-02
Caliptra Core 2.1 ROM Checklist


mhatrevi:

nquarton:
One other change we'll want to make sure is included in 2.1: https://github.com/chipsalliance/caliptra-sw/issues/2282

zhalvorsen:
#2337 should also be included

PR #2681, opened 2025-10-30
kats: Split KATs into initial ROM, initial runtime, all

swenson:
KATs are split into 3:
* Initial KATs in the ROM (for capabilities ROM uses)
* Initial KATs in the runtime (for capabilities that only runtime uses)
* All KATs (used by FIPS self test command)

This helps save space in the ROM and speed up ROM boot times.

We disable some KATs for now to save some space for upcoming changes (see #2679).

We'll re-enable these in future PRs as we get some space savings.

### DIFF
--- a/kat/src/lib.rs
+++ b/kat/src/lib.rs
@@ -41,6 +41,7 @@ pub use aes256ecb_kat::Aes256EcbKat;
 pub use aes256gcm_kat::Aes256GcmKat;
 pub use caliptra_drivers::{CaliptraError, CaliptraResult};
 pub use cmackdf_kat::CmacKdfKat;
+
 pub use ecc384_kat::Ecc384Kat;
 pub use ecdh_kat::EcdhKat;
 pub use hkdf_kat::{Hkdf384Kat, Hkdf512Kat};
@@ -57,14 +58,33 @@ pub use sha512_kat::Sha512Kat;
 
 use caliptra_drivers::cprintln;
 
-/// Execute Known Answer Tests
+/// Execute initial Known Answer Tests for the current environment (ROM or runtime).
 ///
 /// # Arguments
 ///
 /// * `env` - ROM Environment
-pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<()> {
+pub fn execute_initial_kats(env: &mut KatsEnv) -> CaliptraResult<()> {
     cprintln!("[kat] ++");
+    if cfg!(feature = "rom") {
+        execute_rom_kats(env)?;
+    } else {
+        execute_runtime_only_kats(env)?;
+    }
+    cprintln!("[kat] --");
+    Ok(())
+}
 
+/// Execute all Known Answer Tests unconditionally.
+pub fn execute_all_kats(env: &mut KatsEnv) -> CaliptraResult<()> {
+    cprintln!("[kat] ++");
+    execute_rom_kats(env)?;
+    execute_runtime_only_kats(env)?;
+    cprintln!("[kat] --");
+    Ok(())
+}
+
+/// Execute Known Answer Tests for capabilities ROM uses.
+fn execute_rom_kats(env: &mut KatsEnv) -> CaliptraResult<()> {
     cprintln!("[kat] sha1");
     Sha1Kat::default().execute(env.sha1)?;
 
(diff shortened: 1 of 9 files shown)


clundin25 (dismissed):

kat/src/lib.rs
```diff
@@ -80,52 +100,79 @@ pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<()> {
...
 
-    cprintln!("[kat] AES-256-CTR");
-    Aes256CtrKat::default().execute(env.aes)?;
+    // cprintln!("[kat] AES-256-CTR");
+    // Aes256CtrKat::default().execute(env.aes)?;
 
-    cprintln!("[kat] AES-256-GCM");
-    Aes256GcmKat::default().execute(env.aes, env.trng)?;
+    // cprintln!("[kat] AES-256-GCM");
+    // Aes256GcmKat::default().execute(env.aes, env.trng)?;
```
zhalvorsen, line 141: Looks like a few of these are duplicated:
```suggestion
    LmsKat::default().execute(env.sha256, env.lms)?;

    // cprintln!("[kat] HKDF-512");
    // Hkdf512Kat::default().execute(env.hmac, env.trng)?;

    // cprintln!("[kat] KDF-CMAC");
    // CmacKdfKat::default().execute(env.aes)?;

    // cprintln!("[kat] MLDSA87");
    // Mldsa87Kat::default().execute(env.mldsa87, env.trng)?;

    // cprintln!("[kat] AES-256-ECB");
    // Aes256EcbKat::default().execute(env.aes)?;

    // cprintln!("[kat] AES-256-CBC");
    // Aes256CbcKat::default().execute(env.aes)?;

    // cprintln!("[kat] AES-256-CMAC");
    // Aes256CmacKat::default().execute(env.aes)?;

    // cprintln!("[kat] AES-256-CTR");
    // Aes256CtrKat::default().execute(env.aes)?;

    // cprintln!("[kat] AES-256-GCM");
    // Aes256GcmKat::default().execute(env.aes, env.trng)?;
```

rom/dev/README.md
```diff
@@ -368,10 +368,22 @@ ROM performs the following POST tests to ensure that needed cryptographic module
  - SHA1
  - SHA2-256
  - SHA2-384
- - SHA2-384-ACC
+ - SHA2-512
+ - SHA2-512-ACC
```
zhalvorsen, line 372:
```suggestion
 - SHA2-512-ACC
 - SHAKE-256
```

rom/dev/README.md
```diff
@@ -368,10 +368,22 @@ ROM performs the following POST tests to ensure that needed cryptographic module
...
  - ECC-384
- - HMAC-384Kdf
+ - ECDH
```
zhalvorsen:
```suggestion
 - ECDH (conditionally, if not ROM feature)
```

zhalvorsen (dismissed):

kat/src/lib.rs
swenson: Cleaned up.

rom/dev/README.md
```diff
@@ -368,10 +368,23 @@ ROM performs the following POST tests to ensure that needed cryptographic module
  - SHA1
  - SHA2-256
  - SHA2-384
- - SHA2-384-ACC
+ - SHA2-512
+ - SHA2-512-ACC
+ - SHAKE-256
```
zhalvorsen: Sorry, I forgot this piece in my suggestion
```suggestion
 - SHAKE-256 (conditionally, if not ROM feature)
```

zhalvorsen (approved):

clundin25 (approved):

nquarton (approved):
Looks good to me. (Might make sense to add re-enabling all KATs to a 2.1 checklist if it isn't already.)

swenson:
> re-enabling all KATs

Done https://github.com/chipsalliance/caliptra-sw/issues/2575

swenson:
As I was discussing this with @mhatrevi, we realized that the ROM still has the FIPS self-test command. So, we won't really be able to remove the KATs from ROM I don't think?

If that sounds right, I can go ahead and close this.

If we absolutely need ROM space temporarily, we can disable CFI, but maybe I can find some easy wins before that.

nquarton:
> As I was discussing this with @mhatrevi, we realized that the ROM still has the FIPS self-test command. So, we won't really be able to remove the KATs from ROM I don't think?
> 
> If that sounds right, I can go ahead and close this.
> 
> If we absolutely need ROM space temporarily, we can disable CFI, but maybe I can find some easy wins before that.

I think the ROM self test command only needs to be able to run the KATs for the algorithms ROM uses. So I think this still may be okay. Let me run it by John.

nquarton:
> > As I was discussing this with @mhatrevi, we realized that the ROM still has the FIPS self-test command. So, we won't really be able to remove the KATs from ROM I don't think?
> > If that sounds right, I can go ahead and close this.
> > If we absolutely need ROM space temporarily, we can disable CFI, but maybe I can find some easy wins before that.
> 
> I think the ROM self test command only needs to be able to run the KATs for the algorithms ROM uses. So I think this still may be okay. Let me run it by John.

After taking with John, I take this back. It appears it is the case that all of the self-tests need to be called by the on-demand self testing even if ROM doesn't use them as you suggested.

swenson:
@nquarton thanks for confirming. I'll close this since we won't be able to separate KATs into ROM and runtime. If I need to, I'll open a separate PR to disable things for space.

PR #2682, opened 2025-10-30
[feat] Add LDEV ECC and MLDSA cert retrieval commands to ROM

mhatrevi:
This change adds the GET_LDEV_ECC384_CERT and GET_LDEV_MLDSA87_CERT mailbox command support to ROM. These certificates are needed in manufacturing flows where Runtime availability is not always possible.

### DIFF
--- a/common/src/dice.rs
+++ b/common/src/dice.rs
@@ -8,11 +8,181 @@ File Name:
 
 Abstract:
 
-    DICE-related constants.
+    DICE-related
 
 --*/
+
+use caliptra_api::mailbox::{AlgorithmType, GetLdevCertResp, MailboxRespHeader, ResponseVarSize};
+use caliptra_drivers::{
+    CaliptraError, CaliptraResult, Ecc384Signature, Mldsa87Signature, PersistentData,
+};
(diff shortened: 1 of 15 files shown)


common/src/dice.rs
```diff
@@ -8,11 +8,181 @@ File Name:
...
+///
+/// * `usize` - The number of bytes written to `cert`
+#[inline(never)]
+pub fn copy_ldevid_mldsa87_cert(
+    persistent_data: &PersistentData,
+    cert: &mut [u8],
+) -> CaliptraResult<usize> {
+    let tbs = persistent_data
+        .mldsa_ldevid_tbs
+        .get(..persistent_data.fht.mldsa_ldevid_tbs_size.into());
```
swenson, line 184: nit: I know this is just being moved, but the spacing here is quite weird. Is there any way we could fix this?

swenson (commented):
I think we'll also want to update the ROM documentation to specify these commands are available.

Code LGTM.

common/src/dice.rs
swenson: Nevermind, it was just the way the PR diff looked.

swenson (approved):
LGTM

PR #2683, opened 2025-10-30
[TEST] Cert verify warmreset 2.1

meilu-git:
Cert verify warmreset 2.1


meilu-git:
close it as we need to retarget and also merge a few PR into one

PR #2686, opened 2025-10-30
rom: Add subsystem-only ROM

### LINKED ISSUE #2699: rom: Add subsystem-only ROM
# Please include the following information

* Caliptra Version- 2.1
* Caliptra Mode - Both

swenson:
Split off from #2679

This will allow us to save space in the ROM since we won't have to support passive-only flows in subsystem mode and vice versa.

The consequence is that we have to plumb the correct ROM into the correct test now, which is a bit of a chore.

This also switches to using the FPGA ROM when doing subsystem tests, which required a small fix to the emulated DMA controller. This fixes many of the test failures that were in #2679.

### DIFF
--- a/drivers/src/soc_ifc.rs
+++ b/drivers/src/soc_ifc.rs
@@ -547,11 +547,15 @@ impl SocIfc {
     }
 
     pub fn subsystem_mode(&self) -> bool {
-        self.soc_ifc
-            .regs()
-            .cptra_hw_config()
-            .read()
-            .subsystem_mode_en()
+        // Only for ROM it makes sense to hardcode this
+        // For Runtime we want runtime images to work on both caliptra-code and subsystem
+        cfg!(feature = "subsystem")
+            || self
+                .soc_ifc
+                .regs()
+                .cptra_hw_config()
+                .read()
+                .subsystem_mode_en()
     }
 
     pub fn ocp_lock_enabled(&self) -> bool {
--- a/drivers/src/soc_ifc.rs
+++ b/drivers/src/soc_ifc.rs
@@ -548,7 +548,7 @@ impl SocIfc {
 
     pub fn subsystem_mode(&self) -> bool {
         // Only for ROM it makes sense to hardcode this
-        // For Runtime we want runtime images to work on both caliptra-code and subsystem
+        // For Runtime we want runtime images to work on both caliptra-core and subsystem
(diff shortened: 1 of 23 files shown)


clundin25 (dismissed):

drivers/src/soc_ifc.rs
```diff
@@ -547,11 +547,15 @@ impl SocIfc {
     }
 
     pub fn subsystem_mode(&self) -> bool {
-        self.soc_ifc
-            .regs()
-            .cptra_hw_config()
-            .read()
-            .subsystem_mode_en()
+        // Only for ROM it makes sense to hardcode this
+        // For Runtime we want runtime images to work on both caliptra-code and subsystem
```
zhalvorsen: nit: typo 
```suggestion
        // For Runtime we want runtime images to work on both caliptra-core and subsystem
```

zhalvorsen (dismissed):

rom/dev/tests/rom_integration_tests/helpers.rs
```diff
@@ -116,12 +116,12 @@ pub fn change_dword_endianess(data: &mut [u8]) {
     }
 }
 
-pub fn rom_fw_id(subsystem: bool) -> &'static FwId<'static> {
-    if subsystem {
-        firmware::ss_rom_from_env()
-    } else {
-        firmware::rom_from_env()
-    }
+pub fn model_supports_subsystem_mode(subsystem_mode: bool) -> bool {
```
zhalvorsen: nit: I'm worried the name of this function could be misread. When I first read it I read "Does the model support running in subsystem mode?" so `((!subsystem_mode) && fpga_realtime)` seemed incorrect. Maybe a better name would be `model_supports_subsystem_config`,  `*_setting`, or something similar

swenson: Cool. I can rename it. Getting the boolean logic to be readable and naming is the hardest part. :)

zhalvorsen: Totally agree!

zhalvorsen (dismissed):

clundin25 (dismissed):

swenson:
Closing this for now as I think this does not improve space all that much and adds a bigger test burden.

PR #2690, opened 2025-10-31
runtime: Add external mailbox command

### LINKED ISSUE #2357: Option/Experiment for executing large mailbox commands in 2.1
We have a new command, `EXTERNAL_MAILBOX_CMD`

It takes parameters:

* 32-bit command ID
* 64-bit AXI address start
* 32-bit command size

And then the procedure would be:

* Do some checks like if the command ID is known and that the size is no larger than the largest mailbox command we support.
* Schedule a DMA from the AXI address into the AHB FIFO, which we copy to a buffer in DCCM.
* Verify the checksum of the buffer.
* Execute the normal mailbox command and get the response buffer.
* Schedule a DMA from the AHB FIFO to the AXI address, which we fill from the response buffer.
* Complete the EXTERNAL_MAILBOX_CMD command.

This should have a minimal impact on runtime, as we would have to do little if any modifications to the existing commands, and give us a lot of flexibility to use this for additional commands as necessary (instead of creating new commands for 2.1 for each large command).

swenson:
This adds the ability to run mailbox commands from arbitrary AXI addresses by DMAing them into DCCM and then executing.

This is necessary to support SET_AUTH_MANIFEST and FIRMWARE_LOAD (update reset) with a reduced mailbox size. It could potentially also increase performance since the command is executed from DCCM instead of SRAM over AHB.

Fixes #2357

Split off and extended from #2679

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -294,6 +294,7 @@ jobs:
               -E 'package(caliptra-hw-model) and test(tests::test_execution)'
               -E 'package(caliptra-drivers) and test(test_dma_aes)'
               -E 'package(caliptra-drivers) and test(test_ocp_lock)'
+              -E 'package(caliptra-drivers) and test(test_ocp_lock_warm_reset)'
               -E 'package(caliptra-rom) and test(test_capabilities::test_capabilities)'
               -E 'package(caliptra-rom) and test(test_ocp_lock::test_hek_seed_states)'
               -E 'package(caliptra-rom) and test(test_ocp_lock::test_invalid_hek_seed_state)'
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -384,6 +384,12 @@ pub mod driver_tests {
         ..BASE_FWID
     };
 
+    pub const OCP_LOCK_WARM_RESET: FwId = FwId {
+        bin_name: "ocp_lock_warm_reset",
+        features: &["fpga_realtime"],
+        ..BASE_FWID
+    };
+
     pub const DMA_AES: FwId = FwId {
         bin_name: "dma_aes",
         features: &["emu", "fpga_subsystem"],
@@ -550,6 +556,7 @@ pub const REGISTERED_FW: &[&FwId] = &[
     &driver_tests::DMA_SHA384,
     &driver_tests::DMA_SHA384_FPGA,
     &driver_tests::OCP_LOCK,
+    &driver_tests::OCP_LOCK_WARM_RESET,
     &driver_tests::DMA_AES,
     &driver_tests::AXI_BYPASS,
     &rom_tests::ASM_TESTS,
--- a/drivers/test-fw/Cargo.toml
+++ b/drivers/test-fw/Cargo.toml
@@ -217,6 +217,11 @@ path = "src/bin/ocp_lock_tests.rs"
 required-features = ["riscv"]
 
 [[bin]]
+name = "ocp_lock_warm_reset"
+path = "src/bin/ocp_lock_warm_reset_test.rs"
+required-features = ["riscv"]
+
+[[bin]]
 name = "dma_aes"
 path = "src/bin/dma_aes_tests.rs"
 required-features = ["riscv"]
--- a/drivers/test-fw/src/bin/ocp_lock_warm_reset_test.rs
+++ b/drivers/test-fw/src/bin/ocp_lock_warm_reset_test.rs
@@ -1,0 +1,120 @@
+/*++
+Licensed under the Apache-2.0 license.
+
+File Name:
+
+    ocp_lock_warm_reset_test.rs
+
+Abstract:
+
+    File contains test cases for OCP LOCK after a warm reset.
+
+--*/
+
+#![no_std]
+#![no_main]
+
+use caliptra_cfi_lib::CfiCounter;
+use caliptra_drivers::{
+    hmac_kdf, HmacKey, HmacMode, HmacTag, KeyId, KeyReadArgs, KeyUsage, KeyWriteArgs, ResetReason,
+};
+use caliptra_drivers_test_bin::{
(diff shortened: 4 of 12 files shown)


runtime/src/lib.rs
```diff
@@ -177,6 +178,14 @@ fn enter_idle(drivers: &mut Drivers) {
     caliptra_cpu::csr::mpmc_halt_and_enable_interrupts();
 }
 
+fn human_readable_command(bytes: &[u8]) -> Option<&str> {
+    if bytes.len() != 4 || bytes.iter().any(|c| !c.is_ascii_alphanumeric()) {
+        None
+    } else {
+        core::str::from_utf8(bytes).ok()
```
clundin25: This will add about ~900 bytes to the code size, we still have a lot of budget left but we may want to consider removing this if we start running into the limit.

clundin25: (source: https://github.com/chipsalliance/caliptra-sw/commit/f17528a1e7f0f001500c69208b3797261a09f47b) 
Can be found in the size history tool, number is from main

runtime/src/populate_idev.rs
```diff
@@ -14,28 +14,24 @@ Abstract:
...
         if cmd_args.len() > core::mem::size_of::<PopulateIdevEcc384CertReq>() {
             return Err(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY);
         }
 
+        let mut cmd = PopulateIdevEcc384CertReq::default();
+        if cmd_args.len() > cmd.as_bytes().len() {
+            return Err(CaliptraError::RUNTIME_MAILBOX_API_REQUEST_DATA_LEN_TOO_LARGE);
+        }
+        cmd.as_mut_bytes()[..cmd_args.len()].copy_from_slice(cmd_args);
+
```
clundin25, line 34: What prevents zerocopy here?

clundin25 (approved):

runtime/src/lib.rs
jhand2: Can probably implement this in a lot less code with a simple ascii table and only supporting `[a-zA-Z0-9]`

runtime/README.md
```diff
@@ -2402,6 +2406,36 @@ The `exported_cdi` can be created by calling `DeriveContext` with the `export-cd
...
+### EXTERNAL_MAILBOX_CMD
+
+Command Code: `0x4558_544D` ("EXTM")
+
+**Note**: This command is only available in subsystem mode in 2.1+.
+
+Executes a mailbox command located at an AXI address.
+This allows for executing mailbox commands that are larger than the mailbox allows.
+
+This is currently mostly useful for FIRMWARE_LOAD (as part of an update) or SET_AUTH_MANIFEST.
```
jhand2, line 2418: nit: I wonder if we can make this simpler by just adding a FIRMWARE_LOAD_EXT command. Or do you have other future use-cases in mind?

runtime/src/lib.rs
```diff
@@ -229,10 +229,100 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         );
     }
 
+    let mut external_cmd_buffer =
```
jhand2, line 233: Can you refactor this into a helper function? Otherwise the main command handler function is getting pretty messy.

runtime/src/lib.rs
```diff
@@ -229,10 +229,100 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+
+        if !caliptra_common::checksum::verify_checksum(
+            req_hdr.chksum,
+            cmd_id,
+            &cmd_bytes[core::mem::size_of_val(&req_hdr.chksum)..],
+        ) {
+            return Err(CaliptraError::RUNTIME_INVALID_CHECKSUM);
+        }
+
+        if let Some(ascii) = human_readable_command(&cmd_id.to_be_bytes()) {
```
jhand2: nit: Mild preference to do this in a different PR. Seems unrelated.

runtime/src/populate_idev.rs
swenson: The buffer is smaller than the actual type. So the only safe thing to do (AFAIK) is to copy the bytes into a big enough buffer and then zerocopy.

runtime/README.md
swenson: We also need it for SET_AUTH_MANIFEST. Interestingly enough, it also might be more performant to use external commands since they are DMA'd into DCCM.

runtime/src/lib.rs
swenson: The human-readable check was already in the code -- but since we don't want to duplicate this code block, I think it is okay to leave in?

runtime/src/lib.rs
swenson: Strange. I'll tweak it and see if I can get the code size down.

swenson: Switching to `core::str::from_utf8_unchecked` brought it down by 900 bytes or so. Changing to check `[a-zA-Z0-9]` instead of `is_ascii_alphanumeric` had no effect (which makes a little bit of sense).

runtime/src/populate_idev.rs
clundin25: This is sidetracking, but I believe if the type was a DST it would work. The tradeoff would be defining these types twice but it would translate to some real stack savings, given the size of the PQC certificates

runtime/src/lib.rs
```diff
@@ -179,10 +179,11 @@ fn enter_idle(drivers: &mut Drivers) {
 }
 
 fn human_readable_command(bytes: &[u8]) -> Option<&str> {
-    if bytes.len() != 4 || bytes.iter().any(|c| !c.is_ascii_alphanumeric()) {
-        None
+    if bytes.len() == 4 || bytes.iter().all(|c| c.is_ascii_alphanumeric()) {
```
clundin25: Should `||` be `&&`?

swenson: Thanks! Forgot to invert that when I inverted the rest.

clundin25 (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #2700, opened 2025-11-05
emulator: pin to Tock registers 2.2

swenson:
This will match the version of Tock that MCU uses so that MCU can move off nightly. (See https://github.com/chipsalliance/caliptra-mcu-sw/pull/543)


swenson:
Thanks!

PR #2709, opened 2025-11-06
Download Caliptra firmware to MCU SRAM in subsystem mode.

swenson:
Rather than into the mailbox, and support verifying and loading firmware from MCU SRAM using DMA.

We have had to disable a few KATs temporarily for ROM space.

This fixes a couple of bugs around DMA, the SHA accelerator, and endianness to match hardware more closely.

Split off and extended from #2679

(This doesn't actually require the mailbox to be reduced yet, and there are still PRs coming for being flexible in the mailbox size and working with update reset.)

Apologies for the length of this, but hopefully this is at least digestable and reviewable.

### DIFF
--- a/sw-emulator/app/src/main.rs
+++ b/sw-emulator/app/src/main.rs
@@ -13,14 +13,16 @@ Abstract:
 --*/
 
 use caliptra_api_types::{DeviceLifecycle, SecurityState};
-use caliptra_emu_bus::Clock;
+use caliptra_emu_bus::{Clock, Device, Event, EventData};
 use caliptra_emu_cpu::{Cpu, CpuArgs, Pic, RvInstr, StepAction};
+use caliptra_emu_periph::dma::recovery::RecoveryControl;
 use caliptra_emu_periph::soc_reg::DebugManufService;
 use caliptra_emu_periph::{
     CaliptraRootBus, CaliptraRootBusArgs, DownloadIdevidCsrCb, MailboxInternal, MailboxRequester,
     ReadyForFwCb, TbServicesCb, UploadUpdateFwCb,
 };
 use caliptra_hw_model::BusMmio;
+use caliptra_registers::i3ccsr::regs::DeviceStatus0ReadVal;
 use clap::{arg, value_parser, ArgAction};
 use std::fs::File;
 use std::io;
@@ -28,6 +30,7 @@ use std::io::{Read, Write};
 use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::rc::Rc;
+use std::sync::mpsc;
 use tock_registers::interfaces::{ReadWriteable, Readable, Writeable};
 use tock_registers::registers::InMemoryRegister;
 mod gdb;
@@ -48,7 +51,13 @@ const FW_WRITE_TICKS: u64 = 1000;
 const EXPECTED_CALIPTRA_BOOT_TIME_IN_CYCLES: u64 = 20_000_000; // 20 million cycles
 
 // CPU Main Loop (free_run no GDB)
-fn free_run(mut cpu: Cpu<CaliptraRootBus>, trace_path: Option<PathBuf>) {
+fn free_run(
+    mut cpu: Cpu<CaliptraRootBus>,
+    trace_path: Option<PathBuf>,
+    events_to_caliptra: mpsc::Sender<Event>,
+    events_from_caliptra: mpsc::Receiver<Event>,
+) {
+    let mut collected_events_from_caliptra = Vec::new();
     if let Some(path) = trace_path {
         let mut f = File::create(path).unwrap();
         let trace_fn: &mut dyn FnMut(u32, RvInstr) = &mut |pc, instr| {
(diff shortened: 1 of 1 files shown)


rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -19,6 +19,7 @@ use crate::key_ladder;
 use crate::pcr;
 use crate::rom_env::RomEnv;
 use crate::run_fips_tests;
+//use crate::run_fips_tests;
```
zhalvorsen:
```suggestion
```

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -679,6 +712,46 @@ impl FirmwareProcessor {
...
+        let caliptra_base_addr: AxiAddr = soc_ifc.caliptra_base_axi_addr().into();
+
+        // Read manifest from MCU SRAM using DMA directly into manifest buffer
+        let manifest_size_words = manifest_buf.len().div_ceil(4);
+        let manifest_words = unsafe {
+            core::slice::from_raw_parts_mut(
+                manifest_buf.as_mut_ptr() as *mut u32,
+                manifest_size_words,
+            )
+        };
```
zhalvorsen: You should be able to avoid the unsafe here using zerocopy (FYI I didn't try compiling this):
```suggestion
        let (manifest_words, remainder) = <[u32]>::mut_from_prefix(&manifest_buf);
```

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -916,6 +1015,75 @@ impl FirmwareProcessor {
...
+
+        // Load FMC from MCU SRAM
+        let fmc_dest = unsafe {
+            let addr = (manifest.fmc.load_addr) as *mut u8;
+            core::slice::from_raw_parts_mut(addr, manifest.fmc.size as usize)
+        };
+        let fmc_size_words = fmc_dest.len().div_ceil(4);
+        let fmc_words = unsafe {
+            core::slice::from_raw_parts_mut(fmc_dest.as_mut_ptr() as *mut u32, fmc_size_words)
+        };
```
zhalvorsen: nit: I don't think `fmc_dest` adds much here and I would prefer to minimize the amount of unsafe code we have. Similar comment for the Runtime below.

```suggestion
        let fmc_size_words = manifest.fmc.size.div_ceil(4) as usize;
        let fmc_words = unsafe {
            core::slice::from_raw_parts_mut(manifest.fmc.load_addr as *mut u32, fmc_size_words)
        };
```

common/src/verifier.rs
```diff
@@ -33,31 +34,62 @@ pub struct FirmwareImageVerificationEnv<'a, 'b> {
...
-            .ok_or(err)?;
-        Ok(self.sha2_512_384.sha384_digest(data)?.0)
+        if self.image_in_mcu {
+            let dma = FirmwareImageVerificationEnv::create_dma_recovery(self.soc_ifc, self.dma);
+            let result =
+                dma.sha384_mcu_sram(self.sha2_512_384_acc, offset, len, dma::AesDmaMode::None)?;
+            cprintln!(
+                "[rt] SHA384 digest calculated: {}",
+                HexBytes(result.as_bytes())
+            );
```
zhalvorsen: optional: I think we can drop this print to save a little codespace. Mostly because the other digest calculations don't print anything

common/src/verifier.rs
```diff
@@ -66,20 +98,24 @@ impl ImageVerificationEnv for &mut FirmwareImageVerificationEnv<'_, '_> {
...
+
+            if let Some(mut sha_acc_op) = self
+                .sha2_512_384_acc
+                .try_start_operation(ShaAccLockState::NotAcquired)?
+            {
+                sha_acc_op
+                    .digest_384(len, offset, false, &mut digest)
+                    .map_err(|_| digest_failure)?;
+            } else {
+                Err(CaliptraError::KAT_SHA2_512_384_ACC_DIGEST_START_OP_FAILURE)?;
```
zhalvorsen: Is this the right error code? It seems weird it is prefixed with `KAT_`. Same comment for SHA512 below.

drivers/src/dma.rs
```diff
@@ -882,6 +972,62 @@ impl<'a> DmaRecovery<'a> {
...
+        // so that the AXI user is set correctly, but we want the guarantees of the
+        // Sha2_512_384Acc without making that too generic.
+
+        // Lock the SHA accelerator to ensure that the AXI user is set to the DMA user.
+        self.with_sha_acc(|dma_sha| {
+            if dma_sha.lock().read().lock() {
+                cprintln!(
+                    "[dma-image] SHA accelerator lock not acquired by DMA, cannot start operation"
+                );
+                return Err(CaliptraError::RUNTIME_INTERNAL);
```
clundin25: This seems worthy of a unique error code

drivers/src/sha2_512_384acc.rs
```diff
@@ -214,6 +212,58 @@ impl Sha2_512_384AccOp<'_> {
...
+    /// Prepare the SHA accelerator for streaming.
+    ///
+    /// # Arguments
+    ///
+    /// * `dlen` - length of data that will be hashed
+    /// * `maintain_data_endianess` - reorder byte endianess if false, leave as-is if true
+    pub fn stream_start_512(
+        &mut self,
+        dlen: u32,
+        maintain_data_endianess: bool,
```
clundin25: Super nit: Using an enum may make the calling code easier to read:

```rust
sha.stream_start_512(64, StreamEndianness::Native)?;
sha.stream_start_512(64, StreamEndianness::Reorder)?;
```

swenson:
Thanks, I'll address the comments later today.

#2671 shouldn't affect this flow, but we will have to rework it to use the external mailbox command I think (#2690) if it doesn't already take that into account.

rom/dev/src/flow/cold_reset/fw_processor.rs
swenson: Nice catch! I didn't know that trick.

common/src/verifier.rs
swenson: You're right, this is not right. I'll add an error for this.

drivers/src/sha2_512_384acc.rs
swenson: Agreed. I need to be better about using them more liberally.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -916,6 +1008,67 @@ impl FirmwareProcessor {
...
+        );
+
+        // Load Runtime from MCU SRAM
+        let runtime_size_words = manifest.runtime.size.div_ceil(4) as usize;
+        let runtime_words = unsafe {
+            core::slice::from_raw_parts_mut(
+                manifest.runtime.load_addr as *mut u32,
+                runtime_size_words,
+            )
+        };
```
clundin25, line 1064: I think the zerocopy pattern could be used here as well, to avoid unsafe.

swenson: Hmm. I don't think that is applicable here, since `manifest.runtime.load_addr` is a `u32`. I don't think we can or should zerocopy that into a slice.

clundin25: You are right, glanced at this too fast, sorry!

clundin25 (dismissed):

swenson:
Thanks!

PR #2710, opened 2025-11-06
aes: Use u32 instead of u8 for AES IV and tag

swenson:
The GCM IV is a fixed size and the tag is almost always a fixed size, so we can change all of the byte slice logic to use LEArray4xN.

This allows the compiler to generate much more efficient code: we save about 400 bytes in code size in the ROM and about 1300 bytes in the runtime.

Although the types in the mailbox change, the memory layout is identical and bytes are interpreted the same way, so there is no change in behavior.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -10,7 +10,7 @@ use caliptra_image_types::{
 use caliptra_registers::mbox;
 use core::mem::size_of;
 use ureg::MmioMut;
-use zerocopy::{little_endian::U64, FromBytes, Immutable, IntoBytes, KnownLayout, Ref};
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Ref};
 
 /// Maximum input data size for cryptographic mailbox commands.
 pub const MAX_CMB_DATA_SIZE: usize = 4096;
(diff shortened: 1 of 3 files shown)


api/src/mailbox.rs
```diff
@@ -2905,7 +2905,7 @@ impl Response for CmAesGcmEncryptInitResp {}
 pub struct CmAesGcmSpdmEncryptInitReq {
     pub hdr: MailboxReqHeader,
     pub spdm_flags: u32,
-    pub spdm_counter: [u8; 8],
+    pub spdm_counter: U64,
```
clundin25: Do you mind updating the git description for the U64 usage as well?

Not sure how controversial it is to put a non-standard type in this API.

clundin25 (dismissed):

api/src/mailbox.rs
swenson: I'm not sure why `zerocopy` doesn't support native `u64`s. This is basically a wrapper around the native (little endian) `u64`.

mhatrevi (dismissed):

api/src/mailbox.rs
clundin25: It's because a `U64` is a `[u8; 8]` under the hood.

clundin25: Given that, is switching to a `U64` still a code size improvement?

swenson: Oh good call. I'll check if it is worth it, and see if there is something better I can do, otherwise I will revert this part.

swenson: Did some more digging: we can't use `u64` due to alignment, since it would cause padding, and using `#[repr(packed)]` comes with additional headaches we don't want.

Switching to `U64` makes no difference in code size.

But, the new `fn xor_iv()` using `u32`s instead of `u8`s still saves about 150 bytes, so it's worth keeping that code path.

clundin25 (approved):

swenson:
Thanks!

PR #2711, opened 2025-11-06
[update] ROM: Add hardware version validation with minimum 2.0.2 requirement

mhatrevi:
This change introduces hardware version validation in the ROM to ensure compatibility with supported Caliptra generations. The implementation adds an is_supported_hw_version() function that validates against a minimum requirement of version 2.0.2, supporting both the 2.0.x series (from 2.0.2 onwards) and the 2.1.x series (from 2.1.0 onwards). Versions below 2.0.2, including 2.0.0 and 2.0.1, are explicitly rejected to prevent operation on incompatible hardware.


swenson (dismissed):
LGTM

rom/dev/src/main.rs
```diff
@@ -415,3 +427,10 @@ fn validate_trng_config(env: &mut RomEnv) {
         matches!(env.trng, Trng::MfgMode()),
     );
 }
+
+/// Check if the hardware version is supported
+/// Supports 2.0.2+ and 2.1.0+
+fn is_supported_hw_version(cptra_gen: &CptraGeneration) -> bool {
+    !(cptra_gen.major_version() != 2
+        || (cptra_gen.minor_version() == 0 && cptra_gen.patch_version() < 2))
```
clundin25: This expression is difficult to reason about.

Maybe a match statement is clearer? Or using `&&` instead of `||`?

Example:

```rust
match (cptra_gen.major_version(), cptra_gen.minor_version(), cptra_gen.patch_version()) {
  (2, 0, 0..=1) => false,
  _ => true,
}
```

mhatrevi: Good suggestion; updated.

clundin25 (dismissed):

swenson (approved):
LGTM

clundin25 (approved):

DerekWangScaleflux:
@mhatrevi 

What's the reason Caliptra Versions below 2.0.2, including 2.0.0 and 2.0.1, are explicitly rejected ?

PR #2712, opened 2025-11-06
Model: Add feature flag for OCP LOCK

clundin25:
Currently 2.1 subsystem always has OCP LOCK enabled. We want to make this a feature so tests are exercised with OCP LOCK disabled.

We will add a nightly run that enables OCP LOCK, similar to the TRNG feature in 1.x.


clundin25:
@swenson PTAL, thanks!

clundin25:
Closed in favor of https://github.com/chipsalliance/caliptra-sw/pull/2768

PR #2718, opened 2025-11-07
Update reset flows for reduced mailbox

swenson:
This adds support in the update reset flows for the reduced mailbox by using external mailbox commands and staging SRAM for loading the firmware updates.

This requires coordination with the hw-model to support sending in that manner.

We had to disable a KAT for space temporarily.

We also did some code refactoring in tests to prebuild a fixed MCU firmware and SoC manifest to reduce redundant code.

This more or less completes the reduced mailbox flow work. Next we'll actually shrink the mailbox in the emulator and start working on enabling FPGA tests.

### DIFF
--- a/drivers/src/soc_ifc.rs
+++ b/drivers/src/soc_ifc.rs
@@ -484,22 +484,6 @@ impl SocIfc {
         ]
     }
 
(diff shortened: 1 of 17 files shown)


swenson:
Thanks!

PR #2719, opened 2025-11-07
emulator: Shrink mailbox size to 16 KB in subsystem mode

swenson:
There are also some small changes to the firmware drivers to make them able to check that the mailbox size before using it.


mhatrevi (approved):

swenson:
Thanks!

PR #2726, opened 2025-11-10
Dpe cherry picks

clundin25:


runtime/src/debug_unlock.rs
```diff
@@ -64,11 +64,8 @@ impl ProductionDebugUnlock {
         let req = ProductionAuthDebugUnlockReq::read_from_bytes(cmd_bytes)
             .map_err(|_| CaliptraError::RUNTIME_MAILBOX_API_REQUEST_DATA_LEN_TOO_LARGE)?;
 
-        cprintln!("[rt] Starting production debug unlock request");
```
swenson, line 67: Do we want to leave these in 2.x since we're not as space-constrained in runtime?

clundin25: Good point. I will increase the runtime size instead.

zhalvorsen (approved):

zhalvorsen (approved):

mhatrevi (approved):

zhalvorsen (approved):

PR #2728, opened 2025-11-10
Pl0 pl1 reallocation main 2 x

nquarton:


nquarton:
Accidentally opening against main at first seems to have messed some things up

PR #2729, opened 2025-11-10
PL0 PL1 Context Limit Reallocation - cherry pick to main 2.x 

nquarton:


zhalvorsen (approved):

clundin25 (approved):

libcaliptra/src/caliptra_api.c
```diff
@@ -1506,6 +1506,18 @@ int caliptra_authorize_and_stash(struct caliptra_authorize_and_stash_req *req, s
 
     return pack_and_execute_command(&p, async);
 }
+// Authorize and Stash
```
jlmahowa-amd: Comment looks to be copy-pasted from 1497 and not updated.

runtime/src/reallocate_dpe_context_limits.rs
```diff
@@ -0,0 +1,81 @@
...
+            Err(CaliptraError::RUNTIME_REALLOCATE_DPE_CONTEXTS_PL0_LESS_THAN_USED)?
+        }
+        if pl1_context_limit < used_pl1_dpe_context_count as u32 {
+            Err(CaliptraError::RUNTIME_REALLOCATE_DPE_CONTEXTS_PL1_LESS_THAN_USED)?
+        }
+
+        // Update limits in persistent data now that error checking has passed
+        drivers.persistent_data.get_mut().dpe_pl0_context_limit = cmd.pl0_context_limit as u8;
+        drivers.persistent_data.get_mut().dpe_pl1_context_limit =
+            TOTAL_DPE_CONTEXT_LIMIT as u8 - cmd.pl0_context_limit as u8;
```
jlmahowa-amd: Is there a reason you re-do this subtraction instead of using "pl1_context_limit as u8"?

libcaliptra/src/caliptra_api.c
nquarton: Thanks, fixed now

runtime/src/reallocate_dpe_context_limits.rs
nquarton: Yeah that would be better. Best guess is I wrote this before the error checking above. Fixed now

jlmahowa-amd (approved):

PR #2731, opened 2025-11-10
test(fpga): Use correct ROM for FPGA subsystem tests

ArthurHeymans:
Update test_debug_unlock and test_uds_programming to use the FPGA- appropriate ROM when the fpga_subsystem feature is enabled. This ensures tests run with the correct ROM variant for the target environment.

Also remove succeeding tests from filter in CI.

The remaining failing tests are the following ones:
- test(test_wdt_activation_and_stoppage::)
  Reason: WDT is configured by MCU. Now hardcoded
  Solution: core_test MCU ROM handoff block for soc_ifc configuration
- test(test_warm_reset::test_warm_reset_success)
  Reason: Triggering the warm reset in the hardware model (SS_RST_B) does not work...
- test(test_uds_programming::) 
  Reason: Need a way to inform the MCU to trigger this. Same as WDT 
  Solution: core_test MCU ROM handoff block for soc_ifc configuration
- test(test_pmp::test_pmp_enforced)
  Reason: The test FW has 'unsupported instruction mseccfg' at pmp_tests.rs line64
- test(test_debug_unlock::)
  Reason: Same as uds programming and wdt, needs MCU ROM to trigger
  Solution: core_test MCU ROM handoff block for soc_ifc configuration
- test(test_fmcalias_derivation::test_pcr_log)
  Reason: MCU ROM does not set owner_pub_keys_digest_in_fuses, same as above
  Solution: core_test MCU ROM handoff block for soc_ifc configuration
- test(test_fips_hooks::test_fips_hook_exit)
  Reason: Unknown, no debug enabled...

The idea is to have a input block inside the MCU rom guarded by "core_test" which can be binary patched (by finding the struct offset using a magic value) by caliptra tests to set the required fuses and soc_ifc registers.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -27,7 +27,7 @@ package(ureg-systemrdl)
 failure-output = "immediate-final"
 fail-fast = false
 slow-timeout = { period = "30s", terminate-after = 6 }
-retries = 2 # the fpga subsystem is a bit flacky with typically 1 test randomly failing out of 100-200
+retries = 3 # the fpga subsystem is a bit flaky with typically 1 test randomly failing out of 100-200
 
 [profile.fpga-subsystem.junit]
 path = "/tmp/junit.xml"
@@ -80,6 +80,7 @@ store-failure-output = true
 failure-output = "immediate-final"
 fail-fast = false
 slow-timeout = { period = "30s", terminate-after = 6 }
+retries = 3 # the fpga subsystem is a bit flaky with typically 1 test randomly failing out of 100-200
 
 [[profile.nightly.overrides]]
 filter = 'test(test_generate_csr_envelop_stress)'


rom/dev/tests/rom_integration_tests/test_debug_unlock.rs
```diff
@@ -28,7 +28,10 @@ fn test_dbg_unlock_manuf_req_in_passive_mode() {
 
     let dbg_manuf_service = *DbgManufServiceRegReq::default().set_manuf_dbg_unlock_req(true);
 
-    let rom = caliptra_builder::build_firmware_rom(&ROM_WITH_UART).unwrap();
+    let rom = caliptra_builder::build_firmware_rom(firmware::rom_from_env_fpga(cfg!(
+        feature = "fpga_subsystem"
+    )))
+    .unwrap();
```
clundin25, line 34: Why not stick the cfg flag in `firmware::rom_from_env_fpga`?

clundin25 (approved):

rom/dev/tests/rom_integration_tests/test_debug_unlock.rs
swenson: We'd have to add a feature flag for the builder, which doesn't really make sense?

clundin25: An environment variable could make sense.  Ideally this function allows switching ROMs by only tweaking environment variables, instead of updating each call site.

swenson: Environment variables always feel like a leaky abstraction to me -- it feels like the code shouldn't rely on the correct environment variable to be set to generate the right firmware or ROM.

In 2.1, we already have this parameter and method, which is where I think this was copied from.

clundin25: That's a fair point. Should this parameter set grow I think we should push it out of the callsite and into some central code.

It will be hard to manage a test matrix of ROM features + verify that we have good coverage for each variant otherwise.

mhatrevi:
This work in contained in https://github.com/chipsalliance/caliptra-sw/commit/4836f02c9642eb50a86c27965468bd4ef118209e. Abandoning the PR.

PR #2733, opened 2025-11-10
rom: After programming UDS, set response correctly

swenson:
Before we were using write twice in a row, which would erase the fact that the UDS programming was successful.

This is going to be tested in a separate PR in caliptra-mcu-sw on FPGA.


swenson:
Thanks!

PR #2738, opened 2025-11-11
FMC Alias CSR add TCB info - cherry pick to main 2.x

nquarton:
- Adding TCB info to FMC Alias CSR
- Changing path length from 5 to 3 for FMC Alias CSR
- Increasing max FMC Alias CSR size to 768B
- Adding FMC Alias CSR coverage in smoke_test

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -219,7 +219,7 @@ jobs:
           path: /tmp/caliptra-test-firmware
           retention-days: 1
 
-  test_artifacts:
+  test_artifacts_subsystem:
     runs-on: vck190-subsystem-2.0
     needs: [check_cache, build_test_binaries]
     timeout-minutes: 120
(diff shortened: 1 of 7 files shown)


test/tests/caliptra_integration_tests/smoke_test.rs
```diff
@@ -622,6 +622,108 @@ fn smoke_test() {
...
+                            .entries_by_nid(Nid::SERIALNUMBER)
+                            .unwrap_single()
+                            .data()
+                            .as_utf8()
+                            .unwrap()[..],
+                        derive::mldsa_serial_number_str(&fmc_alias_pubkey)
+                    );
+                }
+            }
+            if *algorithm_type == AlgorithmType::Ecc384 {
```
zhalvorsen, line 696: Do we not need to redact these for ML-DSA? Are they not checked the same way? If not, is the assumption that they were generated the same so they should have the same content?

nquarton: To be honest I don't have a great answer for this. I matched the behavior for the other certs. We check everything else for both ECC and MLDSA. But only the ECC version of the der and txt formats are stored and compared.

zhalvorsen: @mhatrevi do you know the history of this?

mhatrevi: Probably a left-over task for 2.1.

zhalvorsen: Okay, created #2753 to track it

PR #2742, opened 2025-11-11
fpga: Fix ROM tests, simplify FPGA ROM flow and try to increase reliability

### LINKED ISSUE #2531: [Recovery boot] The recovery boot process may deadlock when the recovery image is not aligned to 256 bytes
* Caliptra Version(s) 2.x
* Caliptra Mode: Both

Regarding the definition of payload_available (see [Caliptra documentation](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#requirement-for-payload-available-signal-implementation)): it will be asserted when 
(1) the recovery FIFO is full (256 bytes), 
(2) the image activation status is asserted, and the recovery FIFO is not empty during the last transfer.

However, the Caliptra AXI DMA must wait for this signal to copy data from the RI indirect_fifo_data (“the DMA control logic will wait for the payload_available input to trigger”: [CaliptraSS specification](https://github.com/chipsalliance/caliptra-ss/blob/39dd3917923c26515ca02c547807706f797428df/docs/CaliptraSSHardwareSpecification.md#ocp-streaming-boot-payloads)). In addition, the OCP recovery protocol must wait until the Device status becomes pending before issuing the image activation command.
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/rom/src/recovery.rs#L43-L44 

Therefore, when the image size is not aligned to 256 bytes, payload_available will not be asserted, the Caliptra AXI DMA will not complete and this will lead to a deadlock.

The Caliptra ROM will loop forever at https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/drivers/src/dma.rs#L255

### LINKED ISSUE #2625: Current recovery flow requires image size to be 256bytes aligned which don't meet spec's requirement
Hi , 

From current implementation of CC ROM code, the image size should be BLOCK_SIZE(256B) aligned. That's saying for the last image packet which doesn't have 256bytes needs padding to the alignment size. 

Image

 Looks like this implementation doesn’t meet spec’s requirement.

Image

Please help to confirm. Thanks!

Tony

swenson:
These fixes are only in the FPGA ROM, FPGA hardware model, emulator, and tests. There should not be any functional change in the ROM or runtime.

* Use I3C registers directly instead of using I3C packets when possible
* FPGA ROM flow no longer requires MCU FW to be a multiple of 256 bytes. (This should not be an issue for the normal ROM.)
* Shrink default MCU FW to 4 bytes.
* Check Xilinx I3C controller FIFO levels before writing to make sure last packet has been sent.
* Fix warm reset ROM tests to more reliably work
* Update ROM tests to use the right ROM and FW for FPGA
* Update comments in `fpga-subsystem.yml` to include why tests are excluded
* Increase retries for subsystem FPGA tests
* Speed up some IDevID derivation tests
* Emulator fix for SHA512 accelerator during fast warm reset

Hopefully this will increase the reliability when running FPGA subsystem tests?

Fixes #2531.
Fixes #2625.

Includes #2731 for now.


zhalvorsen (dismissed):

swenson:
Still pretty flaky. I might try a few more things today to see what I can do to make this less flaky.

swenson:
Thanks!

mhatrevi (approved):

PR #2763, opened 2025-11-14
Parse junit.xml and add markdown to test result

clundin25:


.github/workflows/fpga-subsystem.yml
```diff
@@ -352,3 +352,24 @@ jobs:
...
+
+      - name: 'Download test results'
+        uses: actions/download-artifact@v4
+        with:
+          name: caliptra-test-results${{ inputs.artifact-suffix }}
+          path: /tmp/junit.xml
+
+      - name: Print results
+        run: |
+          echo "## Test Results\n" >> $GITHUB_STEP_SUMMARY
```
clundin25: TODO: Follow up, remove this \n. 

Skipping now to avoid generating more CI jobs.

.github/workflows/fpga-subsystem.yml
```diff
@@ -352,3 +352,24 @@ jobs:
...
+          ref: ${{ inputs.branch }}
+          submodules: recursive
+
+      - name: 'Download test results'
+        uses: actions/download-artifact@v4
+        with:
+          name: caliptra-test-results${{ inputs.artifact-suffix }}
+          path: /tmp/junit.xml
+
+      - name: Print results
```
swenson, line 373: I think we need to print the test results on failure? I think there is an `if: always()` (or something like that).

You can see on this PR that the results didn't print since the tests failed.

clundin25: Ah whoops, good catch. Added to both YML configs.

swenson: I think it might still be missing from `fpga-subsystem.yml` unless I don't see it?

clundin25: Whoops, missed pushing that version. Now it should be in there

swenson (approved):
LGTM

Issue #2756, opened 2025-11-13
FPGA Subsystem Failing Tests - Set 0


mhatrevi:
test_fmcalias_derivation::test_pcr_log
test_image_validation::test_preamble_owner_pubkey_digest_mismatch
test_image_validation::test_preamble_vendor_lms_pubkey_revocation
test_image_validation::test_preamble_vendor_ecc_pubkey_revocation
test_image_validation::test_preamble_vendor_mldsa_pubkey_revocation
test_image_validation::test_runtime_svn_less_than_fuse_svn

ArthurHeymans:
PR is using https://github.com/chipsalliance/caliptra-mcu-sw/pull/582

mhatrevi:
[feat(hw-model): Add OTP partition offsets and fuse fields #2764](https://github.com/chipsalliance/caliptra-sw/pull/2764) & https://github.com/chipsalliance/caliptra-mcu-sw/pull/582 are done. Closing the issue.

PR #2764, opened 2025-11-14
feat(hw-model): Add OTP partition offsets and fuse fields

ArthurHeymans:
Add support for additional OTP partition fields including:
- SVN partition field offsets (FMC, runtime, SoC manifest)
- Vendor hashes production partition (owner public key hash)
- Vendor revocations production partition (ECC, LMS, MLDSA)
- Runtime firmware SVN and SoC manifest SVN programming

This extends the OTP memory layout to support owner key management and multi-algorithm revocation tracking.

Addresses issue: #2756

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -352,3 +352,25 @@ jobs:
           path: |
             /tmp/junit.xml
             /tmp/nextest-list.json
+
+  print_test_results:
+    runs-on: ubuntu-22.04
+    needs: [test_artifacts_subsystem]
+    if: always()
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5
+        with:
+          ref: ${{ inputs.branch }}
+          submodules: recursive
+
+      - name: 'Download test results'
+        uses: actions/download-artifact@v4
+        with:
+          name: caliptra-test-results${{ inputs.artifact-suffix }}
+          path: /tmp/junit.xml
+
+      - name: Print results
+        run: |
+          echo "## Test Results" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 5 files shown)


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1196,12 +1208,62 @@ impl ModelFpgaSubsystem {
...
+            .copy_from_slice(soc_manifest_svn);
+
+        // Max SOC Manifest SVN (1 byte used)
         println!(
             "Burning fuse for SOC MAX SVN {}",
             fuses.soc_manifest_max_svn
         );
         otp_data[OTP_SVN_PARTITION_SOC_MAX_SVN_FIELD_OFFSET] = fuses.soc_manifest_max_svn;
 
+        // TODO this conflicts with otp_provision.rs, but we need it as it's part of Bootparams
```
ArthurHeymans: @timothytrippel This overrides a setting you tried to set earlier (and is set to default to 1)

timothytrippel: Are you saying this is now set in two places? Can we just set it in one place then?

ArthurHeymans: The value is in the fuses struct which is only passed in the fn boot(). Please evaluate if it is needed in the location you set it.

timothytrippel: Sorry, what I am saying is that instead of setting this again here in the final OTP memory array, on line 1187, can you change:

```rust
let mem = otp_generate_sw_manuf_partition_mem(&OtpSwManufPartition::default())?;
```

to

```rust
let mem = otp_generate_sw_manuf_partition_mem(&OtpSwManufPartition {
    anti_rollback_disable: u32::from(fuses.anti_rollback_disable),
    ..Default::default()
}
```

Then there is no need to set the value here too. Correct?

mhatrevi (approved):

ArthurHeymans:
PR is using https://github.com/chipsalliance/caliptra-mcu-sw/pull/582

Issue #2769, opened 2025-11-15
Fix test_reallocate_dpe_context_limits


clundin25:
# Please include the following information

* 2.x
* Caliptra Mode Subsystem

This test was recently introduced. It will cause a kernel panic on each run, on this line https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs#L128.

This is likely because the axi user is invalid.

clundin25:
Perhaps valid users need to be set?

```
        // Set up the PAUSER as valid for the mailbox (using index 0)
        // self.setup_mailbox_users(boot_params.valid_axi_user.as_slice())
        //     .map_err(ModelError::from)?;
```

nquarton:
MCU SW change (must merge first) - https://github.com/chipsalliance/caliptra-mcu-sw/pull/608
Caliptra SW PR to re-enable tests - https://github.com/chipsalliance/caliptra-sw/pull/2833

PR #2771, opened 2025-11-15
Disable reallocation tests

clundin25:
See https://github.com/chipsalliance/caliptra-sw/issues/2769


zhalvorsen (dismissed):

nquarton (approved):
Sounds good to me. Looks like this is a larger issue with setting the AXI user on SS FPGA. The other ones that utilize that are also disabled in this list. (At least that answers my question of "why wouldn't other set_axi_user tests be failing")

PR #2773, opened 2025-11-16
Update subsystem bitstreams

clundin25:


clundin25:
We will use different bitstreams.

PR #2774, opened 2025-11-16
Refactor Nightly Release

clundin25:
Use a matrix to clean up duplicated jobs.

Doing on 1.x first, then will do the rest. This will be important as the test matrix grows.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -56,321 +56,65 @@ jobs:
           echo "release_ref=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
           echo "Current ref $(git rev-parse HEAD) will receive tag ${TAG_BASE}${INDEX} after tests"
 
-  fpga-1_0-full-suite-etrng-log:
-    name: FPGA Suite (1.0, etrng, log)
+  fpga-tests:
+    name: FPGA Suite (${{ matrix.hw_config.name }}, ${{ matrix.rng.name }}, ${{ matrix.log.name }})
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -56,321 +56,65 @@ jobs:
...
-      rom-logging: false
-      rom-version: "1.0"
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          # The current 1.0.x ROM used in CI will overflow the stack into unused space
+          # We disable the stack overflow check (sw_emu_stack_check_disable) when testing against that ROM
+          - { name: "hw-1.0", rom: "1.0", suffix: "hw-1.0", base_features: "hw-1.0,slow_tests,sw_emu_stack_check_disable" }
+          - { name: "1.1", rom: "1.1", suffix: "hw-1.1", base_features: "slow_tests" }
```
zhalvorsen: The other emulator `config.name` are prefixed with `hw-` should we do the same here?


```suggestion
          - { name: "hw-1.1", rom: "1.1", suffix: "hw-1.1", base_features: "slow_tests" }
```

zhalvorsen (dismissed):
This is so much better!

.github/workflows/nightly-release.yml
clundin25: Hmm good point. I decided to remove the `hw-`. I am not sure why it was added but I imagine it began to lose it's correctness as the code evolved.

I think HW is innacurate, for example the 1.1 run is not running against 1.1 hardware.

PR #2786, opened 2025-11-18
Protect sensitive MMIO UIOs

zhalvorsen:
Accessing any of these MMIOs while subsystem is in reset can cause a kernel panic which makes it very difficult to debug some issues or see test failures in CI. This change puts those accesses behind guards to ensure subsystem is out of reset while accessing them.


swenson (approved):
LGTM; this is a lot less messy than I feared it would be tbh

PR #2788, opened 2025-11-18
OCP LOCK Rom tests: Reported Valid HEK Metadata

clundin25:
This test verifies that HEK is available after reporting a valid HEK seed state.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -109,6 +109,8 @@ pub type DefaultHwModel = ModelFpgaSubsystem;
 
 pub const DEFAULT_APB_PAUSER: u32 = 0x01;
 
+pub type ModelCallback = Box<dyn FnOnce(&mut DefaultHwModel)>;
+
 /// Constructs an HwModel based on the cargo features and environment
 /// variables. Most test cases that need to construct a HwModel should use this
 /// function over HwModel::new_unbooted().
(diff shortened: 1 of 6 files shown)


hw-model/src/lib.rs
```diff
@@ -1250,6 +1254,12 @@ pub trait HwModel: SocManager {
 
         Ok(())
     }
+
+    /// Return control to the model.
+    /// The model can continue with the boot flow that it has yielded to the caller.
+    fn continue_boot(&mut self) -> Result<(), ModelError> {
```
zhalvorsen: There is a pretty common flow for this in other tests (`test_ldev_ecc384_cert` is a decent example). Are you thinking we should replace the current idea of:

1. Get hw-model that has called `boot`
2. Wait for mb ready
3. Send ROM commands
4. Call `helpers::test_upload_firmware`

Off the top of my head, another option could be to add a function that does the above steps, but takes a closure instead of step 3 with operations that need to be done with the ROM mailbox before continuing

clundin25: I think a closure is a good approach.

I was aware of `helpers::test_upload_firmware` but I dislike the pattern. It pushes a lot of logic to each individual test. 

I don't see any reason why each callsite would need to handle this, and it adds duplication and bloat. Pushing logic to tests in this manner makes it difficult to write workflows that run all tests against different common configurations, e.g. itrng, etc.

runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -300,7 +309,9 @@ pub fn start_rt_test_pqc_model(
 // Run a test which boots ROM -> FMC -> test_bin. If test_bin_name is None,
 // run the production runtime image.
 pub fn run_rt_test(args: RuntimeTestArgs) -> DefaultHwModel {
-    run_rt_test_pqc(args, FwVerificationPqcKeyType::LMS)
+    // TODO(clundin): Do we want to use MLDSA by default in 2.1?
```
zhalvorsen, line 343: Yeah, I think we should. I doubt many people are going to use LMS for it. Lets do that in a separate pr

zhalvorsen (dismissed):

PR #2791, opened 2025-11-18
[Runtime] Correctly setting FE base address

mhatrevi:


common/src/uds_fe_programming.rs
```diff
@@ -75,38 +75,45 @@ impl UdsFeProgrammingFlow {
...
     // |          (64 bytes)                 |
     // +-------------------------------------+
-    //
-    // FE Partitions, separate region in fuse controller memory handled by MCU, but at same base as UDS
-    //
-    // +-------------------------------------+ <- uds_seed_dest_base_addr_low()
+    // +-------------------------------------+
+    // |        UDS Digest                   |
+    // |         (8 bytes)                   |
+    // +-------------------------------------+
```
swenson, line 81: For 2.1, we'll need to add an additional 8 bytes for the zeroization fuses.

swenson (approved):
I think this is okay, though caliptra-mcu-sw's cold boot flow will need to change to support this:

https://github.com/chipsalliance/caliptra-mcu-sw/blob/31c117a3b7c8e8193297591efbc8bf03a54334a3/rom/src/rom.rs#L133

Currently, we set the base address to the FE partition if we are programming FE and to UDS otherwise. We can never do both at the same time without resetting, so this wasn't assumed to be a problem.

Either approach is correct, as long as we're consistent with how we set the base address.

ArthurHeymans (approved):

common/src/uds_fe_programming.rs
mhatrevi: Yes, I am aware of the 8 bytes zeroization marker for UDS and each FE partition.

PR #2796, opened 2025-11-18
fpga: Don't pad image to multiple of 256 bytes when sending

swenson:
This restriction was removed from the FPGA ROM but left in the sending code, which is causing some failures when running on the subsystem FPGA.


swenson:
Thanks!

PR #2799, opened 2025-11-19
Move fuse settings from `BootParams` to `InitParams`

zhalvorsen:
The Subsystem model needs to load fuses before being released from reset. Otherwise there is a race condition for the HW model setting OTP memory while MCU ROM reads from it.

The other HW models are changed to cache the fuse value until it is needed during `boot`.


runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -241,11 +246,14 @@ pub fn start_rt_test_pqc_model(
...
-
-    let (vendor_pk_hash, owner_pk_hash) = image_pk_desc_hash(&image.manifest);
+    init_params.fuses = Fuses {
+        fuse_pqc_key_type: pqc_key_type as u32,
+        vendor_pk_hash,
+        owner_pk_hash,
+        soc_manifest_svn: svn_to_bitmap(args.soc_manifest_svn.unwrap_or(0)),
+        soc_manifest_max_svn: args.soc_manifest_max_svn.unwrap_or(127) as u8,
+        ..Default::default()
+    };
```
clundin25, line 256: This looks like it would always override fuses that are passed in, right?

Should this instead happen in the init block on line 237?

zhalvorsen: I had the same thought, but it caused a few tests to fail because they pass in mostly default `init_params`. The logic below hard coded the fuses. So the test would use default fuse values, but need these fuse settings.

clundin25: Okay sounds good, I guess the tests that care about this re-implement some of this common code anyway.

PR #2800, opened 2025-11-19
Read OTP DAI Idle bit and direct access cmd reg offset from generic straps

### LINKED ISSUE #2505: Update UDS Programming sequence
Update values for the following:

    - `DAI_IDLE` bit offset: (`SS_STRAP_GENERIC` register0 >> 16) & 0xFFFF
    - `STATUS` register address: (`SS_STRAP_GENERIC` register0) & 0xFFFF + Fuse Controller's base address.        [STATUS register address retrieval is optional for ROM since it is at a fixed offset from Fuse Controller]
    - `DIRECT_ACCESS_CMD` offset: (`SS_STRAP_GENERIC` register1) & 0xFFFF + Fuse Controller's base address.

mhatrevi:


drivers/src/soc_ifc.rs
```diff
@@ -634,6 +634,14 @@ impl SocIfc {
             MailboxRespHeader::FIPS_STATUS_NON_ZEROIZABLE_KEY
         }
     }
+
+    pub fn otp_dai_idle_bit_num(&self) -> u32 {
+        (self.soc_ifc.regs().ss_strap_generic().at(0).read() >> 16) & 0xFFFF
```
swenson, line 639: If the straps are 0 or not set, do we want to use the default values from the reference OTP controller?

I think otherwise the UDS / FE tests will fail on FPGA.

common/src/uds_fe_programming.rs
```diff
@@ -144,22 +148,38 @@ impl UdsFeProgrammingFlow {
             let fuse_controller_base_addr = soc_ifc.fuse_controller_base_addr();
             let otp_ctrl = DmaOtpCtrl::new(AxiAddr::from(fuse_controller_base_addr), dma);
             let seed_dest_address = self.get_dest_address(soc_ifc);
+            let dai_idle_bit_num = soc_ifc.otp_dai_idle_bit_num();
+            let direct_access_cmd_reg_addr =
+                fuse_controller_base_addr + soc_ifc.otp_direct_access_cmd_reg_offset() as u64;
```
timothytrippel, line 153: whats the point of retrieving the CMD reg offset from the generic straps, but assumming the rest of the registers (ADDR, WDATA_0, WDATA_1) will be at the correct relative offset? 

If you know the base address of the fuse controller, which it looks like you do on 149, shouldn't that be enough?

mhatrevi: The placement of ADDR, WDATA_0, and WDATA_1 at fixed offsets from the CMD register is not an assumption - it’s defined by the specification:
https://github.com/chipsalliance/caliptra-ss/blob/44938924d18bea1316f57bdbab2292364f3b44f0/src/fuse_ctrl/doc/registers.md.

There are a few variable registers between the fuse controller base address and the CMD register, and integrators can choose not to implement them. This means the CMD register’s address can vary by integrator, but once CMD is located, the relative offsets for ADDR and WDATA registers are guaranteed by the spec.

timothytrippel: Yes, I am intimately familiar with the otp_ctrl specification as it stems from the [OpenTitan project](https://github.com/chipsalliance/caliptra-ss/blob/44938924d18bea1316f57bdbab2292364f3b44f0/src/fuse_ctrl/rtl/otp_ctrl.sv#L1). 

I am curious, where in the specification it says the registers that are optional? I seemed to have missed that. Could you share a pointer to those that are optional? I want to make sure to avoid assumptions during provisioning. Also, if this is the case, could you drop a comment in the code pointing to it? I think others might also find this behavior strange and wonder the reason for it.

Similarly, it seems the generic strap signals are quite valuable as there are not many of them, so I found it interesting that this is the use case they were deployed for. Are there other registers in other HW blocks that are optional by the integrator and could cause the memory map to change?

drivers/src/soc_ifc.rs
swenson: Ah, I see that we are setting the straps in ROM in caliptra-mcu-sw. This should be fine then.

common/src/uds_fe_programming.rs
swenson: The registers that are optional (really, variable) is the error registers, which depend on the number of partitions. The integrator can customize the number of partitions, which changes the location of the DAI IDLE bit and the CMD register.

swenson (approved):
LGTM

common/src/uds_fe_programming.rs
timothytrippel: Ahh right. 

In that case, do you need to use both generic straps (0 and 1) @mhatrevi ? Can't you just use one of the generic straps to pass in the "number of partitions", and then compute both:
1. the CMD reg offset and 
2. the IDLE bit offset
from a single input value?

mhatrevi: @timothytrippel Are the question(s) addressed? If yes, can you resolve the comment?

swenson: @timothytrippel these straps specified in the hardware integration spec https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/CaliptraIntegrationSpecification.md#late-binding-interface-signals

timothytrippel: Ah got it. The name "strap_generic" when it has a specific use case was misleading. Thanks for the pointer. Marking resolved.

timothytrippel (approved):

PR #2812, opened 2025-11-19
fpga: Cold reset test can be too fast sometimes

swenson:
Since we step the hw-model to get the MCU booted and the cold reset ROM is so fast, sometimes the output has already been processed before we set up the search term.

We refactor the test to call new_unbooted() and boot() manually so that we can set the search term before boot() so that it will always be captured.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -2007,7 +2007,7 @@ mod tests {
         );
         println!("{init_params_summary:#?}");
 
-        // The Caliptra ROM boots so quickly that sometimes it is done while before we finish returning from booting the MCU
+        // While in boot(), sometimes the test Caliptra ROM has written to the output before we can set the search term in step_until_output().
         // So set the search term manually before calling boot().
         model.output().set_search_term("hii");
         model.boot(BootParams::default()).unwrap();


hw-model/src/lib.rs
```diff
@@ -1994,14 +1994,24 @@ mod tests {
...
+        let init_params_summary = init_params.summary();
+        let mut model = DefaultHwModel::new_unbooted(init_params).unwrap();
+        let hw_rev_id = model.soc_ifc().cptra_hw_rev_id().read();
+        println!(
+            "Using hardware-model {} trng={:?} hw_rev_id={{cptra_generation=0x{:04x}, soc_stepping_id={:04x}}}",
+            model.type_name(), model.trng_mode(),  hw_rev_id.cptra_generation(), hw_rev_id.soc_stepping_id()
+        );
+        println!("{init_params_summary:#?}");
+
+        // The Caliptra ROM boots so quickly that sometimes it is done while before we finish returning from booting the MCU
```
clundin25: nit: Seems there is a typo, do you mind fixing it? Having a hard time understanding this comment

clundin25 (dismissed):

hw-model/src/lib.rs
swenson: I believe this is accurate.

We `step()` until we finish booting MCU in `boot()`. During that time, if it tells Caliptra ROM to start booting, sometimes Caliptra ROM is finished and has already written to the output before we return from `boot()`.

clundin25: Maybe `while` can be dropped? I think that is where I get confused

swenson: Oh I see. "while before". Let me clarify.

zhalvorsen (dismissed):

hw-model/src/lib.rs
swenson: Changed it. Let me know if that makes more sense.

clundin25: That does, thanks!

Issue #2755, opened 2025-11-13
[mcu,jtag] LC transition tests failing on SS FPGA


mhatrevi:
The following tests are failing on FPGA:
- [x] raw unlock --> fixed in:
    - https://github.com/chipsalliance/caliptra-sw/pull/2783
    - https://github.com/chipsalliance/caliptra-mcu-sw/pull/581
- [x] rma unlock --> fixed in:
    - https://github.com/chipsalliance/caliptra-mcu-sw/issues/454 contains FPGA bitstream fix
    - https://github.com/chipsalliance/caliptra-sw/pull/2813
    - https://github.com/chipsalliance/caliptra-mcu-sw/pull/602

timothytrippel:
https://github.com/chipsalliance/caliptra-sw/pull/2783 and https://github.com/chipsalliance/caliptra-mcu-sw/pull/580 should fix when the new bitstreams from https://github.com/chipsalliance/caliptra-mcu-sw/pull/587 and https://github.com/chipsalliance/caliptra-mcu-sw/pull/588 are built and updated in CI.

PR #2813, opened 2025-11-19
[model] expose RMA and scrap PPD strap

timothytrippel:
This exposes the RMA and scrap physical presence detection pin to the subsystem hardware model so tests can set it. This will partially addresses https://github.com/chipsalliance/caliptra-sw/issues/2755.


clundin25 (approved):

clundin25:
Disabled auto merge to see if we can add this to the merge queue.

PR #2822, opened 2025-11-20
Add FPGA-specific ROM (no log) to builder

swenson:
This is needed for the nightly tests. A lot of them override the ROM type to use the UART FPGA ROM, but I think some might use the non-UART log?

This also increases the timeout for the slow tests `test_binaries_are_identical` and `test_generate_csr_envelop_stress`, which were occasionally timing out in nightly.


mhatrevi (approved):

swenson:
Thanks!

timothytrippel (approved):

zhalvorsen (approved):

PR #2832, opened 2025-11-21
test: Fix FW_INFO test to work on subsystem FPGA

swenson:
Needed to use the correct ROM, runtime, and to wait to send the FW_INFO command until after runtime is booted, and to add the SoC manifest and MCU FW if running on the subsystem.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_info.rs
+++ b/runtime/tests/runtime_integration_tests/test_info.rs
@@ -71,7 +71,7 @@ pub fn get_fwinfo(model: &mut DefaultHwModel) -> FwInfoResp {
 
 #[test]
 fn test_fw_info() {
-    let fpga = cfg!(any(feature = "fpga_subsystem"));
+    let fpga = cfg!(feature = "fpga_subsystem");
     let app = if fpga {
         &APP_WITH_UART_FPGA
     } else {


ArthurHeymans (dismissed):

runtime/tests/runtime_integration_tests/test_info.rs
```diff
@@ -71,7 +71,20 @@ pub fn get_fwinfo(model: &mut DefaultHwModel) -> FwInfoResp {
 
 #[test]
 fn test_fw_info() {
+    let fpga = cfg!(any(feature = "fpga_subsystem"));
```
zhalvorsen: Should this be

```Rust
let fpga = cfg!(any(feature = "fpga_realtime", feature = "fpga_subsystem"));
```

If not, you could remove the `any` portion

zhalvorsen (dismissed):

runtime/tests/runtime_integration_tests/test_info.rs
swenson: I had it as that initially, but then realized that fpga_realtime doens't need a special ROM or runtime, so took it out and forgot to remove the any. Will remove.

zhalvorsen (approved):

mhatrevi (approved):

PR #2833, opened 2025-11-22
Re-enabling test_reallocate_dpe_context_limits tests on subsystem FPGA

nquarton:
Depends on https://github.com/chipsalliance/caliptra-mcu-sw/pull/608 to make tests pass


.github/workflows/fpga-subsystem.yml
zhalvorsen: FYI: you'll have to update this reference to take the new MCU ROM change

nquarton: Thanks! (I suspected there might be an additional step like that but I figured I'd like CI try it and dig into it if it failed.)

PR #2835, opened 2025-11-24
fix(ci): Use correct ROM binary for fpga_subsystem tests

ArthurHeymans:
Consolidate ROM selection logic into a single helper function that properly handles fpga_subsystem feature flag. Previously, tests were inconsistently using rom_for_fw_integration_tests() vs rom_for_fw_integration_tests_fpga(), causing incorrect ROM to be loaded for fpga_subsystem builds.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_certs.rs
+++ b/runtime/tests/runtime_integration_tests/test_certs.rs
@@ -6,7 +6,8 @@ use crate::common::{
     get_rt_alias_ecc384_cert, get_rt_alias_mldsa87_cert, run_rt_test, run_rt_test_pqc, DpeResult,
     RuntimeTestArgs, TEST_LABEL,
 };
-use caliptra_builder::firmware::{APP_WITH_UART, FMC_WITH_UART};
+use caliptra_api::SocManager;
+use caliptra_builder::firmware::{APP_WITH_UART, APP_WITH_UART_FPGA, FMC_WITH_UART};
 use caliptra_builder::ImageOptions;
 use caliptra_common::mailbox_api::{
     CommandId, GetIdevCertResp, GetIdevEcc384CertReq, GetIdevEcc384InfoResp, GetIdevMldsa87CertReq,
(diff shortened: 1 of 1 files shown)


swenson:
@ArthurHeymans clippy fixes needed

runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -603,12 +604,19 @@ pub fn test_all_measurement_apis() {
...
-                .to_bytes()
-                .unwrap();
+        let rom = crate::common::rom_for_fw_integration_tests().unwrap();
+        let fw_image = caliptra_builder::build_and_sign_image(
+            &FMC_WITH_UART,
+            &if cfg!(any(feature = "fpga_realtime", feature = "fpga_subsystem")) {
+                APP_WITH_UART_FPGA
+            } else {
+                APP_WITH_UART
+            },
```
zhalvorsen, line 614: For the same reasons, it would be good to have a similar function for runtime as well. We could do that on a follow on though.

runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -189,6 +191,10 @@ pub const fn svn_to_bitmap(svn: u32) -> [u32; 4] {
     ]
 }
 
+pub fn rom_for_fw_integration_tests() -> io::Result<Cow<'static, [u8]>> {
+    caliptra_builder::rom_for_fw_integration_tests_fpga(cfg!(feature = "fpga_subsystem"))
```
zhalvorsen, line 195: Shouldn't this also be set to true for `fpga_realtime`:


```suggestion
    caliptra_builder::rom_for_fw_integration_tests_fpga(cfg!(any(feature = "fpga_realtime", feature = "fpga_subsystem")))
```

zhalvorsen: It doesn't look like it, because the CI all passed. Closing this so we can merge it.

PR #2836, opened 2025-11-24
Fix `test_cryptographic_mailbox::test_derive_stable_key_from_rom`

zhalvorsen:
The tests that stop at ROM don't include a Core FW image, but do include manifest and MCU FW so it was failing while trying to unwrap the Core FW image. This will now return from the subsystem `boot` function if just the Core FW image is excluded.


swenson (approved):
LGTM

ArthurHeymans (approved):

PR #2837, opened 2025-11-24
Separate FMC Cert/CSR Owner and Vendor Info

nquarton:
- Separate owner and vendor fuse info
- Make TCB type names more specific in FMC cert/CSR and RT cert
- Add missing testing for FMC MLDSA CSR


nquarton:
It was brought to my attention that several related changes are needed in the PCR generation which feeds key derivation for the FMC Alias layer. Those changes will be made and added here ASAP.

nquarton:
> It was brought to my attention that several related changes are needed in the PCR generation which feeds key derivation for the FMC Alias layer. Those changes will be made and added here ASAP.

This is included in this PR now. Ideally, there would be shared code to keep the key derivation and TCB infos in-line, but the formatting and grouping for the two are completely different (one is TCB info and one goes into the PCR and PCR log) so there really isn't much of an opportunity to consolidate this.

nquarton:
PR to update spec with these changes is here: https://github.com/chipsalliance/Caliptra/pull/579

nquarton:
Note a change in direction was made here where I need to add the vendor PK index used and SVN from the manifest back into the cert and FMC key derivation.

nquarton:
> Note a change in direction was made here where I need to add the vendor PK index used and SVN from the manifest back into the cert and FMC key derivation.

These have been re-added as well as the bool indicator for the owner pub key being fused. This is ready for review again.

common/src/dice.rs
```diff
@@ -187,24 +183,101 @@ pub fn copy_ldevid_mldsa87_cert(
...
+    ])?;
+    hasher.update(
+        &soc_ifc
+            .fuse_bank()
+            .vendor_lms_pub_key_revocation()
+            .to_le_bytes(),
+    )?;
+    hasher.update(&[
+        soc_ifc.fuse_bank().vendor_mldsa_pub_key_revocation() as u8,
+        soc_ifc.fuse_bank().fw_fuse_svn() as u8,
```
zhalvorsen, line 235: The older version had a concept of an effective fused SVN. Do we want to keep the same concept here? I don't feel too strongly, but figured I'd point it out.

```Rust
    // This info was not saved from ROM so we need to repeat this check
    let fmc_effective_fuse_svn = match env.soc_ifc.fuse_bank().anti_rollback_disable() {
        true => 0_u8,
        false => env.soc_ifc.fuse_bank().fw_fuse_svn() as u8,
    };
```

zhalvorsen (dismissed):

common/src/dice.rs
```diff
@@ -187,24 +183,101 @@ pub fn copy_ldevid_mldsa87_cert(
...
+///
+/// # Returns
+///
+/// * `[u8; 48]` - SHA384 hash of the owner device info
+pub fn gen_fmc_alias_owner_device_info_hash(
+    soc_ifc: &SocIfc,
+    data_vault: &DataVault,
+    sha2_512_384: &mut Sha2_512_384,
+) -> CaliptraResult<[u8; 48]> {
+    // NOTE: The contents of this TCB info and FMC PCR info must stay in sync.
```
zhalvorsen, line 203: Would it make sense to have a common struct with all of the information needed for both? You could add helper functions for both scenarios something like `tcb_info()` and `pcr_info()`. That could help make sure it isn't forgotten in one place or the other

common/src/dice.rs
nquarton: I also don't feel strongly either way. We do capture both anti rollback disable and the SVN required by fuses already so I don't think we are really missing anything if we skip this step. Fabrizio was against much "processing" of the values to get to these inputs and I do see his point with that.

zhalvorsen: Sounds good to me!

zhalvorsen (approved):

zhalvorsen (dismissed):

common/src/dice.rs
nquarton: I agree something like this should be added. I filed an issue: https://github.com/chipsalliance/caliptra-sw/issues/3264

I wanted to take another look at this, but I didn't get the chance.

zhalvorsen: No worries, we can take a look at this in a patch release

PR #2841, opened 2025-11-25
[feat] UDS & FE zeroization

### LINKED ISSUE #2707: UDS & FE FIPS zeroization sequence
* Caliptra Version: 2.1
* Caliptra Mode: Both

mhatrevi:
Closes #2707

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -322,6 +322,7 @@ jobs:
           # -E 'package(caliptra-drivers) and test(test_ocp_lock_warm_reset)'
           # -E 'package(caliptra-drivers) and test(test_ocp_lock)'
           # test(test_version::test_version), test(test_warm_reset::test_warm_reset_version), test(test_fips::test_fips_version) needs bitstream update
+          # test(test_uds_fe::test_uds_programming_granularity_64bit), test(test_uds_fe::test_uds_programming_granularity_32bit) needs jtag for dbg_manuf_service
           COMMON_ARGS=(
               --cargo-metadata="${TEST_BIN}/target/nextest/cargo-metadata.json"
               --binaries-metadata="${TEST_BIN}/target/nextest/binaries-metadata.json"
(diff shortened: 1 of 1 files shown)


common/src/uds_fe_programming.rs
```diff
@@ -241,4 +298,104 @@ impl UdsFeProgrammingFlow {
...
+                } else {
+                    Ok(())
+                }
+            };
+
+            // Step 1: Clear the Partition Zeroization Marker (64-bit)
+            // This step is critical - it masks potential ECC or integrity errors
+            // if the process is interrupted by a power failure
+            execute_dai_cmd(zeroization_marker_address, DAI_CMD_ZEROIZE);
+            verify_cleared(CaliptraError::UDS_FE_ZEROIZATION_MARKER_NOT_CLEARED)?;
```
swenson, line 364: If this fails, I wonder if we should return a non-zero fips status?

rom/dev/src/flow/cold_reset/fmc_alias.rs
```diff
@@ -53,22 +52,6 @@ impl FmcAliasLayer {
         fw_proc_info: &FwProcInfo,
     ) -> CaliptraResult<()> {
         cprintln!("[afmc] ++");
-        cprintln!("[afmc] CDI.KEYID = {}", KEY_ID_ROM_FMC_CDI as u8);
```
swenson, line 56: Are we just deleting these to save space?

rom/dev/src/main.rs
```diff
@@ -75,13 +75,10 @@ pub extern "C" fn rom_entry() -> ! {
     };
 
     if !cfg!(feature = "no-cfi") {
-        cprintln!("[state] CFI Enabled");
```
swenson, line 78: Some of these are potentially useful to ensure we're running with the right hardware and ROM.

rom/dev/src/main.rs
```diff
@@ -92,12 +89,6 @@ pub extern "C" fn rom_entry() -> ! {
     // Check if HW version is supported.
     let cptra_gen = env.soc_ifc.caliptra_generation();
     if !is_supported_hw_version(&cptra_gen) {
-        cprintln!(
```
swenson, line 95: For 2.1 we can eliminate this check entirely, as we don't have a constraint to only run on specific versions.

rom/dev/tests/rom_integration_tests/test_uds_fe.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    // Verify response checksum
+    assert!(caliptra_common::checksum::verify_checksum(
+        resp.hdr.chksum,
+        0x0,
+        &response.as_bytes()[core::mem::size_of_val(&resp.hdr.chksum)..],
+    ));
+
+    // Verify DPE result indicates success
+    assert_eq!(resp.dpe_result, 0); // DPE_STATUS_SUCCESS
```
swenson: One thing that would be interesting to check would be to provision the UDS with random data in the otp_memory, do the zeroization, and check that the value is all 0 or all 1s after.

rom/dev/src/main.rs
mhatrevi: Makes sense. Updated.

common/src/uds_fe_programming.rs
mhatrevi: Update to return a non-zero DPE code.

rom/dev/tests/rom_integration_tests/test_uds_fe.rs
mhatrevi: Added an issue to add this test separately: https://github.com/chipsalliance/caliptra-sw/issues/2848

rom/dev/src/flow/cold_reset/fmc_alias.rs
mhatrevi: Correct

common/src/uds_fe_programming.rs
swenson: Hmm. There might be larger implications to a zeroization failure though. If zeroization fails, then we are in a FIPS failure mode, and we might need to return a FIPS invalid status on anything that provides a service that uses UDS or FE indirectly.

Probably it would be easier to just set a global FIPS status of failure for everything at that point.

mhatrevi: After offline discussion, UDS/FE zeroization will now trigger a ROM shutdown because any cryptographic material derived from UDS/FE becomes invalid.

rom/dev/src/flow/cold_reset/fw_processor.rs
```diff
@@ -682,6 +684,53 @@ impl FirmwareProcessor {
...
+                                Err(_) => 0x1, // InternalError
+                            },
+                        };
+                        resp.populate_chksum();
+                        txn.send_response(resp.as_bytes())?;
+
+                        // Set the zeroization status in non-fatal error register
+                        report_fw_error_non_fatal(u32::from(result.is_err()));
+
+                        // Shutdown after zeroization as UDS and/or FE values and it's derived keys are no longer valid.
```
swenson: nit
```suggestion
                        // Shutdown after zeroization as UDS and/or FE values and its derived keys are no longer valid.
```

common/src/uds_fe_programming.rs
```diff
@@ -241,4 +298,104 @@ impl UdsFeProgrammingFlow {
...
+            // Step 2: Zeroize all data words in the partition
+            let mut addr = config.seed_config.address;
+            let mut words_remaining = config.seed_config.length_bytes / 4;
+
+            while words_remaining > 0 {
+                execute_dai_cmd(addr, DAI_CMD_ZEROIZE);
+
+                // Verify zeroization - should return 0xFFFFFFFF for all bits
+                let rdata_0 = dma.read_dword(AxiAddr::from(config.direct_access_rdata_0_reg_addr));
+                if rdata_0 != 0xFFFFFFFF {
```
swenson, line 375: Question: if the zeroization fails, such as if a bit is stuck at 0, will this error happen? Should we try to zeroize as much as we can before returning an error?

common/src/uds_fe_programming.rs
```diff
@@ -241,4 +298,104 @@ impl UdsFeProgrammingFlow {
...
+            // This step is critical - it masks potential ECC or integrity errors
+            // if the process is interrupted by a power failure
+            execute_dai_cmd(zeroization_marker_address, DAI_CMD_ZEROIZE);
+            verify_cleared(CaliptraError::UDS_FE_ZEROIZATION_MARKER_NOT_CLEARED)?;
+
+            // Step 2: Zeroize all data words in the partition
+            let mut addr = config.seed_config.address;
+            let mut words_remaining = config.seed_config.length_bytes / 4;
+
+            while words_remaining > 0 {
```
swenson: I wonder if this would be easier to read if we used a `for addr in (start_addr..end_addr).step_by(granularity)` since we only use `words_remaining` as a redundant counter?

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -818,7 +818,8 @@ fn test_authorize_and_stash_after_update_reset_skip_auth() {
     assert_eq!(authorize_and_stash_resp.auth_req_result, IMAGE_AUTHORIZED);
 }
 
-#[test]
+//#[test]
```
swenson: Why is this commented out?

mhatrevi: Reason below "// Investigate why this is causing a stack overflow.". We'll investigate this separately.

swenson (approved):
LGTM

swenson (dismissed):
LGTM

swenson (approved):
LGTM

swenson (approved):

PR #2844, opened 2025-11-26
Fix FW update FPGA tests

mlvisaya:
When locking MCU Mailbox SRAM, make sure that mailbox is locked by checking the LOCK value. Make sure that value read is 0 before proceeding to write to MCU Mailbox SRAM.

MCU Reset Request interrupt is triggered twice by Caliptra when the ACTIVATE FW command is processed, 1 to reset the MCU and another to indicate FW is available to ROM. Therefore test should clear these 2 interrupts.

Enable the tests in the pipeline.


ArthurHeymans (approved):

swenson (approved):
LGTM

PR #2858, opened 2025-12-01
Use segmented bitstream for subsystem

clundin25:


ArthurHeymans:
This file is also used for the 2.1 images. Those would be gone then?

clundin25:
> This file is also used for the 2.1 images. Those would be gone then?

Yes, 2.0 and 2.1 will share a base image (built by this job) and pull the "bulk" of the bitstream at runtime.

.github/workflows/fpga-image.yml
```diff
@@ -29,20 +29,16 @@ jobs:
...
             image_variant: core
-          - bitstream_hash: 186dd345e5f051d6d9c2f5119ff05c0563fa4611126acf857de4bd60d6005a80
-            image_variant: subsystem-2.0
-          - bitstream_name: xsa_ss2p0_85443bba.xsa
-            image_variant: subsystem-2.0
-          - bitstream_hash: 9b1bd1df6a4842f5d0481dd817c15cd8b1dd14e03bbb0ccf984c5ba5e57fdecc
-            image_variant: subsystem-2.1
-          - bitstream_name: xsa_ss2p1_6ccc059f.xsa
-            image_variant: subsystem-2.1
+          - bitstream_hash: 8fade301b2d2d4864a8586b5fea2cf8a80a76102afc711a940f66d7df79e30b3
```
timothytrippel, line 38: Might be worth dropping a comment here saying that the base image is built with the this bitstream and the final bitstream is loaded at runtime? if that is the case now?

timothytrippel: @clundin25 will add an xtask flow to show how to reload bitstreams runtime.

Issue #2680, opened 2025-10-29
Caliptra Version in 2.1 branch certificates


clundin25:
The template currently uses `2.0`. We should update the template to `2.1`.

Original comment: https://github.com/chipsalliance/caliptra-sw/pull/2673#discussion_r2475883099

PR #2864, opened 2025-12-01
Update certificate template version to 2.1

mtimkovich:
#2680


x509/build/build.rs
```diff
@@ -57,14 +57,14 @@ fn gen_init_devid_csr(out_dir: &str) {
         .add_basic_constraints_ext(true, 5)
         .add_key_usage_ext(usage)
         .add_ueid_ext(&[0xFF; 17]);
-    let template = bldr.tbs_template("Caliptra 2.0 Ecc384 IDevID");
+    let template = bldr.tbs_template("Caliptra 2.1 Ecc384 IDevID");
```
zhalvorsen: nit: Could we make constants for these strings because a bunch of them are duplicated?

zhalvorsen (dismissed):

x509/build/build.rs
```diff
@@ -94,7 +94,7 @@ fn gen_fmc_alias_csr(out_dir: &str) {
                 },
             }],
         );
-    let template = bldr.tbs_template("Caliptra 2.0 FMC Alias");
+    let template = bldr.tbs_template("Caliptra 2.1 FMC Alias");
```
mhatrevi: Unrelated to this change but this should be "Caliptra 2.1 Ecc384 FMC Alias"

x509/build/build.rs
```diff
@@ -94,7 +94,7 @@ fn gen_fmc_alias_csr(out_dir: &str) {
...
     CodeGen::gen_code("FmcAliasCsrTbs", template, out_dir);
```
mhatrevi: This should be "FmcAliasCsrTbsEcc384"

mhatrevi (changes requested):
Please see the feedback.

mhatrevi (approved):

mhatrevi (changes requested):
The new templates need to be generated and checked in.

mhatrevi (dismissed):

mhatrevi (approved):

PR #2866, opened 2025-12-01
Split bitstream downloader into a library and CLI

clundin25:
This way the caliptra-bitstream-downloader crate can be used from xtask.

### DIFF
--- a/ci-tools/bitstream-downloader/Cargo.toml
+++ b/ci-tools/bitstream-downloader/Cargo.toml
@@ -1,14 +1,15 @@
 # Licensed under the Apache-2.0 license
 
 [package]
-name = "bitstream-downloader"
+name = "caliptra-bitstream-downloader"
 version = "0.1.0"
 edition = "2024"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


ci-tools/bitstream-downloader/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 # Licensed under the Apache-2.0 license
 
 [package]
-name = "bitstream-downloader"
+name = "caliptra-bitstream-downloader"
```
timothytrippel, line 4: It is nice when the library name is the same as the path name (makes it easier to find), i.e. can you keep the name as you had it?

clundin25: In this case I'd prefer to have the caliptra pre-fix, so it's clear in other repos that this is an internal crate

clundin25: I can follow up with a folder rename - going to merge to unblock updating MCU

PR #2867, opened 2025-12-02
fix(test): Use warm_reset_flow and enable tests on FPGA

ArthurHeymans:
Replace warm_reset() calls with warm_reset_flow().unwrap() to use the correct reset functionality. Remove cfg guards that excluded fpga_realtime and fpga_subsystem features, allowing warm reset tests to run on FPGA hardware.


meilu-git:
looks good. and test passed

mhatrevi (approved):

PR #2872, opened 2025-12-02
hw-model: Split OTP in two we can use half for DOT flash

swenson:
The OTP memory is 16384 bytes, but only about ~3300 bytes are used at the moment for the default fuse map.

We need a simulated flash or NV memory for Device Ownership Transfer in MCU. It is much easier to use some of the unused space in the OTP memory than it would be to add another BRAM in the FPGA, wire it up, etc.

So, we limit the OTP space to be only 8192 bytes and reserve the other 8192 bytes for a simulated "flash" storage. (Mostly we need to prevent the OTP provisioning from clearing this storage, so we limit the otp_slice to 8192 bytes.)


mhatrevi (approved):

swenson:
Thanks!

PR #2874, opened 2025-12-02
Always scramble FPGA MAC address

clundin25:
Since the FPGAs share a MAC address, they will step over each other if on the same network.


clundin25:
> I think it's possible to randomize the MAC and save it on the FPGA when it is going through its bootloader, maybe with `saveenv`?
> 
> But LGTM either way.

Good idea, I think it would be good to centralize the FPGA documentation and add that. This method scales well for the CI atleast

timothytrippel (approved):

PR #2887, opened 2025-12-04
Add emulator support for OCP LOCK DOE and fuses HEK Seeds

clundin25:


sw-emulator/lib/periph/src/doe.rs
```diff
@@ -380,15 +418,89 @@ mod tests {
...
+        let mut doe = Doe::new(&clock, key_vault.clone(), soc_reg);
+
+        let mut iv = [0u8; DOE_IV_SIZE];
+        iv.to_big_endian();
+
+        for i in (0..iv.len()).step_by(4) {
+            assert_eq!(
+                doe.write(RvSize::Word, OFFSET_IV + i as RvAddr, make_word(i, &iv))
+                    .ok(),
+                Some(())
```
zhalvorsen: nit: It looks like this was copy/pasted from another test, but instead of converting `Result<(), _>` to `Option<()>` then matching `Some`, it would be cleaner to either use `assert!(*, Ok(()))` or `assert!(*.is_ok())`

sw-emulator/lib/periph/src/doe.rs
zhalvorsen, line 503: I think it would make sense to add the hek_seed to `CaliptraRootBusArgs` as well

clundin25: Hmm I am not sure I understand, why is that?

zhalvorsen: Never mind, looking at it closer I agree it doesn't make sense there

zhalvorsen (approved):

PR #2888, opened 2025-12-04
Add OCP LOCK GET_ALGORITHMS command

clundin25:


runtime/README.md
```diff
@@ -1321,6 +1321,49 @@ Command Code: `0x4143_5446` ("ACTF")
...
+| -------- | -------- | --------------- |
+| chksum   | u32      | Checksum over other input arguments, computed by the caller. Little endian. |
+
+*Table: `GET_ALGORITHMS` output arguments*
+
+| **Name**                 | **Type**              | **Description** |
+| ------------------------ | --------------------- | --------------- |
+| chksum                   | u32                   | Checksum over other output arguments, computed by Caliptra. Little endian. |
+| fips_status              | u32                   | Indicates if the command is FIPS approved or an error. |
+| endorsement_algorithms   | EndorsementAlgorithms | Supported endorsement algorithms. |
```
zhalvorsen: Should we add the reserved words here?

runtime/tests/runtime_integration_tests/test_ocp_lock/test_get_algorithms.rs
```diff
@@ -0,0 +1,59 @@
...
+        );
+        assert_eq!(get_algs_resp.hpke_algorithms, HpkeAlgorithms::all());
+        assert_eq!(get_algs_resp.access_key_sizes, AccessKeySizes::LEN_256);
+    } else {
+        assert_eq!(
+            response.unwrap_err(),
+            ModelError::MailboxCmdFailed(
+                CaliptraError::RUNTIME_OCP_LOCK_UNSUPPORTED_COMMAND.into(),
+            )
+        );
```
zhalvorsen, line 63: I like this negative test pattern!

zhalvorsen (dismissed):

runtime/README.md
clundin25: Oh, nice catch, thank you!

api/src/mailbox.rs
```diff
@@ -631,6 +634,7 @@ impl MailboxReq {
             MailboxReq::CmEcdsaVerify(req) => req.as_bytes_partial(),
             MailboxReq::CmDeriveStableKey(req) => Ok(req.as_bytes()),
             MailboxReq::ReportHekMetadata(req) => Ok(req.as_bytes()),
+            MailboxReq::GetAlgorithms(req) => Ok(req.as_bytes()),
```
bluegate010: Would it make sense to prefix these commands with "Lock" just to collate them together?

runtime/README.md
```diff
@@ -1321,6 +1321,50 @@ Command Code: `0x4143_5446` ("ACTF")
 | chksum         | u32            | Checksum over other output arguments, computed by Caliptra. Little endian. |
 | fips_status    | u32            | Indicates if the command is FIPS approved or an error.                     |
 
+### GET_ALGORITHMS
```
bluegate010: Just flagging that we already have a definition of these mailbox commands at https://github.com/chipsalliance/Caliptra/blob/5a89921fb1ee5d4d7fe4b873ab3540cbd5f1e98c/doc/ocp_lock/lock_spec.ocp#L1290. Should we dedupe?

api/src/mailbox.rs
clundin25: @bluegate010 that sounds great. I plan to feature gate OCP LOCK, so it will make it easier to read.

I'll assume that we can leave the command ID as is, but feel free to ping me if that's not the case.

runtime/README.md
clundin25: Great point. I will make this a pointer to the OCP LOCK spec.

api/src/mailbox.rs
```diff
@@ -494,7 +496,7 @@ pub enum RomMailboxResp {
     CmDeriveStableKey(CmDeriveStableKeyResp),
     CmRandomGenerate(CmRandomGenerateResp),
     CmHmac(CmHmacResp),
-    ReportHekMetaData(ReportHekMetadataReq),
+    ReportHekMetaData(OcpLockReportHekMetadataReq),
```
bluegate010, line 499:
```suggestion
    OcpLockReportHekMetaData(OcpLockReportHekMetadataReq),
```

runtime/README.md
```diff
@@ -1321,6 +1321,12 @@ Command Code: `0x4143_5446` ("ACTF")
 | chksum         | u32            | Checksum over other output arguments, computed by Caliptra. Little endian. |
 | fips_status    | u32            | Indicates if the command is FIPS approved or an error.                     |
 
+## Mailbox commands: OCP LOCK v1.0
+
+These commands are defined in the OCP LOCK v1.0 [specification](https://github.com/chipsalliance/Caliptra/blob/main/doc/ocp_lock/releases/OCP_LOCK_Specification_v1.0_RC2.pdf).
+
+### GET_ALGORITHMS
```
bluegate010, line 1328: Is the idea that there'd be headers here for each implemented command?

api/src/mailbox.rs
clundin25: Whoops, thanks!

I'll fix this in my next PR, so I can merge this one.

runtime/README.md
clundin25: Yes, something to tie the two together. I imagine we will want to flesh this out more closer to release

bluegate010: Ack. Then I guess we add report-hek-metadata later once it's fully implemented.

mhatrevi (approved):

PR #2889, opened 2025-12-04
Remove sw-emulator message from DMA code

ArthurHeymans:
I used this to identify if fpga tests were accidentally using the sw-emulator codepath and it accidentally crept in a PR.


clundin25 (dismissed):

clundin25:
@ArthurHeymans I think this needs to go into `main`. 

We've updated the Branches per https://github.com/chipsalliance/Caliptra/issues/586

PR #2895, opened 2025-12-04
OCP LOCK: Add INITIALIZE_MEK_SECRET

clundin25:


api/src/mailbox.rs
```diff
@@ -188,6 +188,7 @@ impl CommandId {
     // OCP LOCK Commands
     pub const OCP_LOCK_REPORT_HEK_METADATA: Self = Self(0x5248_4D54); // "RHMT"
     pub const OCP_LOCK_GET_ALGORITHMS: Self = Self(0x4741_4C47); // "GALG"
+    pub const OCP_LOCK_INITIALIZE_MEK_SECRET: Self = Self(0x494D_0B53); // "IKMS"
```
bluegate010:
```suggestion
    pub const OCP_LOCK_INITIALIZE_MEK_SECRET: Self = Self(0x494D_4B53); // "IMKS"
```

Looks like we also need to fix the mailbox code in https://github.com/chipsalliance/Caliptra/blob/6115e27ff9e75fbdb5b9ed2e05897b4195394ed7/doc/ocp_lock/lock_spec.ocp#L1652, which uses 0B instead of 4B for whatever reason.

runtime/src/ocp_lock/mod.rs
```diff
@@ -22,4 +115,24 @@ impl OcpLockContext {
...
+        sek: Sek,
+        dpk: Dpk,
+    ) -> CaliptraResult<()> {
+        if !self.hek_available {
+            return Err(CaliptraError::RUNTIME_OCP_LOCK_HEK_UNAVAILABLE);
+        } else {
+            cfi_assert!(self.hek_available);
+        }
+        let epk = Epk::new(hmac, keyvault, trng, sek)?;
+        let mek_secret_seed = epk.derive_mek_secret_seed(dpk)?;
```
bluegate010: Feel free to totally ignore this, but there would be some symmetry if this were done via an MekSecretSeed::new() function.

runtime/src/ocp_lock/mod.rs
```diff
@@ -1,19 +1,112 @@
...
+pub struct Sek(pub [u8; 32]);
+
+#[derive(IntoBytes, KnownLayout, Immutable, ZeroizeOnDrop)]
+/// Represents the DPK type from OCP LOCK.
+pub struct Dpk(pub [u8; 32]);
+
+#[derive(IntoBytes, KnownLayout, Immutable, ZeroizeOnDrop)]
+/// Represents the MEK Secret Seed type from OCP LOCK.
+/// This is constructed from the EPK + DPK.
+pub struct MekSecretSeed(pub Array4x16);
```
bluegate010: According to https://chipsalliance.github.io/Caliptra/ocp-lock/specification/HEAD/pdf/#page=18, MEK secrets are held in KV21. I realize now that that table mentions MEK secrets and MEK seeds, but not MEK secret seeds, which are a precursor to MEK secrets. Naming is fun...

I think we can hold the MEK secret seed in the same KV slot as the MEK secret, KV21.

api/src/mailbox.rs
clundin25: Ah thanks ! I opened https://github.com/chipsalliance/Caliptra/pull/589 to fix the spec.

runtime/src/ocp_lock/mod.rs
clundin25: Oh thanks for clearing this up. I misunderstood! Will fix this!

runtime/src/ocp_lock/mod.rs
clundin25: I like that, will include this in the next rev.

clundin25: I think the only caveat to this change is that the lifetime of the EPK becomes less obvious (EPK erases KV on drop).

I'll think on this some more, but will leave the code as is in the PR for now.

runtime/src/ocp_lock/mod.rs
```diff
@@ -1,19 +1,123 @@
...
+pub struct Dpk(pub [u8; 32]);
+
+/// Represents the MEK Secret from OCP LOCK.
+/// This is constructed from the EPK + DPK.
+///
+/// NOTE: MPKs will be mixed into `MekSecretSeed`
+pub struct MekSecretSeed;
+
+// TODO(clundin): Erase MEK_SECRETS KV on drop? That would require holding a reference to the
+// key vault for the lifetime of `MekSecretSeed`
```
bluegate010, line 35: I think that if we did this we'd need to be sure we hold a reference to the `MekSecretSeed` the whole time we're using the MEK secret derived from it. If we derive the MEK secret into the same KV slot, and then drop `MekSecretSeed`, that would cause the MEK secret to be zeroed prematurely.

We can totally leave it alone for now, but on reflection, since this has no fields and no instance methods, wondering if it even needs to be a struct; could just be a free function. `OcpLockContext::mek_secret_seed` could just be a bool tracking whether the seed is initialized. We could always revisit this later.

runtime/src/ocp_lock/mod.rs
```diff
@@ -1,19 +1,123 @@
...
 
 /// Provides OCP LOCK functionalities.
 pub struct OcpLockContext {
+    /// OCP LOCK is supported on both HW and FW
     available: bool,
+
+    /// HEK is available
+    hek_available: bool,
+
+    /// Holds MEK secret seed, initialized by calling MEK_SECRET_INITIALIZE. Some commands do not work until
```
bluegate010, line 108: Feel free to fix in a follow-up.

```suggestion
    /// Holds MEK secret seed, initialized by calling INITIALIZE_MEK_SECRET. Some commands do not work until
```

runtime/src/ocp_lock/mod.rs
clundin25: Sounds good - I think as the rest of the features land the approach will become clearer.

runtime/src/ocp_lock/mod.rs
clundin25: Will do, thanks!

PR #2905, opened 2025-12-05
chore: Use dpe cargo dependency instead of a submodule

swenson:
Submodules are the worst and aren't necessary since Cargo support git dependencies.

This will reduce the potential for problems caused by accidentally reverting or updating the submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,10 +5,6 @@
 	path = hw/latest/rtl
 	url = https://github.com/chipsalliance/caliptra-rtl
 	branch = main
-[submodule "dpe"]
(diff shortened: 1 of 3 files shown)


zhalvorsen (approved):

swenson:
Sadly, it isn't as easy as this. The fuzzing tests and the tests/dpe_verification Go module also use the submodule, so we would be trading the complexity of us keeping one submodule up-to-date vs. potentially syncing several revisions between Cargo.toml, tests/dpe_verification/go.mod, and some GitHub workflows.

PR #2906, opened 2025-12-05
Fix caliptra-test and add it to fpga_subsystem CI

ArthurHeymans:


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1317,13 +1321,60 @@ impl ModelFpgaSubsystem {
...
         // keys passed in `prod_dbg_unlock_keypairs` field in InitParams.
         println!("Provisioning SW_MANUF partition.");
-        let mem = otp_generate_sw_manuf_partition_mem(&OtpSwManufPartition {
-            anti_rollback_disable: u32::from(self.fuses.anti_rollback_disable),
-            ..Default::default()
-        })?;
+        let mem =
+            otp_generate_sw_manuf_partition_mem(&OtpSwManufPartition {
+                anti_rollback_disable: u32::from(self.fuses.anti_rollback_disable),
+                idevid_cert_attr: self
```
swenson: nit: Would it be easier to do a `zerocopy::transmute!`, reference conversion, or flat map to map words to bytes?

swenson (dismissed):
LGTM, one kind of nit

hw-model/src/model_fpga_subsystem.rs
ArthurHeymans: Done

ArthurHeymans:
abandonned in favor https://github.com/chipsalliance/caliptra-sw/pull/3017 as I have accidentally rebased this first and added more fixes.

PR #2910, opened 2025-12-05
Pull DPE persistent info into a struct

zhalvorsen:
We decided to separate the persistent data into ROM and non-ROM partitions. This change pulls all DPE related information into its own struct which helps to keep this information self contained and will make it easier to move around in following PRs.

The final persistent data will look something like this from @clundin25 's comment below:

```
persistent_data
├── rom
└── runtime
    └── dpe
```

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -56,8 +56,7 @@ pub const CMB_AES_KEY_SHARE_SIZE: u32 = 32;
 pub const DOT_OWNER_PK_HASH_SIZE: u32 = 13 * 4;
 pub const OCP_LOCK_METADATA_SIZE: u32 = 8;
 pub const CLEARED_NON_FATAL_FW_ERROR_SIZE: u32 = 4;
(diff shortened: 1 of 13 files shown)


drivers/src/persistent.rs
```diff
@@ -600,6 +569,25 @@ impl PersistentDataAccessor {
     }
 }
 
+#[cfg(feature = "runtime")]
+#[repr(C)]
+#[derive(IntoBytes, TryFromBytes, KnownLayout, Zeroize)]
+pub struct DpePersistentData {
```
clundin25, line 575: Do you want to make this a part of a `RuntimePersistentData`?

I'm thinking something like:

```
persistent_data
├── rom
└── runtime
    └── dpe
```

That would prevent mixing ROM and Runtime produced data

zhalvorsen: Yeah, that is my end goal. I wanted to move this now as a cleanup before the larger separation

zhalvorsen: I like your text graphic. I put it in the commit message

PR #2918, opened 2025-12-06
OCP LOCK: Add DERIVE_MEK

clundin25:


runtime/src/ocp_lock/mod.rs
```diff
@@ -3,58 +3,115 @@
...
+        )?))
+    }
+}
+
+impl Drop for MekSecret<'_> {
+    // From Spec:
+    //   > Held in the Key Vault and zeroized after each use.
+    fn drop(&mut self) {
+        // We don't set a write or use lock so this should never fail.
+        let _ = self.kv.erase_key(KEY_ID_MEK_SECRETS);
```
bluegate010, line 85: I guess this is one consequence of this application of RAII, that we can't surface errors that occur here. Comments above like "NOTE: This will erase the `MEK_SECRET` key vault" make me wonder if explicitly zeroing secrets would be simpler to reason about. Up to you.

runtime/src/ocp_lock/mod.rs
```diff
@@ -3,58 +3,115 @@
...
 /// NOTE: MPKs will be mixed into `MekSecretSeed`
 pub struct MekSecretSeed;
 
-// TODO(clundin): Erase MEK_SECRETS KV on drop? That would require holding a reference to the
-// key vault for the lifetime of `MekSecretSeed`
 impl MekSecretSeed {
-    /// Consumes the EPK and DPK to produce a `MekSecretSeed`
-    fn new(hmac: &mut Hmac, trng: &mut Trng, epk: Epk, dpk: Dpk) -> CaliptraResult<Self> {
+    /// Consumes `Self` to produce a `MekSecret`.
+    fn derive<'a>(
```
bluegate010:
```suggestion
    fn derive_mek_secret<'a>(
```

runtime/src/ocp_lock/mod.rs
```diff
@@ -3,58 +3,115 @@
...
+/// Represents the MEK Secret from OCP LOCK.
+/// This is constructed from `MekSecretSeed`
+pub struct MekSecret<'a> {
+    kv: &'a mut KeyVault,
+}
+
+impl MekSecret<'_> {
+    /// Consumes `Self` to produce a `MekSeed`.
+    /// NOTE: This will erase the `MEK_SECRET` key vault.
+    fn derive_seed(self, aes: &mut Aes) -> CaliptraResult<MekSeed> {
```
bluegate010:
```suggestion
    fn derive_mek_seed(self, aes: &mut Aes) -> CaliptraResult<MekSeed> {
```
Because we have two seeds at play here, best be clear.

We could think about a different name though, and make the change in the spec. Maybe "MEK secret seed" becomes "Intermediate MEK Secret" or IMS. EPK->IMS->MEK secret->MEK.

runtime/src/ocp_lock/mod.rs
```diff
@@ -142,8 +221,45 @@ impl OcpLockContext {
...
+            return Err(CaliptraError::RUNTIME_OCP_LOCK_MEK_NOT_INITIALIZED);
+        };
+        let mek_secret = mek_secret_seed.derive(hmac, trng, kv)?;
+        let mek_seed = mek_secret.derive_seed(aes)?;
+        let checksum = mek_seed.checksum(aes)?;
+
+        // If `expect_mek_checksum` is all zeros, skip checking that the MEK checksum matches.
+        // `expect_mek_checksum` should match the `checksum` that was derived.
+        // Otherwise we need to report an error.
+        if !constant_time_eq(
```
bluegate010: These checksums aren't secret, so we needn't go out of our way to do a constant-time compare.

runtime/src/ocp_lock/mod.rs
clundin25: > I guess this is one consequence of this application of RAII, that we can't surface errors that occur here.

Unfortunate but should be okay, if for some reason this KV got locked it would require a reset to get things working.

> Comments above like "NOTE: This will erase the MEK_SECRET key vault" make me wonder if explicitly zeroing secrets would be simpler to reason about. Up to you.

I think the error flow is more complicated, since we'd need to check in multiple spots if the KV should be erased. 

I'll leave it for now, but will revisit this to make sure the error handling / drop behavior is reasonable

runtime/src/ocp_lock/mod.rs
clundin25: I like "Intermediate MEK Secret", updated the code, and will update the spec.

runtime/src/ocp_lock/mod.rs
clundin25: Sounds good!

error/src/lib.rs
```diff
@@ -1629,9 +1630,24 @@ impl CaliptraError {
...
+        ),
+        (
+            RUNTIME_OCP_LOCK_INVALID_MEK_SEED_SIZE,
+            0x000E0078,
+            "OCP LOCK Error: Invalid MEK seed size"
+        ),
+        (
+            RUNTIME_OCP_LOCK_MEK_CHKSUM_FAIL,
+            0x000E007A,
+            "OCP LOCK Error: Error in derivation caused MEK mismatch"
```
clundin25, line 1650: Note: I will update this to `0x000E0079` in a separate PR, unless there is other feedback (CI just takes a long time)

runtime/src/ocp_lock/mod.rs
```diff
@@ -142,8 +220,43 @@ impl OcpLockContext {
...
+        };
+        let mek_secret = mek_secret_seed.derive_intermediate_mek_secret(hmac, trng, kv)?;
+        let mek_seed = mek_secret.derive_mek_seed(aes)?;
+        let checksum = mek_seed.checksum(aes)?;
+
+        // If `expect_mek_checksum` is all zeros, skip checking that the MEK checksum matches.
+        // `expect_mek_checksum` should match the `checksum` that was derived.
+        // Otherwise we need to report an error.
+        if expect_mek_checksum.as_bytes() != MekChecksum::default().as_bytes()
+            && expect_mek_checksum.as_bytes() != checksum.as_bytes()
```
zhalvorsen: nit: You could make `MekCheckum` `PartialEq` and `Eq` so you could simplify this statement to:

```Rust
        if expect_mek_checksum != MekChecksum::default()
            && expect_mek_checksum != checksum
```

runtime/tests/runtime_integration_tests/test_ocp_lock/test_derive_mek.rs
```diff
@@ -0,0 +1,217 @@
...
+];
+
+// TODO(clundin): Follow up with the following test cases:
+//
+// These follow ups are pending an independently constructed key ladder / splitting into smaller
+// PRs.
+//
+// * Release MEK and validate MEK contents.
+// * Validate MEK checksum is correct.
+// * MEK and MEK checksum are the same after a warm reset.
```
zhalvorsen, line 27: It would be good to have a hitless update test as well to make sure we don't distort the key when firmware with a different measurement is loaded

runtime/tests/runtime_integration_tests/test_ocp_lock/test_derive_mek.rs
```diff
@@ -0,0 +1,217 @@
...
+            sek: [0xAB; 32],
+            dpk: [0xCD; 32],
+        }),
+        ..Default::default()
+    });
+
+    let mut cmd = MailboxReq::OcpLockDeriveMek(OcpLockDeriveMekReq {
+        hdr: MailboxReqHeader { chksum: 0 },
+        reserved: 0,
+        mek_checksum: [0; 16],
```
zhalvorsen: Should this be the expected checksum? There is a test below with default value. Maybe I'm missing something.


```suggestion
        mek_checksum: EXPECTED_MEK_CHECKSUM,
```

runtime/src/ocp_lock/mod.rs
clundin25: Nice! I keep forgetting about this one :)

runtime/tests/runtime_integration_tests/test_ocp_lock/test_derive_mek.rs
clundin25: A definitely a good call. I've added it to the list.

runtime/tests/runtime_integration_tests/test_ocp_lock/test_derive_mek.rs
clundin25: Yes it should, thanks for catching this!

PR #2919, opened 2025-12-06
Divide persistent data fields by ROM and FW

zhalvorsen:
This makes the firmware specific fields more dynamic while still easily maintaining compatibility with the ROM.

This also moves the DPE information to the bottom of the firmware struct where it can be expanded later.

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -28,6 +28,9 @@ use crate::{FmcAliasCsrs, Mldsa87Signature};
 #[cfg(feature = "runtime")]
 use crate::{pcr_reset::PcrResetCounter, KeyId};
 
+#[cfg(any(feature = "fmc", feature = "runtime"))]
+pub use fw::*;
+
 pub const ECC384_MAX_IDEVID_CSR_SIZE: usize = 512;
 pub const ECC384_MAX_FMC_ALIAS_CSR_SIZE: usize = 768;
 pub const MAN1_SIZE: u32 = 17 * 1024;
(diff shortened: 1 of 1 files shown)


drivers/src/persistent.rs
```diff
@@ -361,167 +353,224 @@ pub struct PersistentData {
 
     // TODO(clundin): For runtime we may want to gate this behind a feature flag.
     pub ocp_lock_metadata: OcpLockMetadata,
```
clundin25, line 354: More of a follow up note, but I think some (maybe all?) of this will be in ROM.

I'll do that in a separate PR

drivers/src/persistent.rs
```diff
@@ -361,167 +353,224 @@ pub struct PersistentData {
...
+    pub rtalias_mldsa_tbs_size: u16,
+    reserved1: [u8; 2],
+    pub rt_dice_mldsa_sign: Mldsa87Signature,
+
+    #[cfg(feature = "runtime")]
+    pub pcr_reset: PcrResetCounter,
+    #[cfg(feature = "runtime")]
+    reserved7: [u8; PCR_RESET_COUNTER_SIZE as usize - size_of::<PcrResetCounter>()],
+
+    #[cfg(not(feature = "runtime"))]
```
clundin25, line 488: Since the whole struct is feature flag gated, can the inline feature flags be removed?

zhalvorsen: These internal feature flags are for FMC because it doesn't include those types

drivers/src/persistent.rs
```diff
@@ -35,28 +37,36 @@ pub const FHT_SIZE: u32 = 2 * 1024;
 pub const IDEVID_MLDSA_PUB_KEY_MAX_SIZE: u32 = 3 * 1024;
 pub const ECC_LDEVID_TBS_SIZE: u32 = 1024;
 pub const ECC_FMCALIAS_TBS_SIZE: u32 = 1024;
+#[cfg(any(feature = "fmc", feature = "runtime"))]
```
swenson: nit/idea: maybe move these fmc/runtime sizes into a separate `mod fw` just to reduce the number of `cfg` blocks needed?

drivers/src/persistent.rs
clundin25: Makes sense, thanks!

drivers/src/persistent.rs
zhalvorsen: That's better. Done

PR #2929, opened 2025-12-08
Re-arrange NTP update location

clundin25:


timothytrippel:
CC: @cfrantz

ci-tools/fpga-image/startup-script.sh
```diff
@@ -37,6 +30,18 @@ elif grep -q "overlay" /proc/mounts; then
     macchanger -r end0 || true
     ip link set dev end0 up
 
+    HOST="google.com"
+    while ! ping -c 1 "$HOST" &> /dev/null; do
```
timothytrippel, line 34: should there be a timeout associated with this loop?

clundin25: The FPGA has 3 mins to boot and connect to GitHub before getting reset by FPGA Boss 

https://github.com/chipsalliance/caliptra-sw/blob/main/ci-tools/fpga-boss/src/main.rs#L417

PR #2931, opened 2025-12-08
rom/runtime: Use a block size of 64 for DMA AXI to AXI transfers

swenson:
Our current code path always uses 256 as the DMA block size, which was the guidance when the code was written in Feburary 2025, but the HW spec changed in March to require a block size of 64 for AXI to AXI transfers.


swenson:
Thanks!

swenson:
Holding, as there might be more restrictions on what scenarios the block size needs to be set to 64 or 256.

swenson:
Closing and will open a new one when we have this right.

PR #2933, opened 2025-12-09
Move stacks to beginning of DCCM

### LINKED ISSUE #1842: Move the stack to the be at the bottom of DCCM for Caliptra 2.0
This way stack overflows trigger a memory fault.

zhalvorsen:
Fixes #1842

This has the benefit of causing a hardware fault if there is a stack
overflow.

To allow persistent data and FW stack to grow in future releases,
persistent data has been moved to the end so all extra bytes in DCCM
will be kept in the middle of memory. This means that all fields added
to persistent data (ROM or FW) will need to be added at the beginning of
the struct.

In a follow on change a marker and version number will be added to
ensure this is handled correctly across warm resets and hitless updates.


clundin25 (approved):
nit: Size optimizations should be in a separate commit (in the context of the PR, doesn't have to be a different PR)

drivers/src/memory_layout.rs
```diff
@@ -22,24 +22,49 @@ use crate::PersistentData;
...
+///
+/// Layout
+///
+/// | Partition       | Field        | Start       | End         | Size                 |
+/// |-----------------|--------------|-------------|-------------|----------------------|
+/// | DCCM            |              | 0x5000_0000 | 0x5004_0000 | 0x4_0000             |
+/// | Stack           |              | 0x5000_0000 | 0x5002_3800 | 0x2_3800             |
+/// |                 | FW Stack     | 0x5000_0000 | 0x5002_3000 | 0x2_3000             |
+/// |                 | ROM Stack    | 0x5000_0000 | 0x5001_2c00 | 0x1_2c00             |
+/// |                 | EStack       | 0x5002_3000 | 0x5002_3400 | 0x400                |
```
swenson, line 36: I worry a little about keeping the memory_layout.rs code, this comment, and the `rom.ld` in sync moving forward.

timothytrippel: is it possible to autogenerate all of these somehow from a single source of truth?

zhalvorsen: I tend to agree, but I had to spent a decent amount of time understanding the memory map with the previous code. Any suggestions? We could adopt something similar to the runtime firmware for ROM and the test harnesses to auto-generate the linker file, but that would still leave two sources of truth

zhalvorsen: Is it okay, if I do this is a later PR?

timothytrippel: Oh yeah absolutely, I was just curious. Perhaps we can file an issue to track this more longer term cleanup.

zhalvorsen: Created #2956 to track

PR #2934, opened 2025-12-09
Refactor hostrunner config

clundin25:
* Make creating an FPGA service a function call.
* Centralize scripts to make them easier to discover.
* Update names of scripts / jobs to be clearer.

### DIFF
--- a/ci-tools/host-runner/hostrunners/kir-1.nix
+++ b/ci-tools/host-runner/hostrunners/kir-1.nix
@@ -1,108 +1,39 @@
 {
   config,
+  lib,
   pkgs,
   user,
-  fpga-boss-script,
+  rtool,
+  fpga-boss,
   ...
 }:
+let
+  fpga_service = import ../fpga-service.nix {
+    inherit
+      pkgs
+      rtool
+      fpga-boss
+      user
+      ;
+  };
+in
 {
-  systemd.user.services.vck-7 = {
-    enable = true;
-    description = "VCK-7 Service";
-    after = [ "network.target" ];
-    wantedBy = [ "multi-user.target" ];
-
-    serviceConfig = {
-      Type = "simple";
-      ExecStart = "${fpga-boss-script}/bin/fpga.sh";
-      Restart = "on-failure";
-      RestartSec = "15s";
-      Environment = [
-        ''ZCU_FTDI="1-1.2"''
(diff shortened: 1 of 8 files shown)


clundin25:
@timothytrippel LMK if I should update the SVL hostrunner.

ci-tools/host-runner/hostrunners/kir-1.nix
```diff
@@ -1,108 +1,39 @@
...
-        export ZCU_FTDI="1-1.3.2"
-        export ZCU_SDWIRE="1-1.3.1"
-
-        caliptra-fpga-boss --zcu104 $ZCU_FTDI --sdwire $ZCU_SDWIRE "$@"
-      '')
-    )
+in
+{
+  config = lib.mkMerge [
+    (fpga_service.mkVckSubsystemJob "caliptra-kir-vck-7" {
```
timothytrippel: nit: should these be ordered in incrementing order (i.e. `vck-0 ... vck-7`) to match the kir-0 config?

ci-tools/host-runner/scripts.nix
```diff
@@ -0,0 +1,80 @@
+{
+  pkgs,
+  rtool,
+  fpga-boss,
+  user,
+  ...
+}:
+{
+  download-image = pkgs.writeShellScriptBin "download-image" ''
```
timothytrippel, line 9: rather than declare these scripts inline, is it possible to do something like:

```
scriptContent = builtins.readFile ./my-script.sh;
myScript = pkgs.writeShellScriptBin "my-command-name" scriptContent;
```

i.e., define the scripts each in their own file in a `scripts/` subdir? might scale better as we define more scripts. WDYT?

timothytrippel (approved):
This is much cleaner, thanks @clundin25 !

Yes, please make this change to the svl config too.

ci-tools/host-runner/hostrunners/kir-1.nix
clundin25: Done!

ci-tools/host-runner/scripts.nix
clundin25: I think it's reasonable but I left if off for now.

zhalvorsen:
This looks a lot better, thanks!

PR #2938, opened 2025-12-09
rom/runtime: Use a block size of 64 for DMA I3C transfers

swenson:
Our current code path always uses 256 as the DMA block size, which was the guidance when the code was written in Feburary 2025, but the HW spec changed in March to change the requirement to be 64 for I3C transfers and 0 otherwise.

We simplify the code path a little and use an i3c parameter instead of passing the block size.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -55,7 +55,7 @@ impl AesDmaMode {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct AxiAddr {
     pub lo: u32,
     pub hi: u32,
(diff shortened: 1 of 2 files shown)


ArthurHeymans (approved):

drivers/src/dma.rs
```diff
@@ -563,6 +572,11 @@ impl<'a> DmaRecovery<'a> {
         mmio.check_error(t)
     }
 
+    fn is_i3c(&self, addr: AxiAddr) -> bool {
+        let i3c = self.recovery_base + Self::INDIRECT_FIFO_DATA_OFFSET;
+        i3c.hi == addr.hi && i3c.lo == addr.lo
```
clundin25: Surprised the `AxiAddr` type doesn't support `PartialEq` & `Eq`

drivers/src/dma.rs
```diff
@@ -116,13 +116,15 @@ impl Add for AxiAddr {
...
 
+/// Used to read from an AXI address and DMA the contents to a variety of targets.
 pub struct DmaReadTransaction {
     pub read_addr: AxiAddr,
     pub fixed_addr: bool,
     pub length: u32,
     pub target: DmaReadTarget,
     pub aes_mode: bool,
     pub aes_gcm: bool,
+    pub i3c_indirect_fifo_data: bool,
```
clundin25: nit: I think the name for this variable is vague on what it does.

Maybe something like?

```
enum FifoMode {
  I3C,
  Other,
}
pub fifo_mode: FifoMode
```

A short comment may also be useful. If I were to set this param, I'd have to read the implementation to grok what it does.

clundin25 (approved):

drivers/src/dma.rs
swenson: Yeah, not sure why I didn't add those. Let me go ahead and fix that and simplify this.

zhalvorsen (approved):

clundin25 (approved):

swenson:
Thanks!

PR #2940, opened 2025-12-09
sw-emulator MCU Register Write Access to caliptra_base, recovery_mci, recovery_ifc, and otp_fc Soc Regs

chrisherczeg:
Add MCU Write Access to SoC Regs:
ss_caliptra_base_addr_l
ss_caliptra_base_addr_h
ss_recovery_mci_base_addr_l
ss_recovery_mci_base_addr_h
ss_recovery_ifc_base_addr_l
ss_recovery_ifc_base_addr_h
ss_otp_fc_base_addr_l
ss_otp_fc_base_addr_h

### DIFF
--- a/sw-emulator/lib/periph/src/soc_reg.rs
+++ b/sw-emulator/lib/periph/src/soc_reg.rs
@@ -734,29 +734,37 @@ struct SocRegistersImpl {
     #[register_array(offset = 0x3c0)]
     fuse_hek_seed: [u32; FUSE_HEK_SEED_SIZE / 4],
 
+    // writable from MCU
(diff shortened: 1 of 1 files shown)


chrisherczeg:
> LGTM
> 
> Ideally we'd have a way to have different access for the SoC vs. Caliptra, but I realize we don't have that fine of granularity in the emulator at the moment.

@swenson agree, I see two ways to approach this. At the bus access level or the protocol level and annotate the transaction with the sender. Former would be more representative of the hardware, latter could be easier to refactor. Would need to scope it out more.

mhatrevi (approved):

PR #2942, opened 2025-12-09
[OCP LOCK]: Gate OCP LOCK commands behind a single command handler

clundin25:
* Ensures all commands check that OCP LOCK is available.
* Ensures that command code is compiled out of the RT firmware if OCP LOCK is not enabled in RT firmware.

### DIFF
--- a/runtime/src/ocp_lock/mod.rs
+++ b/runtime/src/ocp_lock/mod.rs
@@ -1,5 +1,6 @@
 // Licensed under the Apache-2.0 license
 
+use caliptra_api::mailbox::CommandId;
 use caliptra_cfi_lib_git::{cfi_assert, cfi_assert_eq};
 use caliptra_common::keyids::ocp_lock::{KEY_ID_EPK, KEY_ID_HEK, KEY_ID_MEK_SECRETS};
 use caliptra_drivers::{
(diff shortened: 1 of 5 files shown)


zhalvorsen (approved):
Nice! I like that

runtime/src/ocp_lock/mod.rs
```diff
@@ -258,3 +261,29 @@ impl OcpLockContext {
...
 }
+
+/// Entry point for OCP LOCK commands
+pub fn command_handler(
+    cmd_id: CommandId,
+    drivers: &mut Drivers,
+    cmd_bytes: &[u8],
+    resp: &mut [u8],
+) -> CaliptraResult<usize> {
+    if cfg!(feature = "ocp-lock") {
```
zhalvorsen: nit: Could you check for the negative here to return the error? I think it is cleaner to reduce nesting when possible.

```Rust
if !cfg!(feature = "ocp-lock") {
    return Err(CaliptraError::RUNTIME_OCP_LOCK_UNSUPPORTED_COMMAND);
}
```

bluegate010 (approved):

runtime/src/ocp_lock/mod.rs
clundin25: Agreed. I was a bit unsure if the compiler would remove the ocp lock command code but I think your suggestion 
is cleaner and I need to check the assembly output anyway

clundin25: Checked the assembly, and the always returning stripped the OCP LOCK commands from the firmware

zhalvorsen (approved):

PR #2944, opened 2025-12-09
Add a marker and version to persistent data

zhalvorsen:
This will make it easier to evolve in the future and ensure we are interpreting the information correctly on hitless updates.


swenson (approved):
LGTM

Issue #2935, opened 2025-12-09
auth-manifest : tool error for 2.x manifest


jheng-shuo-lin:
- Caliptra Version(s) : 2.x
- Caliptra Mode : Both
- Branch: caliptra-2.1-dev / main-2.x

Hi Caliptra teams,

We are creating an SoC manifest to boot the Caliptra 2.x subsystem MCU runtime, and we are using the  [auth-manifest](https://github.com/chipsalliance/caliptra-sw/tree/main-2.x/auth-manifest)  tool to generate it

However, we may encounter the following error:

<img width="1026" height="109" alt="Image" src="https://github.com/user-attachments/assets/3591e8ab-f802-4fdf-8246-3b6734447e80" />

Even with the --svn option, the command still fails. 
Could you help check this problem? 
I’d also like to confirm whether this is the correct/standard way to create an SoC auth-manifest using the auth-manifest application.
Thanks.

nquarton:
Hello, this is getting fixed in https://github.com/chipsalliance/caliptra-sw/pull/2946

jheng-shuo-lin:
Hello, thanks for your help. It seems to be working now.

PR #2946, opened 2025-12-09
Fixes to auth manifest generation app

nquarton:
Found and fixed 2 bugs when testing generating auth manifests using the app invoked via cargo run (instead of directly calling generate)

### DIFF
--- a/auth-manifest/app/src/main.rs
+++ b/auth-manifest/app/src/main.rs
@@ -38,6 +38,11 @@ fn main() {
                 .value_parser(value_parser!(u32)),
         )
         .arg(
+            arg!(--"svn" <U32> "Manifest Security Version Number")
+                .required(true)
+                .value_parser(value_parser!(u32)),
+        )
+        .arg(
             arg!(--"flags" <U32> "Manifest Flags")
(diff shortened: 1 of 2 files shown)


mtimkovich:
#2935

auth-manifest/app/src/main.rs
mtimkovich, line 92: Should we change these error messages? The args are required so we should never be here, and in the case of this bug it actually obfuscated the real problem (that there was no `--svn` flag).

nquarton: Good question, I looked into this a bit more. You are correct there isn't a path for this message to show up now (you will get a different error about not specifying a require arg). Before, it was also not showing up since the svn arg didn't exist at all (_"Mismatch between definition and access of `svn`. Unknown argument or group id."_). 

The only way I see this show up is if I make the svn arg optional (but still exist) and don't specify it. And in that case, this is the correct message.

So overall I don't think this error message is very necessary but in the event someone did change SVN to optional and make it possible to see it, it seems like the right message. So I would say leaving it as-is makes the most sense to me.

PR #2947, opened 2025-12-10
Reset RECOVERY_CTRL to inactivate state after the image is authenticated

Neal-liu:
Reset RECOVERY_CTRL is missing after authenticating SoC Manifest.

### DIFF
--- a/runtime/src/recovery_flow.rs
+++ b/runtime/src/recovery_flow.rs
@@ -81,6 +81,9 @@ impl RecoveryFlow {
                 drivers.soc_ifc.mci_base_addr().into(),
                 dma,
             );
+            // Reset the RECOVERY_CTRL register Activate Recovery Image field by writing 0x1.
+            dma_recovery.reset_recovery_ctrl_activate_rec_img()?;
+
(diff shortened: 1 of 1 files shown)


clundin25:
@Neal-liu thanks Neal!

`caliptra-2.1-dev` has been migrated to `main` (Please see https://github.com/chipsalliance/Caliptra/issues/586).

Do you mind pointing your PR to `main` instead?

Neal-liu:
Sure!

clundin25:
Hmm sees like this is causing some tests to timeout.

swenson:
Thanks! I'm working on adding a test for this, but I might do that in a separate PR

swenson:
I will pull this commit over to a new PR most likely, as I think this requires a small fix in the hardware model. I'll work on this today.

swenson:
Unfortunately, it looks like this fix causes payload_available to not be asserted in our FPGA. I'm investigating why that is and trying to find a workaround.

swenson:
Thanks @Neal-liu -- I am incorporating your commit in #2966 that includes the additional fix needed for our FPGA to work.

PR #2949, opened 2025-12-10
fpga_subsystem: Set dbg_manuf_service in hw-model

ArthurHeymans:
This requires this and clearing this requires doing this at
the appropriate time so add a loop for probing for ready_for_mb_processing
which also clears the dbg_manuf_service register once IDEVID CSR is generated.

Removing affected tests will be added once #2948 is merged as it would cause conflicts that need
manual resolution anyway.


swenson (approved):
LGTM

clundin25 (approved):

PR #2951, opened 2025-12-10
staging: Separate RomEnv/FmcEnv/RuntimeEnv into crypto and non-crypto parts so that crypto requires KATs before using

swenson:
This is meant to ensure that we cannot use a driver if a KAT has not been run by adding a compile-time protection, requiring KATs to be run before the `RomEnv`, etc., can be created. We thus have to separate the `RomEnv` into two parts: the crypto and the non-crypto parts, and have the KAT initialize the crypto parts.

This will ultimately allow us to remove KATs from ROM for drivers that are not used there.

This provides an example using this pattern with SHA1, one of the simplest drivers.

### DIFF
--- a/kat/src/lib.rs
+++ b/kat/src/lib.rs
@@ -27,7 +27,6 @@ mod hmac_kdf_kat;
 mod kats_env;
 mod lms_kat;
 mod mldsa87_kat;
-mod sha1_kat;
 mod sha256_kat;
 mod sha2_512_384acc_kat;
 mod sha384_kat;
@@ -48,7 +47,6 @@ pub use hmac_kdf_kat::{Hmac384KdfKat, Hmac512KdfKat};
 pub use kats_env::KatsEnv;
 pub use lms_kat::LmsKat;
 pub use mldsa87_kat::Mldsa87Kat;
-pub use sha1_kat::Sha1Kat;
 pub use sha256_kat::Sha256Kat;
 pub use sha2_512_384acc_kat::Sha2_512_384AccKat;
 pub use sha384_kat::Sha384Kat;
@@ -71,8 +69,7 @@ pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<InitializedDrivers> {
     cprintln!("[kat] ++");
 
     cprintln!("[kat] sha1");
-    let mut sha1 = Sha1::default();
-    Sha1Kat::default().execute(&mut sha1)?;
+    let sha1 = Sha1::new()?;
 
     cprintln!("[kat] SHA2-256");
     Sha256Kat::default().execute(env.sha256)?;
--- a/drivers/src/kats/mod.rs
+++ b/drivers/src/kats/mod.rs
@@ -1,0 +1,2 @@
+mod sha1_kat;
+pub use sha1_kat::Sha1Kat;
--- a/drivers/src/kats/sha1_kat.rs
+++ b/drivers/src/kats/sha1_kat.rs
@@ -12,7 +12,7 @@ Abstract:
 
 --*/
 
-use caliptra_drivers::{Array4x5, CaliptraError, CaliptraResult, Sha1};
+use crate::{Array4x5, CaliptraError, CaliptraResult, Sha1};
 
 const EXPECTED_DIGEST: Array4x5 =
     Array4x5::new([0xda39a3ee, 0x5e6b4b0d, 0x3255bfef, 0x95601890, 0xafd80709]);
--- a/drivers/src/lib.rs
+++ b/drivers/src/lib.rs
@@ -36,6 +36,7 @@ pub mod hand_off;
 mod hkdf;
 mod hmac;
 mod hmac_kdf;
+pub mod kats;
 mod key_vault;
 mod kv_access;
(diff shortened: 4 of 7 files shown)


ArthurHeymans:
The rom space is sadly lacking for this :-/ . Maybe removing the explicit KAT from rom should be a part of this?

drivers/src/needs_kat.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// A wrapper around a driver that enforces KAT (Known Answer Test) execution
+/// before allowing access to the underlying driver.
+///
+/// This wrapper ensures that cryptographic operations cannot be performed
+/// without first running a KAT to verify the correctness of the hardware.
+pub struct NeedsKat<T> {
+    driver: T,
+    kat_run: bool,
+}
```
clundin25: I wonder if this invariant can be pushed into the type system? E.g. the driver type can only be instantiated after the KAT is run?

You could make `new` run the kat, or treat `new` as unsafe and add a safe API that runs the KAT before returning the driver object.

swenson: I like that since it means it is a compile-time check instead of a runtime check. It would require a more extensive change to how the `RomEnv`, etc., work, but might be worth it.

swenson: I did some work to split into crypto drivers and non-crypto drivers, and have them initialized by the KAT process itself. Let me know what you think.

(If this approach makes sense, we can start working on moving each driver over and updating the code.)

kat/src/lib.rs
```diff
@@ -55,18 +55,24 @@ pub use sha384_kat::Sha384Kat;
 pub use sha3_kat::Shake256Kat;
 pub use sha512_kat::Sha512Kat;
 
-use caliptra_drivers::cprintln;
+use caliptra_drivers::{cprintln, Sha1};
+
+/// Drivers that have been initialized after KAT execution
+pub struct InitializedDrivers {
+    pub sha1: Sha1,
```
clundin25, line 60: I think whether the KAT being run or not should be pushed into the type itself.

Using `Sha1` as an example:

Approach 1:

```
pub struct Sha1;
impl Sha1 {
    pub fn new() -> Self {
        // Perform KAT test
        let self = Self{};
        Sha1Kat::default().execute(&mut self)?;
        self
    } 
}
```

Approach 2:

```

pub struct RawSha1;
impl RawSha1 {
    /// SAFETY: This type has not performed a KAT test and therefore should be used! See `Sha1` instead.
    unsafe pub fn new() -> Self {
       Self { }
    } 
    pub fn initialize() -> Sha1 {
        let self = Self{};
        Sha1Kat::default().execute(&mut self)?;
        self
    } 
}
pub struct Sha1;
```

I prefer Approach 1 because it guarantees SHA cannot be used without first running a KAT.

Approach 2 would prevent most of the codebase from using SHA without KAT, but still gives the KAT flow an unsafe  flow to use.

WDYT?

clundin25: I will preface this with saying I am unfamiliar with said flow, so I may be missing some relevant bits

swenson: I worry that some of the drivers have multiple KATs, and we really want to avoid duplicated code, since a primary goal of this is to reduce code size. So, for example, having two drivers for AES may require a lot of duplication between the two.

I think the other issue is that the KATs currently live in a crate that depends on the drivers. Refactoring it so the KAT is part of the driver itself could work, but would require more refactoring.

clundin25: > Refactoring it so the KAT is part of the driver itself could work, but would require more refactoring.

That's a good idea, then it's impossible to create the Driver type without first running the KAT, which at compile time guarantees that we have always run the Kat before hand.

swenson: Discussed offline, I think what makes sense will be to do:

```
pub struct Sha1 {
  ...
  _private: (), // prevent instantiation from outside drivers/ package
}
impl Sha1 {
  pub fn new() -> Sha1 {
    let sha1 = unsafe { Sha1::raw_new() };
    sha1_kat(&mut sha1); // KAT in a separate crate still
    sha1
  }
  // *not* pub
  unsafe fn raw_new() -> Sha1 {
    Sha1 { ... } 
  }
  ```
  
So that it is not possible to create the driver without running the KAT.

**Comment by swenson** [comment]
If we could replace each driver with a `LazyCell<Sha1>` (for example), and have the `new()` method run the KAT, then we would not even need to split the `RomEnv` (and other environments). The KATs would take care of themselves. (The KATs would still need to be run as part of a FIPS self-test command, of course).

File: rom/dev/src/rom_env.rs
Comment (swenson) [inline rom/dev/src/rom_env.rs:75]: This needs to be moved out

File: rom/dev/src/rom_env.rs [hunk 3]
```diff
@@ -84,7 +82,7 @@ pub struct RomEnv {
     pub aes: Aes,
```
swenson, line 82: This needs to be moved out

rom/dev/src/rom_env.rs
swenson, line 55: These need to be moved out

PR #2957, opened 2025-12-11
[aes] use u32 instead of u8 for AES all IVs and tags

timothytrippel:
This was originally attempted in #2710, but a couple instances were missed. I noticed this when trying to rev the caliptra-sw repo version in the caliptra-mcu-sw repo.


timothytrippel:
@swenson please confirm this if this is what was intended

PR #2958, opened 2025-12-11
Refactor image verification to use enum for mutually exclusive image sources

ArthurHeymans:
Replace three mutually exclusive fields (image, dma, image_in_mcu) in
FirmwareImageVerificationEnv with a single ImageSource enum. This
enforces mutual exclusivity at the type level, preventing invalid states
where both image sources could be set or the wrong source used with the
wrong flag.

This decreases the rom size by 16 bytes.

### DIFF
--- a/common/src/verifier.rs
+++ b/common/src/verifier.rs
@@ -21,6 +21,14 @@ use zerocopy::{FromBytes, IntoBytes};
 
 use caliptra_drivers::memory_layout::ICCM_RANGE;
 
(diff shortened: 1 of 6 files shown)


clundin25 (approved):

zhalvorsen (approved):
This is great!

swenson (approved):
LGTM

mhatrevi (approved):

PR #2959, opened 2025-12-11
cherry-pick to 2.0: runtime: Use a block size of 64 for DMA I3C transfers (#2938)

swenson:
Our current code path always uses 256 as the DMA block size, which was the guidance when the code was written in Feburary 2025, but the HW spec changed in March to change the requirement to be 64 for I3C transfers and 0 otherwise.

We simplify the code path a little and use an i3c parameter instead of passing the block size.

(cherry picked from commit 9e8e4b15c4756cac9c755109a6a5e83975fd400b)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,14 +1,14 @@
 ##########################################
 #             Code Ownership
 ##########################################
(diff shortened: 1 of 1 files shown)


ArthurHeymans (approved):

swenson:
Thanks!

PR #2960, opened 2025-12-11
[OCP-LOCK]: Set MDK usage to AES

clundin25:
MDK is used to decrypt the MEK


swenson (approved):
LGTM

swenson:
Does this affect a test? I would expect a change like this to be accompanied by a test changing that used it, or was added?

clundin25:
> Does this affect a test? I would expect a change like this to be accompanied by a test changing that used it, or was added?

@swenson the test is not merged, I opened a separate PR since it needs a few others to land beforehand

https://github.com/chipsalliance/caliptra-sw/pull/2961

PR #2961, opened 2025-12-11
[OCP-LOCK]: MEK release and independent verification

clundin25:


runtime/src/ocp_lock/mod.rs
```diff
@@ -256,8 +262,13 @@ impl OcpLockContext {
             return Err(CaliptraError::RUNTIME_OCP_LOCK_MEK_CHKSUM_FAIL);
         }
 
-        // Do not release an MEK with a failing checksum!
-        // TODO(clundin): MEK should be released to KV here.
+        let mek_seed: [u8; 64] = <[u8; 64]>::from(mek_seed.0);
```
clundin25: Should be able to drop the type information from one side. I'll do that in a follow up (unless there is feedback).

zhalvorsen: Seems fine in a follow up. I would have thought that would be a clippy lint, but I guess not

bluegate010: Can we be sure to clean up all copies of the MEK seed from memory?

clundin25: The MEK seed type will be zeroized once dropped [ref](https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/src/ocp_lock/mod.rs#L98-L99).

I will modify the `mek_seed` created on line 265 to be zeroized.

test/src/derive.rs
```diff
@@ -232,6 +251,157 @@ impl IDevId {
...
+        }
+    }
+
+    pub fn add_hek(self, hek_seed: [u32; 8]) -> Self {
+        let mut hek: [u32; 16] = transmute!(hmac512_kdf(
+            swap_word_bytes(&self.cdi).as_bytes(),
+            b"ocp_lock_hek",
+            Some(hek_seed.as_bytes()),
+        ));
+        swap_word_bytes_inplace(&mut hek);
```
zhalvorsen, line 294: optional: it could be nice to have a local helper function for this because it is done so many times

test/src/derive.rs
```diff
@@ -232,6 +251,157 @@ impl IDevId {
...
+        swap_word_bytes_inplace(&mut mek_secret);
+
+        let mek_seed = cmac_kdf(
+            swap_word_bytes(&mek_secret).as_bytes().get(..32).unwrap(),
+            b"ocp_lock_mek_seed",
+            None,
+            4,
+        );
+
+        let key = mek_seed.get(0..32).unwrap().as_bytes();
```
zhalvorsen: nit: If you are going to use `get` and `unwrap` you could also use square brackets. Feel free to update these in a follow up.


```suggestion
        let key = mek_seed[0..32].as_bytes();
```

test/src/derive.rs
clundin25: Going to punt on this for now, will try to clean this up.

test/src/derive.rs
clundin25: Ah good point, did it out of habit.

runtime/src/ocp_lock/mod.rs
clundin25: Updated implementation to not copy the `mek_seed`. Now there should only be one instance that is zeroized on drop.

zhalvorsen (approved):

PR #2962, opened 2025-12-11
Check that MCU commit is in the main-2.1 branch

clundin25:


swenson (approved):
LGTM. I was just noticing that the MCU commit was invalid, so it would be good to add this check.

PR #2790, opened 2025-11-18
runtime: Avoid I3C recovery race by not resetting fifo ctrl

swenson:
We don't need to set the I3C recovery indirect fifo ctrl reset, as this is set by the BMC when setting the recovery image size.

Doing this eliminates a rare race condition that we were occasionally seeing:

* Caliptra sets the recovery status next image index to 1 (SoC manifest)
* BMC reads the recovery status
* BMC writes indirect fifo ctrl to reset and set the next image size
* BMC sends first data block to indirect fifo data
* Caliptra writes indirect fifo ctrl to reset, which drops the first data block

This extra indirect fifo ctrl reset is unnecessary and appears to have been introduced to work around quirks in the fake BMC flow in #2107.

We also change the emulator behavior to load the next image into the recovery registers on writes to the recovery status to clean up these quirks a little.

This addresses one small part of
https://github.com/chipsalliance/caliptra-mcu-sw/issues/572 (but not all of the instability).

PR #2975, opened 2025-12-12
rom/runtime: Re-add reset indirect fifo ctrl before setting next image index

### LINKED ISSUE #2967: Recovery image size race condition
When the BMC is sending the recovery image size, there is a race condition where Caliptra runtime might get to the DMA portion too quickly and read a stale image size before the BMC has written the new image.

Originally reported in #2842 and #2834

### LINKED ISSUE #2969: Update documentation for recovery flow to include INDIRECT_FIFO_CTRL reset
This is needed to be sent by BMC, but is not in the OCP spec I believe. We need to update our docs to reflect this step.

swenson:
We removed the extra reset of indirect fifo ctrl in #2790 and rely on the BMC to do the reset.

However, it was pointed out in #2967 that this causes another race condition: the payload_available signal does not go low without this reset, and depending on when the BMC sends the reset and when Caliptra FW starts trying to download the next recovery image, Caliptra FW might read the wrong size. This was probably why our FPGA did not have trouble downloading small images without #2966.

This should resolve the race condition by ensuring that the indirect fifo ctrl is always reset before setting the next image index.

The BMC can still send its own indirect fifo data reset, but it should no longer be necessary.

Closes #2969.
Closes #2967.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -818,6 +818,15 @@ impl<'a> DmaRecovery<'a> {
         })
     }
 
+    pub fn reset_indirect_fifo_ctrl(&self) -> CaliptraResult<()> {
+        self.with_regs_mut(|regs_mut| {
+            let recovery = regs_mut.sec_fw_recovery_if();
+            recovery
+                .indirect_fifo_ctrl_0()
+                .modify(|val| val.reset(Self::RESET_VAL));
+        })
+    }
+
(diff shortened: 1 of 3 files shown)


zhalvorsen (approved):

mhatrevi (approved):

swenson:
Thanks!

Issue #3012, opened 2025-12-12
DPE ML-DSA (additive, does not replace ECDSA)


zhalvorsen:
This will be done in several step because it will be a decent amount of work

* [x] Update DPE submodule
* [x] Add separate DPE crypto platform support for ML-DSA
* [x] Add a second `INVOKE_DPE` command for ML-DSA
* [ ] Testing

ArchanaMadhavan:
@zhalvorsen We are under the impression that the DPE ML-DSA support is expected to also cater to 2.0.x fw.
Is there a timeline when these commands would be backported to 2.0.x ?
Thanks !

zhalvorsen:
We do plan to back port this to 2.0. I'm not aware of a timeline though. Everyone involved is busy with upcoming MCU releases. Once those are done, we should have more time to re-visit this. That would probably be around the beginning on Q3 2026.

PR #3013, opened 2025-12-12
[dpe] Update to just before adding SVN

zhalvorsen:
Update to a newer version of DPE. This is the first step in the process of upgrading to get to PQC DPE attestations.


zhalvorsen:
Starts work for #3012

dpe
zhalvorsen, line 1: To avoid making this a huge change this points to a small patch off the of `main` history captured in https://github.com/chipsalliance/caliptra-dpe/tree/zhalvorsen/just-before-svn-test-patch

PR #3018, opened 2025-12-15
[dpe] Update DPE to support SVN in DeriveContext

zhalvorsen:
#3012 

This updates the DPE submodule to a version that supports the SVN parameter in DeriveContext. The `StashMeasurement` command takes an SVN. Previously this parameter was unused. Now it will be passed to DPE while creating new contexts.


dpe
zhalvorsen, line 1: To avoid making this a huge change this points to a small patch off the of `main` history captured in https://github.com/zhalvorsen/caliptra-dpe/tree/zhalvorsen/svn-test-patch

timothytrippel (approved):

PR #3020, opened 2025-12-16
feat(hw-model): Reset i3c controller state on cold reset

ArthurHeymans:
Add i3c_not_ready() call during cold_reset to properly reset the i3c
controller state before subsystem reset. This ensures the controller is
in a clean state during reset operations.

### DIFF
--- a/hw-model/src/xi3c/controller.rs
+++ b/hw-model/src/xi3c/controller.rs
@@ -226,6 +226,10 @@ impl Controller {
         Ok(())
     }
 
+    pub fn set_i3c_not_ready(&self) {
+        self.ready.set(false);
+    }
+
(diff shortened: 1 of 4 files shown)


hw-model/src/xi3c/controller.rs
```diff
@@ -226,6 +226,10 @@ impl Controller {
         Ok(())
     }
 
+    pub fn i3c_not_ready(&self) {
+        self.ready.set(false);
+    }
```
clundin25:
```suggestion
    pub fn set_i3c_not_ready(&self) {
        self.ready.set(false);
    }
```

nit: `i3c_not_ready` sounds like a boolean

PR #3022, opened 2025-12-16
[dpe] Update DPE to use `dpe::State`

zhalvorsen:
This change in the DPE submodule will allow two DPE "instances" to share the same state information. That way both the EC and ML-DSA instances can have the same children and measurements without having to duplicate the state for each on every change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -958,6 +958,7 @@ dependencies = [
  "caliptra-test-harness",
  "caliptra_common",
  "cfg-if",
+ "dpe",
  "ufmt",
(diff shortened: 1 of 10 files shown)


dpe
zhalvorsen, line 1: To avoid making this a huge change this points to a small patch off the of main history captured in https://github.com/chipsalliance/caliptra-dpe/tree/zhalvorsen/state-test-patch

clundin25 (approved):

timothytrippel (approved):

PR #3023, opened 2025-12-16
refactor(fw_processor): Remove mbox lock in subsystem mode handling

ArthurHeymans:
Remove unnecessary mailbox transaction locking since subsystem mode no
longer uses mailbox for firmware loading. Replace subsystem_mode()
checks with direct txn presence checks to avoid redundant calls and
reduce code size.

Signed-off-by: Arthur Heymans <arthur.heymans@9elements.com>

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor/mod.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor/mod.rs
@@ -255,14 +255,16 @@ impl FirmwareProcessor {
     /// * `persistent_data` - Persistent data
     ///
     /// # Returns
-    /// * `MailboxRecvTxn` - Mailbox Receive Transaction
+    /// * `Option<MailboxRecvTxn>` - Mailbox Receive Transaction (Some in passive mode, None in subsystem mode)
+    /// * `u32` - Image size in bytes
     ///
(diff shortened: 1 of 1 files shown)


rom/dev/src/flow/cold_reset/fw_processor/mod.rs
```diff
@@ -268,7 +270,7 @@ impl FirmwareProcessor {
         dma: &mut Dma,
         env: &mut KatsEnv,
         persistent_data: &mut PersistentData,
-    ) -> CaliptraResult<(ManuallyDrop<MailboxRecvTxn<'a>>, u32)> {
+    ) -> CaliptraResult<(Option<ManuallyDrop<MailboxRecvTxn<'a>>>, u32)> {
```
swenson, line 275: We should update the documentation for this function to indicate that the MailboxRecvTxn is returned only in passive mode.

And similarly we should update the other functions to indicate that the txn should be None if in subsystem mode, since we are removing the explicit check for subsystem mode now.

swenson (commented):
Code looks good, but I think we should update comments to reflect how the code operates now.

swenson (approved):
LGTM

PR #3024, opened 2025-12-17
rom/runtime: dma: update FIFO wait conditions

Neal-liu:
Change FIFO wait logic to wait for FIFO to be full for every 256 bytes instead of checking only for FIFO empty. This avoids the case where read and write operations may occur simultaneously on the same buffer.

This change has been verified on our FPGA platform and successfully transfers three images in RI bypass mode.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -25,6 +25,7 @@ use caliptra_registers::sha512_acc::RegisterBlock as ShaAccRegisterBlock;
 use core::{cell::Cell, mem::size_of, ops::Add};
 use ureg::{Mmio, MmioMut, RealMmioMut};
 
+const INDIRECT_FIFO_SIZE: u32 = 256;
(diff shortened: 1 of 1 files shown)


drivers/src/dma.rs
```diff
@@ -679,18 +680,19 @@ impl<'a> DmaRecovery<'a> {
...
+                if i % (INDIRECT_FIFO_SIZE as usize / 4) == 0 {
+                    self.with_regs(|r| {
+                        while !r
+                            .sec_fw_recovery_if()
+                            .indirect_fifo_status_0()
+                            .read()
+                            .full()
+                        {}
+                    })?;
+                }
```
ArthurHeymans, line 695: Does this not make it necessary to have images with sizes multiple of 256? Let's see what the CI says about this

Neal-liu: Correct.

PR #3028, opened 2025-12-17
[EMU]: Add OCP LOCK in progress register

clundin25:


sw-emulator/lib/periph/src/soc_reg.rs
```diff
@@ -812,6 +812,9 @@ struct SocRegistersImpl {
     #[register_array(offset = 0x5a0, write_fn = on_write_ss_strap_generic)]
     ss_strap_generic: [u32; 4],
 
+    #[register(offset = 0x54c, write_fn = on_write_ss_ocp_lock_ctrl)]
+    ss_ocp_lock_ctrl: u32,
```
zhalvorsen, line 813: nit: It looks like this register should go up one more to stay in order of address

clundin25: Whoops, fixed!

PR #3037, opened 2025-12-18
[OCP-LOCK]: Add standalone test implementation of HPKE ML-KEM1024-HKD…

clundin25:
…F-SHA-384-AES-256-GCM

This will be used to send and received encrypted messages to caliptra firmware.


test/src/crypto/hpke/test_vector.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[serde(with = "hex::serde")]
+    pub exporter_secret: Vec<u8>,
+    pub encryptions: Vec<Encryption>,
+}
+
+impl HpkeTestArgs {
+    /// Returns the HPKE Test vector for the chosen KEM.
+    pub fn new(kem_id: u16) -> Self {
+        let kdf_id = 2;
+        let aead_id = 2;
```
zhalvorsen, line 56: It looks like the `kdf_id` and `aead_id` from the json file are both 1. Is there a mistake in one of these?

test/src/crypto/hpke/test_vector.rs
```diff
@@ -0,0 +1,62 @@
+// Licensed under the Apache-2.0 license
+
+use serde::{Deserialize, Serialize};
+
+const HPKE_TEST_VECTOR_PQ: &str = include_str!("../../crypto/test_vectors/hpke-pq.json");
```
zhalvorsen, line 7: Can you add a comment here for where this JSON file was taken from or how it was generated?

clundin25: Done!

test/src/crypto/hpke/test_vector.rs
clundin25: This file contains test vectors for all the algorithms defined in https://datatracker.ietf.org/doc/draft-ietf-hpke-pq/03/.

The first entry is using HMAC-256 + AES-128-GCM which have `1` as the id. 

This functions filters the test vectors to the algorithm we are interested in testing.

zhalvorsen: Gotcha, thanks!

PR #3039, opened 2025-12-19
HKDF Expand and Extract implementation

clundin25:
This is an implementation of https://www.rfc-editor.org/rfc/rfc5869 which is a building block for HPKE.


bluegate010:
Could we reuse implementations across https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/hkdf.rs?

clundin25:
> Could we reuse implementations across https://github.com/chipsalliance/caliptra-sw/blob/main/drivers/src/hkdf.rs?

Yes, didn't see that file. Closing this

Issue #2950, opened 2025-12-10
Encrypted FW proposal


swenson:
# Encrypted Firmware Proposal

Potential solution for #2227

Encrypted firmware is officially supported with AES GCM only.
Other AES modes are potentially possible using Caliptra, but the AES-DMA hardware only supports AES GCM, so decrypting firmware with other AES modes would have performance issues.

There are 2 methods that Calitpra will support for decrypting firmware:

1. New `CM_AES_GCM_DECRYPT_DMA` command (no special boot procedure for encrypted firmware). This is most applicable to non-MCU firmware, but can be used to decrypt MCU firmware as well, with some caveats (see below).
2. New ROM command: `RI_DOWNLOAD_ENCRYPTED_FIRMWARE`. Alternatively: we could add a parameter to `RI_DOWNLOAD_FIRMWARE` that indicates that Caliptra RT needs to not activate MCU's firmware after downloading.

All of these methods utilize the existing cryptographic mailbox keys (CMKs) for handling the keys, so any keys used for decrypting firmware will need to be loaded or derived by MCU ROM / unencrypted FMC using `CM_IMPORT`, `CM_DERIVE_STABLE_KEY`, etc.

## CM_AES_GCM_DECRYPT_DMA

This is a new runtime cryptographic mailbox command that will do in-place decryption of an AXI address range.

Request parameters:

* CMK (128 bytes)
* AES IV (12 bytes)
* AES tag (16 bytes)
* SHA384 of encrypted data
* (Maybe: SHA384 of decrypted data (48 bytes))
* AXI address (64 bits / 8 bytes)
* Length (32 bits)
* AAD length (32 bits)
* AAD (0..=4095 bytes)

This will require two passes: first, we Caliptra will use the DMA to check the SHA384 of the existing data and ensure it matches, and then do the in-place decryption using the AES mode of the DMA.

The SHA384 is necessary since we may be decrypting firmware that is only a portion of the firmware that was checked in the SoC manifest.

**Question**: Should we check that the AXI address + length falls inside an SoC component from the SoC manifest? Maybe even take that SoC manifest entry ID as a parameter?

**Question**: Should we require the SHA384 of the decrypted data? The AES tag is a 128-bit hash, effectively, but it has some binding weaknesses.

**Question**: Double check if there is an alignment requirement for AES DMA.

**Question**: If the tag fails to validate, we've just decrypted some data, which is not ideal. It's better if it is has to match

[...]

PR #3043, opened 2025-12-22
Add support for encrypted firmware flow via DMA

ArthurHeymans:
This PR introduces the necessary changes to support an encrypted firmware loading flow via the recovery interface. It addresses https://github.com/chipsalliance/caliptra-sw/issues/2950 by allowing the system to boot in an encrypted firmware mode where MCU activation is deferred until decryption is performed.

(This PR is ROM only -- we will move the rest of the changes  into a separate PR)

### Key Changes
*   **Mailbox API**: Added `RI_DOWNLOAD_ENCRYPTED_FIRMWARE` and `CM_AES_GCM_DECRYPT_DMA` command identifiers and their associated request/response structures.
*   **Persistent Data**: Introduced `BootMode` to distinguish between `Normal` and `EncryptedFirmware` boots, stored in `RomPersistentData`.
*   **ROM Changes**: Updated the firmware processor to handle the new encrypted download command and persist the boot mode.
*   **Runtime Changes**:
    *   Implemented `aes_256_gcm_decrypt_dma` in the cryptographic mailbox to perform in-place decryption of MCU firmware using DMA.
    *   Updated the recovery flow to skip MCU activation when `EncryptedFirmware` boot mode is detected.
*   **HW Model**: Added support for encrypted boot parameters and a new `read_payload_from_ss_staging_area` method to facilitate testing firmware contents in MCU SRAM.
*   **Testing**: Added integration tests to verify the end-to-end encrypted firmware flow, including key import, DMA decryption, and tag verification. These are currently emulator only

### Implementation Details
*   The `CM_AES_GCM_DECRYPT_DMA` command performs a two-pass operation on AXI memory: a first pass to verify the SHA384 hash of the encrypted data, and a second pass for in-place AES-GCM decryption.
*   This command is restricted to scenarios where the ROM has explicitly set the boot mode to `EncryptedFirmware`.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -126,6 +126,9 @@ impl CommandId {
     // The download firmware from recovery interface command.
     pub const RI_DOWNLOAD_FIRMWARE: Self = Self(0x5249_4644); // "RIFD"
 
+    // The download encrypted firmware from recovery interface command.
+    pub const RI_DOWNLOAD_ENCRYPTED_FIRMWARE: Self = Self(0x5249_4645); // "RIFE"
+
     // The get IDevID ECC CSR command.
     pub const GET_IDEV_ECC384_CSR: Self = Self(0x4944_4352); // "IDCR"
 
--- a/drivers/src/lib.rs
+++ b/drivers/src/lib.rs
@@ -130,7 +130,7 @@ pub use persistent::IDEVID_CSR_ENVELOP_MARKER;
 #[cfg(feature = "runtime")]
 pub use persistent::{AuthManifestImageMetadataList, ExportedCdiEntry, ExportedCdiHandles};
 pub use persistent::{
-    Ecc384IdevIdCsr, FuseLogArray, InitDevIdCsrEnvelope, Mldsa87IdevIdCsr, OcpLockFlags,
+    BootMode, Ecc384IdevIdCsr, FuseLogArray, InitDevIdCsrEnvelope, Mldsa87IdevIdCsr, OcpLockFlags,
     PcrLogArray, PersistentData, PersistentDataAccessor, RomPersistentData, StashMeasurementArray,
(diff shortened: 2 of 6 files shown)


api/src/mailbox.rs
```diff
@@ -108,6 +108,9 @@ impl CommandId {
     // The download firmware from recovery interface command.
     pub const RI_DOWNLOAD_FIRMWARE: Self = Self(0x5249_4644); // "RIFD"
 
+    // The download encrypted firmware from recovery interface command.
+    pub const RI_DOWNLOAD_ENCRYPTED_FIRMWARE: Self = Self(0x5249_4645); // "RIFE"
```
swenson, line 130: We need to document these new commands in the ROM and runtime READMEs.

We will also need to add a section to the runtime README noting that we added support for decrypting firmware.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2370,6 +2371,140 @@ impl Commands {
...
+
+        let computed_digest = dma_recovery.sha384_image(
+            &mut drivers.sha2_512_384_acc,
+            axi_addr,
+            length,
+            AesDmaMode::None,
+        )?;
+
+        let computed_digest_bytes: [u8; 48] = computed_digest.into();
+        if computed_digest_bytes != cmd.encrypted_data_sha384 {
```
swenson: nit: I always prefer to do constant-time comparison of hashes, even though it is overkill.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2370,6 +2371,140 @@ impl Commands {
...
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
```
swenson: We should probably have a unique error code for this.

swenson (commented):
Great work! I think this is exactly what I had in mind. Just a few minor things.

swenson:
I think this needs to be rebased. I think for now, let's split this into a ROM PR and a runtime PR, please

PR #3047, opened 2025-12-30
Add GET_PCR_LOG command

sunguk-bin-fad:
(Caliptra1.x)

Add GET_PCR_LOG mailbox command and corresponding API

### DIFF
--- a/runtime/src/pcr.rs
+++ b/runtime/src/pcr.rs
@@ -120,7 +120,7 @@ impl GetPcrLogCmd {
             .data
             .get_mut(..len)
             .ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
-        data.copy_from_slice(&pcr_log);
+        data.copy_from_slice(pcr_log);
 
         Ok(MailboxResp::GetPcrLog(get_pcr_log_resp))
     }


runtime/src/pcr.rs
```diff
@@ -98,3 +98,22 @@ impl ExtendPcrCmd {
         Ok(MailboxResp::default())
     }
 }
+
+pub struct GetPcrLogCmd;
+impl GetPcrLogCmd {
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[inline(never)]
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_bytes: &[u8]) -> CaliptraResult<MailboxResp> {
+        let pcr_log = drivers.persistent_data.get().pcr_log.as_bytes();
```
zhalvorsen: Could you take just the bytes that are valid (i.e. check the number of logs and only copy those). It would will be easier to parse on the receiver side if that is the case.

```suggestion
    let next_available = drivers.persistent_data.get().fht.pcr_log_index as usize;
    let Some(pcr_logs) = drivers.persistent_data.get().pcr_log.get(..next_available) else {
        return Err(CaliptraError::RUNTIME_PCR_INVALID_INDEX);
    }
    let pcr_log = pcr_logs.as_bytes();
```

runtime/src/pcr.rs
```diff
@@ -98,3 +98,22 @@ impl ExtendPcrCmd {
...
+        let len = pcr_log.len();
+
+        let mut get_pcr_log_resp = GetPcrLogResp {
+            hdr: MailboxRespHeader::default(),
+            data_size: len as u32,
+            data: [0u8; GetPcrLogResp::DATA_MAX_SIZE],
+        };
+        get_pcr_log_resp.data[..len].copy_from_slice(&pcr_log[..len]);
```
zhalvorsen: This might introduce a panic. Can you switch this to use get to make sure?


```suggestion
        let data = get_pcr_log_resp.data.get_mut(..len).ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
        data.copy_from_slice(&pcr_log);
```

zhalvorsen:
Thanks for doing this!

runtime/src/pcr.rs
sunguk-bin-fad: Thanks for suggestion. Fixed.

runtime/src/pcr.rs
sunguk-bin-fad: Fixed

sunguk-bin-fad:
@zhalvorsen 
Can you review this PR? Also, could you take a look at the build_and_test issue?

zhalvorsen:
It looks like you are going over the codespace limit. Sometimes the easiest way to fix this is to remove some less useful print statements.

zhalvorsen (approved):

sunguk-bin-fad:
> It looks like you are going over the codespace limit. Sometimes the easiest way to fix this is to remove some less useful print statements.

There is no print statements in my code. It seems to exceed the codespace when fips-test-hooks is enabled. Does Caliptra WG have a plan for runtime FW optimization, or do you think it's OK to merge this as is?

zhalvorsen:
Sorry, but we won't be able to merge this without CI passing. Everyone is really busy working on 2.x right now. We don't have any plans for a new 1.x release for a few months. The best option I'd recommend at the moment is to search through the `runtime` directory and find `cprintln` calls that are unnecessary. It looks like you are only going over by 113 bytes. Some of the best candidates are printing error messages when the returned error code indicates the same thing.

Let me know if you would like me to help out.

zhalvorsen:
A recent change was made to free up a bunch of codespace. I rebased your branch, it should be good now.

zhalvorsen:
Thank you for your contributions @sunguk-bin-fad!

PR #3051, opened 2026-01-02
Fix failed FW verification during FW update (2.1)

mlvisaya:
During FW update, the downloaded Caliptra FW image is verified through the FIRMWARE_VERIFY mailbox command. For 2.1, the downloaded firmware image is in the external staging memory, not in Mailbox or MCU SRAM.However currently, the image verifier only verifies the image in MCU SRAM or mailbox, not from external staging memory.

To fix this, a new parameter is passed to the verifier to specify where the image to be verified resides. The value of the enum can either be Mailbox, or Staging. If the image is coming from staging then the axi address need to be supplied.

Note that for cold boot recovery flow, the image is temporarily downloaded to MCU SRAM, so the ROM will verify the image in the MCU SRAM, so the image source is Staging with the axi address of the MCU SRAM.

Other changes:
- The EXTERNAL_TEST_SRAM (which corresponds to the FPGA staging peripheral) address has been updated to make it the same as the FPGA staging address. This is to simplify the code so that both emulator and FPGA code can use the same address for staging.
- Some optimization on the code is made to make the ROM size smaller. In the update-reset file, the staging_addr has been retrieved as AxiAddr but passed as u64, and then later on re-parsed as AxiAddr. To make it simpler, we just pass it as AxiAddr and bypass the unnecessary conversions.

### DIFF
--- a/common/src/verifier.rs
+++ b/common/src/verifier.rs
@@ -21,6 +21,15 @@ use zerocopy::{FromBytes, IntoBytes};
 
 use caliptra_drivers::memory_layout::ICCM_RANGE;
 
+/// Source describing where the firmware image is located for verification.
+#[derive(Copy, Clone)]
+pub enum ImageSource<'b> {
+    /// Image resides in mailbox FIFO.
+    Mailbox { data: &'b [u8] },
+    /// Image resides in staging memory reachable over AXI.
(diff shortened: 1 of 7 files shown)


mlvisaya:
wrong base

PR #3053, opened 2026-01-03
Fix failed FW verification during FW update (2.1)

mlvisaya:
During FW update, the downloaded Caliptra FW image is verified through the FIRMWARE_VERIFY mailbox command. For 2.1, the downloaded firmware image is in the external staging memory, not in Mailbox or MCU SRAM.However currently, the image verifier only verifies the image in MCU SRAM or mailbox, not from external staging memory.

Added a type for image_source for verification which can include the Axi address of the image to be verified.

Note that for cold boot recovery flow, the image is temporarily

Other changes:
- The EXTERNAL_TEST_SRAM (which corresponds to the FPGA staging memory peripheral) address has been updated to make it the same as the FPGA staging address. This is to simplify the code so that both emulator and FPGA code can use the same address for staging.

### DIFF
--- a/common/src/verifier.rs
+++ b/common/src/verifier.rs
@@ -16,7 +16,6 @@ use caliptra_drivers::*;
 use caliptra_image_types::*;
 use caliptra_image_verify::ImageVerificationEnv;
 use core::ops::Range;
-use dma::AesDmaMode;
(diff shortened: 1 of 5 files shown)


swenson (approved):
LGTM

mlvisaya:
thanks!

PR #3054, opened 2026-01-03
Separate Sha1 from RomEnv/FmcEnv/RuntimeEnv into non-crypto part

mtimkovich:
Other crypto engines to move in future PRs.

### DIFF
--- a/drivers/src/sha1.rs
+++ b/drivers/src/sha1.rs
@@ -23,12 +23,27 @@ const SHA1_MAX_DATA_SIZE: usize = 1024 * 1024;
 
 pub type Sha1Digest<'a> = &'a mut Array4x5;
 
-#[derive(Default)]
 pub struct Sha1 {
     compressor: Sha1Compressor,
+    _private: (), // prevent instantiation from outside drivers/ package
 }
 
 impl Sha1 {
+    /// Create a new, initialized and KATted SHA1 driver.
+    pub fn new() -> CaliptraResult<Self> {
+        let mut sha1 = Self::raw_new();
+        let kat = crate::kats::Sha1Kat {};
+        kat.execute(&mut sha1)?;
+        Ok(sha1)
+    }
+
+    fn raw_new() -> Self {
+        Self {
+            compressor: Sha1Compressor::default(),
+            _private: (),
+        }
+    }
+
     /// Initialize multi step digest operation
     ///
     /// # Returns
--- a/fmc/src/fmc_env.rs
+++ b/fmc/src/fmc_env.rs
@@ -26,10 +26,8 @@ use caliptra_registers::{
 };
 
 /// Hardware Context
+/// Contains only the non-cryptographic drivers and those needed before KAT execution
 pub struct FmcEnv {
-    // SHA1 Engine
-    pub sha1: Sha1,
-
     // SHA2-256 Engine
     pub sha256: Sha256,
 
@@ -84,7 +82,6 @@ impl FmcEnv {
         )?;
 
         Ok(Self {
-            sha1: Sha1::default(),
             sha256: Sha256::new(Sha256Reg::new()),
             sha2_512_384: Sha2_512_384::new(Sha512Reg::new()),
             sha2_512_384_acc: Sha2_512_384Acc::new(Sha512AccCsr::new()),
@@ -98,5 +95,53 @@ impl FmcEnv {
             persistent_data: PersistentDataAccessor::new(),
             mldsa: Mldsa87::new(AbrReg::new()),
         })
+    }
+}
+
+/// Full Hardware Context
+/// Contains all drivers needed after KAT execution
+pub struct FmcEnvFips {
+    /// Non-crypto environment (embedded)
+    pub non_crypto: FmcEnv,
+
+    // SHA1 Engine (initialized by KATs)
+    pub sha1: Sha1,
+}
+
+impl FmcEnvFips {
+    /// Create FmcEnvFips from non-crypto environment and initialized drivers
+    pub fn from_non_crypto(
+        non_crypto: FmcEnv,
+        initialized: caliptra_kat::InitializedDrivers,
+    ) -> Self {
+        Self {
+            non_crypto,
+            sha1: initialized.sha1,
+        }
+    }
+
+    /// Get a mutable reference to the non-crypto environment
+    pub fn non_crypto_mut(&mut self) -> &mut FmcEnv {
+        &mut self.non_crypto
+    }
+
+    /// Get an immutable reference to the non-crypto environment
+    pub fn non_crypto(&self) -> &FmcEnv {
+        &self.non_crypto
+    }
+}
+
+// Provide transparent access to non-crypto fields via Deref
+impl core::ops::Deref for FmcEnvFips {
+    type Target = FmcEnv;
+
+    fn deref(&self) -> &Self::Target {
+        &self.non_crypto
+    }
+}
+
+impl core::ops::DerefMut for FmcEnvFips {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.non_crypto
     }
 }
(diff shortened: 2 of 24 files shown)


drivers/src/sha1.rs
```diff
@@ -23,12 +23,27 @@ const SHA1_MAX_DATA_SIZE: usize = 1024 * 1024;
...
-#[derive(Default)]
 pub struct Sha1 {
     compressor: Sha1Compressor,
+    _private: (), // prevent instantiation from outside drivers/ package
 }
 
 impl Sha1 {
+    /// Create a new, initialized and KATted SHA1 driver.
+    pub fn new() -> CaliptraResult<Self> {
+        let mut sha1 = unsafe { Self::raw_new() };
```
clundin25: I think you should include a safety comment for all the unsafe blocks, otherwise some linters will complain

fmc/src/fmc_env.rs
```diff
@@ -26,10 +26,8 @@ use caliptra_registers::{
 };
 
 /// Hardware Context
-pub struct FmcEnv {
-    // SHA1 Engine
-    pub sha1: Sha1,
-
+/// Contains only the non-cryptographic drivers and those needed before KAT execution
+pub struct FmcEnvNonCrypto {
```
clundin25: Nit but you could use a generic to promote code sharing

```rust
trait KatState;
struct KatMissing; // <- Use this type before Kat's are called
struct KatComplete; // <- Use this type after Kat's are called
impl KatState for KatComplete;
impl KatState for KatMissing;

struct FmcEnv<KatState> {
  _priv: PhantomData<KatState>
}

impl FmcEnv<KatState> { 
  /// Only provide accessors for non-crypto drivers
}

impl FmcEnv<KatComplete> { 
  /// Includes additional accessors for crypto drivers
}
````

File: fmc/src/fmc_env.rs [hunk 3]
```diff
@@ -100,3 +97,53 @@ impl FmcEnv {
...
+    pub non_crypto: FmcEnvNonCrypto,
+
+    // SHA1 Engine (initialized by KATs)
+    pub sha1: Sha1,
+}
+
+impl FmcEnv {
+    /// Create FmcEnv from non-crypto environment and initialized drivers
+    /// # Safety
+    /// The KATs have been run.
```
clundin25:
```suggestion
    /// KATS MUST be run before calling this function.
```

fmc/src/fmc_env.rs
```diff
@@ -100,3 +97,53 @@ impl FmcEnv {
...
+    pub unsafe fn from_non_crypto(
+        non_crypto: FmcEnvNonCrypto,
+        initialized: caliptra_kat::InitializedDrivers,
```
clundin25: Does this have to be unsafe? Doesn't `InitializedDrivers` require the Kat to be run ?

swenson: Ah, maybe this is okay now. I think I originally made this `unsafe` just as more of a warning, but if it only accepts initialized things, then it's probably fine.

drivers/src/sha1.rs
zhalvorsen: fyi: I created [an issue](https://github.com/chipsalliance/caliptra-sw/issues/3093) to track adding the clippy lint for `cailptra-sw`

fmc/src/fmc_env.rs
zhalvorsen: This might be overly nit-picky, but `FmcEnvNonCrypto` feel like a misnomer because it has a lot of crypto drivers inside of it

mtimkovich: You're correct but I believe the plan is to eventually move all of the crypto drivers out of this struct.

zhalvorsen: Oh, got it! Sorry, I forgot this was happening piecemeal. Sounds good, thanks!

rom/dev/src/rom_env.rs
```diff
@@ -115,3 +112,52 @@ impl RomEnv {
...
+    /// Create RomEnv from non-crypto environment and initialized drivers
+    pub fn from_non_crypto(
+        non_crypto: RomEnvNonCrypto,
+        initialized: caliptra_kat::InitializedDrivers,
+    ) -> Self {
+        Self {
+            non_crypto,
+            sha1: initialized.sha1,
+        }
+    }
```
clundin25, line 136: Does it makes sense to make this a method of `RomEnvNonCrypto`? 

Then it could be renamed to `initialize`

rom/dev/src/rom_env.rs
```diff
@@ -115,3 +112,52 @@ impl RomEnv {
...
+
+    /// Get a mutable reference to the non-crypto environment
+    pub fn non_crypto_mut(&mut self) -> &mut RomEnvNonCrypto {
```
clundin25: nit: But `NonCrypto` seems a bit confusing. Maybe `RomEnv` should be `RomEnvFips` and `RomEnvNonCrypto` should be `RomEnv`?

CC @swenson @zhalvorsen

swenson: That sounds reasonable to me.

rom/dev/src/rom_env.rs
mtimkovich: Yeah, I like that a lot

mtimkovich: I tried this out I ended up with a `deref recursion limit reached` error I wasn't able to solve. I'll try to add this in a later PR.

fmc/src/fmc_env.rs
mtimkovich: This ended up being trickier to implement than I anticipated, I think I'll have to save this for a future PR as well.

fmc/src/fmc_env.rs
```diff
@@ -100,3 +97,51 @@ impl FmcEnv {
         })
     }
 }
+
+/// Full Hardware Context
+/// Contains all drivers needed after KAT execution
+pub struct FmcEnv {
+    /// Non-crypto environment (embedded)
+    pub non_crypto: FmcEnvNonCrypto,
```
clundin25: There is a mixture of terms now, it would be good to have a unified naming scheme for this (i.e the FIPs naming from my previous common)

rom/dev/src/flow/cold_reset/ldev_id.rs
```diff
@@ -331,15 +330,16 @@ impl LocalDevIdLayer {
                 sig.zeroize();
             })?;
 
-        let _pub_x: [u8; 48] = (&ecc_pub_key.x).into();
-        let _pub_y: [u8; 48] = (&ecc_pub_key.y).into();
-        cprintln!("[ldev] PUB.X = {}", HexBytes(&_pub_x));
-        cprintln!("[ldev] PUB.Y = {}", HexBytes(&_pub_y));
+        // Commented to save on ROM space.
```
clundin25: Can you add a tracking issue for re-enabling this to the comment?

fmc/src/fmc_env.rs
mtimkovich: Whoops, good catch

rom/dev/src/flow/cold_reset/ldev_id.rs
mtimkovich: These lines are removed on the main branch now, I think the commented section should just be cut entirely.

fmc/src/fmc_env.rs
```diff
@@ -100,3 +97,51 @@ impl FmcEnv {
         })
     }
 }
+
+/// Full Hardware Context
+/// Contains all drivers needed after KAT execution
```
ArthurHeymans, line 102: Is running KAT in FMC desired? Does it make sense to uncoditionally run them again in both FMC and runtime?

rom/dev/src/rom_env.rs
```diff
@@ -115,3 +112,52 @@ impl RomEnv {
         })
     }
 }
+
+/// Full ROM Context
+/// Contains all drivers needed after KAT execution
+pub struct RomEnvFips {
+    /// Non-crypto environment (embedded)
+    pub non_crypto: RomEnv,
```
ArthurHeymans, line 120: The naming non_crypto is as it contains a lot of crypto related engines.

mhatrevi:
Closing this PR in lieu of https://github.com/chipsalliance/caliptra-sw/pull/3190

PR #3072, opened 2026-01-07
Make LMS key generation and signing tools public

swenson:
This is meant so that MCU xtask can use these to generate signed firmware without duplicating this code.

The key generation code was missing in any case, so added that and some tests for the methods.


zhalvorsen (approved):

swenson:
Thanks!

clundin25:
@swenson sorry this needs a rebase. I re-worked the CI jobs and the required jobs have shifted.

PR #3073, opened 2026-01-08
[OCP-LOCK]: HPKE ML-KEM 1024 firmware implementation

### LINKED ISSUE #3032: [HPKE] - MLKEM1024
Implement the ML-KEM-1024 - HKDF-384 - AES-256-GCM HPKE algorithm.

clundin25:
Outstanding tasks:
* Fix panics - It's likely this code may panic. Once this driver is linked into runtime we can clean these up. Since the current code is not linked no panic symbols will be introduced.
* Optimize memory usage - There are a few large objects being placed on the stack (Notably ML-KEM keys and ciphers). We need to tweak the code to avoid any copies.

This resolves #3032

### DIFF
--- a/rom/dev/src/flow/cold_reset/mod.rs
+++ b/rom/dev/src/flow/cold_reset/mod.rs
@@ -72,14 +72,16 @@ impl ColdResetFlow {
         fht::initialize_fht(env);
 
         // Execute IDEVID layer
-        let mut idevid_layer_output = InitDevIdLayer::derive(env)?;
-        let ldevid_layer_input = dice_input_from_output(&idevid_layer_output);
-
-        // Execute LDEVID layer
-        let result = LocalDevIdLayer::derive(env, &ldevid_layer_input);
-        idevid_layer_output.zeroize();
-        let mut ldevid_layer_output = result?;
-        let fmc_layer_input = dice_input_from_output(&ldevid_layer_output);
+        let mut dice_out = InitDevIdLayer::derive(env)?;
+        let dice_in = dice_input_from_output(&dice_out);
+        let mut dice_out = {
+            // Execute LDEVID layer
+            let res = LocalDevIdLayer::derive(env, &dice_in);
+            dice_out.zeroize();
+            res?
+        };
+        let dice_in = dice_input_from_output(&dice_out);
+        // Generate the CMB AES key
 
         // Generate the CMB AES key
         generate_cmb_aes_key(env)?;
@@ -88,8 +90,8 @@ impl ColdResetFlow {
         let mut fw_proc_info = FirmwareProcessor::process(env)?;
 
         // Execute FMCALIAS layer
-        let result = FmcAliasLayer::derive(env, &fmc_layer_input, &fw_proc_info);
-        ldevid_layer_output.zeroize();
+        let result = FmcAliasLayer::derive(env, &dice_in, &fw_proc_info);
+        dice_out.zeroize();
         fw_proc_info.zeroize();
         result?;
 


clundin25:
@clundin25: TODO: Remove empty commit before merging.

rom/dev/src/flow/cold_reset/mod.rs
bluegate010, line 1: Just to confirm, these are cosmetic changes right?

clundin25: Yes, it just hints to the compiler that we are done with `idevid_layer_output`.

I'll check with others if this is an OK change. I understand this is battle tested code, so we may want to find the savings elsewhere.

drivers/src/hpke/kem/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Uses the decapsulation key to produce a shared secret key from a ciphertext.
+    fn decap(
+        &mut self,
+        enc: &EncapsulatedSecret<NENC>,
+        dk: &DecapsulationKey<NSK>,
+    ) -> CaliptraResult<SharedSecret<NSECRET>>;
+}
+
+/// Shared Secret produced by `encap`.
+pub struct SharedSecret<const NSECRET: usize> {
```
zhalvorsen, line 45: This should derive `ZeroizeOnDrop` so it gets cleared when we are done with it

drivers/src/hpke/kem/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+
+pub use mlkem::*;
+
+use zerocopy::{FromBytes, Immutable, KnownLayout};
+
+use crate::Trng;
+
+use super::suites::KemIdExt;
+
+// TODO(clundin): Leaving the trait in for now but I am not sure we really need it.
```
zhalvorsen, line 16: Yeah, if we plan to support several options this will be really nice. But, something similar was done with the LMS keys which IMO ends up making it harder to figure out which variant we actually use.

drivers/src/hpke/kem/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Implements KEM functionality for HPKE
+pub trait Kem<const NSK: usize, const NENC: usize, const NPK: usize, const NSECRET: usize> {
+    /// The extended Kem id, fed into the labeled derive function to expand `ikm`.
+    const KEM_ID_EXT: KemIdExt;
+
+    /// Derives a KEM keypair from the `ikm` seed.
+    fn derive_key_pair(
+        &mut self,
+        ikm: &[u8; NSK],
+    ) -> CaliptraResult<(EncapsulationKey<NPK>, DecapsulationKey<NSK>)>;
```
clundin25, line 27: note: I will most likely move the decap key to an associated type in a follow up.

drivers/src/hpke/encryption_context.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///
+    /// Should this error occur, integrators should rotate the HPKE key pair by calling
+    /// the `OCP_LOCK_ROTATE_HPKE_KEY` mailbox command.
+    pub fn increment_sequence(&mut self) -> CaliptraResult<()> {
+        if let Some(sequence) = self.sequence_number.checked_add(1) {
+            self.sequence_number = sequence;
+            Ok(())
+        } else {
+            Err(CaliptraError::RUNTIME_DRIVER_HPKE_SEQ_EXHAUSTED)
+        }
```
zhalvorsen: nit: this might be slightly more readible with an `ok_or`:

```suggestion
        let sequence = self.sequence_number.checked_add(1)
            .ok_or(CaliptraError::RUNTIME_DRIVER_HPKE_SEQ_EXHAUSTED)?;
        self.sequence_number = sequence;
        Ok(())
```

drivers/src/hpke/encryption_context.rs
```diff
@@ -0,0 +1,178 @@
...
+        aes: &mut Aes,
+        trng: &mut Trng,
+        aad: &[u8],
+        pt: &[u8],
+        ct: &mut [u8],
+    ) -> CaliptraResult<[u8; Aes256GCM::NT]> {
+        let nonce = self
+            .base_nonce
+            .xor_with_sequence_count(self.sequence_number)
+            .into();
```
zhalvorsen: nit: can you use the `compute_nonce` function from above here or does it end up looking uglier with mapping the different types?

rom/dev/src/main.rs
```diff
@@ -117,18 +117,20 @@ pub extern "C" fn rom_entry() -> ! {
...
-            "Yes"
-        } else {
-            "No"
-        }
-    );
-
-    if env.soc_ifc.ocp_lock_enabled() {
-        cprintln!("[ROM] OCP-LOCK Supported");
-    }
+    // TODO(clundin): Re-add after more code savings
```
timothytrippel: Can you link these to issues? so we can track them properly

rom/dev/src/flow/cold_reset/mod.rs
```diff
@@ -72,14 +72,16 @@ impl ColdResetFlow {
...
+        let mut dice_out = InitDevIdLayer::derive(env)?;
+        let dice_in = dice_input_from_output(&dice_out);
+        let mut dice_out = {
+            // Execute LDEVID layer
+            let res = LocalDevIdLayer::derive(env, &dice_in);
+            dice_out.zeroize();
+            res?
+        };
+        let dice_in = dice_input_from_output(&dice_out);
+        // Generate the CMB AES key
```
timothytrippel, line 84: nit: duplicate comment?

drivers/src/hpke/kdf.rs
```diff
@@ -0,0 +1,224 @@
...
+        &mut self,
+        trng: &mut Trng,
+        suite_id: &CipherSuite,
+        salt: &[u8],
+        label: &[u8],
+        ikm: &[u8],
+    ) -> CaliptraResult<[u8; Hmac384::NH]> {
+        let labeled_ikm = [
+            &b"HPKE-v1"[..],
+            &b"HPKE"[..],
```
zhalvorsen, line 86: It looks like the spec doesn't have this parameter in the `LabeledExtract` or `LabeledExpand` pseudocode, or am I reading it wrong?

drivers/src/hpke/kdf.rs
```diff
@@ -0,0 +1,224 @@
...
+        let key_schedule_context = {
+            // TODO(clundin): This may panic. Fix this once the code is linked to the main binary.
+            let mut key_schedule_context = [0; Self::KEY_SCHEDULE_CONTEXT_LEN];
+            let (mode, rem) = <u8>::mut_from_prefix(&mut key_schedule_context).unwrap();
+            *mode = Mode::BASE.0;
+            let (psk, rem) = <[u8; Hmac384::NH]>::mut_from_prefix(rem).unwrap();
+            psk.clone_from_slice(&psk_id_hash);
+            let info = <[u8; Hmac384::NH]>::mut_from_bytes(rem).unwrap();
+            info.clone_from_slice(&info_id_hash);
+            key_schedule_context
```
zhalvorsen: nit: would this be easier with packed struct using zerocopy? It would effectively accomplish the same thing, but could be easier to read. You'd probably want to add a static assert to make sure there isn't any padding.

```Rust
[repr(packed)]
struct KeyScheduleCtx {
    mode: u8,
    psk: [u8; Hmac384::NH],
    info: [u8; Hmac384::NH],
}
const _: () = assert!(size_of::<KeyScheduleCtx>() == 1 + (2 * Hmac384::NH));

drivers/src/hpke/encryption_context.rs
clundin25: Much nicer, thanks!

drivers/src/hpke/encryption_context.rs
clundin25: Oh whoops, yes that should use `compute_nonce` :) thanks for catching that !

drivers/src/hpke/kdf.rs
clundin25: Relevant text:

> The value of suite_id depends on where the KDF is used; it is assumed implicit from the implementation and not passed as a parameter. If used inside a KEM algorithm, suite_id MUST start with "KEM" and identify this KEM algorithm; if used in the remainder of HPKE, it MUST start with "HPKE" and identify the entire ciphersuite in use. See sections [Section 4.1](https://datatracker.ietf.org/doc/html/draft-ietf-hpke-hpke-02#dhkem) and [Section 5.1](https://datatracker.ietf.org/doc/html/draft-ietf-hpke-hpke-02#encryption-context) for details.

Source: https://datatracker.ietf.org/doc/html/draft-ietf-hpke-hpke-02#name-cryptographic-dependencies

zhalvorsen: Gotcha! Thanks!

drivers/src/hpke/kem/mod.rs
zhalvorsen: Discussed offline, this is a public value

drivers/src/hpke/kdf.rs
clundin25: Oh great idea! Much cleaner

rom/dev/src/flow/cold_reset/mod.rs
clundin25: Whoops, thanks!

PR #3081, opened 2026-01-09
[hw] rev i3c-core submodule and regenerate regs

timothytrippel:
This revs the i3c-core submodule to match the `v1p4-patch` upstream tag used in the caliptra-ss submodule. This also regenerates the register definitions.

PR is large because it is all autogenerated code. No manual changes were made.

### DIFF
--- a/hw/latest/registers/src/i3ccsr.rs
+++ b/hw/latest/registers/src/i3ccsr.rs
@@ -961,9 +961,11 @@ impl<TMmio: ureg::Mmio> StdbyctrlmodeBlock<TMmio> {
             )
         }
     }
-    /// Read value: [`u32`]; Write value: [`u32`]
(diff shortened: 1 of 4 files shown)


swenson (approved):
Looks okay, though I am worried about all of those offset changes of DAI, etc., since that might mean that hw-model or ROM might need some changes to reflect the new offsets if we start running on new FPGA bitstreams.

timothytrippel:
These offsets should correspond to the current FPGA bitstreams loaded in CI. The commit of the i3c-core submodule of the caliptra-ss submodule from which the subsystem bitstream loaded in CI was built did not match the commit of the i3c-core submodule pulled into this repo (the version in this repo was behind). This brings the two in sync. We should really have CI checks in place to make sure all of these are in sync. Does that make sense? or am I misunderstanding your concern?

PR #3082, opened 2026-01-09
cherry-pick: Make LMS key generation and signing tools public (#3072)

swenson:
This is meant so that MCU xtask can use these to generate signed firmware without duplicating this code.

The key generation code was missing in any case, so added that and some tests for the methods.

(cherry picked from commit 5b924e11bd55909406fa135fa571a6cc03322230)


swenson:
Thanks!

PR #3083, opened 2026-01-09
sw-emulator: Remove DMA block size panic

swenson:
This check will cause problems when testing the 2.0.0 ROM with the latest emulator and firmware, since the 2.0.0 ROM uses a block size of 256.

This does work on the real hardware in the cases the ROM uses.

This change was originally part of #2938


mhatrevi (approved):

swenson:
Thanks!

PR #3085, opened 2026-01-09
cherry-pick to 2.0: sw-emulator: Remove DMA block size panic (#3083)

swenson:
This check will cause problems when testing the 2.0.0 ROM with the latest emulator and firmware, since the 2.0.0 ROM uses a block size of 256.

This does work on the real hardware in the cases the ROM uses.

This change was originally part of #2938

(cherry picked from commit 6fb5d4ab638b875235f077aedccce2e6b29b47f6)


swenson:
Thanks!

PR #3088, opened 2026-01-09
[OCP-LOCK]: Implement Preconditioned AES

clundin25:
This is a building block used in a few commands.

Also include OCP LOCK code for the panic test check.

### DIFF
--- a/drivers/src/preconditioned_aes.rs
+++ b/drivers/src/preconditioned_aes.rs
@@ -1,0 +1,159 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_error::{CaliptraError, CaliptraResult};
+
+use crate::{
+    hmac_kdf, Aes, AesGcmIv, AesKey, Array4x16, Hmac, HmacKey, HmacMode, HmacTag, KeyReadArgs,
+    KeyUsage, KeyWriteArgs, LEArray4x3, LEArray4x4, LEArray4x8, Trng,
+};
+
+use zerocopy::{FromBytes, IntoBytes};
+
+pub struct PreconditionedAesEncryptionResult {
+    pub salt: LEArray4x3,
+    pub iv: LEArray4x3,
+    pub tag: LEArray4x4,
+}
+
+/// Implements Pre-conditioned AES Encrypt as described in
+/// OCP LOCK version 1.0 RC2, Section "Preconditioned AES".
+/// https://github.com/chipsalliance/Caliptra/issues/603 will align the specification with this
+/// implementation.
+///
+/// NOTE: This function will overwrite `key`'s keyvault contents.
+#[allow(clippy::too_many_arguments)]
+pub fn preconditioned_aes_encrypt(
+    aes: &mut Aes,
+    hmac: &mut Hmac,
+    trng: &mut Trng,
+    key: HmacKey,
+    label: &[u8],
+    aad: &[u8],
+    plaintext: &[u8],
+    ciphertext: &mut [u8],
+) -> CaliptraResult<PreconditionedAesEncryptionResult> {
+    let salt: LEArray4x3 = {
+        // Convert 48 bytes of entropy to 12.
+        let seed = trng.generate()?;
+        let seed = seed
+            .as_bytes()
+            .get(..12)
+            .and_then(|seed| <[u32; 3]>::read_from_bytes(seed).ok())
+            .ok_or(CaliptraError::RUNTIME_DRIVER_PRECONDITIONED_AES_ENCRYPT_ERROR)?;
+        LEArray4x3::new(seed)
+    };
+
+    let mut output_array = Array4x16::default();
+    let tag = match key {
+        HmacKey::Key(kv) => HmacTag::Key(KeyWriteArgs {
+            id: kv.id,
+            usage: KeyUsage::default().set_aes_key_en(),
+        }),
+        HmacKey::Array4x12(_) | HmacKey::Array4x16(_) => HmacTag::Array4x16(&mut output_array),
+        HmacKey::CsrMode() => {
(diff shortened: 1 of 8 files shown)


drivers/src/preconditioned_aes.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use zerocopy::{FromBytes, IntoBytes};
+
+pub struct PreconditionedAesEncryptionResult {
+    pub iv: LEArray4x3,
+    pub tag: LEArray4x4,
+}
+
+/// Implements Pre-conditioned AES Encrypt as described in
+/// https://chipsalliance.github.io/Caliptra/ocp-lock/specification/HEAD/?content_ref=initializes+random+hpke+keypairs+for+each+supported+algorithm+and+assigns+handles+to+them+reported+via+enumerate_hpke_handles#sec:preconditioned-aes.
```
bluegate010: I would instead say something like "OCP LOCK version 1.0 RC2, Section X.Y.

drivers/src/preconditioned_aes.rs
```diff
@@ -0,0 +1,78 @@
...
+    aes: &mut Aes,
+    trng: &mut Trng,
+    key: AesKey,
+    label: &[u8],
+    aad: &[u8],
+    salt: &[u8; 12],
+    plaintext: &[u8],
+    ciphertext: &mut [u8],
+) -> CaliptraResult<PreconditionedAesEncryptionResult> {
+    let subkey = cmac_kdf(aes, key, label, Some(salt), 2)?;
```
bluegate010: The `32` on line 30 and the `2` on this line are related - can we make at least one of them a named constant?

drivers/src/preconditioned_aes.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Implements Pre-conditioned AES Encrypt as described in
+/// https://chipsalliance.github.io/Caliptra/ocp-lock/specification/HEAD/?content_ref=initializes+random+hpke+keypairs+for+each+supported+algorithm+and+assigns+handles+to+them+reported+via+enumerate_hpke_handles#sec:preconditioned-aes.
+#[allow(clippy::too_many_arguments)]
+pub fn preconditioned_aes_encrypt(
+    aes: &mut Aes,
+    trng: &mut Trng,
+    key: AesKey,
+    label: &[u8],
+    aad: &[u8],
+    salt: &[u8; 12],
```
bluegate010: Would it make sense to instead have this function randomly generate the salt?

(If there's a concern about test vectors, a common pattern is to have two functions, one that takes a salt (used for testing against deterministic vectors) and one that randomly generates a salt and then calls the first one (used everywhere else).)

drivers/src/preconditioned_aes.rs
clundin25: Will do !

drivers/src/preconditioned_aes.rs
clundin25: Will do!

drivers/src/preconditioned_aes.rs
clundin25: That was my initial attempt but wasn't sure how to handle building the AAD.

I suppose it is better to pass in ``key_type` || `salt` || `metadata_len` || `metadata`` to this function?

Alternatively, is dropping `salt` from the AAD a reasonable idea?

bluegate010: I don't think salt is part of AAD in the spec; it doesn't need to be, since if the salt changes, the subkey changes and the GCM tag will be wrong.

runtime/tests/runtime_integration_tests/test_panic_missing.rs
```diff
@@ -1,14 +1,16 @@
...
+use caliptra_builder::firmware::{APP_WITH_UART, APP_WITH_UART_OCP_LOCK};
 
 #[test]
 fn test_panic_missing() {
-    let rt_elf = caliptra_builder::build_firmware_elf(&APP_WITH_UART).unwrap();
-    let symbols = caliptra_builder::elf_symbols(&rt_elf).unwrap();
-    if symbols.iter().any(|s| s.name.contains("panic_is_possible")) {
-        panic!(
-            "The caliptra RT contains the panic_is_possible symbol, which is not allowed. \
+    for elf_id in [&APP_WITH_UART, &APP_WITH_UART_OCP_LOCK] {
```
zhalvorsen, line 6: Maybe put this in a separate test so if it is only in OCP LOCK it can be easier to trace. Alternatively, you could add an indicator to the panic message

drivers/src/preconditioned_aes.rs
```diff
@@ -0,0 +1,155 @@
...
+use zerocopy::{FromBytes, IntoBytes};
+
+pub struct PreconditionedAesEncryptionResult {
+    pub salt: LEArray4x3,
+    pub iv: LEArray4x3,
+    pub tag: LEArray4x4,
+}
+
+/// Implements Pre-conditioned AES Encrypt as described in
+/// OCP LOCK version 1.0 RC2, Section "Preconditioned AES"
```
bluegate010: Maybe note here that this doesn't quite implement that as-written in the spec? We can unify once we cut a new version.

clundin25: Opened https://github.com/chipsalliance/Caliptra/issues/603 to track this. If I end up needing to make more changes I'll add a comment to link to it

zhalvorsen (dismissed):

runtime/tests/runtime_integration_tests/test_panic_missing.rs
clundin25: That's a good idea, I'll do that in a follow up! I should've split this up from the start

clundin25: Added the `elf_id` to the panic

PR #3095, opened 2026-01-12
[OCP-LOCK]: HPKE Generation & Enumeration

clundin25:
* Generate HPKE context on reset
* Add OCP_LOCK_ENUMERATE_HPKE_HANDLES command

This resolves https://github.com/chipsalliance/caliptra-sw/issues/2994

### DIFF
--- a/drivers/src/hpke/mod.rs
+++ b/drivers/src/hpke/mod.rs
@@ -9,8 +9,8 @@ use kem::{
     MlKemEncapsulationKey,
 };
 use suites::CipherSuite;
-use zerocopy::IntoBytes;
-use zeroize::ZeroizeOnDrop;
+use zerocopy::transmute;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use crate::Trng;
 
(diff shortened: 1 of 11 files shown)


drivers/src/hpke/mod.rs
```diff
@@ -72,6 +95,54 @@ pub trait Hpke<
     fn serialize_public_key(&self, kem: &mut Self::K<'_>) -> CaliptraResult<EncapsulationKey<NPK>>;
 }
 
+// High level struct for https://datatracker.ietf.org/doc/html/draft-ietf-hpke-hpke-02.
+pub struct HpkeContext {
+    /// ML-KEM HPKE Context
+    /// Stored as a result to lazily surface HPKE generation failures on first use
+    ml_kem: CaliptraResult<HpkeMlKemContext>,
+    ml_kem_handle: HpkeHandle,
```
bluegate010: I feel like we should try and be algorithm-agnostic here, since there will be two other kinds of KEMs we need to track.

What about storing the private key as an enum?

drivers/src/hpke/mod.rs
```diff
@@ -72,6 +95,54 @@ pub trait Hpke<
...
+    ctx: &'a HpkeContext,
+    index: usize,
+}
+
+impl Iterator for HpkeContextIter<'_> {
+    type Item = (HpkeHandle, CipherSuite);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let item = match (&self.index, &self.ctx.ml_kem) {
+            (0, Ok(_)) => Some((self.ctx.ml_kem_handle.clone(), CipherSuite::ML_KEM_1024)),
```
bluegate010: This confuses me a bit. If the index is incremented past zero, we will never return the ML-KEM handle?

clundin25: Once ECDH and hybrid are implemented those will be returned as well, then once the iterator reaches `3` it will stop yielding value

drivers/src/hpke/mod.rs
clundin25: That's a good idea, maybe something like

```
enum HpkeEntry { 
  MlKem { handle: HpkeHandle, context: HpkeMlKemContext },
}
```

I think there will only be one handle available per algorithm at all times, correct? 

That made me think it may also be reasonable to have an explicit variable. Compressing them to the same enum may make it harder to mix handles up, however

bluegate010: That should work. Consider s/Context/PrivateKey/, since that's really what it is.

bluegate010 (approved):

zhalvorsen (approved):

Issue #3090, opened 2026-01-12
Add CM_SHA one-shot command for ROM


swenson:
This helps with implementing verifying signatures and hashes against Vendor PK hash, etc., from MCU ROM, without MCU having to have its own hash ability.

For simplicity, this should be a single command that hashes its input with either SHA384 or SHA512, but does not use contexts or support continuing hashes.

PR #3097, opened 2026-01-12
rom: Add CM_SHA command

swenson:
This is a one-shot, ROM-only command meant to make some Device Ownership Transfer-like flows easier from MCU ROM or the SoC manager.

For example, if MCU has something like a Vendor PK hash burned into fuses and needs to match an ECDSA or MLDSA public key against it as part of DOT recovery, then it would need a way in the MCU ROM to compute SHA384/512 hashes, either using Caliptra or with a hardware or software SHA capability in MCU. Since Caliptra is meant to be the only cryptography capabilities for MCU, this means we need at least a simple SHA command offered by Caliptra ROM.

(This could potentially be worked around by using the existing CM_HMAC, and having something similar to a Vendor PK hash using HMACs, but this is cumbersome and would need to be computed per-device, since the only HMAC keys available to Caliptra ROM are stable keys, which are device-specific.)

This is targeted for 2.1 and 2.0.1 ROMs.

### DIFF
--- a/rom/dev/README.md
+++ b/rom/dev/README.md
@@ -676,11 +676,12 @@ ROM supports the following set of commands before handling the FW_DOWNLOAD comma
 10. **ECDSA384_SIGNATURE_VERIFY**: This command verifies ECDSA384 signatures for Device Ownership Transfer or other flows. [ECDSA384_SIGNATURE_VERIFY](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#ecdsa384_signature_verify)
 11. **MLDSA87_SIGNATURE_VERIFY**: This command verifies MLDSA87 signatures for Device Ownership Transfer or other flows. [MLDSA87_SIGNATURE_VERIFY](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#mldsa87_signature_verify)
 12. **CM_RANDOM_GENERATE**: This command returns random numbers from Caliptra's RNG for Device Ownership Transfer or other flows. [CM_RANDOM_GENERATE](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#cm_random_generate)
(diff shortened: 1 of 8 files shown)


swenson:
Fixes #3090.

Probably blocked on ROM size at the moment.

rom/dev/README.md
```diff
@@ -676,11 +676,12 @@ ROM supports the following set of commands before handling the FW_DOWNLOAD comma
...
-13. **GET_LDEV_ECC384_CERT**: This command fetches an LDevID ECC384 certificate signed by the ECC384 IDevID private key. [GET_LDEV_ECC384_CERT](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime#get_ldev_ecc384_cert)
-14. **GET_LDEV_MLDSA87_CERT**: This command fetches an LDevID MLDSA87 certificate signed by the MLDSA87 IDevID private key. [GET_LDEV_MLDSA87_CERT](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime#get_ldev_mldsa87_cert)
-15. **INSTALL_OWNER_PK_HASH**: This command saves the owner public key hash to persistent data. [INSTALL_OWNER_PK_HASH](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime#install_owner_pk_hash)
+13. **CM_SHA**: This ROM-only command (ROM 2.0.1+ only) computes a SHA-384 or SHA-512 hash of input data in a single operation. This is useful for MCU ROM to verify signatures and hashes against Vendor PK hash without needing its own hash implementation. Unlike the runtime CM_SHA_INIT/CM_SHA_UPDATE/CM_SHA_FINAL commands, this is a one-shot operation that does not support streaming or contexts. See [CM_SHA](#cm_sha) below for details.
```
zhalvorsen, line 679: nit: Should we remove the reference to the 2.0 ROM here?

swenson: I kind of like to keep these references, just to keep docs mostly the same when possible.

PR #3098, opened 2026-01-13
[OCP-LOCK]: Generate MEK

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3005

### DIFF
--- a/runtime/src/ocp_lock/mod.rs
+++ b/runtime/src/ocp_lock/mod.rs
@@ -1,23 +1,26 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_api::mailbox::CommandId;
+use caliptra_api::mailbox::{CommandId, WrappedKey};
 use caliptra_cfi_lib_git::{cfi_assert, cfi_assert_eq};
-use caliptra_common::keyids::ocp_lock::{KEY_ID_EPK, KEY_ID_HEK, KEY_ID_MEK_SECRETS};
+use caliptra_common::keyids::ocp_lock::{KEY_ID_EPK, KEY_ID_HEK, KEY_ID_MDK, KEY_ID_MEK_SECRETS};
 use caliptra_drivers::{
     cmac_kdf, hmac_kdf,
     hpke::{HpkeContext, HpkeContextIter, HpkeHandle},
+    preconditioned_aes::preconditioned_aes_encrypt,
     Aes, AesKey, AesOperation, Dma, Hmac, HmacKey, HmacMode, HmacTag, KeyReadArgs, KeyUsage,
     KeyVault, KeyWriteArgs, LEArray4x16, LEArray4x8, SocIfc, Trng,
 };
 use caliptra_error::{CaliptraError, CaliptraResult};
 
 use enumerate_hpke_handles::EnumerateHpkeHandles;
+use generate_mek::GenerateMekCmd;
 use rotate_hpke_key::RotateHpkeKeyCmd;
-use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+use zerocopy::{transmute, FromBytes, Immutable, IntoBytes, KnownLayout};
 use zeroize::ZeroizeOnDrop;
 
 mod derive_mek;
 mod enumerate_hpke_handles;
+mod generate_mek;
 mod get_algorithms;
 mod initialize_mek_secret;
 mod rotate_hpke_key;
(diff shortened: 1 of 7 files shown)


runtime/src/ocp_lock/mod.rs
```diff
@@ -36,8 +39,99 @@ pub struct Dpk(pub [u8; 32]);
 /// NOTE: MPKs will be mixed into `IntermediateMekSecret`
 pub struct IntermediateMekSecret;
 
+/// Represents a `WrappedMek`
+///
+/// OCP LOCK version 1.0 RC2, Section "Common mailbox types".
+pub struct WrappedMek {
+    salt: [u8; 12],
+    iv: [u8; 12],
+    encrypted_key: [u8; 64],
```
zhalvorsen: nit: you could use Self::KEY_LEN here for consistency

```suggestion
    encrypted_key: [u8; Self::KEY_LEN],
```

runtime/src/ocp_lock/mod.rs
```diff
@@ -276,6 +428,27 @@ impl OcpLockContext {
 
         Ok(checksum)
     }
+
+    /// Generate an MEK
+    ///
+    /// NOTE: This operation will consume `intermediate_secret` and erase the MEK secret key vault on
+    /// completion.
+    #[allow(clippy::too_many_arguments)]
```
zhalvorsen: I think this clippy allow can be dropped

runtime/src/ocp_lock/mod.rs
```diff
@@ -40,8 +46,99 @@ pub struct Dpk(pub [u8; 32]);
...
+
+impl WrappedMek {
+    const KEY_TYPE: u16 = 0x03;
+    const KEY_LEN: usize = 64;
+}
+
+impl TryFrom<WrappedMek> for WrappedKey {
+    type Error = CaliptraError;
+    fn try_from(value: WrappedMek) -> Result<Self, Self::Error> {
+        let mut cipher_text_and_auth_tag = [0; 80];
```
bluegate010: Uber-nit, else in my head I read this as "cipher, text, and auth tag"
```suggestion
        let mut ciphertext_and_auth_tag = [0; 80];
```

runtime/src/ocp_lock/mod.rs
```diff
@@ -40,8 +46,99 @@ pub struct Dpk(pub [u8; 32]);
...
+        let (cipher, rem) = <[u8; 64]>::mut_from_prefix(cipher_text_and_auth_tag.as_mut_bytes())
+            .map_err(|_| CaliptraError::RUNTIME_OCP_LOCK_FAILED_TO_CONVERT_WRAPPED_KEY)?;
+        *cipher = value.encrypted_key;
+
+        let tag = <[u8; 16]>::mut_from_bytes(rem)
+            .map_err(|_| CaliptraError::RUNTIME_OCP_LOCK_FAILED_TO_CONVERT_WRAPPED_KEY)?;
+        *tag = value.tag;
```
bluegate010: Could these errors be promoted to panics with copy_from_slice?
```suggestion
        ciphertext_and_auth_tag[0..64].copy_from_slice(value.encrypted_key);
        ciphertext_and_auth_tag[64..].copy_from_slice(value.tag);
```

runtime/src/ocp_lock/mod.rs
```diff
@@ -62,6 +159,27 @@ impl IntermediateMekSecret {
         )?;
         Ok(MekSecret { kv })
     }
+    /// Consumes `Self` to produce a wrapped `MekSecret`.
```
bluegate010: The returned value looks like it's just a derived key, not a 'wrapped' key. Should this just read 'produce an `MekSecret`?

zhalvorsen: Similarly would it be easier to read from a call site if the function was named something like `wrapping_mek_secret`?

runtime/src/ocp_lock/mod.rs
clundin25: Ah I have been using zerocopy too much :) Yes much cleaner.

runtime/src/ocp_lock/mod.rs
clundin25: Done!

runtime/src/ocp_lock/mod.rs
clundin25: Done!

runtime/src/ocp_lock/mod.rs
clundin25: Done!

runtime/src/ocp_lock/mod.rs
clundin25: Done!

PR #3100, opened 2026-01-13
[OCP-LOCK]: HPKE Key Rotation

clundin25:


drivers/src/hpke/mod.rs
```diff
@@ -134,6 +140,31 @@ impl HpkeContext {
...
+        &mut self,
+        trng: &mut Trng,
+        hpke_handle: &HpkeHandle,
+    ) -> CaliptraResult<Option<HpkeHandle>> {
+        for handle in self.priv_keys.iter_mut() {
+            match handle {
+                HpkePrivateKey::MlKem {
+                    ref mut handle,
+                    ref mut context,
+                } if handle == hpke_handle => {
```
bluegate010, line 158: Optional: each enum variant will include a handle, so I wonder if it would make sense to store the handle in a wrapper struct that holds the handle and the enum.

drivers/src/hpke/mod.rs
```diff
@@ -134,6 +140,31 @@ impl HpkeContext {
             index: 0,
         }
     }
+
+    /// Searches for a matching `HpkeHandle`. If it is found a new HPKE seed is generated and a new
+    /// handle is returned
+    ///
+    /// If `hpke_handle` does not match any existing `HpkeHandle`, `None` is returned
```
bluegate010: Minor: seems like this could just be an error condition and we could return RUNTIME_OCP_LOCK_UNKNOWN_HPKE_HANDLE in that case.

drivers/src/hpke/mod.rs
clundin25: I think the enum captures this. The context types will vary so I am not sure a struct will work

drivers/src/hpke/mod.rs
clundin25: Sounds good - split things up a bit too much

drivers/src/hpke/mod.rs
bluegate010: I was meaning that, with the current scheme, you have to reach into the enum to find out the entry's handle, where instead you could have,

```
struct Entry {
  handle: HpkeHandle,
  private_key: HpkePrivateKey,
}

for entry in self.priv_keys.iter_mut() {
  if entry.handle != hpke_handle {
    continue;
  }

  match entry.private_key {
    HpkePrivateKey::MlKem { ref mut context } => { *context = HpkeMlKemContext::generate(trng)?; }
  }

  entry.handle = self.handle_cursor.generate_handle();
  Ok(entry.handle)
}
```

clundin25: I'm happy to make the change but I prefer the "style" of the enum, since I can use pattern matching with it

bluegate010: In my example HpkePrivateKey is still an enum that uses pattern matching. The only difference is that it no longer holds the handle, since the handle's format/representation doesn't vary between keypair algorithms.

But it's fine either way.

bluegate010 (approved):

PR #3101, opened 2026-01-13
[EMU]: Emit stack usage when stack info is set

clundin25:
Additionally, add stack information to the Caliptra driver tests.


sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -300,6 +306,17 @@ pub struct Cpu<TBus: Bus> {
 impl<TBus: Bus> Drop for Cpu<TBus> {
     fn drop(&mut self) {
         if let Some(stack_info) = &self.stack_info {
+            if let Ok(mut file) = File::create("/tmp/caliptra-stack-usage.txt") {
```
zhalvorsen, line 309: Is it guaranteed that this file will only be used in one thread at a time? Or should we use something like `tempfile`?

clundin25: I think this is mostly a nice to have utility for creating optimizations. Future work to bind this to a test case would be useful but out of scope for this change

PR #3104, opened 2026-01-13
Remove manifest2 from persistent_data

zhalvorsen:
This changes all references that used manifest2 to use a stack variable. This frees up 17KB of space in persistent data. This would be enough to add expand DPE to use 64 contexts.


swenson (approved):
LGTM

swenson (approved):
LGTM

PR #3116, opened 2026-01-14
Add ML-DSA RT alias key to FHT

### LINKED ISSUE #3106: [ROM] - Add ML-DSA RT Alias Key to FirmwareHandoffTable
Currently there is just an ECC Key in the FHT: https://github.com/chipsalliance/caliptra-sw/blob/8c277d9f95d738f6154832c9888409f2bc0276e7/drivers/src/hand_off.rs#L203

We need to add an ML-DSA key and populate in the FMC.

zhalvorsen:
Fixes #3106

The ML-DSA RT alias key was missing from the FirmwareHandoffTable. This adds the missing field so it can be used by DPE and OCP LOCK for endorsement.


zhalvorsen:
Should this bump the FHT version number?

runtime/src/disable.rs
```diff
@@ -91,7 +93,7 @@ impl DisableAttestationCmd {
     #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
     fn generate_dice_key(drivers: &mut Drivers) -> CaliptraResult<()> {
         let key_id_rt_cdi = Drivers::get_key_id_rt_cdi(drivers)?;
-        let key_id_rt_priv_key = Drivers::get_key_id_rt_priv_key(drivers)?;
+        let key_id_rt_priv_key = Drivers::get_key_id_rt_ecc_priv_key(drivers)?;
         let pub_key = drivers.ecc384.key_pair(
             Ecc384Seed::Key(KeyReadArgs::new(key_id_rt_cdi)),
```
clundin25, line 96: Is this missing an ML-DSA equivalent function?

zhalvorsen: It is, but that pub key it is replacing also isn't kept in persistent data, but I think we could get away with lazy generating it, because it is so big. Do you think I should rename the function to emphasize it isn't doing anything with the ML-DSA keys?

clundin25: I think a comment explaining why ML-DSA is omitted is enough

zhalvorsen: I realized, it was also populating the key vault with a new zeroized key. It is now populating that value for ML-DSA as well. I put a comment about not populating the public key for ML-DSA

zhalvorsen: It turns out that derivation caused an exception because of a stack overflow. It shouldn't be very deep into the stack at that point. If so, we will need a major refactor to get DPE working. In any case, we will need to revisit the stack for runtime FW and I'd like to get this in early to be included in the ROM release. I have it commented out for now and will take a closer look later.

runtime/src/disable.rs
```diff
@@ -109,6 +113,23 @@ impl DisableAttestationCmd {
             .fht
             .rt_dice_ecc_pub_key = pub_key;
 
+        // TODO: Derive the MLDSA Key Pair. This is causing a stack overflow and
```
timothytrippel: can we tie this TODO to an issue that is tracked? i.e. something like "//TODO(#xyz) ..."

timothytrippel (approved):

runtime/src/disable.rs
zhalvorsen: Done

PR #3145, opened 2026-01-17
[mbox] Separate external commands from mbox commands

zhalvorsen:
By separating these flows a little bit, we are able to gain back 35KB of stack space for internal mbox commands because the `external_cmd_buffer` is only used with external commands.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -227,8 +227,8 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
 
     // Get the command bytes
     let req_packet = Packet::get_from_mbox(drivers)?;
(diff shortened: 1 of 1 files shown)


drivers/src/riscv.rs
```diff
@@ -0,0 +1,18 @@
+// Licensed under the Apache-2.0 license.
+
+/// Return the current stack pointer
+///
+/// Useful while debugging stack usage
+#[inline(always)]
+pub fn stack_pointer() -> usize {
```
swenson: unused? also scares me a bit :)

zhalvorsen: Haha I'll keep it as a personal tool. It was just helpful to pinpoint where big chunks of stack are being used.

zhalvorsen: Removed

PR #3146, opened 2026-01-18
fpga: ModelFpgaSubsystem supports flash-based booting (2.1)

swenson:
We implement the flash logic in the model itself, backed by an array that can be provisioned by tests.
    
This is used only by MCU tests for now.
    
This also fixes a bug in the FPGA-only ROM that was gating on the the INDIRECT_FIFO_STATUS empty when reading from the I3C fifo, and instead gates on `payload_available`, which maps closer to how the standard DMA transfer works.
    
We also ensure that recovery images in FPGA to multiple of 256 bytes before loading them into the recovery interface to work around hardware issues with the `payload_available` bit (which only goes high if writes are always 256 bytes in size).
    
We have temporarily disabled flash until new bitstreams are available.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -679,18 +679,10 @@ impl<'a> DmaRecovery<'a> {
 
         #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
         {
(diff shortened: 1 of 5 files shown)


swenson:
Thanks!

zhalvorsen (approved):

PR #3148, opened 2026-01-20
[EMU]: OCP LOCK feature changes

clundin25:
This makes the emulator work for a larger set of OCP LOCK tests. (test_derive_mek does not yet pass).

* Add OCP LOCK runtime tests to emulator PR check.
* Store last seen MEK for convenience (Not used yet).
* Clean up fuse management (Hek seed fuses were not being set / erased)

### DIFF
--- a/sw-emulator/lib/periph/src/aes_clp.rs
+++ b/sw-emulator/lib/periph/src/aes_clp.rs
@@ -114,6 +114,8 @@ pub struct AesClp {
     op_key_write_complete_action: Option<ActionHandle>,
 
     key: Rc<RefCell<Option<[u8; 32]>>>,
+    /// A copy of the latest released MEK for an easy backdoor to the MEK.
+    latest_mek: Vec<u8>,
     key_destination: Rc<RefCell<AesKeyReleaseOp>>,
 }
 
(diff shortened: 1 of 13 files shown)


sw-emulator/lib/periph/src/aes_clp.rs
```diff
@@ -114,6 +114,8 @@ pub struct AesClp {
     op_key_write_complete_action: Option<ActionHandle>,
 
     key: Rc<RefCell<Option<[u8; 32]>>>,
+    /// A copy of the latest released MEK for an easy backdoor to the MEK.
+    pub latest_mek: Vec<u8>,
```
zhalvorsen: Looks like the rest of the fields are private. Should this be made private and add accessor functions instead?

zhalvorsen (approved):

sw-emulator/lib/periph/src/aes_clp.rs
clundin25: Oh good point - I'll do this in a follow up since this joined the merge queue

zhalvorsen (approved):

PR #3149, opened 2026-01-20
ci: Add nightly 2.0 job and fix test-matrix

swenson:
The nightly 2.0 job is copied almost verbatim from the 2.x job, but using the `caliptra-2.0` branch instead of `main`, and with the appropriate '2-x' names changed to '2-0'.

The test-matrix code no longer worked, so I fixed that as well. It is currently still only configured for 1.x. I originally intended to add test matrices for 2.0 and 2.x as well, but it looks like it will be messy to combine all of the test results, so I saved that for future work.


swenson:
Going to move this into caliptra-2.0 as workflow_dispatch only.

PR #3150, opened 2026-01-20
[2.0 Release]: Rename release tags to not conflict with the latest 2.x

clundin25:
tag

The 2.0 release branch should use 2.0 instead of 2.x


clundin25:
Actually, maybe we don't need this. I'll think on it some more.

PR #3153, opened 2026-01-20
[OCP-LOCK]: Rewrap MPK

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3000.

### DIFF
--- a/runtime/src/ocp_lock/mod.rs
+++ b/runtime/src/ocp_lock/mod.rs
@@ -2,14 +2,16 @@
 
 use caliptra_api::mailbox::{CommandId, WrappedKey, OCP_LOCK_WRAPPED_KEY_MAX_METADATA_LEN};
 use caliptra_cfi_lib_git::{cfi_assert, cfi_assert_eq};
-use caliptra_common::keyids::ocp_lock::{KEY_ID_EPK, KEY_ID_HEK, KEY_ID_MDK, KEY_ID_MEK_SECRETS};
+use caliptra_common::keyids::ocp_lock::{
+    KEY_ID_EPK, KEY_ID_HEK, KEY_ID_LOCKED_MPK_ENCRYPTION_KEY, KEY_ID_MDK, KEY_ID_MEK_SECRETS,
+};
 use caliptra_drivers::{
     cmac_kdf, hmac_kdf,
     hpke::{
         aead::Aes256GCM, suites::CipherSuite, EncryptionContext, HpkeContext, HpkeContextIter,
         HpkeHandle, Receiver,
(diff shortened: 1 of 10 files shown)


runtime/src/ocp_lock/rewrap_mpk.rs
```diff
@@ -0,0 +1,101 @@
...
+        let metadata = current_locked_mpk
+            .metadata
+            .get(..current_locked_mpk.metadata_len as usize)
+            .ok_or(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        let current_ak_ct = cmd
+            .sealed_access_key
+            .ak_ciphertext
+            .get(..cmd.sealed_access_key.access_key_len as usize)
+            .and_then(|ct| <[u8; AccessKey::<Current>::KEY_LEN]>::ref_from_bytes(ct).ok())
```
bluegate010, line 48: > .ok()

Is it ok (heh) to just call this? The documentation says this just discards any error.

clundin25: Maybe a nit for the documentation but it sounds a little confusing. This essentially converts the `Err` to `None`. `and_then` requires an `Option` and we want to use our own `CalitpraError` anyway, so we convert the `Err` to `None` and map `None` to the error we want to raise

bluegate010: Ah gotcha.

runtime/src/ocp_lock/mod.rs
```diff
@@ -486,6 +576,89 @@ impl<'a> Epk<'a> {
...
+            HmacKey::Key(KeyReadArgs::new(KEY_ID_EPK)),
+            Self::LOCKED_MPK_ENCRYPTION_KEY_LABEL,
+            Some(current_access_key.as_bytes()),
+            trng,
+            HmacTag::Key(KeyWriteArgs::new(
+                KEY_ID_LOCKED_MPK_ENCRYPTION_KEY,
+                KeyUsage::default().set_hmac_key_en(),
+            )),
+            HmacMode::Hmac512,
+        )?;
```
zhalvorsen, line 602: Could you add a helper function for this? There are a lot of hard coded parameters that could get messy if we need to change any of them. Feel free to do it in a follow up.

zhalvorsen (approved):

runtime/src/ocp_lock/mod.rs
clundin25: Will do - Thanks !

PR #3154, opened 2026-01-20
sw-emulator: Fully support all 255 interrupts in PIC

swenson:
And add tests that they actually work.

There were a few bugs that limited us to 31 interrupts (which is how many Caliptra supports), even though MCU supports all 255.


swenson:
Thanks!

Issue #3152, opened 2026-01-20
[OCP-LOCK]: Add support for decrypting an MPK in the test mirror of the OCP LOCK key ladder


clundin25:
This type let's us build a parallel but exact replica of Caliptra's key ladder (assuming we know which secrets are in use).

https://github.com/chipsalliance/caliptra-sw/blob/main/test/src/derive.rs#L259 we need to update this type to support decrypting MPKs to better validate test cases for some of the OCP LOCK commands.

PR #3155, opened 2026-01-21
[OCP-LOCK]: Add a derive implementation for decrypting MPKs

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3152.

### DIFF
--- a/test/src/crypto.rs
+++ b/test/src/crypto.rs
@@ -22,10 +22,13 @@ use ml_kem::{
 
 use serde::{Deserialize, Serialize};
 use serde_json;
+use zerocopy::{transmute, IntoBytes};
 
 pub mod hpke;
 
 use hpke::Hpke;
+
+use crate::{swap_word_bytes, swap_word_bytes_inplace};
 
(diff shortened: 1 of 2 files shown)


test/src/crypto.rs
```diff
@@ -601,3 +604,110 @@ fn hpke_mlkem_1024() {
...
+}
+
+pub fn preconditioned_aes_decrypt(
+    key: &[u8],
+    label: &[u8],
+    aad: &[u8],
+    salt: &[u8],
+    iv: &[u8],
+    tag: &[u8],
+    ciphertext: &[u8],
```
zhalvorsen, line 679: optional: could use the PreconditionedAesEncryptionResult here so there are fewer `&[u8]` parameters, it might be easy to mix up some of those parameters and the compiler won't give any indication

zhalvorsen (approved):

test/src/crypto.rs
clundin25: Ah yea that's a good point. I'll do this in a follow up unless the merge queue fails

PR #3162, opened 2026-01-21
rom: Add CM_SHA command

swenson:
This is a one-shot, ROM-only command meant to make some Device Ownership Transfer-like flows easier from MCU ROM or the SoC manager.

For example, if MCU has something like a Vendor PK hash burned into fuses and needs to match an ECDSA or MLDSA public key against it as part of DOT recovery, then it would need a way in the MCU ROM to compute SHA384/512 hashes, either using Caliptra or with a hardware or software SHA capability in MCU. Since Caliptra is meant to be the only cryptography capabilities for MCU, this means we need at least a simple SHA command offered by Caliptra ROM.

(This could potentially be worked around by using the existing CM_HMAC, and having something similar to a Vendor PK hash using HMACs, but this is cumbersome and would need to be computed per-device, since the only HMAC keys available to Caliptra ROM are stable keys, which are device-specific.)

Note: this supports computing the SHA of the entire mailbox contents (minus the command head), so can also be used for stashing.

This is targeted for 2.1 and 2.0.1 ROMs.

(Cherry picked and adapted from #3097)

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -14,6 +14,11 @@ use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Ref};
 
 /// Maximum input data size for cryptographic mailbox commands.
 pub const MAX_CMB_DATA_SIZE: usize = 4096;
+/// Mailbox size in bytes (256 KB).
+pub const MAILBOX_SIZE: usize = 256 * 1024;
+/// Maximum input size for CM_SHA command (mailbox size minus header overhead).
+/// The overhead consists of: MailboxReqHeader (4 bytes) + hash_algorithm (4 bytes) + input_size (4 bytes) = 12 bytes.
+pub const MAX_CM_SHA_INPUT_SIZE: usize = MAILBOX_SIZE - 12;
 /// Maximum output size for AES GCM encrypt or decrypt operations.
 pub const MAX_CMB_AES_GCM_OUTPUT_SIZE: usize = MAX_CMB_DATA_SIZE + 16;
 /// Context size for CMB SHA commands.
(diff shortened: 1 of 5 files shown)


mhatrevi:
>         cprintln!("[idev] ECC SIG.S = {}", HexBytes(&_sig_s));

Can remove these as well.

---
Refers to: rom/dev/src/flow/cold_reset/idev_id.rs:369 in f242b00. [](commit_id = f242b00bfa3fec0bccd75729873f7e99a4bcae06, deletion_comment = False)

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -363,9 +363,6 @@ impl InitDevIdLayer {
 
         let _pub_x: [u8; 48] = key_pair.pub_key.x.into();
         let _pub_y: [u8; 48] = key_pair.pub_key.y.into();
```
mhatrevi: Can remove these.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -363,9 +363,6 @@ impl InitDevIdLayer {
...
-        cprintln!("[idev] ECC PUB.X = {}", HexBytes(&_pub_x));
-        cprintln!("[idev] ECC PUB.Y = {}", HexBytes(&_pub_y));
-
         let _sig_r: [u8; 48] = (&sig.r).into();
         let _sig_s: [u8; 48] = (&sig.s).into();
         cprintln!("[idev] ECC SIG.R = {}", HexBytes(&_sig_r));
```
mhatrevi: Can remove these.

swenson: done

rom/dev/src/flow/cold_reset/idev_id.rs
swenson: done

swenson:
Thanks!

PR #3163, opened 2026-01-22
[OCP-LOCK]: Add metadata to firmware persistent data

clundin25:
The VEK is lazily generated & only overwritten on a cold reset. This means we need to store the VEK state in persistent data.


drivers/src/persistent.rs
```diff
@@ -282,14 +283,21 @@ pub struct DOT_OWNER_PK_HASH {
...
     pub active_hek_seed_slots: u16,
     pub hek_seed_state: u16,
     pub hek_available: bool,
     reserved: [u8; 1],
 }
 
+#[derive(TryFromBytes, IntoBytes, KnownLayout, Zeroize)]
+#[repr(C)]
+pub struct OcpLockMetadataFirmware {
+    pub vek_available: bool,
```
zhalvorsen: Would it make sense to turn this into `bitflags`?

clundin25: Oh that's a good idea! Updated!

zhalvorsen (approved):

PR #3166, opened 2026-01-22
Move okref to a separate common crate

zhalvorsen:
Starts the discussed concept of common firmware libraries. This will be used to optimize stack usage in `caliptra-dpe`.

### DIFF
--- a/common/okref/src/lib.rs
+++ b/common/okref/src/lib.rs
@@ -1,4 +1,5 @@
 // Licensed under the Apache-2.0 license
+#![cfg_attr(not(test), no_std)]
 
 /// This function makes it easier to efficiently deal with Result types with
 /// large `Ok` variants.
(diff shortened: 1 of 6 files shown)


common/okref/src/lib.rs
```diff
@@ -1,12 +1,13 @@
...
+#![cfg_attr(not(test), no_std)]
 
 /// This function makes it easier to efficiently deal with Result types with
 /// large `Ok` variants.
 ///
 /// Consider this code:
 ///
 /// ```
-/// use caliptra_drivers::okref;
+/// use caliptra_okref::okref;
```
bluegate010: Uber-nit: this line could be deleted from this example code block as it's not used until the second block.

zhalvorsen: Good point! Removed

Issue #3001, opened 2025-12-12
ENABLE_MPK command


clundin25:

clundin25:
Implementation with tests locally. Blocked on re-wrap MPK PR.

PR #3168, opened 2026-01-23
[OCP-LOCK]: Implement ENABLE_MPK

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3001.

### DIFF
--- a/runtime/src/ocp_lock/enable_mpk.rs
+++ b/runtime/src/ocp_lock/enable_mpk.rs
@@ -1,0 +1,136 @@
+// Licensed under the Apache-2.0 license
+
+use crate::{mutrefbytes, Drivers};
+
+use caliptra_api::mailbox::{
+    MailboxRespHeader, OcpLockEnableMpkReq, OcpLockEnableMpkResp, WrappedKey,
+};
+use caliptra_cfi_derive_git::cfi_impl_fn;
+
+use caliptra_common::keyids::ocp_lock::{KEY_ID_HEK, KEY_ID_VEK};
+use caliptra_drivers::{
+    hmac_kdf,
+    hpke::{aead::Aes256GCM, HpkeHandle},
+    HmacKey, HmacMode, HmacTag, KeyReadArgs, KeyUsage, KeyWriteArgs, OcpLockFlags,
+};
+use caliptra_error::{CaliptraError, CaliptraResult};
+
+use zerocopy::{FromBytes, IntoBytes};
+
+use super::{AccessKey, Current, LockedMpk, Sek, Vek};
+
+pub struct EnableMpkCmd;
+impl EnableMpkCmd {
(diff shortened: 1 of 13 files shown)


runtime/src/ocp_lock/mod.rs
```diff
@@ -286,6 +300,45 @@ impl TryFrom<&WrappedKey> for LockedMpk {
...
+        ciphertext_and_auth_tag[EnabledMpk::KEY_LEN..EnabledMpk::KEY_LEN + Aes256GCM::NT]
+            .copy_from_slice(&value.tag);
+
+        Ok(Self {
+            key_type: EnabledMpk::KEY_TYPE,
+            salt: value.salt,
+            metadata_len: value.metadata_len,
+            key_len: EnabledMpk::KEY_LEN as u32,
+            iv: value.iv,
+            cipher_text_and_auth_tag: ciphertext_and_auth_tag,
```
zhalvorsen, line 335: nit about something else. I had to look at this for a second to figure out why this line had to be separated. Can you switch the `cipher_text_and_auth_tag` in `WrappedKey` to `ciphertext_and_auth_tag`?

runtime/src/ocp_lock/enable_mpk.rs
```diff
@@ -0,0 +1,136 @@
...
+        )?;
+        drivers
+            .persistent_data
+            .get_mut()
+            .fw
+            .ocp_lock_metadata
+            .flags
+            .set(OcpLockFlags::VEK_AVAILABLE, true);
+        // Mark the VEK as available to the OCP LOCK context.
+        drivers.ocp_lock_context.vek = Some(Vek);
```
bluegate010, line 133: Not necessarily for this PR but would it make sense to fold `drivers.ocp_lock_context` into persistent_data's ocp_lock_metadata?

If persistent_data doesn't have the VEK, does the VEK survive across warm resets?

zhalvorsen: I had a similar thought, it would be nice to not have to keep track of it in two different places. The one in `drivers.ocp_lock_context` doesn't survive a warm reset

runtime/src/ocp_lock/mod.rs
clundin25: Thanks for catching this, for some reason my brain keeps reaching for this (or it reincarnated from a previous patch). I will fix this in my next PR.

runtime/src/ocp_lock/enable_mpk.rs
clundin25: I agree, it's a little bit ugly at the moment. I wanted to keep updating persistent data out of the `ocp_lock` module, but similarly catch any logical errors.

I believe this is the only code path that depends on the VEK, so there is likely some opportunity to coalesce / clean up the code.

I think I will do a "larger" refactoring after I finish landing the series of patches I have locally

PR #3169, opened 2026-01-23
ci: update matrix web builder for caliptra 2.x

leongross:
- add verbose logging
- update nextest-metadata crate to parse `SKIPPED_DEFAULT_FILTER`
- add README.md
- add clap cli
- update `publish-website` workflow
- remove website publishing for version 1.x workflow


ci-tools/test-matrix/src/main.rs
```diff
@@ -155,14 +163,56 @@ struct TestRun {
...
+        //     --message-format json > /tmp/nextest-list-${{ matrix.partition }}.json
+        let json_filename = archive
+            .file_names()
+            .find(|f| f.starts_with("nextest-list") && f.ends_with(".json"))
+            .ok_or("No nextest-list*.json file found in zip archive")?
+            .to_string();
+
+        std::fs::write(format!("{}_debug.zip", json_filename), bytes)?;
+
+        let json = zip_extract_file(&mut archive, &json_filename)
```
clundin25, line 174: nit: Seems like this is now extracting multiple files?

```suggestion
        let json = zip_extract_files(&mut archive, &json_filename)
```

ci-tools/test-matrix/src/main.rs
```diff
@@ -155,14 +163,56 @@ struct TestRun {
...
+        let junit_suites = junit::TestSuites::from_xml(&junit_xml)
+            .map_err(|e| format!("Failed to parse {}: {}", xml_name, e))?;
+        log::info!(
+            "{}/{}: {} suites",
+            name,
+            xml_name,
+            junit_suites.test_suites.len()
+        );
+
+        std::fs::write(format!("{}_valid.zip", json_filename), bytes)?;
```
clundin25: Is this for debugging?

clundin25 (commented):
This looks great, thanks! Also thanks for adding docs + debug info!

ci-tools/test-matrix/src/main.rs
leongross: Yes, that was for triaging the differences in the zip files; this will be removed.

ci-tools/test-matrix/src/main.rs
leongross: The `zip_extract_file` function calls `zip.by_name(.)`, which searches for one file by the provided string as name, and returns its contents, so I think the naming applies here.

leongross (commented):
Since the workflow artifacts are retrieved based on the workflow file, in the case of 2.x this is `nightly-release.yml`, I propose to pass this via argv to the tool to make it compatible to potential future workflow files of interest without modifying the code.

LMK what your thoughts on that are.

clundin25:
> Since the workflow artifacts are retrieved based on the workflow file, in the case of 2.x this is `nightly-release.yml`, I propose to pass this via argv to the tool to make it compatible to potential future workflow files of interest without modifying the code.
> 
> LMK what your thoughts on that are.

@leongross LGTM but I'd prefer `clap` be used so this tool has the same UX as the others used in this repo.

leongross:
> > Since the workflow artifacts are retrieved based on the workflow file, in the case of 2.x this is `nightly-release.yml`, I propose to pass this via argv to the tool to make it compatible to potential future workflow files of interest without modifying the code.
> > LMK what your thoughts on that are.
> 
> @leongross LGTM but I'd prefer `clap` be used so this tool has the same UX as the others used in this repo.

I agree, I will migrate the CLI to clap.

.github/workflows/nightly-release.yml
```diff
@@ -195,3 +195,14 @@ jobs:
         with:
           name: release-info-2-x
           path: /tmp/release-info-2-x
+
+  publish-website:
+    name: Publish https://caliptra-sw.github.io/
+    needs:
+      create-release-2-x
+    uses: chipsalliance/caliptra-sw/.github/workflows/publish-website.yml@main
```
clundin25: nit: Now that this flow is on the same branch we can drop `@main`

.github/workflows/publish-website.yml
```diff
@@ -3,6 +3,14 @@ name: Publish website
 on:
   workflow_dispatch:
   workflow_call:
+    inputs:
+      # github workflow file to get the artifacts for to generate the website matrix
```
clundin25: nit: en example could be useful, e.g.


```suggestion
      # github workflow file to get the artifacts for to generate the website matrix
      # For example, ".github/workflows/nightly-release.yml"
```

.github/workflows/publish-website.yml
```diff
@@ -3,6 +3,14 @@ name: Publish website
...
+      workflow_file:
+        required: true
+        type: string
+      www_out:
+        type: string
+        default: /tmp/www
```
clundin25, line 16: This is interesting, do you have a use case in mind?

.github/workflows/publish-website.yml
```diff
@@ -12,21 +20,26 @@ jobs:
...
         uses: actions/checkout@v3
   
       - name: Generate HTML
         run: |
+          mkdir -p ${{inputs.www_out}}
           cd ci-tools/test-matrix
-          mkdir /tmp/www
-          CPTRA_WWW_OUT=/tmp/www cargo run --release 
 
+          cargo run --release -- \
```
clundin25, line 37: I am getting an error that the `1.85` toolchain is too old. Let's see what happens after merging, but this may need a manual override, e.g. `cargo +1.93 run`

ci-tools/test-matrix/src/main.rs
```diff
@@ -13,9 +10,13 @@ use octocrab::{
 };
 use serde::Serialize;
 use zip::result::ZipError;
+use clap::Parser;
 
 mod html;
 mod junit;
+mod cli;
```
clundin25, line 17: It seems this module has not been pushed?

clundin25:
@leongross I tried to check this PR out locally and compile the code, but saw several compiler errors.

ci-tools/test-matrix/Cargo.toml
```diff
@@ -4,15 +4,25 @@
 name = "test-matrix"
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.93"
```
leongross, line 7: > @leongross I tried to check this PR out locally and compile the code, but saw several compiler errors.

I bumped the Rust version to update the nextest version, maybe that causes the issues on your system. On my machine, I cannot observe any compilation issues:

```
$ rustc --version
rustc 1.93.0 (254b59607 2026-01-19)
$ cargo version
cargo 1.93.0 (083ac5135 2025-12-15)
```

Try running
```
$ rustup override set 1.93.0
```

clundin25: I posted a log here: https://gist.github.com/clundin25/97fe1fd1bbc92ec9a64569d21511e0dc

Do you mind following these steps? I suspect something isn't pushed.

ci-tools/test-matrix/src/main.rs
leongross: That was my bad, I amended it to the last commit.

clundin25 (dismissed):
This LGTM! Thank you @leongross  !

One last request, can you remove the existing publish in the 1.x workflow in this PR?

ci-tools/test-matrix/src/main.rs
```diff
@@ -13,9 +10,13 @@ use octocrab::{
...
+
+const NUM_RUNS: usize = 6;
```
clundin25, line 19: Is it possible to avoid hard coding this?

leongross: As far as I understand it, this value was chosen quite arbitrarily to have a sufficiently large set size but not pull all worklow runs.

Do you want this to be configurable via the CLI interface, or do you propose to inline it and remove the constant?

clundin25: Ah I see, I thought this was a new constant. My apologies!

clundin25:
@leongross remove this section https://github.com/chipsalliance/caliptra-sw/blob/main/.github/workflows/nightly-release-1.x.yml#L187-L194 and rebase the PR then I think this can be merged.

I am not sure what will happen if we have two separate workflows that are publishing the website.

clundin25:
@leongross LGTM! Let's merge this after the next release.

clundin25 (approved):

PR #3171, opened 2026-01-23
sw-emulator: Fully support all 255 interrupts in PIC (#3154)

swenson:
And add tests that they actually work.

There were a few bugs that limited us to 31 interrupts (which is how many Caliptra supports), even though MCU supports all 255.

(cherry picked from commit 17fa701e8d082606601c5b025915ef286b80b3ad)


swenson:
Thanks!

swenson:
Thanks!

PR #3173, opened 2026-01-23
Set RESET_REASON first before clearing FW_EXEC_CTRL (2.1)

mlvisaya:
During hitless update, when processing the firmware activate command, the MCU FW_EXEC_CTRL bit is cleared, this will initiate an MCU request, however, the MCU RESET_REASON is updated after the MCU reset request. MCU might still read the previous reset reason instead of HITLESS_UPDATE reason.

This change ensures that the reset reason is set first before clearing FW_EXEC_CTRL.


swenson (approved):
LGTM

Issue #3170, opened 2026-01-23
Add support for ML-DSA HPKE Key Endorsement


clundin25:
https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/src/ocp_lock/endorse_hpke_pubkey.rs#L66 needs to be updated to sign HPKE pub keys using the runtime ML-DSA key.

PR #3175, opened 2026-01-23
[OCP-LOCK]: Add HPKE Pub Key Endorsement by MLDSA

clundin25:
This signs the HPKE pub key with the Caliptra run time ML-DSA key.

This resolves https://github.com/chipsalliance/caliptra-sw/issues/3170.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -49,7 +49,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 36 * 1024; // Must be 4k aligned
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 163 * 1024;
+pub const RUNTIME_SIZE: u32 = 167 * 1024;
 
 pub use memory_layout::{EXTRA_MEMORY_ORG, PERSISTENT_DATA_ORG};
 pub use wdt::{restart_wdt, start_wdt, stop_wdt, WdtTimeout};
--- a/runtime/src/ocp_lock/endorse_hpke_pubkey.rs
+++ b/runtime/src/ocp_lock/endorse_hpke_pubkey.rs
@@ -10,7 +10,7 @@ use caliptra_cfi_derive_git::cfi_impl_fn;
 
 use caliptra_common::{
     crypto::{Crypto, PubKey},
-    dice::ecc384_cert_from_tbs_and_sig,
+    dice::{ecc384_cert_from_tbs_and_sig, mldsa87_cert_from_tbs_and_sig},
     x509,
 };
 use caliptra_drivers::hpke::{
@@ -21,7 +21,8 @@ use caliptra_drivers::hpke::{
 use caliptra_error::{CaliptraError, CaliptraResult};
 
 use caliptra_x509::{
-    NotAfter, NotBefore, OcpLockMlKemCertTbsEcc384, OcpLockMlKemCertTbsEcc384Params,
+    OcpLockMlKemCertTbsEcc384, OcpLockMlKemCertTbsEcc384Params, OcpLockMlKemCertTbsMlDsa87,
+    OcpLockMlKemCertTbsMlDsa87Params,
 };
 use zerocopy::FromBytes;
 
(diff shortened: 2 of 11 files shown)


common/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 36 * 1024; // Must be 4k aligned
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 160 * 1024;
```
clundin25: Note for reviewers, this change is about ~3KiB of code space. The other 4 is from the outstanding OCP LOCK commands that are stacked in my local branch.

runtime/src/ocp_lock/endorse_hpke_pubkey.rs
```diff
@@ -131,4 +135,59 @@ impl EndorseHpkePubkeyCmd {
             _ => Err(CaliptraError::RUNTIME_OCP_LOCK_UNKNOWN_KEM_ALGORITHM)?,
         }
     }
+
+    fn create_mldsa_endorsed_certificate(
```
bluegate010, line 139: Looks like this code is missing the HPKEIdentifiers extension?

clundin25: The `HPKEIdentifier` is embedded during template generation [src](https://github.com/chipsalliance/caliptra-sw/blob/main/x509/build/build.rs#L283-L287).

That said, the tests should verify it's presence.

At a glance, I am not sure the OID for the `HPKEIdentifier extension` is correct anymore, either there is some drift or I use the wrong OID earlier: https://github.com/chipsalliance/caliptra-sw/blob/main/x509/build/x509.rs#L518.

Will see if I can re-trace my steps.

clundin25: Here is a sample cert: https://gist.github.com/clundin25/ef2f5fceecb691a60d80e7961bd6598f

clundin25: Okay - Source for OID is the TCG MEK-MPA spec. The oid is correct as of v1 rc1.

bluegate010: Looks like we're hardcoding [ML-KEM-1024](https://github.com/chipsalliance/caliptra-sw/blob/1ed1a6c3aefcd27ec245e8029918c2115ccd963b/x509/build/build.rs#L284), will we need to update that when we add other KEM algorithms?

clundin25: (Not sure why the GitHub links aren't resolving to the exact lines)

There is also a template for `ECDH_P384`: https://github.com/chipsalliance/caliptra-sw/blob/1ed1a6c3aefcd27ec245e8029918c2115ccd963b/x509/build/build.rs#L291-L311.

A template for hybrid `ML-KEM & ECDH_P384` is missing however

bluegate010: Oh I see. Nevermind then.

clundin25:
> Approval pending tests that cover the HpkeIdentifiers extension.

@bluegate010 I have added test coverage for the `HpkeIdentifiers` extension. This exposed a bug, where each ID was being encoded as `Sequence { INTEGER }`. I have fixed the encoding to remove the nested `Sequence` and updated the OCP LOCK certificate templates.

bluegate010:
> > Approval pending tests that cover the HpkeIdentifiers extension.
> 
> @bluegate010 I have added test coverage for the `HpkeIdentifiers` extension. This exposed a bug, where each ID was being encoded as `Sequence { INTEGER }`. I have fixed the encoding to remove the nested `Sequence` and updated the OCP LOCK certificate templates.

Thanks!

runtime/src/ocp_lock/endorse_hpke_pubkey.rs
```diff
@@ -131,4 +135,59 @@ impl EndorseHpkePubkeyCmd {
...
+        drivers: &mut Drivers,
+        pub_key: &[u8],
+        kem_id: &KemId,
+        cert_buf: &mut [u8],
+    ) -> CaliptraResult<usize> {
+        let not_before = NotBefore::default();
+        let not_after = NotAfter::default();
```
zhalvorsen: Should these be set by the image manifest similar to https://github.com/chipsalliance/caliptra-sw/blob/fc971ba5f7435be68964ae49b3a759287c0e34e7/runtime/src/drivers.rs#L922C12-L922C34?

clundin25: I think that is correct - I'll send a follow up patch

PR #3176, opened 2026-01-24
drivers: Refactor MLKEM / MLDSA drivers to use common AbrReg

### LINKED ISSUE #3107: [Drivers] - Instantiate ML-KEM & ML-DSA Drivers in a closure that takes a mutable reference to the ABR Register
Both types can't take ownership of the ABR registers. They will have to be lazily created to prevent contention on the ABR interface.

swenson:
That is borrowed in a closure to avoid both drivers having a copy of AbrReg.

Most uses can be replaced with the closures, but there are some tricky borrowing situations where it is easier to borrow the underlying AbrReg itself.

Fixes #3107.

### DIFF
--- a/runtime/src/fips.rs
+++ b/runtime/src/fips.rs
@@ -103,27 +103,29 @@ pub mod fips_self_test_cmd {
             runtime: rt,
         };
 
-        let mut venv = FirmwareImageVerificationEnv {
-            sha256: &mut env.sha256,
-            sha2_512_384: &mut env.sha2_512_384,
-            sha2_512_384_acc: &mut env.sha2_512_384_acc,
-            soc_ifc: &mut env.soc_ifc,
-            ecc384: &mut env.ecc384,
-            mldsa87: &mut env.mldsa87,
-            data_vault: &env.persistent_data.get().rom.data_vault,
-            pcr_bank: &mut env.pcr_bank,
-            image_source,
-            persistent_data: &env.persistent_data.get(),
-        };
+        env.abr.with_mldsa87(|mut mldsa87| {
+            let mut venv = FirmwareImageVerificationEnv {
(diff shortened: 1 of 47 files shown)


zhalvorsen (dismissed):

clundin25 (dismissed):
This is awesome, thank you !

runtime/src/fips.rs
```diff
@@ -141,42 +143,44 @@ pub mod fips_self_test_cmd {
 
     /// Execute KAT for cryptographic algorithms implemented in H/W.
     fn execute_kats(env: &mut Drivers) -> CaliptraResult<()> {
-        let mut kats_env = caliptra_kat::KatsEnv {
-            // sha256
-            sha256: &mut env.sha256,
+        env.abr.with_mldsa87(|mut mldsa87| {
```
zhalvorsen, line 146: Do we not have an ML-KEM KAT?

swenson: Wait, you're right, we don't.

swenson: Filed this as an issue #3278. Thanks for pointing that out!

runtime/src/ocp_lock/endorse_hpke_pubkey.rs
```diff
@@ -0,0 +1,198 @@
+// Licensed under the Apache-2.0 license
```
clundin25: Looks like the merge conflict got messed up, this file was deleted & shouldn't be re-added

swenson: Oh that's my bad. I'll delete it.

zhalvorsen (approved):

swenson:
Thanks!

PR #3186, opened 2026-01-26
add common eat encoder crate

parvathib:
Add common EAT crate for encoding OCP signed EAT structures.


clundin25:
@parvathib do you mind signing the CLA?

PR #3188, opened 2026-01-26
skip the fuse enforcement in the case the lifecycle state is unprovisioned in set_auth_manifest

### LINKED ISSUE #3134: SET_AUTH_MANIFEST should not check pqc_key_type fuse in unprovisioned mode
SET_AUTH_MANIFEST checks the pqc_key_type fuse in unprovisioned mode.

https://github.com/chipsalliance/caliptra-sw/blob/c5298f664d43321a4365dcd872ae891fddd5527e/runtime/src/set_auth_manifest.rs#L727-L729

This is not consistent with the image verifier which does not enforce this is set in unprovisioned mode.

https://github.com/chipsalliance/caliptra-sw/blob/c5298f664d43321a4365dcd872ae891fddd5527e/image/verify/src/verifier.rs#L111-L117

https://github.com/chipsalliance/caliptra-sw/blob/c5298f664d43321a4365dcd872ae891fddd5527e/common/src/verifier.rs#L445-L450

SET_AUTH_MANIFEST should also skip this check in unprovisioned mode and the error code should be updated to be unique from the image_verifier failure.

meilu-git:
also use unique error code insetad of IMAGE_VERIFIER_ERR_INVALID_PQC_KEY_TYPE_IN_FUSE

### DIFF
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -1729,6 +1729,21 @@ impl CaliptraError {
             0x000E007E,
             "Runtime Error: RT current PCR validation failed"
         ),
+        (
+             RUNTIME_AUTH_MANIFEST_INVALID_PQC_KEY_TYPE_IN_FUSE,
+            0x000E008A,
+            "Runtime Error: Auth manifest invalid PQC key type in fuse"
+        ),
+               (
+             RUNTIME_AUTH_MANIFEST_INVALID_PQC_KEY_TYPE,
+            0x000E008B,
+            "Runtime Error: Auth manifest invalid PQC key type"
+        ),
+        (
+            RUNTIME_AUTH_MANIFEST_PQC_KEY_TYPE_MISMATCH,
+            0x000E008C,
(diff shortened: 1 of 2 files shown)


nquarton (dismissed):
Looks right to me, thanks!

error/src/lib.rs
```diff
@@ -1729,6 +1729,21 @@ impl CaliptraError {
             0x000E007E,
             "Runtime Error: RT current PCR validation failed"
         ),
+        (
+             RUNTIME_AUTH_MANIFEST_INVALID_PQC_KEY_TYPE_IN_FUSE,
+            0x000E008A,
+            "Runtime Error: Auth manifest invalid PQC key type in fuse"
+        ),
+               (
```
zhalvorsen: It looks like you need to run `cargo fmt`

zhalvorsen (dismissed):

nquarton (approved):

zhalvorsen (approved):

Issue #3004, opened 2025-12-12
TEST_ACCESS_KEY command


clundin25:

clundin25:
Implemented, just blocked on merging other features.

PR #3192, opened 2026-01-27
 [OCP-LOCK]: Implement TEST_ACCESS_KEY

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3004.

### DIFF
--- a/runtime/src/ocp_lock/mod.rs
+++ b/runtime/src/ocp_lock/mod.rs
@@ -13,8 +13,10 @@ use caliptra_drivers::{
         HpkeHandle, Receiver,
     },
     preconditioned_aes::{preconditioned_aes_decrypt, preconditioned_aes_encrypt},
-    Aes, AesKey, AesOperation, Dma, Hmac, HmacKey, HmacMode, HmacTag, KeyReadArgs, KeyUsage,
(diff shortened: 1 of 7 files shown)


runtime/src/ocp_lock/mod.rs
```diff
@@ -1147,6 +1151,42 @@ impl OcpLockContext {
...
+    /// Tests an access key and returns `SHA2-384(metadata || decrypted access key || nonce)` to
+    /// provide proof
+    #[allow(clippy::too_many_arguments)]
+    pub fn test_access_key(
+        &mut self,
+        aes: &mut Aes,
+        hmac: &mut Hmac,
+        trng: &mut Trng,
+        sha2: &mut Sha2_512_384,
+        kv: &mut KeyVault,
```
zhalvorsen, line 1163: Can be done in a follow up. But I'm thinking it might be good to just pass the `Drivers` to all of these functions so we can reduce the number of arguments. What do you think?

clundin25: I am not sure if that will work, since `Drivers` owns this context object.

You could wrap the registers in a struct, but all you are really buying is a reduction in function params

zhalvorsen: Oh, you're right. Never mind

PR #3193, opened 2026-01-27
sw-emulator: Don't panic if watchdog timers are set to large values

swenson:
This is one of the few places in our code base that could cause a timer panic by setting a value larger than `i64::MAX` based on user input.

This means that theoretically the emulator will sleep for an incorrect amount of time if the number of ticks starts getting close to `i64::MAX`, but this edge case is virtually impossible in practice, as we would have to run the emulator for approximately 200,000+ years (at 1 MHz) to trigger this behavior.

(We're seeing some panics in test runs. I don't think this is the root cause, but we might as well fix these issues while we're here.)


swenson:
Thanks!

PR #3197, opened 2026-01-28
fpga: Clear MCI generic input wires on cold reset and add delay

swenson:
The MCI generic wires allowed the MCU ROM to continue loading fuses even if we were in the middle of writing them. So we clear it to prevent race conditions.

The debug manuf service register should also be written before we allow MCU ROM to write fuses.

We also limit how long we wait for MCU ROM to boot to 100,000,000 cycles so that we don't wait forever if it is hung.


swenson:
Thanks!

PR #3201, opened 2026-01-28
Set FW_EXEC_CTRL value before resetting MCU (2.1)

mlvisaya:
The value of FW_EXEC_CTRL controls the behavior of MCU reset.

Currently, if the FW_EXEC_CTRL[2] is cleared, it will trigger an MCI interrupt to MCU which will cause the MCU to request MCI for a reset. The MCI emulator peripheral will then check the value of FW_EXEC_CTRL[2]. If the value is 1, it will immediately reboot, but if the value is 0, then the MCU will halt until FW_EXEC_CTRL[2] is set back to 1.

This sequence causes the MCI emulator peripheral to read the old value of FW_EXEC_CTRL resulting to the incorrect behavior.

The FW_EXEC_CTRL bit value must be set first before triggering the MCI interrupt so that MCI emulator peripheral will read the latest value.

### DIFF
--- a/sw-emulator/lib/periph/src/soc_reg.rs
+++ b/sw-emulator/lib/periph/src/soc_reg.rs
@@ -1420,14 +1420,14 @@ impl SocRegistersImpl {
         if index >= self.ss_generic_fw_exec_ctrl.len() {
             return Err(BusError::StoreAccessFault);
         }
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM

PR #3205, opened 2026-01-28
Move AES KATs to drivers and run lazily on first use

ArthurHeymans:
This change implements lazy KAT execution for AES operations, similar to
the SHA1 KAT changes in PR #3190. The key changes are:

- Move AES KAT files from kat/src/ to drivers/src/kats/
- Add AesKatState bitflags to track which KATs have been executed
- Each AES operation (ECB, CBC, CTR, CMAC, GCM) now runs its KAT on
  first use via ensure_*_kat() helper methods
- Internal *_impl() methods are pub(crate) so KAT tests can call them
  without triggering recursive KAT checks
- Runtime FIPS self-test explicitly runs all AES KATs to ensure
  unconditional KAT execution for FIPS compliance

The CmacKdfKat remains in kat/src/ and continues to use env.aes since
it tests the CMAC KDF function which internally calls aes.cmac().


drivers/src/aes.rs
```diff
@@ -595,6 +667,22 @@ impl Aes {
         plaintext: &[u8],
         ciphertext: &mut [u8],
         tag_size: usize,
+    ) -> CaliptraResult<(AesGcmIvBlock, AesGcmTag)> {
+        self.ensure_gcm_kat(trng)?;
```
zhalvorsen: Is there are reason this is done on each driver call instead of the Sha1 strategy of not being able to get the driver at initialization until the KATs pass?

ArthurHeymans: yes we only use AES-GCM in rom and this way the other KAT get optimized away

ArthurHeymans:
> I like the approach of running KATs only as needed (which we can see helps shrink the ROM by ~2KB) and the fact that this is a less drastic change in terms of the diff size, but I worry that we're moving further away from the compile-time guarantees that the Sha1 PR did, since there is no protection that if we modified the code in the future to add a new function or accidentally lost the KAT in a rebase, we would have no easy way to tell that KATs were not run.
> 
> That said, I might be okay with this approach for 2.1.0 in the interest of time, with the intention of moving toward a compile-time mechanism for 2.1.1 / 2.2 timeframes?
> 
> What do others think?

We could also add it to new() but use a feature flag to guard things depending on feature flag rom?

zhalvorsen:
We could create a separate struct `RomAes` that only has the ROM functionality we need. It could live nested in the regular `Aes` object. Then we get the compile time guarantees as well as the code savings.

zhalvorsen:
I guess if it is only AES-GCM, we could call it `AesGcm`

kat/src/lib.rs
```diff
@@ -104,30 +92,30 @@ pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<InitializedDrivers> {
     cprintln!("[kat] HKDF-512");
     Hkdf512Kat::default().execute(env.hmac, env.trng)?;
 
-    cprintln!("[kat] KDF-CMAC");
-    CmacKdfKat::default().execute(env.aes)?;
+    // ROM uses AesGcm::new() which runs its own GCM and CMAC-KDF KATs at construction time,
+    // so we only run all AES KATs for non-ROM builds (e.g., runtime).
+    #[cfg(not(feature = "rom"))]
```
swenson, line 103: I think we should make this a positive feature (`runtime`) rather than negative, that way if we accidentally include the `rom` feature in `runtime`, then we won't suddenly be missing KATs. (This could happen due to feature flag unification.)

We could do this in a follow-up issue/PR though. We don't need to block this one.

swenson (dismissed):
LGTM. Great work!

runtime/src/fips.rs
```diff
@@ -177,6 +177,11 @@ pub mod fips_self_test_cmd {
         };
 
         caliptra_kat::execute_kat(&mut kats_env)?;
+
+        // Note: AES KATs are run lazily on first use of each AES mode.
```
zhalvorsen: nit: this comment is no longer true with your recent update right?

rom/dev/src/rom_env.rs
```diff
@@ -78,18 +79,43 @@ pub struct RomEnv {
...
+        if soc_ifc.debug_locked() {
+            let mut wdt_timeout_cycles = soc_ifc.wdt1_timeout_cycle_count();
+            if wdt_timeout_cycles == 0 {
+                wdt_timeout_cycles = 1;
+            }
+            caliptra_common::wdt::start_wdt(
+                &mut soc_ifc,
+                WdtTimeout::from(core::num::NonZeroU64::new(wdt_timeout_cycles).unwrap()),
+            );
+        }
```
zhalvorsen: It looks like this is copy-pasted from `rom/dev/src/wdt.rs`, is there a reason we don't use it directly?


```suggestion
        crate::wdt::start_wdt(&mut soc_ifc);
```

kat/src/lib.rs
```diff
@@ -104,30 +92,30 @@ pub fn execute_kat(env: &mut KatsEnv) -> CaliptraResult<InitializedDrivers> {
...
+        cprintln!("[kat] AES-ECB");
+        caliptra_drivers::kats::execute_ecb_kat(env.aes)?;
+        cprintln!("[kat] AES-CBC");
+        caliptra_drivers::kats::execute_cbc_kat(env.aes)?;
+        cprintln!("[kat] AES-CTR");
+        caliptra_drivers::kats::execute_ctr_kat(env.aes)?;
+        cprintln!("[kat] AES-CMAC");
+        caliptra_drivers::kats::execute_cmac_kat(env.aes)?;
+        cprintln!("[kat] AES-GCM");
+        caliptra_drivers::kats::execute_gcm_kat(env.aes, env.trng)?;
```
zhalvorsen, line 116: `caliptra_drivers::kats::execute_gcm_kat(env.aes, env.trng)?;` and `caliptra_drivers::kats::execute_cmackdf_kat(env.aes)?;` still need to be run in this function because it gets called by the `SELF_TEST` command in ROM per FIPS requirements

drivers/src/aes.rs
```diff
@@ -645,6 +654,38 @@ impl Aes {
         Ok(())
     }
 
+    /// Internal implementation of AES-256-GCM encrypt (called by KAT and AesGcm wrapper)
```
zhalvorsen: Are these needed anymore now that it isn't checking KAT status on each call?

drivers/src/aes.rs
```diff
@@ -188,7 +190,15 @@ fn wait_for_idle(aes: &caliptra_registers::aes::RegisterBlock<ureg::RealMmioMut<
...
 impl Aes {
+    /// Create a new AES driver.
     pub fn new(aes: AesReg, aes_clp: AesClpReg) -> Self {
+        if cfg!(feature = "rom") {
+            panic!("Do not use in ROM!");
+        }
+        Self { aes, aes_clp }
+    }
+
+    pub(crate) fn new_gcm(aes: AesReg, aes_clp: AesClpReg) -> Self {
```
zhalvorsen: I think this can be private to this file:


```suggestion
    fn new_gcm(aes: AesReg, aes_clp: AesClpReg) -> Self {
```

zhalvorsen:
Thanks Arthur, this is awesome!

PR #3207, opened 2026-01-28
runtime: kek_next_iv needs to be initialized to a random value

### LINKED ISSUE #3203: [runtime] kek_next_iv is never initialized to a random value in cryptographic mailbox
https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/src/cryptographic_mailbox.rs#L104

We initialize context_next_iv twice, but not kek_next_iv

swenson:
In the cryptographic mailbox, we were accidentally setting the context IV twice instead of initializing the KEK next IV.

This shouldn't be too much of an issue, since the KEK is generated randomly on each boot, so there is not a chance that the IV is reused with the same key.

Fixes #3203


mhatrevi (approved):

swenson:
Thanks!

PR #3216, opened 2026-01-29
[dpe] Update DPE submodule to latest commit

zhalvorsen:


.cargo/config.toml
```diff
@@ -1,6 +1,5 @@
 # Licensed under the Apache-2.0 license
 [env]
-ARBITRARY_MAX_HANDLES = "32"
```
clundin25, line 3: What's the reason behind this?

zhalvorsen: In https://github.com/chipsalliance/caliptra-dpe/pull/498 we set the default DPE configuration to Caliptra so we can drop this variable because it is already set to 32 contexts by default

clundin25: Gotcha - thanks!

PR #3217, opened 2026-01-29
[dpe] Remove DPE submodule

zhalvorsen:
This change removes the DPE submodule from the caliptra-sw repository and updates the Cargo.toml and Go module files to reference the DPE code directly from its Git repository. This simplifies dependency management by eliminating the need for a submodule while ensuring that the DPE code is still accessible for building and testing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,9 +160,6 @@ convert_case = "0.6.0"
 cms = "0.2.2"
 ctr = "0.9.2"
 der = { version = "0.7.10", features = ["derive", "alloc"] }
-dpe = { path = "dpe/dpe", default-features = false, features = ["p384", "arbitrary_max_handles"] }
-crypto = { path = "dpe/crypto", default-features = false }
-platform = { path = "dpe/platform", default-features = false }
 elf = "0.7.2"
 fips204 = "0.4.6"
 fslock = "0.2.1"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -221,6 +218,16 @@ sec1 = { version = "0.7.3" }
 itertools = { version = "0.12.0", default-features = false }
 riscv = "0.13.0"
 
+# DPE dependencies; keep git revs in sync alongside test/dpe_verification/go.mod
```
clundin25, line 221: We should add a small check to validate this.

Opened https://github.com/chipsalliance/caliptra-sw/issues/3220 to track.

Cargo.toml
```diff
@@ -221,6 +218,16 @@ sec1 = { version = "0.7.3" }
...
+dpe = { git = "https://github.com/chipsalliance/caliptra-dpe", rev = "2729a0d492577c3d0a27f529c004b678e55c0702", default-features = false, features = ["p384", "arbitrary_max_handles"] }
+crypto = { git = "https://github.com/chipsalliance/caliptra-dpe", rev = "2729a0d492577c3d0a27f529c004b678e55c0702", default-features = false }
+platform = { git = "https://github.com/chipsalliance/caliptra-dpe", rev = "2729a0d492577c3d0a27f529c004b678e55c0702", default-features = false }
+
+# local DPE dependency; useful when developing
+# dpe = { path = "../caliptra-dpe/dpe", default-features = false, features = ["p384", "arbitrary_max_handles"] }
+# crypto = { path = "../caliptra-dpe/crypto", default-features = false }
+# platform = { path = "../caliptra-dpe/platform", default-features = false }
```
clundin25, line 229: mini nit: Don't love committing commented stuff but I understand that uncommenting it might be faster vs swapping the above to a path.

swenson: This is my fault since I established this pattern in caliptra-mcu-sw I think :) especially since `path` dependencies have a different syntax that relies on remembering the entire path, rather than just linking to the parent directory, it's a real pain to have to rediscover this every time you are doing local development.

swenson (approved):
LGTM

Cargo.toml
zhalvorsen: That's a good idea, thanks!

Cargo.toml
zhalvorsen: I'm usually not a fan either, but I agree it makes it really easy to switch back and forth between local development. Let's keep it for now, we can always pull it out later

jhand2:
Sweet. Thanks for fixing this :)

PR #3218, opened 2026-01-29
Add common EAT encoder for several OCP EAT structures

parvathib:
Add common EAT encoder for several OCP EAT structures


common/eat/decoder/decode.py
```diff
@@ -0,0 +1,1314 @@
+#!/usr/bin/env python3
```
swenson: Could we remove these for now? They are still present in the MCU history I believe if we need them. But if they aren't part of an actively used test (and preferably converted to Rust), I'd prefer to just not have them so we don't forget about them and have them go stale.

parvathib:
> These changes are all from MCU, right? So it's not like these have never been reviewed? It's just a large PR :)

These are from the reviewed changed from MCU + clippy fixes in unit tests which caliptra-sw CI reported.

clundin25:
> These are from the reviewed changed from MCU + clippy fixes in unit tests which caliptra-sw CI reported.

It would be good to track updating MCU to have a similar clippy lint

common/eat/decoder/decode.py
parvathib: removed the python decoder

swenson (approved):
LGTM

Issue #2991, opened 2025-12-12
GET_STATUS command


clundin25:

clundin25:
Assigned to @GwangbaeCHOI (Implementation complete, working on tests).

clundin25:
@GwangbaeCHOI to send a PR soon, tests + implementation complete locally.

clundin25:
Completed by @GwangbaeCHOI in https://github.com/chipsalliance/caliptra-sw/pull/3219

PR #3219, opened 2026-01-29
[OCP-LOCK] Add encryption engine related commands

GwangbaeCHOI:
* Add GET_STATUS command (#2991)
* Add CLEAR_KEY_CACHE command (#2992)
* Add UNLOAD_MEK command (#3009)

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -62,6 +62,14 @@ pub const OCP_LOCK_WRAPPED_KEY_MAX_INFO_LEN: usize = 256;
 /// Currently the largest algorithm is the hybrid ML-KEM + P-384.
 pub const OCP_LOCK_MAX_ENC_LEN: usize = 1665;
 
+/// Metadata size for encryption engine interaction supported by OCP LOCK
+pub const OCP_LOCK_ENCRYPTION_ENGINE_METADATA_SIZE: usize = 20;
+const _: () = assert!(OCP_LOCK_ENCRYPTION_ENGINE_METADATA_SIZE % size_of::<u32>() == 0);
+
+/// Aux size for encryption engine interaction supported by OCP LOCK
+pub const OCP_LOCK_ENCRYPTION_ENGINE_AUX_SIZE: usize = 32;
+const _: () = assert!(OCP_LOCK_ENCRYPTION_ENGINE_AUX_SIZE % size_of::<u32>() == 0);
+
 #[derive(PartialEq, Eq)]
 pub struct CommandId(pub u32);
 
@@ -226,6 +234,9 @@ impl CommandId {
     pub const OCP_LOCK_REWRAP_MPK: Self = Self(0x5245_5750); // "REWP"
     pub const OCP_LOCK_ENABLE_MPK: Self = Self(0x524D_504B); // "RMPK"
     pub const OCP_LOCK_TEST_ACCESS_KEY: Self = Self(0x5441_434B); // "TACK"
+    pub const OCP_LOCK_CLEAR_KEY_CACHE: Self = Self(0x434C_4B43); // "CLKC"
+    pub const OCP_LOCK_GET_STATUS: Self = Self(0x4753_5441); // "GSTA"
+    pub const OCP_LOCK_UNLOAD_MEK: Self = Self(0x554D_454B); // "UMEK"
 
     pub const REALLOCATE_DPE_CONTEXT_LIMITS: Self = Self(0x5243_5458); // "RCTX"
 }
@@ -380,6 +391,9 @@ pub enum MailboxResp {
     OcpLockRewrapMpk(OcpLockRewrapMpkResp),
     OcpLockEnableMpk(OcpLockEnableMpkResp),
     OcpLockTestAccessKey(OcpLockTestAccessKeyResp),
+    OcpLockGetStatus(OcpLockGetStatusResp),
+    OcpLockClearKeyCache(OcpLockClearKeyCacheResp),
+    OcpLockUnloadMek(OcpLockUnloadMekResp),
 }
 
 pub const MAX_RESP_SIZE: usize = size_of::<MailboxResp>();
@@ -456,6 +470,9 @@ impl MailboxResp {
             MailboxResp::OcpLockRewrapMpk(resp) => Ok(resp.as_bytes()),
             MailboxResp::OcpLockEnableMpk(resp) => Ok(resp.as_bytes()),
             MailboxResp::OcpLockTestAccessKey(resp) => Ok(resp.as_bytes()),
+            MailboxResp::OcpLockGetStatus(resp) => Ok(resp.as_bytes()),
+            MailboxResp::OcpLockClearKeyCache(resp) => Ok(resp.as_bytes()),
+            MailboxResp::OcpLockUnloadMek(resp) => Ok(resp.as_bytes()),
         }
     }
 
@@ -530,6 +547,9 @@ impl MailboxResp {
             MailboxResp::OcpLockRewrapMpk(resp) => Ok(resp.as_mut_bytes()),
             MailboxResp::OcpLockEnableMpk(resp) => Ok(resp.as_mut_bytes()),
             MailboxResp::OcpLockTestAccessKey(resp) => Ok(resp.as_mut_bytes()),
+            MailboxResp::OcpLockGetStatus(resp) => Ok(resp.as_mut_bytes()),
+            MailboxResp::OcpLockClearKeyCache(resp) => Ok(resp.as_mut_bytes()),
+            MailboxResp::OcpLockUnloadMek(resp) => Ok(resp.as_mut_bytes()),
         }
     }
 
@@ -671,6 +691,9 @@ pub enum MailboxReq {
     ExternalMailboxCmd(ExternalMailboxCmdReq),
     FeProg(FeProgReq),
     ReallocateDpeContextLimits(ReallocateDpeContextLimitsReq),
+    OcpLockGetStatus(OcpLockGetStatusReq),
+    OcpLockClearKeyCache(OcpLockClearKeyCacheReq),
+    OcpLockUnloadMek(OcpLockUnloadMekReq),
 }
 
 pub const MAX_REQ_SIZE: usize = size_of::<MailboxReq>();
@@ -762,6 +785,9 @@ impl MailboxReq {
             MailboxReq::ExternalMailboxCmd(req) => Ok(req.as_bytes()),
             MailboxReq::FeProg(req) => Ok(req.as_bytes()),
             MailboxReq::ReallocateDpeContextLimits(req) => Ok(req.as_bytes()),
+            MailboxReq::OcpLockGetStatus(req) => Ok(req.as_bytes()),
+            MailboxReq::OcpLockClearKeyCache(req) => Ok(req.as_bytes()),
+            MailboxReq::OcpLockUnloadMek(req) => Ok(req.as_bytes()),
         }
     }
 
@@ -851,6 +877,9 @@ impl MailboxReq {
             MailboxReq::ExternalMailboxCmd(req) => Ok(req.as_mut_bytes()),
             MailboxReq::FeProg(req) => Ok(req.as_mut_bytes()),
             MailboxReq::ReallocateDpeContextLimits(req) => Ok(req.as_mut_bytes()),
+            MailboxReq::OcpLockGetStatus(req) => Ok(req.as_mut_bytes()),
+            MailboxReq::OcpLockClearKeyCache(req) => Ok(req.as_mut_bytes()),
+            MailboxReq::OcpLockUnloadMek(req) => Ok(req.as_mut_bytes()),
         }
     }
 
@@ -946,6 +975,9 @@ impl MailboxReq {
             MailboxReq::OcpLockRewrapMpk(_) => CommandId::OCP_LOCK_REWRAP_MPK,
             MailboxReq::OcpLockEnableMpk(_) => CommandId::OCP_LOCK_ENABLE_MPK,
             MailboxReq::OcpLockTestAccessKey(_) => CommandId::OCP_LOCK_TEST_ACCESS_KEY,
+            MailboxReq::OcpLockGetStatus(_) => CommandId::OCP_LOCK_GET_STATUS,
+            MailboxReq::OcpLockClearKeyCache(_) => CommandId::OCP_LOCK_CLEAR_KEY_CACHE,
+            MailboxReq::OcpLockUnloadMek(_) => CommandId::OCP_LOCK_UNLOAD_MEK,
         }
     }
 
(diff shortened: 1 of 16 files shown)


sw-emulator/lib/periph/src/dma.rs
```diff
@@ -501,7 +501,6 @@ impl Dma {
             let data = match self.fifo.pop_front() {
                 Some(b) => b,
                 None => {
-                    self.status0.reg.write(Status0::ERROR::SET);
```
GwangbaeCHOI, line 504: When accumulated data in the DMA emulator's FIFO are not enough at the first polling, this line triggers a runtime error rather than waiting for next polling. So, it should be dropped to properly emulate the DMA.

runtime/src/ocp_lock/mod.rs
```diff
@@ -1256,6 +1264,118 @@ impl OcpLockContext {
...
+    }
+
+    macro_rules! walltime_to_mtime {
+        ($walltime_ms:expr, $period_ps:expr) => {
+            ms_to_ps!($walltime_ms as u64).div_ceil($period_ps as u64)
+        };
+    }
+
+    let period = match clock_period {
+        0 => 2500u64,
```
GwangbaeCHOI, line 1325: To avoid division by 0. the default value 2500 is computed from Caliptra's validated frequency, 400MHz

runtime/src/ocp_lock/mod.rs
```diff
@@ -871,7 +879,8 @@ pub struct OcpLockContext {
 
 impl OcpLockContext {
     pub fn new(soc_ifc: &SocIfc, trng: &mut Trng, hek_available: bool) -> CaliptraResult<Self> {
-        let available = cfg!(feature = "ocp-lock") && soc_ifc.ocp_lock_enabled();
+        let available =
+            cfg!(feature = "ocp-lock") && soc_ifc.ocp_lock_enabled() && soc_ifc.subsystem_mode();
```
GwangbaeCHOI, line 914: Add checking if Caliptra is running in subsystem mode. I guess it was one of OCP LOCK requirements.

api/src/mailbox.rs
```diff
@@ -4797,6 +4824,77 @@ impl Request for ZeroizeUdsFeReq {
...
+
+impl Response for OcpLockGetStatusResp {}
+
+// CLEAR_KEY_CACHE
+#[repr(C)]
+#[derive(Debug, Default, IntoBytes, FromBytes, Immutable, KnownLayout, PartialEq, Eq)]
+pub struct OcpLockClearKeyCacheReq {
+    pub hdr: MailboxReqHeader,
+    pub reserved: u32,
+    pub rdy_timeout: u32,
```
clundin25: I think we will need to add this to the spec, do you mind opening an issue to track it so we don't forget?

drivers/src/dma.rs
```diff
@@ -1110,3 +1110,103 @@ impl<'a> DmaOtpCtrl<'a> {
...
+            mek_size,
+        }
+    }
+
+    fn write_array_fifo(&self, transaction: DmaWriteTransaction, data: &[u8]) {
+        self.dma.flush();
+        self.dma.setup_dma_write(transaction);
+        let iterator = data
+            .chunks_exact(4)
+            .map(|chunk| u32::from_le_bytes(chunk.try_into().unwrap()));
```
clundin25: This _may_ panic, if so you'll need to add an error code & return it.

drivers/src/dma.rs
```diff
@@ -1110,3 +1110,103 @@ impl<'a> DmaOtpCtrl<'a> {
...
+                    write_addr,
+                    fixed_addr: false,
+                    length: size,
+                    origin: DmaWriteOrigin::AhbFifo,
+                    aes_mode: false,
+                    aes_gcm: false,
+                };
+                self.write_array_fifo(write_transaction, mek);
+            }
+            None => panic!("MEK size is \"None\""),
```
clundin25: Panics are forbidden in firmware. Does `mek_size` need to be an option, or can we require a value to always exist?

error/src/lib.rs
```diff
@@ -2590,7 +2595,17 @@ impl CaliptraError {
             UDS_FE_PROGRAMMING_ZEROIZATION_FAILED,
             0xa006_0005,
             "UDS FE Zeroization Failed"
-        )
+        ),
+        (
+            OCP_LOCK_ENGINE_TIMEOUT,
+            0x4C45_544F,
+            "OCP LOCK Encryption Engine Timeout"
```
clundin25: I have a TODO to update the spec here, but rather than use the error codes from the spec, can you coalesce them so we can efficiently use the `u32` range?

I haven't looked too closely at where these are used but my guess is you can fit them in the `DRIVERS_OCP_LOCK` range?

runtime/src/ocp_lock/clear_key_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+pub struct ClearKeyCacheCmd;
+impl ClearKeyCacheCmd {
+    #[inline(never)]
+    pub(crate) fn execute(
+        drivers: &mut Drivers,
+        cmd_bytes: &[u8],
+        resp: &mut [u8],
+    ) -> CaliptraResult<usize> {
+        if cmd_bytes.len() != size_of::<OcpLockClearKeyCacheReq>() {
+            Err(CaliptraError::RUNTIME_INVALID_REQUEST_LENGTH)?;
```
clundin25: You can use `CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS` to align with the other commands

runtime/src/ocp_lock/clear_key_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        let cmd = OcpLockClearKeyCacheReq::ref_from_bytes(cmd_bytes)
+            .map_err(|_| CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?;
+        let soc_ifc = &drivers.soc_ifc;
+
+        // Convert timeout into mtime
+        let (ready_mtimeout, cmd_mtimeout) =
+            timeout_to_mtime(cmd.rdy_timeout, cmd.cmd_timeout, soc_ifc.get_clock_period());
```
clundin25: Do we need to set some timeout ceiling in case of a malicious command?

runtime/src/ocp_lock/clear_key_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+        let start_mtime = soc_ifc.get_timestamp();
+
+        // Prepare an encryption engine DMA
+        let addr = soc_ifc.ocp_lock_get_key_release_addr();
+        let dma_encryption_engine = DmaEncryptionEngine::new(addr.into(), &drivers.dma, None);
+
+        // Clear pending done bit
+        let mut clear_ctrl = EncryptionEngineCtrl(0);
+        clear_ctrl.set_done_bit(true);
```
clundin25: nit: Maybe this could be a helper function for `EncryptionEngineCtrl` so this could look like:


```suggestion
        let clear_ctrl = EncryptionEngineCtrl:::clear();
```

runtime/src/ocp_lock/clear_key_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+            &dma_encryption_engine,
+            soc_ifc,
+            start_mtime,
+            target_mtimeout,
+        )?;
+
+        // Write Zeroization command
+        let mut execution_ctrl = EncryptionEngineCtrl(0);
+        execution_ctrl.set_command_code(EncryptionEngineCommandCode::Zeroize.into());
+        execution_ctrl.set_execute_bit(true);
```
clundin25: Similar nit, a helper function could help readability:
```suggestion
        let zeroize_ctrl = EncryptionEngineCtrl::zeroize_command(); // `zeroize` is an overloaded function, so not sure on a great name.
```

runtime/src/ocp_lock/clear_key_cache.rs
clundin25: Thinking on it some more, maybe these could even be helper functions for the `DmaEncryptionEngine`?

runtime/src/ocp_lock/mod.rs
clundin25: Sounds good - thanks!

runtime/src/ocp_lock/mod.rs
```diff
@@ -1256,6 +1265,118 @@ impl OcpLockContext {
...
+        ($time_ms:expr) => {
+            $time_ms * 1_000_000_000u64
+        };
+    }
+
+    macro_rules! walltime_to_mtime {
+        ($walltime_ms:expr, $period_ps:expr) => {
+            ms_to_ps!($walltime_ms as u64).div_ceil($period_ps as u64)
+        };
+    }
```
clundin25: I think the could may be easier to read by either using functions or types instead of macros.

Having types for the units of time prevents bugs caused from mixing them up.

runtime/src/ocp_lock/mod.rs
clundin25: Do you mind adding this comment to the code as well?

runtime/src/ocp_lock/mod.rs
```diff
@@ -1256,6 +1265,118 @@ impl OcpLockContext {
...
+}
+
+fn create_error_code_from_ctrl(value: u32) -> CaliptraError {
+    let ctrl = EncryptionEngineCtrl(value);
+    let mut error_byte = EncryptionEngineError(0);
+
+    error_byte.set_ready_bit(ctrl.ready_bit());
+    error_byte.set_error_code(ctrl.error_code());
+
+    CaliptraError(CaliptraError::OCP_LOCK_ENGINE_ERR.0 | (error_byte.0 as u32))
```
clundin25: Ah we may want to include a comment in `error_lib.rs` that the lower 8 bits are reserved for engine error codes, so we don't accidentally duplicate the error.

Probably should add "OCP_LOCK_ENGINE_RESERVED_ERROR_X" for each bit

runtime/src/ocp_lock/mod.rs
```diff
@@ -1256,6 +1265,118 @@ impl OcpLockContext {
...
+        if elapsed_mtime > timeout_mtime {
+            Err(CaliptraError::OCP_LOCK_ENGINE_TIMEOUT)?
+        }
+
+        let ctrl = EncryptionEngineCtrl(dma.read_ctrl());
+        if ctrl.done_bit() {
+            return Ok(ctrl);
+        }
+    }
+}
```
clundin25: nit: You could pass a closure for the "complete" condition to combine these functions.

Maybe something similar to the model code here: https://github.com/chipsalliance/caliptra-sw/blob/main/hw-model/src/lib.rs#L859-L863

runtime/tests/runtime_integration_tests/test_ocp_lock/mod.rs
```diff
@@ -573,3 +579,34 @@ fn get_enabled_mpk(
...
 }
+
+#[cfg(test)]
+struct TestConfig {
+    subsystem_mode: bool,
+    ocp_lock_en: bool,
+}
+
+#[cfg(test)]
+fn init_model(config: TestConfig) -> DefaultHwModel {
```
clundin25: This seems similar to `boot_ocp_lock_runtime`, is it incompatible with the tests?

clundin25: `validate_ocp_lock_response` may also be a useful helper function

clundin25:
@GwangbaeCHOI looks like the subsystem FPGAs are crashing

drivers/src/dma.rs
```diff
@@ -1110,3 +1110,103 @@ impl<'a> DmaOtpCtrl<'a> {
...
+        }
+        self.dma.wait_for_dma_complete();
+    }
+
+    pub fn read_ctrl(&self) -> u32 {
+        let read_addr = self.key_release_base + Self::OCP_LOCK_ENCRYPTION_ENGINE_CTRL_OFFSET;
+        self.dma.read_dword(read_addr)
+    }
+
+    pub fn write_ctrl(&self, ctrl: &[u8]) {
```
zhalvorsen: Lets take a u32 here so it isn't accidentally misused:

```suggestion
    pub fn write_ctrl(&self, ctrl: u32) {
```

zhalvorsen: Similarly, we should only accept the size of the register for all of the `write_*` functions. It could be really easy for a caller to accidentally do a buffer overflow because this is run by the DMA engine and won't get caught by normal Rust memory protections.

drivers/src/dma.rs
```diff
@@ -1110,3 +1110,103 @@ impl<'a> DmaOtpCtrl<'a> {
...
+            fixed_addr: false,
+            length: Self::OCP_LOCK_ENCRYPTION_ENGINE_AUX_SIZE,
+            origin: DmaWriteOrigin::AhbFifo,
+            aes_mode: false,
+            aes_gcm: false,
+        };
+        self.write_array_fifo(write_transaction, aux);
+    }
+
+    pub fn write_mek(&self, mek: &[u8]) {
```
zhalvorsen: I might be misunderstanding this use, but shouldn't this only be done by the KeyVault?

runtime/src/ocp_lock/mod.rs
zhalvorsen: Alternatively you could report the error code in `soc_ifc.set_fw_extended_error`. That is how DPE error codes work.

runtime/src/ocp_lock/unload_mek.rs
```diff
@@ -0,0 +1,97 @@
...
+        let mut execution_ctrl = EncryptionEngineCtrl(0);
+        execution_ctrl.set_command_code(EncryptionEngineCommandCode::UnloadMek.into());
+        execution_ctrl.set_execute_bit(true);
+        dma_encryption_engine.write_ctrl(&execution_ctrl.0.to_le_bytes());
+
+        // Wait the execution to be done
+        let output_ctrl = wait_encryption_engine_done(
+            &dma_encryption_engine,
+            soc_ifc,
+            start_mtime,
```
zhalvorsen: We should probably restart the `start_mtime` so it isn't cumulative with the `rdy_timeout`

runtime/src/ocp_lock/mod.rs
GwangbaeCHOI: Sure, I did it

api/src/mailbox.rs
GwangbaeCHOI: I think this is already defined in OCP LOCK spec. Do you think we also need to add this to Caliptra RT spec?

drivers/src/dma.rs
GwangbaeCHOI: I see. I updated into a different way.

drivers/src/dma.rs
GwangbaeCHOI: This was actually a test code that I've forgotten to remove. Dropped in the updated commit.

error/src/lib.rs
GwangbaeCHOI: I updated it. Then, we might need to update OCP LOCK spec to change its error code.

runtime/src/ocp_lock/clear_key_cache.rs
GwangbaeCHOI: Sure. updated.

runtime/src/ocp_lock/clear_key_cache.rs
GwangbaeCHOI: Yes, that sounds better. I moved encryption engine related helper functions to `DmaEncryptionEngine`.

runtime/src/ocp_lock/mod.rs
GwangbaeCHOI: Updated

runtime/src/ocp_lock/clear_key_cache.rs
GwangbaeCHOI: I guess it will be quite hard to select a threshold value. I think we can bring this thing to OCP LOCK meeting this month to make a decision.

runtime/src/ocp_lock/clear_key_cache.rs
GwangbaeCHOI: Updated.

runtime/src/ocp_lock/mod.rs
GwangbaeCHOI: I changed to use `soc_ifc.set_fw_extended_error` to report the encryption engine error code. I guess this approach will be more consistent. But we now need to update the OCP LOCK spec. I will write an issue along with the error code changes after merging this PR.

runtime/tests/runtime_integration_tests/test_ocp_lock/mod.rs
GwangbaeCHOI: Yes, they work. I changed to use them and dropped these ones.

drivers/src/dma.rs
GwangbaeCHOI: That's right. I forgot to remove test codes.

runtime/src/ocp_lock/unload_mek.rs
GwangbaeCHOI: Correct. I updated them.

api/src/mailbox.rs
clundin25: I see a `cmd_timeout` field but I don't see a `rdy_timeout` in the spec. 

I opened https://github.com/chipsalliance/Caliptra/issues/618 to track. I have a list of other changes that need to be upstreamed, will add them there as well.

runtime/src/ocp_lock/clear_key_cache.rs
clundin25: Maybe we could use a best guess for now while we wait on the group? 

Say 1 hour? Happy to let you decide on the value.

The command is only calleable by the PL0 user, so should be mostly safe.

clundin25 (approved):
This looks great, thanks @GwangbaeCHOI !

I think we should add an upper timeout bound & a test case for it. I'd be okay with punting it to a follow up PR.

The test case may be a bit tricky.

runtime/src/ocp_lock/clear_key_cache.rs
GwangbaeCHOI: Sure, I set the max timeout to be 1 hour. We can update the value later once it is fixed by the group.

runtime/src/ocp_lock/mod.rs
GwangbaeCHOI: Updated

drivers/src/dma.rs
GwangbaeCHOI: Yes, that will be safer. I updated the code

drivers/src/dma.rs
```diff
@@ -1110,3 +1114,232 @@ impl<'a> DmaOtpCtrl<'a> {
...
+            error_byte.set_error_code(ctrl.error_code());
+
+            Some(error_byte.0)
+        }
+    }
+
+    fn wait_until_or_timeout<T1, T2>(
+        &self,
+        soc_ifc: &SocIfc,
+        start_mtime: u64,
```
zhalvorsen: nit: I think it would be a little cleaner to remove the start parameter and keep the value internal to the function

Same would apply to wait_ready and wait_done

Feel free to do this in a follow up

zhalvorsen (dismissed):

clundin25:
@GwangbaeCHOI looks like this just needs to apply the filter to `test_clear_key_cache_ready_timeout` as well.

```
#[cfg(not(any(feature = "fpga_realtime", feature = "fpga_subsystem")))]
#[test]
fn test_clear_key_cache_ready_timeout() {
```

clundin25 (dismissed):

GwangbaeCHOI:
@clundin25, @zhalvorsen: I just realized that I was referring an old version of OCP LOCK spec, which had `rdy_timeout` as input, while the latest version dropped it. I updated the codes to reflect the latest version.

drivers/src/dma.rs
GwangbaeCHOI: Sounds good. I updated it.

zhalvorsen (approved):

PR #3225, opened 2026-01-29
tests: Add RI_DOWNLOAD_ENCRYPTED_FIRMWARE test

swenson:
This adds a test that the code in #3043 operates like a normal RI_DOWNLOAD_FIRMWARE for now.

We also update the ROM messages to reflect which kind of boot is being done.


swenson:
Thanks!

PR #3227, opened 2026-01-30
Write granularity_64 to its new register.

mtimkovich:
`uds_fuse_granularity64` lives in `cptra_hw_config` register now and this updates sw accordingly.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1840,11 +1840,16 @@ impl HwModel for ModelFpgaSubsystem {
         }));
 
         // Set generic input wires.
-        m.set_generic_input_wires(&[0, 0]);
-        m.set_mci_generic_input_wires(&[0, 0]);
+        // The FPGA has generic input wire bit 0, word 0 set to HW_CONFIG fuse granularity,
+        // where 0 = 64, 1 = 32
+        let input_wires = [(!params.uds_fuse_row_granularity_64 as u32) << 0, 0];
+        println!(
+            "Setting input wires {:x} {:x}",
+            input_wires[0], input_wires[1]
+        );
+        m.set_generic_input_wires(&input_wires);
 
-        // TODO(mtimkovich): uds_fuse_row_granularity_64 should be written to cptra_hw_config
-        // here but that register is readonly.
+        m.set_mci_generic_input_wires(&[0, 0]);
 
         println!("Set itrng divider");
         // Set divisor for ITRNG throttling


drivers/src/soc_ifc.rs
```diff
@@ -587,9 +587,8 @@ impl SocIfc {
     }
 
     pub fn uds_fuse_row_granularity_64(&self) -> bool {
-        let config_val = self.soc_ifc.regs().cptra_generic_input_wires().read()[0];
-        // Bit 31 = 0 ==> 64-bit granularity; 1 ==> 32-bit granularity
-        ((config_val >> 31) & 1) == 0
+        let config_val = self.soc_ifc.regs().cptra_hw_config().read();
```
mhatrevi, line 590: This is incorrect. Per the documentation here: [https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.CPTRA_HW_CONFIG](https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.CPTRA_HW_CONFIG)

Fuse_Granularity 
Fuse row granularity. 0: 64-bits 1: 32-bits

Also, keep the comment what 0 and 1 mean w.r.t granularity.

hw-model/src/model_emulated.rs
```diff
@@ -218,13 +218,17 @@ impl HwModel for ModelEmulated {
             if params.ocp_lock_en {
                 hw_config |= 1 << 6
             }
+            if params.uds_fuse_row_granularity_64 {
```
mhatrevi, line 222: This logic is also reversed. For a fuse granularity of 64 bits, the bit should be set to 0; for 32-bit granularity, it should be set to 1.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1640,6 +1640,17 @@ impl ModelFpgaSubsystem {
         }
     }
 
+    fn set_fuse_granularity(&mut self, granularity_64: bool) {
```
mhatrevi: This is also reversed.

hw-model/src/model_fpga_realtime.rs
```diff
@@ -552,9 +552,11 @@ impl HwModel for ModelFpgaRealtime {
         println!("AXI reset");
         m.axi_reset();
 
+        // TODO(mtimkovich): uds_fuse_row_granularity_64 should be written to cptra_hw_config
+        // here but that register is readonly.
```
mhatrevi, line 556: Remove this TODO.

swenson: Done.

swenson:
I tested this on FPGA

swenson (approved):
LGTM

PR #3234, opened 2026-02-02
Restore Verilator support for Caliptra 2.0

ArthurHeymans:
Update the Verilator integration to work with Caliptra 2.0 RTL changes:

- Replace APB interface with AXI subordinate interface
- Add Adams Bridge (ABR) memory SRAMs for ML-DSA/ML-KEM support
  using proper 1R1W dual-port RAMs (abr_1r1w_ram, abr_1r1w_be_ram)
- Add cptra_csr_hmac_key input signal support
- Update security_state from 4-bit to security_state_t struct
- Rename ready_for_fw_push to ready_for_mb_processing
- Add ECC signal connections for ICCM/DCCM memory interfaces
- Add Axi4PC.sv empty stub for AXI Protocol Checker
- Add --timing flag and warning suppressions to Makefile

Tested with ML-KEM driver tests which now pass.


swenson (commented):
This looks great. I suspect we'll need to make some changes to the hardware model to support this mode, and this will only support core tests (not subsystem). But this would be great if we can get some tests executing with this.

swenson (dismissed):
LGTM

zhalvorsen (dismissed):

zhalvorsen (dismissed):

swenson (approved):

Issue #3177, opened 2026-01-24
OCP LOCK refactor list


clundin25:
Tracker for deferred items

- [Tracking VEK state is split in two locations](https://github.com/chipsalliance/caliptra-sw/pull/3168#discussion_r2723502776)
- Erase KDF Subkey in preconditioned-AES
- Address nit feedback in https://github.com/chipsalliance/caliptra-sw/pull/3180.
- Address feedback from https://github.com/chipsalliance/caliptra-sw/pull/3175#discussion_r2733501951.

clundin25:
Closing as done, will skip the `serialize` nit

PR #3238, opened 2026-02-02
[drivers]: Zeroize Preconditioned AES Subkeys

clundin25:
This is an action item tracked in https://github.com/chipsalliance/caliptra-sw/issues/3177.


zhalvorsen (dismissed):

common/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 36 * 1024; // Must be 4k aligned
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 167 * 1024;
+pub const RUNTIME_SIZE: u32 = 180 * 1024;
```
clundin25, line 52: Big allocation because it's annoying to keep hitting & causes lots of merge conflicts when we do.

zhalvorsen (approved):

PR #3240, opened 2026-02-02
rom: Document where CSRNG window size comes from

swenson:
The 2048 number is confusing at a glance since the NIST spec requires 512 or 1024.


zhalvorsen (approved):

swenson:
Thanks!

PR #3242, opened 2026-02-03
sw-emulator: Warm reset resets entropy_src module

### LINKED ISSUE #3241: sw-emulator: Warm reset should reset entropy_src module to be module_enable false
I checked on our FPGA, and on warm reset, the entropy_src comes back as module_enable false.

swenson:
Fixes #3241

Verified this matches the FPGA behavior and the expected behavior with the HW team.


swenson:
Thanks!

PR #3246, opened 2026-02-03
[dpe] Add support for 64 DPE contexts

### LINKED ISSUE #2354: Consider increasing number of DPE contexts supported in 2.x
As discussed in Caliptra Firmware WG, we should consider the number of DPE contexts supported from 24 to a higher number, as we have more DCCM to work with.

zhalvorsen:
Fixes #2354

This increases the number of DPE contexts from 32 to 64. This has been a common request from the community.

This increases the worst case EC-P384 CSR size to `10501` bytes and ML-DSA CSR to `22050` bytes.

This increases dpe::State from `4492` bytes to `9228` bytes.


swenson (approved):
LGTM

zhalvorsen:
Thanks Chris, let's merge this after the ROM release gets tagged though in case it clashes with the memory layout change in Nicks PR.

runtime/src/certify_key_extended.rs
```diff
@@ -59,10 +60,13 @@ impl CertifyKeyExtendedCmd {
         let certify_key_cmd = CertifyKeyCmd::ref_from_bytes(&cmd.certify_key_req[..]).or(Err(
             CaliptraError::RUNTIME_DPE_COMMAND_DESERIALIZATION_FAILED,
         ))?;
-        let resp = with_dpe_env(drivers, dmtf_device_info, None, |env| {
+        let resp = &{
```
zhalvorsen, line 63: Note to reviewers: I was cleaning this up for stack usage and realized I was accidentally using the wrong error type using `with_dpe_env` all callers were expecting a `DpeErrorCode`, but got a `CaliptraError`. Somehow they look similar enough the rest of the code didn't notice

timothytrippel (approved):

PR #3247, opened 2026-02-03
sw-emulator: Halted CPUs should still process events

swenson:
Otherwise memory writes during, e.g., hitless fw updates, can be queued until after the CPU resets, which can cause the CPU to try to execute invalid instructions.


swenson:
Thanks!

Issue #2868, opened 2025-12-02
[RFC] Dynamic power handling of entropy source


GwangbaeCHOI:
# Dynamic power handling of entropy source
This RFC outlines entropy source handling for better power consumptions within Caliptra integrated devices.

## Scope
Document update: a short text about `entropy_src` handling might be added to the Caliptra RT specification.
Security posture: This change will be impacted by NIST SP 800-90A. The implementation will be done complying NIST SP 800-90A requirement regarding entropy source.
Memory consumption: This change will slightly increase the code size, but it will be negligible since they are only few lines for condition checking and changing HW configurations.
Runtime: Execution time will be slightly increase as every trng generate call will be led by reseed counter check, but it will be negligible.

## Rationale
Power consumption is one of important metrics for any type of device. This change will improve power profile without compromising anything. The OpenTitan CSRNG is designed to support such functionality to improve power profile of devices (https://opentitan.org/book/hw/ip/csrng/doc/programmers_guide.html#running-csrng-with-entropy_src-disabled).
The Caliptra core is currently configuring `RESEED_INTERVAL` to be `0xFFFFFFFF`, which means that once the CSRNG needs to be reseeded after 2<sup>32</sup> requests. 

When `entropy_src` is enabled, it keeps generating fresh entropy with the underlying noise source whether it is used or not. So, this computation is completely unnecessary, and this is the point where we can improve the power profile without losing anything. Moreover, it will result in relaxing hardware requirements on the noise source to be connected to `entropy_src` since the noise source will not be fully loaded anymore with this change.

## Implementation Tradeoffs
The OpenTitan CSRNG provides two options:
1. regular, non-deterministic mode; and
2. fully deterministic mode.

Since the Caliptra internal TRNG is instantiated in non-deterministic mode, it would be better to take the first option as we can at least avoid using all-zero seed which can be happened with fully deterministic mode in the worst case scenario.

In order to simplify implementation, we would like to touch `entropy_src` in the boot flow. 
With the first option, we are foreseeing the following changes within the Caliptra RT implementation:
1. Once the Caliptra RT is operational, the Caliptra RT disables `entropy_src` by using `MODULE_ENABLE` register of `entropy_src`.
2. On every CSRNG `Generate` command call, the Caliptra RT checks if `RESEED_COUNTER` is *close enough* to `RESEED_INTERVAL`. If so, the Caliptra RT enables `entropy_src`.
[...]

PR #3256, opened 2026-02-04
drivers: Allow configuration of entropy_src for low-entropy sources

swenson:
This addresses some entropy source threshold settings related to if a
hardware entropy source might have low entropy. Previously, some of the
thresholds and settings were hardcoded, which could be problematic if a
source turns out to have very low entropy.

We add a few additional protections:

* Check repcnts (as well as repcnt) fail count when checking entropy_src
  errors
* Store a snapshot of all entropy configuration parameters during cold
  boot, which is used during warm reset to restore these from ROM
  persistent data, rather than re-reading the values from registers (as
  some of the registers are not protected from modification at runtime, so
  could have manipulated)
* Allow setting bypass mode if a configuration bit is set. This allows
  conditioning to be bypassed, for example, if the raw entropy is being
  evaluated
* Update ROM documentation for these new late-binding strap bits used

A small unrelated documentation was also added:
* Update ROM documentation to document how we use output wires as a
  UART, as this appeared missing.

Also updates the emulator and tests for these new changes:

Add proper emulation of entropy_src continuous health testing to detect
health check failures that occur after boot passes. Previously, the
emulator only consumed entropy during seed conditioning, missing runtime
failures.

Key changes:
- Add simulate_continuous_health_testing() to consume entropy when
  main_sm_state is read, simulating background health testing
- Fix alert threshold comparison: failures >= alert_threshold (not > 0)
- Test TWO consecutive windows during boot (matching RTL behavior)
- Add boot_instantiate_complete flag to avoid consuming entropy needed
  for seed conditioning in tests with limited entropy
- Add repcnts (repetition count symbol) health check support
- Include repcnts failures in alert_fail_counts register

Add integration tests for the entropy_src health check emulation:
- test_csrng_runtime_health_failure: Verifies runtime health check detection
- test_csrng_repcnts: Tests repetition count symbol health check
- test_csrng_entropy_config_warm_reset: Tests config persistence across reset

Co-authored-by: Arthur Heymans <arthur.heymans@9elements.com>

### DIFF
--- a/drivers/src/csrng.rs
+++ b/drivers/src/csrng.rs
@@ -21,7 +21,9 @@ Abstract:
     (CSRNG) peripheral.
 
 --*/
-use crate::{wait, CaliptraError, CaliptraResult};
+use crate::persistent::EntropyConfiguration;
+use crate::soc_ifc::reset_reason;
+use crate::{wait, CaliptraError, CaliptraResult, PersistentDataAccessor, ResetReason};
 use caliptra_registers::csrng::CsrngReg;
 use caliptra_registers::entropy_src::{self, regs::AlertFailCountsReadVal, EntropySrcReg};
 use caliptra_registers::soc_ifc::{self, SocIfcReg};
@@ -54,8 +56,15 @@ impl Csrng {
         csrng: CsrngReg,
         entropy_src: EntropySrcReg,
         soc_ifc: &SocIfcReg,
+        persistent_data: PersistentDataAccessor,
     ) -> CaliptraResult<Self> {
-        Self::with_seed(csrng, entropy_src, soc_ifc, Seed::EntropySrc)
+        Self::with_seed(
+            csrng,
+            entropy_src,
+            soc_ifc,
+            Seed::EntropySrc,
+            persistent_data,
+        )
     }
 
     /// # Safety
@@ -80,6 +89,7 @@ impl Csrng {
         entropy_src: EntropySrcReg,
         soc_ifc: &SocIfcReg,
         seed: Seed,
+        persistent_data: PersistentDataAccessor,
     ) -> CaliptraResult<Self> {
         const FALSE: u32 = MultiBitBool::False as u32;
         const TRUE: u32 = MultiBitBool::True as u32;
@@ -91,7 +101,8 @@ impl Csrng {
         // If already enabled, assume it was configured correctly by a previous call.
         if e.module_enable().read().module_enable() == FALSE {
             // Configure entropy_src
-            set_health_check_thresholds(e, soc_ifc.regs());
+            let entropy_cfg = read_entropy_configuration(&soc_ifc.regs(), persistent_data);
+            set_health_check_thresholds(e, entropy_cfg);
 
             e.conf().write(|w| {
                 w.fips_enable(TRUE)
@@ -99,18 +110,23 @@ impl Csrng {
                     .threshold_scope(TRUE)
                     .rng_bit_enable(FALSE)
             });
+
+            // We allow the SoC to set bypass mode so that entropy can be
+            // characterized directly, without passing through conditioning.
+            if (soc_ifc.regs().ss_strap_generic().at(2).read() >> 31) & 1 == 1 {
+                e.entropy_control().modify(|w| w.es_type(TRUE));
+            }
             e.module_enable().write(|w| w.module_enable(TRUE));
             check_for_alert_state(result.entropy_src.regs())?;
 
             // Lock entropy_src configuration if not in debug mode.
             // Per security model: ROM programs once, then locks permanently.
             // - SW_REGUPD: When cleared, configuration registers become read-only
-            // - ME_REGWEN: When cleared, MODULE_ENABLE becomes read-only
             // In debug mode (debug_locked == false), leave unlocked for characterization.
+            // We leave the module enable able to be turned off for potential power savings in runtime.
             if soc_ifc.regs().cptra_security_state().read().debug_locked() {
                 let e = result.entropy_src.regs_mut();
-                e.sw_regupd().write(|w| w.sw_regupd(false));
-                e.me_regwen().write(|w| w.me_regwen(false));
+                e.sw_regupd().modify(|w| w.sw_regupd(false));
             }
         }
 
@@ -217,7 +233,7 @@ fn check_for_alert_state(
             ALERT_HANG => {
                 let alert_counts = entropy_src.alert_fail_counts().read();
 
-                if alert_counts.repcnt_fail_count() > 0 {
+                if alert_counts.repcnt_fail_count() > 0 || alert_counts.repcnts_fail_count() > 0 {
                     return Err(CaliptraError::DRIVER_CSRNG_REPCNT_HEALTH_CHECK_FAILED);
                 }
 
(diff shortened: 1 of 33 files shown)


drivers/src/csrng.rs
```diff
@@ -91,14 +100,21 @@ impl Csrng {
...
                     .entropy_data_reg_enable(FALSE)
-                    .threshold_scope(TRUE)
+                    .threshold_scope(FALSE) // check thresholds invdividuslly rather than summing them
                     .rng_bit_enable(FALSE)
             });
+
+            // We allow the SoC to set bypass mode so that entropy can be
+            // characterized directly, without passing through conditioning.
+            if (soc_ifc.regs().ss_strap_generic().at(2).read() >> 31) & 1 == 1 {
+                e.entropy_control().modify(|w| w.es_type(TRUE));
```
ekarabu, line 117: We want it to be False. We do not want to bypass the conditioner.
https://opentitan.org/book/hw/ip/entropy_src/doc/registers.html#entropy_control--es_type.

If you do not want a partial write on this register. You can keep their default value of 0x99. 0x99 is {FALSE, FALSE}.

SoC needs to solve this characterization problem with a different approach.

swenson:
Interestingly, this is causing FPGA errors on our subsystem, such as DRIVER_CSRNG_ADAPTP_HEALTH_CHECK_FAILED. I will have to dig into this a bit more.

drivers/src/csrng.rs
```diff
@@ -372,76 +406,143 @@ fn send_command(csrng: &mut CsrngReg, command: Command) -> CaliptraResult<()> {
     }
 }
 
+fn read_entropy_configuration(
+    soc_ifc: &soc_ifc::RegisterBlock<ureg::RealMmio>,
+    mut persistent_data: PersistentDataAccessor,
+) -> EntropyConfiguration {
+    // Some of the entropy config registers are not lockable,
+    // so we keep them in persistent storage so that they cannot
+    // be maliciously later.
```
timothytrippel: " be maliciously later." --> " be maliciously modified later."?

drivers/src/csrng.rs
```diff
@@ -372,76 +406,143 @@ fn send_command(csrng: &mut CsrngReg, command: Command) -> CaliptraResult<()> {
...
+    // Use 75% and 25% of the 2048 bit FIPS window size for the default HI and LO thresholds
+    // respectively.
+    //
+    // This window value of 2048 comes from the OpenTitan documentation, since two noise
+    // channels are used. https://opentitan.org/book/hw/ip/entropy_src/index.html#description
+    const ADAPTP_WINDOW_SIZE_BITS: u32 = 2048;
+    const ADAPTP_DEFAULT_HI: u32 = 3 * (ADAPTP_WINDOW_SIZE_BITS / 4);
+    const ADAPTP_DEFAULT_LO: u32 = ADAPTP_WINDOW_SIZE_BITS / 4;
+
+    // TODO: What to do if HI <= LO?
```
timothytrippel: In this case, I think it seems sensible to use the defaults. Wdyt?

drivers/src/csrng.rs
```diff
@@ -372,76 +406,143 @@ fn send_command(csrng: &mut CsrngReg, command: Command) -> CaliptraResult<()> {
...
+    };
+
+    // Configure Repetition Count Test threshold
+
+    // The Repetition Count test fails if:
+    //  * An RNG wire repeats the same bit THRESHOLD times in a row.
+    // See section 4.4.1 of NIST.SP.800-90B for more information of about this test.
+
+    // If the SOC doesn't specify a repcnt threshold, use this default, which assumes a min-entropy of 1.
+    const DEFAULT_REPCNT_THRESHOLD: u32 = 41;
```
timothytrippel, line 438: Where does a threshold of 41 come from?

drivers/src/csrng.rs
```diff
@@ -91,14 +100,21 @@ impl Csrng {
...
             // Configure entropy_src
-            set_health_check_thresholds(e, soc_ifc.regs());
+            let entropy_cfg = read_entropy_configuration(&soc_ifc.regs(), persistent_data);
+            set_health_check_thresholds(e, entropy_cfg);
 
             e.conf().write(|w| {
                 w.fips_enable(TRUE)
                     .entropy_data_reg_enable(FALSE)
-                    .threshold_scope(TRUE)
+                    .threshold_scope(FALSE) // check thresholds invdividuslly rather than summing them
```
timothytrippel: After some digging, I think you want to either:
1. set THRESHOLD_SCOPE to True and RNG_BIT_ENABLE to False, or
2. make THRESHOLD_SCOPE, RNG_BIT_ENABLE, and RNG_BIT_SEL configurable by the SoC (ideally by fuses after characterization, though that would be up to the integrator), to make the health checks more robust in the face of manufacturing variability

@vogelpi can perhaps confirm (or deny hehe) my thinking here.

Regardless, if you set `threshold_scope` to `FALSE`, the you need to set `RNG_BIT_ENABLE` to `TRUE` I believe.

rom/dev/README.md
```diff
@@ -71,6 +71,25 @@ Following are the main FUSE & Architectural Registers used by the Caliptra ROM f
...
+| CPTRA_I_TRNG_ENTROPY_CONFIG_0    | [15:0]  | Adaptive Proportion test high threshold (default: 1536). The test fails if any window has more than this threshold of 1's. |
+| CPTRA_I_TRNG_ENTROPY_CONFIG_0    | [31:16] | Adaptive Proportion test low threshold (default: 512). The test fails if any window has less than this threshold of 1's. |
+| CPTRA_I_TRNG_ENTROPY_CONFIG_1    | [15:0]  | Repetition Count test threshold (default: 41). The test fails if an RNG wire repeats the same bit this many times in a row. |
+| CPTRA_I_TRNG_ENTROPY_CONFIG_1    | [31:16] | Alert threshold (default: 2). Number of health check failures before an alert is triggered. |
+
+**Notes:**
+- If any threshold value is set to 0, the ROM uses the default value specified above.
+- These configuration values are stored in persistent storage after first read to prevent malicious modification.
+- In debug mode (`debug_locked == false`), entropy source configuration registers remain unlocked for characterization.
+- In production mode, ROM locks the entropy source configuration after programming to prevent modification.
```
timothytrippel, line 91: I see #3194 performs this locking. Do we think however that maybe we should NOT lock:
- ME_REGWEN
- SW_REGUPD 
in the ROM to potentially support https://github.com/chipsalliance/caliptra-sw/issues/2868 in a future FW release?

It seems like setting these locks could be a runtime mailbox command. Wdyt?

rom/dev/README.md
```diff
@@ -71,6 +71,25 @@ Following are the main FUSE & Architectural Registers used by the Caliptra ROM f
...
 | CPTRA_OWNER_PK_HASH             | 384          | Owner ECC and LMS or MLDSA Public Key Hash              |
 
+### Entropy Source Configuration Registers
+
+The ROM configures the entropy source (CSRNG) during initialization using the following registers:
+
+| Register                         | Field/Bits    | Description                                             |
+| :------------------------------- | :------------ | :------------------------------------------------------ |
+| SS_STRAP_GENERIC[2]              | [15:0]  | Health test window size for FIPS mode (default: 512). This is the window size for all health tests when entropy is tested in FIPS mode. |
+| SS_STRAP_GENERIC[2]              | [31]    | Entropy bypass mode. When set to 1, enables bypass mode (`es_type`) to allow entropy characterization directly without passing through conditioning. |
```
timothytrippel, line 81: It might be worth adding some language to the RTL integration guide that states how an integrator may choose to wire these up to fuses or SW programmable CSRs. Wdyt?

drivers/src/csrng.rs
vogelpi: I recommend to go with @timothytrippel's second proposal, i.e., make these settings configurable by the SoC.

Regarding the THRESHOLD_SCOPE, there seems to be some misunderstanding. This does not have to do to with how test failures are counted towards to the alert threshold (see commit message), but rather how the test results are computed / how the configured health test thresholds are applied. When RNG_BIT_ENABLE is set to true, THRESHOLD_SCOPE has no effect.

If RNG_BIT_ENABLE is set to false, i.e., if all four channels of the noise source are used, the THRESHOLD_SCOPE setting only affects the Adaptive proportion test and the Markov test.
- If THRESHOLD_SCOPE is false, the **maximum** result of the individual, line-based tests is taken and compared against the threshold. This allows detecting failures of individual noise source channels.
- If THRESHOLD_SCOPE is high, the individual, line-based test results are **summed up** and then compared against the threshold. This allows lowering the likelihood for coincidental test failures (higher alpha).

Please note that the value of THRESHOLD_SCOPE needs to be considered when defining the health test threshold values.

drivers/src/csrng.rs
swenson: I'm okay with that. This comment was left in from before this PR. The only reason I could see it would be useful to set HI <= LO would be testing to ensure that there was an adaptp failure.

drivers/src/csrng.rs
swenson: No clue. You'd have to ask the original authors of this driver.

rom/dev/README.md
swenson: I think only ME_REGWEN is necessary for #2868, which is why I took that out. I'll update the docs.

SW_REGUPD was locked at the request of @ekarabu.

I think specifically he wanted them locked in ROM so that runtime could not be tricked somehow into changing them or leaving them unlocked.

rom/dev/README.md
swenson: I think that's a great idea.

drivers/src/csrng.rs
timothytrippel: Thanks @vogelpi. We should probably update the docs here https://opentitan.org/book/hw/ip/entropy_src/doc/registers.html#conf--threshold_scope to include this description. I think it is clearer.

rom/dev/README.md
timothytrippel: Ah yes you are correct I believe. If your goal is to just be able to toggle the enablement of the entropy_src block on/off but leave the thresholds and configurations locked then you can leave SW_REGUPD locked I believe.

drivers/src/csrng.rs
vogelpi: That's a good idea. I'll take care of that.

swenson: I think we will add some additional configuration flexibility in a 2.1.1 / 2.0.2 ROM, rather than trying to rush it in today. I'll focus on these changes. I've filed an issue to track this: https://github.com/chipsalliance/caliptra-sw/issues/3266

rom/dev/README.md
swenson: Filed an issue https://github.com/chipsalliance/caliptra-rtl/issues/1187

timothytrippel (approved):

drivers/src/csrng.rs
```diff
@@ -217,7 +233,7 @@ fn check_for_alert_state(
             ALERT_HANG => {
                 let alert_counts = entropy_src.alert_fail_counts().read();
 
-                if alert_counts.repcnt_fail_count() > 0 {
+                if alert_counts.repcnt_fail_count() > 0 || alert_counts.repcnts_fail_count() > 0 {
```
zhalvorsen, line 236: Maybe dumb question, but should we use the alert_threshold here?

drivers/src/csrng.rs
```diff
@@ -372,76 +388,140 @@ fn send_command(csrng: &mut CsrngReg, command: Command) -> CaliptraResult<()> {
...
+    // use the cached config only if we aren't in cold reset (assuming it has been configured)
+    if !cold_reset && entropy_cfg.configured != 0 {
+        return entropy_cfg;
+    }
+
+    // Configure alert threshold from CPTRA_ITRNG_ENTROPY_CONFIG_1[31:16]
+    // Default alert threshold value
+    const DEFAULT_ALERT_THRESHOLD: u32 = 2;
+
+    let alert_threshold = soc_ifc.cptra_i_trng_entropy_config_1().read().rsvd();
```
zhalvorsen: Why isn't this documented in the register reference?

zhalvorsen, line 409: Should we update the register reference to include this field?

drivers/src/csrng.rs
swenson: I thought the alert threshold is what controls when a fail count is triggered? Or perhaps it is used for the interrupt triggering (which we don't alert on).

drivers/src/csrng.rs
swenson: Added a comment here about that https://github.com/chipsalliance/caliptra-rtl/issues/1187

drivers/src/csrng.rs
swenson: If desired, we can make that sort of change (and possibly even using the interrupt state check instead of checking these counts directly) in a follow-up for 2.1.1 / 2.0.2, but I think it is too late for this release.

swenson:
Thanks!

drivers/src/csrng.rs
zhalvorsen: Ack, I'm not familiar enough to say. We can revisit it later if we need to

drivers/src/csrng.rs
zhalvorsen: Great, thanks!

drivers/src/csrng.rs
swenson: Discussed this offline. The code is correct, and allows this to be set via strap.

PR #3257, opened 2026-02-04
sw-emulator: Halted CPUs should still process events 2.0 (#3247)

swenson:
Otherwise memory writes during, e.g., hitless fw updates, can be queued until after the CPU resets, which can cause the CPU to try to execute invalid instructions.

(cherry picked from commit 8f764b6dfc83369293cb0a3fd748842f02e751c4)


swenson:
Thanks!

PR #3259, opened 2026-02-04
sw-emulator: Don't fire saved internal timers during reset

swenson:
These internal timers might be saved up during hitless update, where the MCU processor is halted. When it comes out of reset, the timers would immediately try to fire, which could trigger the wrong code to be executed.


swenson:
Thanks!

PR #3282, opened 2026-02-06
Adding markdown for release process

nquarton:


docs/release_checklist.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Update version numbers in:
+  - builder/src/version.rs
+- Update expected values in:
+  - test/tests/fips_test_suite/common.rs
+- Update versions in the appropriate README files:
+  - rom/dev/README.md
+  - fmc/README.md
+  - runtime/README.md
+
```
swenson, line 21: Also need to regenerate the frozen sums?

docs/release_checklist.md
```diff
@@ -0,0 +1,74 @@
...
+- ROM:
+  - rom-x.y.z
+- FMC:
+  - fmc-x.y.z
+- Runtime Firmware:
+  - rt-x.y.z
+
+### Push Tags
+
+- Push the new git tag(s) to origin.
```
swenson, line 63: I am not 100% sure sure this is necessary? If we rename the release above, then we can change move the nightly tag to the release tag.

docs/release_checklist.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Update GitHub Release
+
+Once the workflow passes:
+
+- Update the existing GitHub release:
+  - Update the release title (e.g. ROM-1.2.3)
+  - Uncheck "Set as a pre-release"
+  - Check "Set as the latest release"
+- Select "Update release"
```
swenson, line 44: I think we should also try to compile release notes as part of updating the release.

swenson:
On ROM releases, we should also update the nightly job to add it to the regression test matrix.

docs/release_checklist.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+## 1. Update Versions in caliptra-sw
+
+Push version changes to the caliptra-sw repository.
+
+### Required Updates
+
+- Update version numbers in:
+  - builder/src/version.rs
```
swenson, line 14: [‎builder/test_data/default_image_options.toml‎](https://github.com/chipsalliance/caliptra-sw/pull/3287/changes#diff-abb61bf502795fe88343f151d8a3455c448816582b614b21117d586a6209a896) might also need to be bumped

docs/release_checklist.md
swenson: Turns out no, we have to push the tag manually

PR #3285, opened 2026-02-08
nightly release: Set the correct ROM env for FIPS tests

swenson:
And make the nightly tests a little more future-proof for ROM version bumps.

Tested here: https://github.com/chipsalliance/caliptra-sw/actions/runs/21797350465


swenson:
Thanks!

PR #3290, opened 2026-02-09
[HPKE]: Add a test implementation of DH(P-384,SHA-384)

clundin25:
Since there is no official test vector for
DH(P-384,SHA-384-HKDF-SHA-384-AES-256-GCM, create the test vectors by:

* Generating them with Python
* Running them against an existing Golang library


test/src/crypto/test_vectors/hpke-p384.json
zhalvorsen, line 1: Can you add a README.md to this directory that explains how you generated these test vectors?

clundin25: Will do in a follow up!

PR #3292, opened 2026-02-10
runtime: Add KATs for ML-KEM

### LINKED ISSUE #3278: runtime: ML-KEM KATs missing

swenson:
These are run when the ML-KEM driver is instantiated, similar to how the AES KATs are run.

For now, I pulled these from NIST, which adds about 13 KB to the runtime size.

If we want, we could instead store just the seeds, ciphertext, and output, and store just the hashes of the expected encapsulation and decapsulation keys, which will save most of that additional space.

Also did a slight refactoring of the `cprintln!`s in some of the KATs so it is more obvious when they are called in the runtime.

Fixes #3278


drivers/src/kats/mlkem1024_kat.rs
```diff
@@ -0,0 +1,454 @@
...
+
+pub fn execute_mlkem1024_kat(mlkem: &mut MlKem1024) -> CaliptraResult<()> {
+    cprintln!("[kat] MLKEM1024");
+    kat_keygen(mlkem)?;
+    kat_encaps_decaps(mlkem)?;
+    Ok(())
+}
+
+/// Test key pair generation against NIST test vectors.
+fn kat_keygen(mlkem: &mut MlKem1024) -> CaliptraResult<()> {
```
zhalvorsen, line 406: Should these functions also be called from `kat::execute_kat`?

swenson: They are: https://github.com/chipsalliance/caliptra-sw/pull/3292/changes#diff-78e358fae0c228f69e7d0098fbc92c1af923990114b5ce725ae050607585d34bR110

zhalvorsen: Oh sorry somehow I missed that

swenson:
(ROM hashes changed since we removed the ML-KEM driver from the ROM and change print statements. No real functional changes though.)

zhalvorsen (approved):

swenson:
Thanks! Going to go ahead and merge but happy to revisit this if there are additional things we need to do.

PR #3293, opened 2026-02-10
driver test files modified with cavp tests

ravithakurAMD:
Adding CAVP tests to driver test code


drivers/test-fw/src/bin/sha1_tests.rs
```diff
@@ -74,11 +76,58 @@ fn test_kat() {
     );
 }
 
+fn test_acvp(){
```
zhalvorsen, line 79: Are these more KAT style tests? I thought CAVP tests were dynamic. They send a random test suite and you upload the responses to their server.

ravithakurAMD: That is correct. Adding file IO to these tests was causing build errors. As a workaround, I have a python script that reads the vectors and "writes" it to the rust files before running the test. I log the terminal output and generate my response to report back to NIST.

zhalvorsen: Would it be possible to put the JSON files in the `target` directory and in the test read the vectors with the `serde_json` crate? We could check in a default set that gets copied in to the same `target` path if there isn't anything there when the `build.rs` runs.

PR #3295, opened 2026-02-10
feat(aes): Add KAT execution to AES driver initialization

ArthurHeymans:
Since fe88c48cc1db87b8b40781fd2b11d28c4a796fcb AES KATs are lazily run
to reduce rom size. This means they have yet to run in runtime so add
them in the initialization path.

Execute non-GCM KATs (ECB, CBC, CTR, CMAC) during AES driver
construction. Change Aes::new() to return CaliptraResult to propagate
KAT failures. GCM and CMAC-KDF KATs remain executed by AesGcm::new().

Signed-off-by: Arthur Heymans <arthur.heymans@9elements.com>

### DIFF
--- a/drivers/src/aes.rs
+++ b/drivers/src/aes.rs
@@ -191,11 +191,21 @@ fn wait_for_idle(aes: &caliptra_registers::aes::RegisterBlock<ureg::RealMmioMut<
 #[allow(clippy::too_many_arguments)]
 impl Aes {
     /// Create a new AES driver.
-    pub fn new(aes: AesReg, aes_clp: AesClpReg) -> Self {
+    ///
+    /// Runs the non-GCM KATs (ECB, CBC, CTR, CMAC) at construction time.
+    /// GCM and CMAC-KDF KATs are run by `AesGcm::new()` instead.
+    pub fn new(aes: AesReg, aes_clp: AesClpReg) -> CaliptraResult<Self> {
         if cfg!(feature = "rom") {
             panic!("Do not use in ROM!");
         }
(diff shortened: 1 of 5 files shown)


zhalvorsen (dismissed):

drivers/src/aes.rs
```diff
@@ -191,11 +191,21 @@ fn wait_for_idle(aes: &caliptra_registers::aes::RegisterBlock<ureg::RealMmioMut<
...
             panic!("Do not use in ROM!");
         }
-        Self { aes, aes_clp }
+        let mut aes = Self { aes, aes_clp };
+        if cfg!(feature = "runtime") {
+            crate::kats::execute_ecb_kat(&mut aes)?;
+            crate::kats::execute_cbc_kat(&mut aes)?;
+            crate::kats::execute_ctr_kat(&mut aes)?;
+            crate::kats::execute_cmac_kat(&mut aes)?;
+        }
```
zhalvorsen, line 207: It would probably be good to put this as the else of the previous feature check, in case we ever want to use it in FMC

ArthurHeymans: I used runtime because it was messing up driver unit tests as it needs CFI enabled before running KAT (iirc). WDYT of a new feature that FMC and runtime select, to have the logic of using this code in cargo rather than in code?

zhalvorsen: That would be fine with me

zhalvorsen: Feel free to do it in a follow up

PR #3298, opened 2026-02-11
dma: prevent buffer overflow in DMA read

Neal-liu:
Limit the DMA read length based on the actual destination buffer size.
Previously, the length was solely determined by image_size_bytes, which could exceed the buffer capacity and lead to a buffer overflow.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -693,10 +693,11 @@ impl<'a> DmaRecovery<'a> {
 
         #[cfg(not(any(feature = "fpga_realtime", feature = "fpga_subsystem")))]
         {
+            let len = (image_size_bytes as usize / 4).min(buffer.len());
             let read_transaction = DmaReadTransaction {
                 read_addr: addr,
                 fixed_addr: true,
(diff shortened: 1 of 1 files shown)


drivers/src/dma.rs
```diff
@@ -693,10 +693,11 @@ impl<'a> DmaRecovery<'a> {
 
         #[cfg(not(any(feature = "fpga_realtime", feature = "fpga_subsystem")))]
         {
+            let len = (image_size_bytes as usize / 4).min(buffer.len());
```
swenson, line 696: If the `buffer.len()` is too small, we would prefer to return an error instead, which will trigger a fatal error.

If the buffer is not large enough, then that is a bug in our code, as this code path is only used to load the SoC manifest.

We wouldn't want to accidentally leave data in the I3C FIFO.

swenson:
Oh, it looks like we did this fix as part of https://github.com/chipsalliance/caliptra-sw/pull/3343

swenson:
Thanks!

Neal-liu:
Thanks for the help.

PR #3302, opened 2026-02-11
api: Ensure mailbox api is panic-free

### LINKED ISSUE #2978: caliptra-api mailbox exec functions are not panic-free
These should be panic-free so that MCU ROM, etc. can use them.

swenson:
Fixes some slicing and unwrapping that resulted in panics being generated, which makes the api/ mailbox methods suitable for use in embedded systems that restrict panics.

Adds a test to keep this invariant moving forward.

Fixes #2978


swenson:
Thanks!

PR #3303, opened 2026-02-11
api: Ensure mailbox api is panic-free (1.x)

swenson:
Fixes some slicing and unwrapping that resulted in panics being generated, which makes the api/ mailbox methods suitable for use in embedded systems that restrict panics.

Adds a test to keep this invariant moving forward.


swenson:
Thanks!

PR #3305, opened 2026-02-11
[drivers]: Fix memory alignment error in `Array4xN`

clundin25:
The Rust specification states that the alignment of a slice `[T; N]` is the same as `T`.

This means that the alignment of a `&[u8; N]` is the alignment of a `u8` (1). The compiler is incorrectly optimizing this code to use a `lw` operation, which requires that the address is word aligned.

This can cause `Array4xN::from` to trigger a memory alignment fault if `&[u8; N]` was not aligned to a word.

This change prevents the optimizer from loading a word from the input slice, and instead loads one byte at a time.

Before:
```
4000af34 <_ZN16caliptra_drivers5array17u8_to_u32_be_impl17h62081b73de219b38E>:
4000af34:   04058613  addi    a2,a1,64
4000af38:   4194      lw  a3,0(a1) <===== a1 is not word aligned!
 000af3a:   6986d693  rev8    a3,a3
```

After:
```
4000b00a <_ZN16caliptra_drivers5array17u8_to_u32_be_impl17h9fdad0f2d6c8a41eE>:
4000b00a:   058d                addi    a1,a1,3
4000b00c:   03050813            addi    a6,a0,48
4000b010:   ffd5c683            lbu a3,-3(a1) <==== Now loading u32 one
byte at a time.
4000b014:   ffe5c703            lbu a4,-2(a1)
4000b018:   fff5c783            lbu a5,-1(a1)
4000b01c:   0005c603            lbu a2,0(a1)
```

### DIFF
--- a/drivers/src/array.rs
+++ b/drivers/src/array.rs
@@ -115,9 +115,15 @@ unsafe fn u8_to_u32_be_impl<const W: usize, const B: usize>(
     src: &[u8; B],
 ) {
     let dest = dest.as_mut_ptr() as *mut u32;
+    let src_ptr = src.as_ptr();
     for i in 0..W {
-        dest.add(i)
-            .write(u32::from_be_bytes(src[i * 4..][..4].try_into().unwrap()));
+        let p = src_ptr.add(i * 4);
+        // Volatile read to prevent optimization to unaligned word load
+        let b0 = core::ptr::read_volatile(p);
+        let b1 = core::ptr::read_volatile(p.add(1));
+        let b2 = core::ptr::read_volatile(p.add(2));
+        let b3 = core::ptr::read_volatile(p.add(3));
+        dest.add(i).write(u32::from_be_bytes([b0, b1, b2, b3]));
     }
 }
 


swenson (approved):
LGTM

drivers/src/array.rs
```diff
@@ -115,9 +115,15 @@ unsafe fn u8_to_u32_be_impl<const W: usize, const B: usize>(
     src: &[u8; B],
 ) {
     let dest = dest.as_mut_ptr() as *mut u32;
+    let src_ptr = src.as_ptr();
     for i in 0..W {
-        dest.add(i)
-            .write(u32::from_be_bytes(src[i * 4..][..4].try_into().unwrap()));
+        let p = src_ptr.add(i * 4);
```
swenson, line 120: Discussed on Slack, but the `lw` is correct since we have `unaligned-scalar-mem` turned on.

But the problem is that not all memory allows that. In this case, the value is coming from the mailbox SRAM, which doesn't allow unaligned access.

So, we could also fix this by cloning the underlying value first to ensure it is copied to DCCM. This should be much smaller.

Prior to https://github.com/chipsalliance/caliptra-sw/pull/2126 we would always copy requests to DCCM, which is why this was not a problem before.

PR #3309, opened 2026-02-12
Issue #6: include metadata info for FPGA image

sileno-rocco-soc:
Adding metadata info into FPGA image.
Creating a new file "/etc/caliptra-build-info" to store the metadata, still need to add some more info in the future.

Note: first diff of "if BUILD_DEV_IMAGE" is just an adjust of identation

### DIFF
--- a/ci-tools/fpga-image/build.sh
+++ b/ci-tools/fpga-image/build.sh
@@ -48,10 +48,10 @@ if [[ -z "${SKIP_DEBOOTSTRAP}" ]]; then
   chroot out/rootfs bash -c 'echo "[Time]" > /etc/systemd/timesyncd.conf'
   chroot out/rootfs bash -c 'echo "NTP=time.google.com" >> /etc/systemd/timesyncd.conf'
 
-if [[ "$BUILD_DEV_IMAGE" == "true" ]]; then
+  if [[ "$BUILD_DEV_IMAGE" == "true" ]]; then
       echo "Adding developer keys to image"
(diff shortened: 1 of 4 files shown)


clundin25:
Can you modify [fpga.yml](https://github.com/chipsalliance/caliptra-sw/blob/main/.github/workflows/fpga.yml) and [fpga-subsystem.yml](https://github.com/chipsalliance/caliptra-sw/blob/main/.github/workflows/fpga-subsystem.yml) to check for the presence of this file, and if it exists print it?

This should happen at the beginning of the "test_artifacts" job

ci-tools/fpga-image/build.sh
```diff
@@ -130,6 +130,22 @@ chroot out/rootfs systemctl enable startup-script.service
 
 cp out/io-module.ko out/rootfs/home/runner/io-module.ko
 
+
```
clundin25, line 133: Can you also include the commit this was built with? I think GitHub has some variables in it's environment:

https://docs.github.com/en/actions/reference/workflows-and-actions/variables

clundin25:
Also, can you modify this file to also trigger on "pull-request" events, with a similar file filter as the push command: https://github.com/chipsalliance/caliptra-sw/blob/main/.github/workflows/fpga-image.yml#L5-L10?

We don't want this to trigger on every pull request (uses a lot of compute) but if the FPGA image build is modified, it's useful to see if the build "works" in the scope of the pull request.

clundin25 (approved):
Let me know if you were able to verify the file. This looks great, thanks!

PR #3315, opened 2026-02-13
[dpe] Add ML-DSA DPE Crypto trait

zhalvorsen:
This will be used when invoking DPE commands with the ML-DSA profile.

This also simplifies the DPE command handling logic so `execute` is only called once instead of per command.

### DIFF
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -18,36 +18,65 @@ use caliptra_common::keyids::{
     KEY_ID_DPE_CDI, KEY_ID_DPE_PRIV_KEY, KEY_ID_EXPORTED_DPE_CDI, KEY_ID_TMP,
 };
 use caliptra_drivers::{
-    hmac_kdf,
+    hmac_kdf, okref,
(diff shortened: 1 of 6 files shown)


runtime/src/dpe_crypto.rs
```diff
@@ -332,33 +493,18 @@ impl Crypto for DpeCrypto<'_> {
...
+            priv_key,
+            pub_key,
+        )
     }
 }
+
+enum Signer<'a> {
+    Ec(&'a mut Ecc384),
+    Mldsa(&'a mut Mldsa87),
+}
```
clundin25, line 514: Do we need a check that this matches the `SignatureType`? I suppose not, our tests should catch any issue at runtime.

clundin25 (approved):

runtime/src/dpe_crypto.rs
zhalvorsen: Good idea, I'll add that check to the match statements

zhalvorsen: Done

timothytrippel (approved):

PR #3317, opened 2026-02-13
[HPKE]: Generate test vectors from JSON during build

clundin25:
This is much easier to maintain then manually converting the test vectors into Rust slices.


drivers/test-fw/build.rs
```diff
@@ -12,9 +12,171 @@ Abstract:
...
+                hex_to_byte_array(&e.aad)
+            ));
+            out.push_str(&format!("            ct: {},\n", hex_to_byte_array(&e.ct)));
+            out.push_str(&format!(
+                "            nonce: {},\n",
+                hex_to_byte_array(&e.nonce)
+            ));
+            out.push_str(&format!("            pt: {},\n", hex_to_byte_array(&e.pt)));
+            out.push_str("        },\n");
+        }
```
clundin25, line 90: Forgot to clean this up into a string template. Will do in a follow up.

PR #3321, opened 2026-02-13
[HPKE]: Refactor for additional implementations

clundin25:
* Use GAT to make KEM interface driver agnostic 
* Move registers into GAT context to avoid borrowing a driver for too long
* Make KDF interface flexible so it can also be used during KEM operations (Needed for P-384).
* Put KEM in HPKE interface to allow supplying an external KEM.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -49,7 +49,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 36 * 1024; // Must be 4k aligned
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 184 * 1024;
+pub const RUNTIME_SIZE: u32 = 190 * 1024;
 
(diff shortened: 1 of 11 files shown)


common/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub use pmp::lock_datavault_region;
 pub const FMC_ORG: u32 = 0x40000000;
 pub const FMC_SIZE: u32 = 36 * 1024; // Must be 4k aligned
 pub const RUNTIME_ORG: u32 = FMC_ORG + FMC_SIZE;
-pub const RUNTIME_SIZE: u32 = 184 * 1024;
+pub const RUNTIME_SIZE: u32 = 190 * 1024;
```
zhalvorsen, line 52: fyi: I am also bumping this, but to 200 in https://github.com/chipsalliance/caliptra-sw/pull/3326

clundin25: @zhalvorsen you'll get a merge conflict. Want to bump it with an extra 10K of space? I'll need some for the P-384 implementation.

zhalvorsen: Sure, I'll bump mine to 210

PR #3324, opened 2026-02-13
[OCP-LOCK]: Add certificate template for hybrid certificates

clundin25:
Additionally, re-implements `cert.rs` with `x509-cert` to workaround OpenSSL limitations.

### DIFF
--- a/x509/build/x509.rs
+++ b/x509/build/x509.rs
@@ -44,6 +44,18 @@ pub const TCG_DICE_KP_IDENTITY_LOC: &str = "2.23.133.5.4.100.7";
 pub const TCG_DICE_KP_ATTEST_LOC: &str = "2.23.133.5.4.100.9";
 pub const TCG_DICE_KP_ECA: &str = "2.23.133.5.4.100.12";
 
+// TODO(clundin): Double check this oid https://github.com/chipsalliance/caliptra-sw/issues/3322.
+// From https://datatracker.ietf.org/doc/draft-ietf-lamps-pq-composite-kem/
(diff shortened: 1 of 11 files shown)


x509/build/x509.rs
```diff
@@ -280,6 +306,58 @@ impl Default for MlKem1024Key {
...
+    // Not really just an EC key but we don't really care about private keys for KEM types.
+    ecc384: Ecc384AsymKey,
+    pub_key: Vec<u8>,
+}
+
+impl AsymKey for HybridP384MlKem1024Key {
+    fn priv_key(&self) -> &PKey<Private> {
+        // For template generation only the public key matters (at least for the subject key).
+        // Return the ECC key to make the compiler happy but this is effectively a NOP.
+        self.ecc384.priv_key()
```
zhalvorsen: Would it make more sense reading this later if instead we put an `unimplemented` macro here?

clundin25: You are absolutely right, that is much clearer.

x509/build/x509.rs
```diff
@@ -280,6 +306,56 @@ impl Default for MlKem1024Key {
...
 }
 
+pub struct HybridP384MlKem1024Key {
+    pub_key: Vec<u8>,
+}
+
+impl AsymKey for HybridP384MlKem1024Key {
+    fn priv_key(&self) -> &PKey<Private> {
+        // For template generation only the public key matters (at least for the subject key).
+        // Return the ECC key to make the compiler happy but this is effectively a NOP.
```
zhalvorsen: This comment doesn't make sense anymore. Lets drop the second line.

```suggestion
        // For template generation only the public key matters (at least for the subject key).
```

clundin25: Whoops, done

PR #3326, opened 2026-02-14
[dpe] Add ML-DSA DPE command

zhalvorsen:
This adds a new command for executing DPE with the ML-DSA profile. Modifies the original DPE commmand handler to take an enum of which profile to execute with. An environment will be created based on which profile was selected allowing most of the code to stay common.

There are many commands that will be transparent for which profile was used, but some are crucial. For example, calling the `Sign` command will sign the given data with the correct profile.

This currently only tests `GetProfile`. More testing and additional functionality will be added in subsequent PRs.


api/src/mailbox.rs
```diff
@@ -101,7 +101,8 @@ impl CommandId {
     pub const LMS_SIGNATURE_VERIFY: Self = Self(0x4C4D5632); // "LMV2"
     pub const MLDSA87_SIGNATURE_VERIFY: Self = Self(0x4d4c5632); // "MLV2"
     pub const STASH_MEASUREMENT: Self = Self(0x4D454153); // "MEAS"
-    pub const INVOKE_DPE: Self = Self(0x44504543); // "DPEC"
+    pub const INVOKE_DPE_ECC384: Self = Self(0x44504543); // "DPEC"
+    pub const INVOKE_DPE_MLDSA87: Self = Self(0x4450454D); // "DPEM"
```
clundin25, line 105: I think we will have to update the runtime spec. Is that being tracked anywhere?

clundin25: I spoke too soon :)

PR #3327, opened 2026-02-16
Removing manifest hash from RT Alias CDI and adding info to CCIV

nquarton:
- Removing manfiest hash from RT alias CDI
- Renaming MBVP to CCIV
- Adding info from manifest to CCIV hash
  - PL0 PAUSER
  - FW bundle flags
  - FMC load address
  - FMC entry point
  - RT FW load address
  - RT FW entry point


nquarton:
Changes are needed to update CCIV and maintain a current and journey measurement

PR #3328, opened 2026-02-17
Add runtime command for encrypted boot decryption

ArthurHeymans:
This PR adds support for encrypted firmware boot in subsystem mode, allowing MCU firmware to be loaded encrypted and decrypted in-place via DMA.

### Encrypted Firmware Boot Mode
- Added `BootMode::EncryptedFirmware` to distinguish encrypted firmware boot from normal boot
- Added `encrypted_boot` flag to `BootParams` to enable encrypted firmware mode
- MCU ROM checks GPIO bit 28 (`mci_generic_input_wires[1]`) to determine which recovery interface command to send:
  - When set: sends `RI_DOWNLOAD_ENCRYPTED_FIRMWARE` 
  - When cleared: sends `RI_DOWNLOAD_FIRMWARE` (normal flow)
- Runtime skips MCU firmware activation when in `EncryptedFirmware` mode, allowing MCU ROM to decrypt first

### CM_AES_GCM_DECRYPT_DMA Command
- New mailbox command (`0x434D4444` / "CMDD") for in-place AES-256-GCM decryption via DMA
- Performs two-pass DMA operation:
  1. First pass: SHA384 verification of encrypted data at AXI address
  2. Second pass: In-place AES-GCM decryption with tag verification
- Restricted to `EncryptedFirmware` boot mode only
- Used by MCU ROM to decrypt firmware after import via `CM_IMPORT`

### HwModel Trait Updates
- Added `read_payload_from_ss_staging_area()` for reading back decrypted firmware
- Added `decrypt_encrypted_mcu_firmware()` with default implementation that orchestrates:
  - AES key import via `CM_IMPORT`
  - Writing encrypted data to staging area
  - In-place decryption via `CM_AES_GCM_DECRYPT_DMA`
  - GCM tag verification
- Added `upload_firmware_rri_encrypted()` for encrypted firmware upload flow

### Implementation Details
- Updated DMA driver to support AES-GCM mode with single continuous transfer (required for maintaining cipher state)
- Updated MCU subsystem commit to bb7ba2a81a3cfc5cb7bf6195f910723a05f88af3
- Fixed AES-GCM DMA word ordering to use little-endian in emulator
- Added comprehensive integration tests for encrypted firmware flow
- Use the same hardcoded IV and KEY in MCU ROM & test

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -49,7 +49,7 @@ permissions:
   contents: read
 
 env:
-  CALIPTRA_MCU_COMMIT: d74a1d7cf8b0144bd5394fe8d9baf65c833c8144
+  CALIPTRA_MCU_COMMIT: 02ea798304ccccff8e6b5a065781b3d5ed38b118 
 
 jobs:
   check_cache:
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1069,6 +1069,11 @@ impl ModelFpgaSubsystem {
                 // do the indirect fifo thing
                 println!("Recovery image available; writing blocks");
 
+                // Store the original image length before padding.
+                // The firmware uses this length for DMA transfer and SHA-384 digest,
+                // so it must reflect the actual image size, not the padded size.
+                self.recovery_ctrl_len = image.len();
+
                 // Recovery images must be padded to a multiple of 256 bytes
                 // or the last chunk will not finish.
                 let image = if image.len() % 256 == 0 {
@@ -1078,8 +1083,6 @@ impl ModelFpgaSubsystem {
                     image.resize(image.len().next_multiple_of(256), 0);
                     image
                 };
-
-                self.recovery_ctrl_len = image.len();
                 self.recovery_ctrl_written = false;
 
                 self.recovery_fifo_blocks = image.chunks(256).map(|chunk| chunk.to_vec()).collect();
@@ -2007,6 +2010,14 @@ impl HwModel for ModelFpgaSubsystem {
         } else {
             val & !(1 << 29)
         };
+        // Set encrypted boot wire on mci_generic_input_wires[1] bit 28.
+        // MCU ROM (core_test) checks this bit to decide whether to send
+        // RI_DOWNLOAD_ENCRYPTED_FIRMWARE instead of RI_DOWNLOAD_FIRMWARE.
+        let val = if boot_params.encrypted_boot {
+            val | 1 << 28
+        } else {
+            val & !(1 << 28)
+        };
         gpio.set(val);
 
         const MAX_WAIT_FOR_CPTRA_BOOT_GO_CYCLES: u64 = 100_000_000;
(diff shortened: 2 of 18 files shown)


.github/workflows/fpga-subsystem.yml
```diff
@@ -49,7 +49,7 @@ permissions:
   contents: read
 
 env:
-  CALIPTRA_MCU_COMMIT: 982ca4f02b7e9fefb58850657246b0c8f28e5b53
+  CALIPTRA_MCU_COMMIT: 9d75c88ab722381341034feffb0926e3da1238ff
```
swenson: I don't see a PR for the MCU changes yet. Maybe we can merge that PR first to avoid the MCU commit in main failure?

runtime/README.md
```diff
@@ -41,6 +41,19 @@ The Runtime Firmware main function SHALL perform the following on cold boot rese
 * Initialize any SRAM structures used by Runtime Firmware
 * Upload the firwmare to the Manufacturer Control Unit (2.0, susbystem mode only)
 
+#### Encrypted Firmware Support (2.0, subsystem mode only)
```
swenson: Technically this could work in 2.0 without DMA assistance, but let's limit it to 2.1 for documentation.

```suggestion
#### Encrypted Firmware Support (2.1, subsystem mode only)
```

runtime/src/recovery_flow.rs
```diff
@@ -121,6 +122,28 @@ impl RecoveryFlow {
...
+            // Check if firmware was loaded encrypted - if so, skip MCU activation
+            // MCU ROM will decrypt the firmware and send CM_ACTIVATE_FIRMWARE command
+            let boot_mode = drivers.persistent_data.get().rom.boot_mode;
+            if boot_mode == BootMode::EncryptedFirmware {
+                cprintln!("[rt] Encrypted firmware boot mode - skipping MCU activation");
+
+                // Write MCU firmware image size to ss_generic_fw_exec_ctrl[1] so MCU ROM
+                // can read it for decryption (it needs to know ciphertext length).
+                let mut fw_ctrl = [0u32; 4];
+                drivers.soc_ifc.get_ss_generic_fw_exec_ctrl(&mut fw_ctrl);
```
swenson: I don't think we can use `ss_generic_fw_exec_ctrl` in this manner -- this is used to control reset lines of SoC components, so writing these bytes to it could have bad side effects.

We'll need to find some other way to communicate the size, like adding a mailbox command that can be used just to get this value (or can be used to get size of any SoC manifest component?).

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -2256,6 +2303,24 @@ impl HwModel for ModelFpgaSubsystem {
         Ok(mci_base_addr + 0xc00000)
     }
 
+    fn read_payload_from_ss_staging_area(&mut self, len: usize) -> Result<Vec<u8>, ModelError> {
+        let staging_offset = 0xc00000_usize / 4; // Convert to u32 offset since mci.ptr is *mut u32
+        let staging_ptr = unsafe { self.mmio.mci().unwrap().ptr.add(staging_offset) };
+
+        let mut result = Vec::with_capacity(len);
+        let num_words = (len + 3) / 4;
```
swenson: nit: Use `div_ceil`

auth-manifest/types/src/lib.rs
```diff
@@ -165,6 +165,7 @@ pub struct AuthManifestImageMetadata {
     pub component_id: u32,
     pub classification: u32,
     pub flags: u32, // ImageMetadataFlags(image_source, ignore_auth_check)
+    pub size: u32,
```
swenson: We'll need to update the documentation for this in the README.

swenson (commented):
Great work! Almost there I think. Just a few comments.

swenson (commented):
Looks good. We'll wait for the MCU commit to be updated before merging this though.

PR #3331, opened 2026-02-17
emu: Add peripheral event processing during halt

ternst-nv:
While diagnosing an issue causing update reset test to fail in subsystem, it was determined peripheral events were being halted while the MCU core was halted.  This is incorrect as the SRAM and MCI peripherals can and must still process events even while the core is in the halt state.


swenson:
This is to the wrong branch, and I think this was already done in #3247 and #3257. Closing for now.

PR #3332, opened 2026-02-17
[dpe] Update DPE submodule

### LINKED ISSUE #3061: [BUG] Inconsistent Use of flagsMask vs. integrityRegisters Between Caliptra ROM and DPE
Dice Attestation Architecture v1.2 replaces the field `flagsMask [10] IMPLICIT OperationalFlagsMask OPTIONAL` (used in v1.1)with `integrityRegisters [10] IMPLICIT IrList OPTIONAL` in the `DiceTcbInfo` structure.
Caliptra ROM currently follows the v1.1 layout and sets `flagsMask` at **index 10** for the FMC alias DiceTcbInfo. In contrast, the DPE implementation uses `integrityRegisters` in the DiceTcbInfo structure of its certificates.

The current field index mismatch will break DICE certificate parsers and verifiers. 

The errata for Dice Attestation Architecture v1.2 will likely maintain backward compatibility by relocating `integrityRegisters` to index 11, preserving the v1.1 meaning of index 10. 
A potential fix is to update the implementation (in DPE) to adopt this corrected index assignment so both components remain aligned with the spec and avoid breaking existing parsers.

zhalvorsen:
This updates the submodule to the newest version of the `runtime-1.2` branch which adds two changes to DPE certificates:

**Fix IntegrityRegisters certificate index**

Fixes #3061

**Remove criticality flags when always false**

Per https://github.com/chipsalliance/caliptra-dpe/issues/513 DPE certificates were incorrectly including criticality flags with the default value of false.

This also updates the .gitignore to a difference made in the 2.x branch to help with Cargo target directories not at the base. Mostly used when rust-analyzer has a separate directory so it doesn't clash with the main build.


nquarton (approved):
I don't know if my review means much here but thanks for fixing this!

PR #3340, opened 2026-02-19
[drivers]: Reset ecc384 internal error register after reading error

clundin25:


drivers/src/ecc384.rs
```diff
@@ -225,7 +225,7 @@ impl Ecc384 {
             // error_global_intr_r is RO
```
zhalvorsen, line 225: Can you update this comment. I'm not sure about previous hardware, but it isn't RO in 2.1 https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.ecc_reg.intr_block_rf.error_internal_intr_r

zhalvorsen (approved):

drivers/src/ecc384.rs
clundin25: I think you mixed up with https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.ecc_reg.intr_block_rf.error_global_intr_r

PR #3343, opened 2026-02-19
Ensure that SoC manifest and runtime are a multiple of 256 bytes

swenson:
This is required by the recovery interface, so we might as well enforce it when building the image.

This also adds a fix in the DMA driver to ensure that we don't overflow the buffer, as noted
https://github.com/chipsalliance/caliptra-mcu-sw/issues/723.

Fixes https://github.com/chipsalliance/caliptra-mcu-sw/issues/723.


swenson:
Thanks!

PR #3350, opened 2026-02-19
Ensure that SoC manifest and runtime are a multiple of 256 bytes (2.0) (#3343)

swenson:
This is required by the recovery interface, so we might as well enforce it when building the image.

(cherry picked from commit c1bb966841b79eabfa707e158b6256221a9b00bd)


swenson:
Thanks!

PR #3353, opened 2026-02-19
hw-model: Add flash boot support for AXI recovery bypass loading

### LINKED ISSUE #3323: hw-model: Add tests for flash (AXI recovery bypass) loading
The MCU ROM now has the ability to load from the hw-model FPGA flash into the recovery interface via AXI bypass mode in 2.1.

We should add a test for this in the caliptra-sw that

- Extends the emulator to have a minimal flash controller (maybe the same as model_fpga_subsystem.rs)
- Sets up the flash partition with the images
- Runs a test with the right wire set to tell MCU ROM to flash boot

swenson:
Add support for flash-based firmware loading in the FPGA subsystem model, where MCU ROM reads images from the flash controller and loads them into the recovery interface via AXI bypass mode.

Changes:
- Add `flash_image.rs` module to create flash image layout for MCU ROM
- Modify `model_fpga_subsystem.rs` `boot()` to skip BMC/I3C recovery when `flash_boot` is enabled, letting MCU ROM handle loading from flash
- Add `smoke_test_flash_boot` integration test

Tested as working on my local 2.1 subsystem FPGA.

We have to make a small change in the recovery flow to wait for payload_done to not be asserted to avoid a race condition in the MCU ROM flash boot flow, but this should not affect the I3C streaming boot flow.

Fixes #3323

### DIFF
--- a/hw-model/src/flash_image.rs
+++ b/hw-model/src/flash_image.rs
@@ -1,0 +1,203 @@
+// Licensed under the Apache-2.0 license
+
+//! Flash image builder for creating flash images in the MCU ROM format.
+//!
+//! The flash image format consists of:
+//! - A [`FlashHeader`] with magic number "FLSH", version, image count, and checksum
+//! - An array of [`ImageHeader`] entries describing each image
+//! - The image data itself, padded to 256-byte alignment
+//!
+//! This matches the format expected by the MCU ROM's flash boot path in caliptra-mcu-sw.
+
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+pub const CALIPTRA_FMC_RT_IDENTIFIER: u32 = 0x0000_0000;
+pub const SOC_MANIFEST_IDENTIFIER: u32 = 0x0000_0001;
+pub const MCU_RT_IDENTIFIER: u32 = 0x0000_0002;
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, FromBytes, IntoBytes, Immutable, KnownLayout)]
+pub struct FlashHeader {
+    pub magic: [u8; 4],
+    pub version: u16,
+    pub image_count: u16,
+    pub image_headers_offset: u32,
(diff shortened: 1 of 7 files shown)


swenson:
I figured out the FPGA test failure here: https://github.com/chipsalliance/caliptra-mcu-sw/pull/941

Though the MCU FW we set here seems to generate an illegal instruction exception (it shouldn't; I'll debug separately), which then causes an interrupt crash loop.

After that PR is merged, I will update the commit here and we should be able to pass all of the FPGA tests again.

swenson:
I've done some more research, and the illegal instruction exception is occurring due to the update reset overwriting the MCU SRAM as we're executing from it. I think this is probably okay for now, as MCU doesn't have a good way to respond to interrupts to know to reset in our "default" firmware.

I'm investigating the failure in the new flash test failure on the FPGA, which was not happening for me locally initially, and seems to be related to recent FW changes.

hw-model/src/flash_image.rs
```diff
@@ -0,0 +1,201 @@
...
+//!
+//! This matches the format expected by the MCU ROM's flash boot path in caliptra-mcu-sw.
+
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+pub const CALIPTRA_FMC_RT_IDENTIFIER: u32 = 0x0000_0000;
+pub const SOC_MANIFEST_IDENTIFIER: u32 = 0x0000_0001;
+pub const MCU_RT_IDENTIFIER: u32 = 0x0000_0002;
+
+const HEADER_VERSION: u16 = 0x0001;
```
clundin25: nit: Could make it an associated constant with the header type. I think the pattern could also apply to the identifier constants.

swenson:
@clundin25 I think following up and making this the default would make a lot of sense. This is quite a bit faster based on my testing.

swenson:
Thanks!

PR #3354, opened 2026-02-19
runtime: Add CM_SHAKE256_INIT/UPDATE/FINAL cryptographic mailbox comm…

### LINKED ISSUE #3184: Add cryptographic mailbox command for SHAKE256

swenson:
…ands

Implements multi-step SHAKE256 hashing via the cryptographic mailbox, following the init/update/final pattern used by the existing CM_SHA commands.

Since the SHA3 hardware cannot save/restore Keccak state, the context returned to callers is an encrypted random session token (u128). On INIT, a random token is generated from TRNG and stored in the SHA3 driver. On UPDATE/FINAL the driver verifies the token matches the active session, preventing replay attacks and detecting if another SHA3 operation disturbed the hardware state between calls. Any new SHA3 driver operation will erase the current token.

Fixes #3184.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -960,6 +960,7 @@ dependencies = [
  "platform",
  "rand",
  "sha2",
+ "sha3 0.10.8",
  "spki 0.7.3",
  "ufmt",
  "ureg",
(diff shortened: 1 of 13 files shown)


swenson:
(Converting to draft while I debug why this is failing on the core FPGA at least.)

swenson:
Thanks!

PR #3356, opened 2026-02-20
[DPE]: Reduce monomorphization in x509 code

clundin25:
Some heavy functions took `impl` as an argument leading to a lot of code bloat.


clundin25:
This is included in https://github.com/chipsalliance/caliptra-sw/pull/3357

PR #3359, opened 2026-02-20
Add taggedOID as type choice for evidence-id field of concise-ev-map

parvathib:
As per the latest [concise-evidence.cddl](https://github.com/TrustedComputingGroup/dice-coev/blob/main/concise-evidence.cddl), the `ce.evidence-id` value type has been extended to support a tagged OID. This enables the verifier to identify the attester and its attesting environment.
The update also introduces reusable TaggedOID and TaggedUUID definitions, and adds interoperable CBOR encoding tests.

### DIFF
--- a/common/eat/src/cbor.rs
+++ b/common/eat/src/cbor.rs
@@ -32,6 +32,8 @@ pub mod tag {
     pub const COSE_SIGN1: u64 = 18;
     pub const CWT: u64 = 61;
     pub const SELF_DESCRIBED_CBOR: u64 = 55799;
+    pub const OID: u64 = 111;
+    pub const UUID: u64 = 37;
 }
 
 /// Construct a CBOR initial byte from major type and additional info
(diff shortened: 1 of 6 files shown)


common/eat/src/cbor.rs
```diff
@@ -296,6 +296,139 @@ impl<'a> CborEncoder<'a> {
...
+    }
+}
+
+impl CborEncodable for TaggedOid<'_> {
+    /// Encode the tagged OID to CBOR
+    ///
+    /// Produces: tag(111) followed by byte string containing the X.690 encoded OID value
+    fn encode(&self, encoder: &mut CborEncoder) -> Result<(), EatError> {
+        // Tag 111 for OID (RFC 8949 Section 3.4.5.3)
+        encoder.encode_tag(111)?;
```
clundin25: nit: `111` should be a constant

common/eat/src/cbor.rs
```diff
@@ -296,6 +296,139 @@ impl<'a> CborEncoder<'a> {
...
+    ///
+    /// # Arguments
+    /// * `uuid` - 16-byte UUID value in binary format
+    ///
+    /// # Note
+    /// The caller is responsible for ensuring the UUID is exactly 16 bytes.
+    /// This is validated during encoding.
+    pub const fn new(uuid: &'a [u8]) -> Self {
+        Self { uuid }
+    }
```
clundin25: Alternatively you can push that invariant into the type system:

```suggestion
    /// # Note
    /// The caller is responsible for ensuring the UUID is exactly 16 bytes.
    /// This is validated during encoding.
    pub const fn new(uuid: &'a [u8; 16]) -> Self {
        Self { uuid }
    }
```

common/eat/src/cbor.rs
```diff
@@ -296,6 +296,139 @@ impl<'a> CborEncoder<'a> {
...
+    /// # Errors
+    /// Returns `EatError::EncodingError` if the UUID is not exactly 16 bytes.
+    fn encode(&self, encoder: &mut CborEncoder) -> Result<(), EatError> {
+        // Validate UUID length (must be exactly 16 bytes per RFC 9562)
+        if self.uuid.len() != 16 {
+            return Err(EatError::EncodingError);
+        }
+
+        // Tag 37 for UUID (RFC 9562 Section 4)
+        encoder.encode_tag(37)?;
```
clundin25: Same nit

common/eat/src/cbor.rs
parvathib: defined a constant

common/eat/src/cbor.rs
parvathib: defined a constant

common/eat/src/cbor.rs
```diff
@@ -296,6 +298,131 @@ impl<'a> CborEncoder<'a> {
...
+
+impl<'a> TaggedUuid<'a> {
+    /// Create a new tagged UUID
+    ///
+    /// # Arguments
+    /// * `uuid` - 16-byte UUID value in binary format
+    ///
+    /// # Note
+    /// The caller is responsible for ensuring the UUID is exactly 16 bytes.
+    /// This is validated during encoding.
```
zhalvorsen: nit: this comment is a bit stale now

zhalvorsen (approved):

clundin25 (dismissed):

common/eat/src/cbor.rs
parvathib: Right! Removed it.

PR #3365, opened 2026-02-20
runtime: Reuse some common code in AES GCM / SPDM mbox

swenson:
There was some duplicate code in the AES GCM and SPDM cryptographic mailbox commands that this consolidates a little to save space.

This also fixes a bug I noticed where the "reserved" space in the AES GCM (non-SPDM) flags were being interpreted as the SPDM flags (i.e., version), which would trigger us to try to derived a key using HMAC. I think this was left over from the initial implementation of AES SPDM commands, but was never documented as being supported, so I just removed that behavior.

### DIFF
--- a/runtime/src/cryptographic_mailbox.rs
+++ b/runtime/src/cryptographic_mailbox.rs
@@ -1069,6 +1069,78 @@ impl Commands {
         resp.partial_len()
     }
 
+    /// Common GCM init logic for encrypt and decrypt.
+    /// When `cmd_iv` is `None`, generates a random IV and increments the usage counter (encrypt).
+    /// When `cmd_iv` is `Some`, uses the provided IV (decrypt).
+    #[inline(never)]
+    fn gcm_init_common(
+        drivers: &mut Drivers,
+        cmk_bytes: &[u8],
+        flags: u32,
+        aad: &[u8],
+        cmd_iv: Option<LEArray4x3>,
+    ) -> CaliptraResult<(EncryptedAesGcmContext, [u32; 3])> {
+        let encrypted_cmk = EncryptedCmk::ref_from_bytes(cmk_bytes)
+            .map_err(|_| CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+
+        let cmk = drivers.cryptographic_mailbox.decrypt_cmk(
+            &mut drivers.aes,
+            &mut drivers.trng,
+            encrypted_cmk,
+        )?;
+
+        if cmk.length as usize > cmk.key_material.len() {
+            Err(CaliptraError::RUNTIME_CMB_INVALID_KEY_USAGE_AND_SIZE)?;
+        }
+
+        let iv_arr;
+        let key_usage: CmKeyUsage = CmKeyUsage::from(cmk.key_usage as u32);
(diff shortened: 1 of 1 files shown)


runtime/src/cryptographic_mailbox.rs
```diff
@@ -1130,15 +1131,38 @@ impl Commands {
...
+    pub(crate) fn aes_256_gcm_encrypt_init(
+        drivers: &mut Drivers,
+        cmd_bytes: &[u8],
+        resp: &mut [u8],
+    ) -> CaliptraResult<usize> {
+        if cmd_bytes.len() > core::mem::size_of::<CmAesGcmEncryptInitReq>() {
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+        let mut cmd = CmAesGcmEncryptInitReq::default();
+        cmd.as_mut_bytes()[..cmd_bytes.len()].copy_from_slice(cmd_bytes);
```
zhalvorsen: Can we use `zerocopy::try_from_prefix` here? or `zerocopy::ref_from_prefix`?

runtime/src/cryptographic_mailbox.rs
```diff
@@ -1215,6 +1232,34 @@ impl Commands {
...
+    pub(crate) fn aes_256_gcm_spdm_encrypt_init(
+        drivers: &mut Drivers,
+        cmd_bytes: &[u8],
+        resp: &mut [u8],
+    ) -> CaliptraResult<usize> {
+        if cmd_bytes.len() > core::mem::size_of::<CmAesGcmSpdmEncryptInitReq>() {
+            Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?;
+        }
+        let mut cmd = CmAesGcmSpdmEncryptInitReq::default();
+        cmd.as_mut_bytes()[..cmd_bytes.len()].copy_from_slice(cmd_bytes);
```
zhalvorsen: Same zerocopy comment here

zhalvorsen (dismissed):

runtime/src/cryptographic_mailbox.rs
swenson: Let me try.

clundin25 (dismissed):

runtime/src/cryptographic_mailbox.rs
swenson: Yep! That works, and if we use `try_ref_from_prefix` then we can save an additional 100+ bytes. Definitely going to do that. (We could probably make that change elsewhere in a follow-up PR.)

zhalvorsen (approved):

runtime/src/cryptographic_mailbox.rs
zhalvorsen: Nice! I didn't expect it to be that beneficial. I just thought it would be better from a readability/maintainability perspective.

swenson: Ah, wait, I spoke too soon. The code compiled, but I realized that the GCM requests are variable-length, so we have to use copy_from_slice :(

swenson:
Thanks!

PR #3369, opened 2026-02-22
runtime: Shrink MLKEM 1024 KAT with SHA512

swenson:
Replace ~11KB of test vectors with SHA512 digests (~352 bytes total). Follow the same pattern used by the MLDSA KAT.


drivers/src/kats/mlkem1024_kat.rs
```diff
@@ -27,426 +32,102 @@ const KEYGEN_SEED_D: LEArray4x8 = LEArray4x8::new([
...
+const KAT_SK_DIGEST: Array4x16 = Array4x16::new([
+    0x8A0C1713, 0x45B885B9, 0xE11A7E63, 0x026178E8, 0x529282FA, 0xEFFA790F, 0xA5AB2B25, 0x88399494,
+    0x51FD3214, 0xD686BB49, 0x76432EC1, 0x2B40DE7E, 0x4E68A595, 0x260B7E9E, 0x5B7A522B, 0x4D0F05AA,
 ]);
 
 pub fn execute_mlkem1024_kat(mlkem: &mut MlKem1024) -> CaliptraResult<()> {
     cprintln!("[kat] MLKEM1024");
-    kat_keygen(mlkem)?;
-    kat_encaps_decaps(mlkem)?;
+    let mut sha2 = unsafe { Sha2_512_384::new(Sha512Reg::new()) };
```
zhalvorsen, line 69: What do you think of passing the SHA-512 driver to this function as well? It could be easier to make sure the SHA kats are run first, especially if we change it to do the self-tests on driver creation.

Another benefit could be depending on how we end up doing periodic self tests in 1766 it could be easier to justify thread safety of the drivers.

swenson: This is the same pattern as we have in MLDSA -- I think it would be good to pass the sha driver to both, but maybe let's fix both if we can use that pattern.

zhalvorsen: Sounds good to me

PR #3375, opened 2026-02-24
[HPKE]: Implement MLKEM1024-P384 Hybrid KEM

clundin25:
This resolves https://github.com/chipsalliance/caliptra-sw/issues/3034.

### DIFF
--- a/drivers/test-fw/src/bin/hpke_tests.rs
+++ b/drivers/test-fw/src/bin/hpke_tests.rs
@@ -17,12 +17,13 @@ Abstract:
 use caliptra_cfi_lib::CfiCounter;
 use caliptra_drivers::{
     hpke::{
-        self,
         kem::{
-            EncapsulationKey, Kem, MlKem, MlKemContext, MlKemEncapsulatedSecret,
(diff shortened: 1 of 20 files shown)


drivers/test-fw/src/bin/hpke_tests.rs
```diff
@@ -34,11 +35,29 @@ use zerocopy::{FromBytes, IntoBytes};
...
+];
+
 test_suite! {
     test_ml_kem_1024_test_vector,
     test_ml_kem_1024_self_talk,
     test_p384_self_talk,
     test_p384_test_vector,
+    test_hybrid_test_vector,
+    test_hybrid_self_talk,
+    // Keep this test case last
```
zhalvorsen, line 60: Is this comment still true after the changes to the driver to clear the registers on an error?

clundin25: It's an issue with the emulator, FPGA clears the error. Add a note to follow up with a fix in https://github.com/chipsalliance/caliptra-sw/issues/3177

PR #3376, opened 2026-02-24
runtime: Add MLKEM1024 cryptographic mailbox commands

ArthurHeymans:
Add CM_MLKEM_KEY_PAIR, CM_MLKEM_ENCAPSULATE, and CM_MLKEM_DECAPSULATE
mailbox commands following the same patterns as the existing MLDSA and
ECDSA commands. Seeds (seed_d || seed_z, 64 bytes) are imported as a
CMK with CmKeyUsage::Mlkem. Decapsulate uses the hardware's combined
keygen_decapsulate operation to avoid materializing the 3168-byte
decapsulation key in memory.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -51,6 +51,15 @@ const _: () = assert!(CMB_SHAKE256_CONTEXT_PLAINTEXT_SIZE + 12 + 16 == CMB_SHAKE
 /// Maximum digest size for SHAKE256 (64 bytes).
 pub const SHAKE256_MAX_DIGEST_BYTE_SIZE: usize = 64;
 
+/// ML-KEM-1024 Encapsulation Key size in bytes.
+pub const MLKEM1024_ENCAPS_KEY_SIZE: usize = 1568;
+/// ML-KEM-1024 Ciphertext size in bytes.
+pub const MLKEM1024_CIPHERTEXT_SIZE: usize = 1568;
+/// ML-KEM-1024 Shared Key size in bytes.
+pub const MLKEM1024_SHARED_KEY_SIZE: usize = 32;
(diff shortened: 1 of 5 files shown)


api/src/mailbox.rs
```diff
@@ -220,6 +229,9 @@ impl CommandId {
     pub const CM_ECDSA_VERIFY: Self = Self(0x434D_4556); // "CMEV"
     pub const CM_DERIVE_STABLE_KEY: Self = Self(0x494D_4453); // "CMDS"
     pub const CM_SHA: Self = Self(0x434D_5348); // "CMSH"
+    pub const CM_MLKEM_KEY_PAIR: Self = Self(0x434D_4C4B); // "CMLK"
```
swenson: Nit: could we rename this operation to KeyGen to match the FIPS spec?

runtime/README.md
```diff
@@ -1805,6 +1805,71 @@ Command Code: `0x434D_4556` ("CMEV")
...
+| chksum      | u32      |                                          |
+| encaps_key  | u8[1568] | ML-KEM-1024 encapsulation key            |
+
+*Table: `CM_MLKEM_ENCAPSULATE` output arguments*
+| **Name**    | **Type** | **Description**                          |
+| ----------- | -------- | ---------------------------------------- |
+| chksum      | u32      |                                          |
+| fips_status | u32      | FIPS approved or an error                |
+| ciphertext  | u8[1568] | ML-KEM-1024 ciphertext                   |
+| shared_key  | u8[32]   | Shared secret key                        |
```
swenson: I think for the encaps/decaps, they should return the shared key as a CMK, just like ECDH does.

I think they will also need to add the key usage as an input parameter to both then.

clundin25 (approved):

zhalvorsen (approved):

PR #3377, opened 2026-02-24
Move x509 cert gen to a commmand instead of a build script

swenson:
Using a build script with a feature flag to generate temporary files that then have to be copied over to the main source directory is cumbersome. This replaces that procedure with a simple binary that does the same job.

We also added a CI step that ensures that the x509 cert templates don't need to be regenerated. (Previously this was in a series of unit tests, but with the refactoring it was easier to move to a CI test.)


swenson:
Thanks!

Issue #3379, opened 2026-02-24
Use `xtask` for important workspace flows


zhalvorsen:
There are a lot of makefiles, build scripts, and binaries that have accrued over the years, many of which are no longer maintained and don't work anymore.

We should move this functionality to `xtask` so it is easier to maintain and co-locates all tools so they are easier to find. This would be similar to the flows used in https://github.com/chipsalliance/caliptra-mcu-sw.

The following are tools that should be moved to xtask:

* [ ] Certificate template generator
* [ ] ImageBuilder
* [ ] Register generation
* [ ] Print binary size
* [ ] Release flow
* [ ] Fix linter bit rot
    * Some feature flags (fpga_subsystem) are not getting captured by the current lint step
* [ ] Header check
* [ ] Key management
* [ ] Bloat
* [ ] Objdump
* [ ] Size history
* [ ] Build all FW ELFs
* [ ] Build nextest archive to FPGA testing
* [ ] Add `clippy` execution for firmware with target `riscv32imc-unknown-none-elf`

PR #3382, opened 2026-02-24
Add `xtask` binary to the workspace

zhalvorsen:
Functionality will be added in a future changes. This adds a `clippy` command as the first functionality, which will run clippy on all workspace members (copied from `caliptra-mcu-sw`).

### DIFF
--- a/xtask/src/clippy.rs
+++ b/xtask/src/clippy.rs
@@ -1,0 +1,34 @@
+// Licensed under the Apache-2.0 license
+
+use anyhow::{bail, Result};
+use log::info;
+use std::process::Command;
+
+use crate::PROJECT_ROOT;
+
+pub(crate) fn clippy() -> Result<()> {
+    clippy_all()?;
+    Ok(())
+}
+
+fn clippy_all() -> Result<()> {
+    info!("Running: cargo clippy");
+    let args = vec![
+        "clippy",
+        "--locked",
+        "--all-targets",
+        "--",
+        "-D",
+        "warnings",
+    ];
+    let status = Command::new("cargo")
+        .current_dir(&*PROJECT_ROOT)
+        .args(args)
+        .env("RUSTFLAGS", "-Dwarnings")
+        .status()?;
+
+    if !status.success() {
+        bail!("cargo clippy failed");
+    }
+    Ok(())
+}
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,0 +1,56 @@
+// Licensed under the Apache-2.0 license
+
+use clap::{Parser, Subcommand};
+use log::LevelFilter;
+use simple_logger::SimpleLogger;
+use std::{
+    path::{Path, PathBuf},
+    sync::LazyLock,
+};
+
+mod clippy;
+mod precheckin;
+
+#[derive(Parser)]
+#[command(version, about, long_about = None)]
+struct Xtask {
+    #[command(subcommand)]
+    xtask: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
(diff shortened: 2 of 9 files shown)


xtask/src/clippy.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use crate::PROJECT_ROOT;
+
+pub(crate) fn clippy() -> Result<()> {
+    clippy_all()?;
+    Ok(())
+}
+
+fn clippy_all() -> Result<()> {
+    println!("Running: cargo clippy");
```
clundin25: Since this is a fresh implementation it may be nice to use a log shim so we can have more fine-grained control over the xtask logs

xtask/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+#[command(version, about, long_about = None)]
+struct Xtask {
+    #[command(subcommand)]
+    xtask: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Run clippy on all targets
+    Clippy,
```
clundin25, line 24: nit: Maybe call it "lint" or "precheck" and include the other re-req checks like:

* FMT
* License headers
* Clippy

Don't have to add all of it at once but I think it's useful to start with a generic command name.

xtask/src/clippy.rs
zhalvorsen: That would be nice. Added `log` and `simpler_logger` to the binary

xtask/src/main.rs
zhalvorsen: Added another command `Precheckin` that currently only calls clippy. More functionality will be added later

clundin25: Ah good idea, it would probably be annoying to run the full pre-check sweet if you want just the linter

xtask/src/clippy.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub(crate) fn clippy() -> Result<()> {
+    clippy_all()?;
+    Ok(())
+}
+
+fn clippy_all() -> Result<()> {
+    info!("Running: cargo clippy");
+    let mut args = vec!["clippy", "--locked", "--all-targets"];
+    args.extend(["--", "-D", "warnings"]);
```
clundin25: nit: seems like this could just be one line

xtask/src/clippy.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::PROJECT_ROOT;
+
+pub(crate) fn clippy() -> Result<()> {
+    clippy_all()?;
+    Ok(())
+}
+
+fn clippy_all() -> Result<()> {
+    info!("Running: cargo clippy");
+    let mut args = vec!["clippy", "--locked", "--all-targets"];
```
ternst-nv: Should we include a duplicate call of clippy with `--target=riscv32imc-unknown-none-elf`.  That will catch any code which is only compiled for riscv32.

zhalvorsen: I could be wrong, but shouldn't that be covered by the `--all-targets`?

ternst-nv: The documention looks to be a bit scarce, but it looks to be only all target types (binary, libary, doc, etc.).   Not all target tuples. 

https://doc.rust-lang.org/stable/clippy/usage.html

zhalvorsen: Ah yeah, you're right. I gave it a shot, but it will need to be more targeted for specific bins/libs because it doesn't have `std` and it tries to build binaries that require it. Lets do this in a follow on. I added to the tracking issue.

zhalvorsen:
This starts work for #3379

PR #3387, opened 2026-02-25
[OCP-LOCK]: Replace HPKE Endorsement API with GET HPKE API

clundin25:
The OCP LOCK WG has elected to sign the HPKE pub keys with the MCU firmware DPE Context instead of the Caliptra Runtime Alias key.

The certificates will now be created by Caliptra MCU.


runtime/tests/runtime_integration_tests/test_ocp_lock/mod.rs
```diff
@@ -335,108 +319,113 @@ fn verify_hpke_pub_key_with_algo(
...
-        }
-        _ => {
-            panic!(
-                "Unverified HPKE alg: {:?} IDs: {} {} {}",
-                expected_hpke_alg, kem_id, kdf_id, aead_id
-            );
-        }
-    }
-}
+// TODO(clundin): Re-eanable once MCU firmware supports creating an endorsement.
```
zhalvorsen, line 338: I think you can remove this  because I don't think we are doing any tests in `caliptra-sw` with legit MCU firmware so either it will need into the MCU repo or we'll also need to add support in our testing infrastructure to provide MCU firmware

clundin25: I will do the plumbing necessary so we can test it here (and maybe in MCU as well)

PR #3389, opened 2026-02-26
[EAT] Use naming consistently and add mldsa87 support for cose

parvathib:
Keep the CSR EAT (ATTESTED_CSR) name consistent with the mailbox command and add mldsa87 support for cose.


fdamato (commented):
LGTM

PR #3390, opened 2026-02-26
Invert CFI feature from "no-cfi" to "cfi"

clundin25:
Features that remove code in Rust are an anti-pattern. This leads to the "no-cfi" build being broken and clunky.

This feature gates "cfi" and makes it a default feature.

A constant check is included in the ROM, FMC, and runtime to make sure they were compiled with CFI. This check can be removed in a follow up, but is proof that CFI is still included.


swenson (commented):
This would be great if we can get it in!

swenson (approved):
LGTM

clundin25:
> I'm curious why the ROM changed hash. It looks like it should be functionally the same

Ah missed the common crate. Fixed

zhalvorsen:
I noticed the Size History job is also reporting 0 byte change in all the images so this should be identical to what we had before just a lot more readable

PR #3391, opened 2026-02-26
[dpe] Add optional ML-DSA response over DMA

zhalvorsen:
The worst case scenario DPE certificate/CSR is too large to fit in the Caliptra core mailbox in subsystem mode. This change adds an optional external ML-DSA response DMA address to the ML-DSA variant of the request. This allows executing all DPE commands without the need to split the certificate/CSR into multiple parts and reassemble it on the caller side.

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -380,6 +380,20 @@ impl Dma {
         self.wait_for_dma_complete();
     }
 
+    /// Writes an arbitrary length buffer to the given AXI address
+    ///
+    /// # Arguments
+    ///
+    /// * `write_addr` - Address to write to
+    /// * `buffer`  - Target location to write from
+    pub fn write_buffer(&self, write_addr: AxiAddr, buffer: &[u32]) {
+        // TODO(zhalvorsen): figure out why one shot is only writing ~1KB in the emulator and switch
+        // to using a single DMA transaction instead of writing a dword at a time.
+        for (i, write_val) in buffer.iter().enumerate() {
+            self.write_dword(write_addr + (i as u32 * 4), *write_val);
+        }
+    }
+
     /// Write a 32-bit word to the specified address
     ///
     /// # Arguments
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -632,7 +632,7 @@ pub enum MailboxReq {
     GetLdevMldsa87Cert(GetLdevMldsa87CertReq),
     StashMeasurement(StashMeasurementReq),
     InvokeDpeEcc384Command(InvokeDpeReq),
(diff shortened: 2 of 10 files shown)


drivers/src/dma.rs
```diff
@@ -380,6 +380,20 @@ impl Dma {
...
+    /// # Arguments
+    ///
+    /// * `write_addr` - Address to write to
+    /// * `buffer`  - Target location to write from
+    pub fn write_buffer(&self, write_addr: AxiAddr, buffer: &[u32]) {
+        // TODO(zhalvorsen): figure out why one shot is only writing ~1KB in the emulator and switch
+        // to using a single DMA transaction instead of writing a dword at a time.
+        for (i, write_val) in buffer.iter().enumerate() {
+            self.write_dword(write_addr + (i as u32 * 4), *write_val);
+        }
```
zhalvorsen, line 394: @swenson @ArthurHeymans you two are more familiar with the DMA. I originally had this as the code below, but in the emulator it was only writing ~1KB of the 10KB payload. Do you know why that would be the case?

```Rust
        let write_transaction = DmaWriteTransaction {
            write_addr,
            fixed_addr: false,
            // Length is in bytes.
            length: buffer.len() as u32 * 4,
            origin: DmaWriteOrigin::AhbFifo,
            aes_mode: false,
            aes_gcm: false,
        };
        self.flush();
        self.setup_dma_write(write_transaction);
        self.with_dma(|dma| {
            for write_data in buffer.iter() {
                let max_fifo_depth = dma.cap().read().fifo_max_depth();
                while max_fifo_depth == dma.status0().read().fifo_depth() {}

                dma.write_data().write(|_| *write_data);
            }
        });
        self.wait_for_dma_complete();
```

swenson: I am guessing it has to do with https://github.com/chipsalliance/caliptra-sw/blob/main/sw-emulator/lib/periph/src/dma.rs#L202

I think in general we didn't test the emulator AHB FIFO path in the manner in which you're using it.

If you try to push more than 1 KB before it is read, it should just ignore it and set an error: https://github.com/chipsalliance/caliptra-sw/blob/main/sw-emulator/lib/periph/src/dma.rs#L405

But I would expect the fifo depth to reflect the current status.

zhalvorsen: That looks very related! Thanks! I'll play with that in a follow on PR. Maybe there is a bug in the FIFO code

clundin25 (dismissed):

timothytrippel (dismissed):

drivers/src/dma.rs
zhalvorsen: Fixed in #3423

PR #3393, opened 2026-02-26
drivers: handle mailbox FSM error state and unexpected DataReady

### LINKED ISSUE #718: Mailbox driver needs to handle error states correctly.
While fixing some driver test cases to pass with the latest RTL, I noticed that there are cases where the SoC can force the mailbox state machine into an error state. The driver does not handle this, and the uC firmware can get stuck after SoC bad behavior (noticed during a uC->SoC transaction, but there are likely other scenarios).

Also, we need better hardware documentation in this area.

swenson:
Fix mailbox driver to detect and recover from the MBOX_ERROR FSM state during uC→SoC transactions. When the SoC misbehaves (e.g., writing to execute register when it doesn't own the lock), the FSM enters the error state. Previously, is_response_ready() would loop forever.

Also, while we're here, we handle the unexpected DataReady case, where additional data is attempted to be sent through the mailbox after the initial response. We don't support this kind of flow in the firmware today, so we have abort the transaction in this case.

Fixes #718


swenson:
Thanks!

PR #3396, opened 2026-02-26
Add release script and job

clundin25:
This automates the release process defined in
https://github.com/chipsalliance/caliptra-sw/pull/3282.

### DIFF
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -1,0 +1,44 @@
+name: Caliptra Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The release tag to deploy, formatted as component-major.minor.patch (e.g. fmc-2.0.0)'
+        required: true
+        type: string
+      commit:
+        description: 'The commit reference to checkout and tag'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+  deploy-release:
+    name: Deploy Release
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.commit }}
+          submodules: 'true'
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config --global user.name "GitHub CI"
+          git config --global user.email "username@users.noreply.github.com"
+
+          git fetch --depth 1 origin ${{ inputs.commit }}
+          git reset --hard ${{ inputs.commit }}
+
+          git remote rename origin release-repo
+
+      - name: Run xtask release deploy
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          cargo xtask release deploy ${{ inputs.tag }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,29 @@
+# rom-2.1.0
+
+## Caliptra ROM 2.1.0 Release Notes
+
+Release notes for changes introduced since ROM 2.0.0 (62c8009d) through a72a76f2 on the main branch.
+
+### Features
+- Support for smaller 16 KB mailbox size in subsystem mode (#2719)
+- OCP LOCK preparation (most of the features will be in the runtime)
+- UDS & FE zeroization (#2841)
+- CM_SHA command (#3097)
+- RI_DOWNLOAD_ENCRYPTED_FIRMWARE command to support encrypted MCU firmware booting (#3043, #3225)
+- Fixes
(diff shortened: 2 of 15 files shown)


.github/workflows/deploy-release.yml
```diff
@@ -0,0 +1,48 @@
...
+          git config --global user.email "username@users.noreply.github.com"
+
+          git fetch --depth 1 origin ${{ inputs.commit }}
+          git reset --hard ${{ inputs.commit }}
+
+          git remote rename origin release-repo
+
+      - name: Install required packages
+        run: |
+          sudo apt-get update -qy && sudo apt-get install -y libftdi1-dev libusb-1.0-0-dev golang-1.21-go
```
swenson: Why do we need libusb and libftdi?

xtask/src/update_frozen_images.rs
```diff
@@ -0,0 +1,111 @@
...
+        Ok(())
+    };
+
+    build_rom(&rom_with_log_path, "--rom-with-log")?;
+    build_rom(&rom_no_log_path, "--rom-no-log")?;
+
+    Ok(())
+}
+
+pub fn update_frozen_images() -> Result<()> {
```
swenson, line 51: It would also be nice to pull the release ROMs out and see if they match (other than the git commit). I do this manually usually.

.github/workflows/deploy-release.yml
clundin25: I don't think we do, I brought this in to match the `build-test` job but the `fpga-boss` tool is no longer in this repo (nor was it one of the Cargo workspace members) so it should almost certainly be safe to remove.

xtask/src/update_frozen_images.rs
clundin25: Do you mind expanding a bit more:

* Is this for a release flow or for the frozen check (Sounds like release)
* For ROM release or for all releases?

I guess we should include metadata on what the _right_ ROMs should be. Might be easier to solve by including more data with the FROZEN ROM then building logic to guess

nquarton, line 51: I'm not fully following why this step would be needed unless version numbers were actually modified by this process. I think they are just being verified, right?

So, this should already be forced to match in the CI when the PR was submitted, right?

xtask/src/release.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    let release_files = ReleaseRelevantFiles::new()?;
+    match tag.component.as_str() {
+        "rom" => {
+            check_rom(&tag, &release_files)?;
+        }
+        "fmc" => {
+            check_fmc(&tag, &release_files)?;
+        }
+        "fw" => {
```
nquarton, line 30: This ("fw") doesn't match the error message ("rt"). But also we did use "rt-x.y.z" for all the 1.x tags. I don't have a strong preference between "fw" or "rt" though.

xtask/src/update_frozen_images.rs
clundin25: I think you are right, this should already pass, but I included it as a check as the frozen ROM should match the released commit

xtask/src/release.rs
clundin25: Oh man, thanks for calling this out. I guess since 2.0 used "FW" we should probably use it, but this deserves a wider discussion.

xtask/src/update_frozen_images.rs
swenson: When we do a release, I usually like to do a hexdump or binary diff and check that the only differences from the current branch frozen ROM is the commit.

And yeah, might be easier to move that logic elsewhere.

xtask/src/release.rs
```diff
@@ -0,0 +1,499 @@
...
+            "Unknown component '{}'. Expected 'rom', 'fmc', or 'rt'",
+            tag.component
+        ),
+    }
+
+    check_changelog(&tag.release_name())?;
+    check_frozen_images()?;
+
+    let meta = GitHubReleaseManager::new(&tag)?;
+    check_nightly_workflow(&meta).await?;
```
zhalvorsen: nit: Should we make `check_nightly_workflow` and `create_github_release` member functions of `GitHubReleaseManager`?

CHANGELOG.md
```diff
@@ -0,0 +1,37 @@
+# fw-2.1.0
+
+## Caliptra FW 2.1.0 Release Notes
+
+### Features
+
+TODO
```
zhalvorsen: Should we remove the `fw-2.1.0` section until we have real content so we don't accidentally firmware with `TODO` as the features :)

clundin25: Fair :D removed

xtask/src/release.rs
clundin25: Done!

PR #3402, opened 2026-02-27
Modify AuthorizeAndStash command to not skip stash by default

### LINKED ISSUE #3364: [bug] Stash MCU measurement
The call to `AuthorizeAndStash` for MCU firmware accidentally contains the `SKIP_STASH` flag. This needs to be removed so we can measure the firmware into DPE.

clundin25:
* Also, modified the TCI type for the MCU FW measurement.
* Added a test that checks for the presence of the MCU FW TCB node

This resolves https://github.com/chipsalliance/caliptra-sw/issues/3364.

### DIFF
--- a/runtime/src/stash_measurement.rs
+++ b/runtime/src/stash_measurement.rs
@@ -26,31 +26,33 @@ use dpe::{
 };
 use zerocopy::{FromBytes, IntoBytes};
 
+const MCU_TCI_TYPE: u32 = u32::from_be_bytes(*b"MCFW");
+
 pub struct StashMeasurementCmd;
 impl StashMeasurementCmd {
     #[cfg_attr(feature = "cfi", cfi_impl_fn)]
     #[inline(never)]
+    /// This function MUST ONLY be called by Caliptra.
+    /// Mailbox commands MUST use the `execute` function.
     pub(crate) fn stash_measurement(
         drivers: &mut Drivers,
         metadata: &[u8; 4],
         measurement: &[u8; 48],
         svn: u32,
+        caller_privilege_level: PauserPrivileges,
+        locality: u32,
     ) -> CaliptraResult<DpeErrorCode> {
         let dpe_result = {
-            let caller_privilege_level = drivers.caller_privilege_level();
-            match caller_privilege_level {
-                // Only PL0 can call STASH_MEASUREMENT
-                PauserPrivileges::PL0 => (),
-                PauserPrivileges::PL1 => {
-                    return Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL);
-                }
-            }
+            // Check for MCU FW ID and swap it's TCI type
+            let tci_type = if metadata == &[2, 0, 0, 0] {
+                MCU_TCI_TYPE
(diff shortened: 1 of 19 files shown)


clundin25:
A separate follow up should migrate this from AuthorizeAndStash` to directly calling DPE.

zhalvorsen (dismissed):

runtime/tests/runtime_integration_tests/test_debug_unlock.rs
```diff
@@ -119,7 +119,7 @@ fn test_dbg_unlock_prod_success() {
     let crypto = Crypto::default();
     let digest = from_hw_format(&crypto.sha384_digest(&mcu_fw).unwrap());
     let metadata = vec![AuthManifestImageMetadata {
-        fw_id: 2,
+        fw_id: u32::from_le_bytes(*b"MCFW"),
```
zhalvorsen: I think we have typically been using `from_be_bytes` for human readable ascii words

clundin25: No longer relevant per discussion in https://github.com/chipsalliance/caliptra-mcu-sw/pull/1014

zhalvorsen (dismissed):

swenson (dismissed):
LGTM

runtime/src/stash_measurement.rs
parvathib, line 89: Can we split the current **SKIP_STASH** behavior into two independent flags, e.g. **SKIP_DPE** and **SKIP_PCR_EXTEND**?
MCU journey measurements don’t need PCR31 extension since they’re already tracked via the DPE context. 
Conversely, when the MCU issues **AuthorizeAndStashReq** for SoC measurements, it may want to skip DPE context creation but still extend to PCR31. 
A single SKIP_STASH flag makes this hard to express cleanly. Separating the flags would make the intent explicit and support both use cases.

clundin25: Seems reasonable to me, but deserves a wider discussion. We must also decide whether we want this for 2.1.0, or not.

cc @zhalvorsen

zhalvorsen: That sounds reasonable to me too. We can do it for 2.1.1.

It looks like https://github.com/chipsalliance/caliptra-mcu-sw/pull/1013 also proposes some other ways we should stash this measurement. It should be pretty transparent to do in 2.1.1. The only differences would be SVN (should be 0 anyway) and PCR values.

swenson: for backwards compatibility, I'd propose we keep this flag as-is with its current meaning, and add new behavior as additional flags. So, if someone wanted just to stash to PCR 31 and not to create a context, they could set SKIP_STASH = 1 (current flag) PCR31_STASH = 1 (new flag), for example.

zhalvorsen (dismissed):

zhalvorsen (approved):

runtime/src/stash_measurement.rs
clundin25: Looks like this fell through the cracks, I opened: https://github.com/chipsalliance/caliptra-sw/issues/4051

PR #3403, opened 2026-02-27
[dpe] Add HW model support for large DPE responses in subsystem mode

zhalvorsen:
This adds HW model support for large DPE responses over DMA. And duplicates the `test_invoke_dpe_sign_and_certify_key_cmds` test to run in subsystem mode, which will trigger the large DPE response code paths in the HW model.

### DIFF
--- a/hw-model/src/model_fpga_realtime.rs
+++ b/hw-model/src/model_fpga_realtime.rs
@@ -715,10 +715,6 @@ impl HwModel for ModelFpgaRealtime {
         todo!()
     }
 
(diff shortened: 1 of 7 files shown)


hw-model/src/model_fpga_realtime.rs
```diff
@@ -715,10 +715,18 @@ impl HwModel for ModelFpgaRealtime {
         todo!()
     }
 
+    fn staging_physical_address(&mut self) -> Result<u64, ModelError> {
+        Err(ModelError::SubsystemSramError)
+    }
+
```
clundin25: nit: could add this as the default trait implementation

PR #3406, opened 2026-02-27
[ATTESTED_CSR] Add Attested CSR support for RT alias key

### LINKED ISSUE #3253: Add mailbox command support for GET_ATTESTED_CSR for RT alias

parvathib:
Fixes #3253

### DIFF
--- a/runtime/src/attested_csr/rt_alias.rs
+++ b/runtime/src/attested_csr/rt_alias.rs
@@ -1,0 +1,109 @@
+// Licensed under the Apache-2.0 license
+
+use super::MAX_CSR_SIZE;
+use crate::Drivers;
+use caliptra_common::{crypto::PubKey, x509};
+use caliptra_drivers::{KeyReadArgs, Mldsa87Seed, Mldsa87SignRnd};
+use caliptra_error::{CaliptraError, CaliptraResult};
+use caliptra_x509::{
+    Ecdsa384CsrBuilder, Ecdsa384Signature, MlDsa87CsrBuilder, MlDsa87Signature,
+    RtAliasCsrTbsEcc384, RtAliasCsrTbsEcc384Params, RtAliasCsrTbsMlDsa87,
+    RtAliasCsrTbsMlDsa87Params,
+};
+use zerocopy::IntoBytes;
+
+/// Generate an RT Alias ECC384 CSR, signed with the RT Alias ECC private key.
+pub fn generate_rt_alias_ecc_csr(
+    drivers: &mut Drivers,
+    csr_buf: &mut [u8; MAX_CSR_SIZE],
+) -> CaliptraResult<usize> {
+    let pub_key = drivers.persistent_data.get().rom.fht.rt_dice_ecc_pub_key;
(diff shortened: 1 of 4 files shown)


runtime/tests/runtime_integration_tests/test_attested_csr/test_rt_alias.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// Generate an ECC384 DPE leaf cert using CertifyKey and return it as X509.
+fn get_ecc_dpe_leaf_cert(model: &mut DefaultHwModel) -> X509 {
+    let certify_key_cmd = CertifyKeyCommandNoRef::new(CreateCertifyKeyCmdArgs {
+        profile: CaliptraDpeProfile::Ecc384,
+        handle: ContextHandle::default(),
+        label: TEST_LABEL,
+        flags: CertifyKeyFlags::empty(),
+        format: CertifyKeyCommand::FORMAT_X509,
+    });
```
zhalvorsen, line 23: nit: Most of these are the default values. You can simplify this (and the ml-dsa function) with something like this instead:


```suggestion
    let certify_key_cmd = CertifyKeyCommandNoRef::new(CreateCertifyKeyCmdArgs {
        profile: CaliptraDpeProfile::Ecc384,
        ..Default::default()
    });
```

runtime/src/attested_csr/rt_alias.rs
```diff
@@ -0,0 +1,106 @@
...
+        .build(csr_buf)
+        .ok_or(CaliptraError::RUNTIME_ATTESTED_CSR_EAT_ENCODING_ERROR)
+}
+
+/// Generate an RT Alias ML-DSA-87 CSR, signed with the RT Alias ML-DSA private key.
+pub fn generate_rt_alias_mldsa_csr(
+    drivers: &mut Drivers,
+    csr_buf: &mut [u8; MAX_CSR_SIZE],
+) -> CaliptraResult<usize> {
+    let pub_key = Drivers::get_key_id_rt_mldsa_pub_key(drivers)?;
```
zhalvorsen: It would be good to use okref here because the public key is so huge:


```suggestion
    let pub_key = Drivers::get_key_id_rt_mldsa_pub_key(drivers);
    let pub_key = okref(pub_key)?;
```

runtime/src/attested_csr/rt_alias.rs
```diff
@@ -0,0 +1,106 @@
...
+    };
+
+    let tbs = RtAliasCsrTbsMlDsa87::new(&params);
+
+    // Sign the TBS with RT Alias ML-DSA private key
+    // ML-DSA handles its own internal hashing, so pass raw TBS data
+    let rt_seed = Drivers::get_key_id_rt_mldsa_keypair_seed(drivers)?;
+    let key_args = KeyReadArgs::new(rt_seed);
+
+    let signature = drivers.mldsa87.sign_var(
```
zhalvorsen: I would use okref here as well

parvathib: done

PR #3413, opened 2026-03-02
hw-model: Do a full AXI reset on cold reset for subsystem

swenson:
And clear the logs and exit status properly.
    
This forces a couple of things that were causing flaky tests:
    
* Logs are flushed. Sometimes MCU would have just written a fatal error
  to its log, including the exit code, and this would not be picked up
  until cold_reset() finished and read the logs, which would cause the
  test to unexpectedly fail.
* Subsystem sometimes just does not come out of reset. No MCU logs are
  shown and the OTP status is 0, which seems to indicate some sort of
  hardware failure.

Also updated the DEFAULT_MCU_FW to not trigger an illegal instruction
exception immediately by fixing the endianness of the infinite loop
instruction.


zhalvorsen (approved):

swenson:
Thanks!

PR #3415, opened 2026-03-02
[dpe] Fail early for response sizing

zhalvorsen:
This change will make sure the DPE commands fail during execution instead of after execution during the transport layer. When dealing with a non-default DPE context, the context handle changes after each operation. If a DPE operation succeeds but fails at the transport layer, the caller could lose access to the context.

### DIFF
--- a/runtime/src/invoke_dpe.rs
+++ b/runtime/src/invoke_dpe.rs
@@ -12,14 +12,14 @@ Abstract:
 
 --*/
 
-use crate::{ec_dpe_env, mldsa_dpe_env, mutrefbytes, AxiAddr, Drivers, PauserPrivileges};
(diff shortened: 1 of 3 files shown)


runtime/src/invoke_dpe.rs
```diff
@@ -99,6 +99,23 @@ impl InvokeDpeCmd {
             return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
         }
 
+        // Trim the response buffer to the correct size. If the response doesn't fit, it will fail
+        // during DPE execution and not at the transport layer. This is especially important for DPE
+        // handle rotation so the caller doesn't lose the handle.
+        let mbox_resp = if !drivers.soc_ifc.subsystem_mode() {
```
clundin25: nit: I think the logic may be easier to follow if you swap this conditional

PR #3417, opened 2026-03-02
[dpe] Add a test for worst case scenario certs and CSRs

zhalvorsen:
ML-DSA CSR can get quite large. This adds a test to make sure both ECC384 and ML-DSA DPE profiles can retrieve the largest possible certificates and CSRs.


swenson (dismissed):
LGTM

clundin25 (dismissed):

PR #3424, opened 2026-03-04
Warm reset debug unlocked workaround

nquarton:
Re-derive dummy FMC key pair on warm reset in debug unlocked mode. This is required because the key vault is reset to debug values on the warm reset but an FMC key pair is still needed to sign RT alias. This key will not match the FMC cert generated by ROM on the cold boot but allows the boot to proceed.


sunguk-bin-fad:
Thanks for doing this. I'll check this on our platform

PR #3426, opened 2026-03-04
[dpe] Add ML-DSA support to `CertifyKeyExtended`

zhalvorsen:
The main `InvokeDpe` interface has been updated to support ML-DSA, but this command is another way to access DPE. This change adds support.

The only other access is through `StashMeasurement` which is unaffected by which profile it is called by so support does not need to be added for that command.

This also fixes the order of the fields for the command description in the runtime README.md.


runtime/tests/runtime_integration_tests/test_certify_key_extended.rs
```diff
@@ -52,45 +58,95 @@ fn test_dmtf_other_name_validation_fail() {
...
+    let resp = model
+        .mailbox_execute(u32::from(cmd_id), cmd.as_bytes().unwrap())
         .map_err(|e| anyhow!("Failed to execute mailbox command: {:?}", e))?
         .ok_or(anyhow!("Did not receive a response from mailbox command"))?;
-    resp.resize(size_of::<CertifyKeyExtendedResp>(), 0);
+    check_header_checksum(&resp)?;
+
+    let mut certify_key_extended_resp = CertifyKeyExtendedResp::default();
+    if external_response {
+        println!("Reading staging addr");
```
clundin25: nit: Looks like a debug comment that was left in

runtime/tests/runtime_integration_tests/test_certify_key_extended.rs
```diff
@@ -52,45 +58,95 @@ fn test_dmtf_other_name_validation_fail() {
...
-#[test]
-fn test_dmtf_other_name_extension_present() {
-    let mut model = run_rt_test(RuntimeTestArgs::default());
+fn execute_certify_key_extended_cmd(
+    model: &mut impl HwModel,
+    profile: CaliptraDpeProfile,
+    flags: CertifyKeyExtendedFlags,
+) -> anyhow::Result<CertifyKeyResp> {
+    let (flags, axi_info) = if model.subsystem_mode() && profile == CaliptraDpeProfile::Mldsa87 {
+        println!("Getting staging addr");
```
clundin25: similar nit

clundin25 (dismissed):

runtime/tests/runtime_integration_tests/test_certify_key_extended.rs
zhalvorsen: Good catch, thanks!

PR #3429, opened 2026-03-04
Warm reset debug unlocked workaround

nquarton:
Re-derive dummy FMC key pairs on warm reset in debug unlocked mode. This is required because the key vault is reset to debug values on the warm reset but FMC key pairs are still needed to sign RT alias. These keys will not match the FMC cert generated by ROM on the cold boot but allows the boot to proceed.

### DIFF
--- a/fmc/src/flow/rt_alias.rs
+++ b/fmc/src/flow/rt_alias.rs
@@ -161,26 +161,47 @@ impl RtAliasLayer {
     ///
     /// * `DiceInput` - DICE Layer Input
     fn dice_input_from_hand_off(env: &mut FmcEnv) -> CaliptraResult<DiceInput> {
-        let ecc_auth_pub = HandOff::fmc_ecc_pub_key(env);
-        let ecc_auth_sn = x509::subj_sn(&mut env.sha256, &PubKey::Ecc(&ecc_auth_pub))?;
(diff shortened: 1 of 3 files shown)


fmc/src/flow/rt_alias.rs
```diff
@@ -579,4 +600,54 @@ impl RtAliasLayer {
...
+
+        let mldsa_key_pair = env
+            .abr
+            .with_mldsa87(|mut mldsa87| {
+                Crypto::mldsa87_key_gen(
+                    &mut mldsa87,
+                    &mut env.hmac,
+                    &mut env.trng,
+                    fmc_cdi,
+                    b"alias_rt_mldsa_key",
```
zhalvorsen:
```suggestion
                    b"alias_mldsa_ecc_key",
```

zhalvorsen (dismissed):

fmc/src/flow/rt_alias.rs
nquarton: Ah good catch. Copy-paste error...

I think we want "alias_fmc_mldsa_key" though right?

zhalvorsen: Ha ha ha whoops thanks!

nquarton: No worries at all. It got the point across. Thanks again!

PR #3444, opened 2026-03-05
runtime: Allow AES DMA decrypting always

swenson:
I think we want to allow decrypting firmware (or any other decrypting) even if we are not in the encrypted boot mode, for instance, if other SoC images need to be decrypted, or we simply want to leverage the AES DMA capabilities.


zhalvorsen (approved):

swenson:
Thanks!

PR #3447, opened 2026-03-06
[release]: Use OCP LOCK enabled firmware

clundin25:


nquarton:
No problem with the default being with OCP lock. But maybe there should be an option at the builder level to also build the FW without.

PR #3452, opened 2026-03-06
Bump ml-dsa version to fix build issues with external dependencies

chris-hinson:
When hw-model is sourced as an external dependency, dep resolution is re-run and the ml-dsa dep for sw-emulator/periph is resolved to a newer release candidate (0.1.0-rc7) than the version in Cargo.lock (rc.0).  This rc bump has breaking api changes, specifically the SigningKey::<MlDsa87>::decode function was removed.

Attempting to pin the dependency to rc.0 introduces further depdency conflicts, so instead I chose to bump the dep to rc.7 (newest at time of writing), and pin this dep version so there cannot be further surprises with breaking api changes in release candidates. I also replaced the removed ::decode function with the new from_seed method to fix the compilation error from the version bump.

This commit also applies the same patch as PR#3725 on the main-2.x branch, bringing the same dependency feature resolution fix when caliptra-sw is sourced as a dependency into the main branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,7 +20,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


clundin25:
@chris-hinson can you fix the CI check? Looks like you need to commit the Cargo.lock changes

chris-hinson:
@clundin25 Just updated the commit to fix an issue with the replaced function and committed the new cargo.lock.  Not thrilled with the diff size of the new Cargo.lock, but I made best efforts to ensure the only lines affected are from the ml-dsa version bump.

zhalvorsen:
Looks like you need to run `cargo fmt`

clundin25 (approved):

PR #3453, opened 2026-03-06
Add new integration test for hw-model sourced as a dependency

chris-hinson:
Adds "Rust HW_Model Build Integration Test" to the existing integration tests section.  This test creates a new blank cargo project, sources the hw-model as a dependency and attempts to build the project.

This test is intended to catch issues with dependency + feature resolution that dont present when building in isolation but will be breaking for downstream consumers of hw-model if not caught

NOTE: this test currently fails on main and will continue to fail until #3452 is merged.  It is intended as a check of the sorts of behavior 3452 is fixing.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -229,6 +229,16 @@ jobs:
         run: |
           (cd test/dpe_verification && make run)
 
+      - name: Rust HW_Model Build Integration Test
+        run: |
+          mkdir /tmp/dep-test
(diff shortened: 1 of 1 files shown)


clundin25:
LGTM

clundin25:
Thinking on this some more:

* We probably need some agreement on how to treat braking changes for the hw-model.
* This may be better as a crate so we can pull in other crates that are externally consumed.

PR #3460, opened 2026-03-09
[dpe] Add comment that initialization is profile agnostic

zhalvorsen:
It was confusing why DPE was only intitialized for the ECC P384 profile. This adds a comment to clarify it doesn't need to be done for the ML-DSA profile.

Also adds minor changes to the changelog.


swenson (approved):
LGTM

PR #3470, opened 2026-03-11
[xtask]: Release script fixes

clundin25:
Need to loop through multiple pages since caliptra-sw has so many releases.

### DIFF
--- a/xtask/src/release.rs
+++ b/xtask/src/release.rs
@@ -380,21 +380,16 @@ impl GitHubReleaseManager {
         );
         let runs: WorkflowRuns = self.crab.get(url, None::<&()>).await?;
 
-        let run = runs.workflow_runs.into_iter().next().ok_or_else(|| {
-            anyhow!(
-                "No nightly workflow run found for commit {}",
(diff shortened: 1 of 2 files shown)


xtask/src/release.rs
```diff
@@ -380,21 +380,16 @@ impl GitHubReleaseManager {
         );
         let runs: WorkflowRuns = self.crab.get(url, None::<&()>).await?;
 
-        let run = runs.workflow_runs.into_iter().next().ok_or_else(|| {
-            anyhow!(
-                "No nightly workflow run found for commit {}",
+        let passed_nightly = runs
+            .workflow_runs
+            .iter()
+            .any(|run| run.conclusion == Some("success".to_owned()));
```
zhalvorsen, line 386: Will this be true if any of the workflows were successful or _the_ workflow was successful?

clundin25: This solves the case where a retried job succeeds. Previously it was just checking if the first run passed

zhalvorsen: Oh awesome, thanks!

PR #3478, opened 2026-03-11
[EAT] Fix version-map encoding in measurement-values-map

parvathib:


common/eat/src/ocp_profile/concise_evidence.rs
```diff
@@ -78,9 +78,43 @@ impl CborEncodable for IntegrityRegisterEntry<'_> {
     }
 }
 
+/// Version map per CoRIM CDDL:
+///   version-map = {
+///     &(version: 0) => text
+///     ? &(version-scheme: 1) => $version-scheme
+///   }
+#[derive(Debug, Clone, Copy)]
+pub struct VersionMap<'a> {
```
clundin25, line 87: I think we should consider removing `Copy` from these types to avoid accidental copies, but that's out of scope for this

parvathib: I'm using static initialization patterns which need `Copy`. Since all values are references, the copy is trivial operation. But I'll surely look into this in details and see if we are not accidentally doing any copies.

clundin25: Gotcha, feel free to ignore. Thanks!

PR #3479, opened 2026-03-11
docs: Fix markdown tables, update links

swenson:
Fix tables where header/separator rows and data rows had different numbers of | delimiters, which breaks mdbook rendering. Changes are minimal — only adding missing trailing (or leading) | to make rows consistent within each table.

Also fixes some links to the old main-2.x branch, which is no longer updated.


swenson:
Thanks!

PR #3482, opened 2026-03-11
Clear OTP on new FPGA instances

amit-kumarh:
OTP persists across test runs, causing intermittent failures in MCU tests. Clear it out when creating an unbooted machine to avoid this

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1795,6 +1795,10 @@ impl HwModel for ModelFpgaSubsystem {
         println!("Putting subsystem into reset");
         m.set_subsystem_reset(true);
 
+        // Declaring this vec! gets LLVM to emit a memcpy. Otherwise, writes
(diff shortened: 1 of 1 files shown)


zhalvorsen (dismissed):

clundin25 (dismissed):

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1795,6 +1795,9 @@ impl HwModel for ModelFpgaSubsystem {
         println!("Putting subsystem into reset");
         m.set_subsystem_reset(true);
 
+        // we need LLVM to emit a memcpy here, otherwise we get a SIGBUS fault...
```
timothytrippel: nit: could you elaborate this comment with some more details here, i.e. the scenario that requires this and why this fix works? Just in case someone hits this again or stumbles on upon something similar somewhere else with FPGA testing

timothytrippel (dismissed):

PR #3487, opened 2026-03-12
rom: Randomize AES ENTROPY_IF_SEED on reset

swenson:
I confirmed that this needs to be seeded by FW on reset, so we seed it from the entropy source when the AES driver is instantiated (so, on ROM reset, and also during runtime FW startup).

This register is used to randomize the HW masking of the AES engine, which is a countermeasure against timing and power analysis.

The RDL description of these registers states:

> After reset and whenever firmware wants to reseed the Trivium stream cipher primitive, it has to write every register once. The order in which the registers are written doesn't matter. Upon writing the last register, the provided 288-bit value is loaded into the Trivium primitive.
>
> It's fine to write the registers while AES is busy and even while it's performing a a reseed operation of the internal PRNGs via the EDN interface.
>
> Note: Upon reset, the state of the Trivium primitive is initialized to a netlist constant. The primitive thus always generates the same output after reset. It is the responsibility of firmware to provide a new state seed after reset.


swenson:
Converting to a draft for now. Waiting on guidance from the HW team to know if we should make this happen in ROM-only or if it is okay to happen in RT.

swenson:
Either the second write to the ENTROPY_IF_SEED registers will be ignore, or the value will be mixed into the Trivium state. In either case, it is safe to keep the second write to them in RT. Marking this as ready for review again.

swenson:
Thanks!

PR #3489, opened 2026-03-12
Adding rt-1.2.4 release notes

nquarton:


clundin25:
@nquarton I think this should be on the caliptra-1.x branch

nquarton:
> @nquarton I think this should be on the caliptra-1.x branch

Ah right, I didn't think through your change from this morning all the way. Makes sense that it corrects that too. I'll change this over. Thanks!

clundin25 (approved):

PR #3495, opened 2026-03-16
rom: Randomize AES ENTROPY_IF_SEED on reset (#3487) (2.0)

swenson:
I confirmed that this needs to be seeded by FW on reset, so we seed it from the entropy source when the AES driver is instantiated (so, on ROM reset, and also during runtime FW startup).

This register is used to randomize the HW masking of the AES engine, which is a countermeasure against timing and power analysis.

The RDL description of these registers states:

> After reset and whenever firmware wants to reseed the Trivium stream cipher primitive, it has to write every register once. The order in which the registers are written doesn't matter. Upon writing the last register, the provided 288-bit value is loaded into the Trivium primitive.
>
> It's fine to write the registers while AES is busy and even while it's performing a a reseed operation of the internal PRNGs via the EDN interface.
>
> Note: Upon reset, the state of the Trivium primitive is initialized to a netlist constant. The primitive thus always generates the same output after reset. It is the responsibility of firmware to provide a new state seed after reset.

(cherry picked from commit ae1f7f293f5a0277af42d8cf2910b705c9313e59)


swenson:
Thanks!

PR #3497, opened 2026-03-17
hw-model: Support setting the exact lifecycle state (2.0)

swenson:
Rather than relying on the coarser device security state.

This allows fuller testing on FPGA for differnet lifecycle transitions.


timothytrippel (approved):

swenson:
Thanks!

PR #3498, opened 2026-03-17
hw-model: Support setting the exact lifecycle state

swenson:
Rather than relying on the coarser device security state.

This allows fuller testing on FPGA for differnet lifecycle transitions.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -468,6 +468,7 @@ pub struct ModelFpgaSubsystem {
     saved_num_prod_dbg_unlock_pk_hashes: u32,
     saved_ocp_lock_en: bool,
     saved_security_state: SecurityState,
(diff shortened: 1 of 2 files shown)


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -1502,6 +1504,14 @@ impl ModelFpgaSubsystem {
     }
 
     pub fn init_otp(&self, security_state: Option<&SecurityState>) -> Result<(), Box<dyn Error>> {
+        self.init_otp_with_lc_override(security_state, None)
+    }
+
+    pub fn init_otp_with_lc_override(
```
timothytrippel, line 1510: Curious: what happens if both a security state and LC state are provided but they don't match? is this possible?

swenson: I think the Lifecycle state would override in what is written to the OTP, and would be what would take effect in the FPGA.

swenson:
Thanks!

PR #3501, opened 2026-03-17
Fix SVN dropped from ROM-stashed measurements in initialize_dpe()

### LINKED ISSUE #3500: initialize_dpe() drops SVN from ROM-stashed measurements during replay
## Summary

`runtime/src/drivers.rs` → `initialize_dpe()` hardcodes `svn: 0` when replaying ROM measurement log entries into DPE, even though `MeasurementLogEntry` stores the SVN provided by the caller. This means measurements stashed during ROM phase lose their SVN in DPE — affecting CDI derivation, X.509 certificates, and anti-rollback guarantees.

## Branch

`main` branch of caliptra-sw (commit `083d7946`), which uses caliptra-dpe at commit `cfc9a71` (on caliptra-dpe `main`).

**Note:** The `main-2.x` branch pins an older DPE (`f56f66e`, `runtime-1.2` branch) that does not have SVN in `DeriveContextCmd`, so this bug only manifests on the `main` branch which uses the newer DPE with SVN support.

### The Bug — `runtime/src/drivers.rs` lines 713-726

ROM measurement log replay in `initialize_dpe()` hardcodes `svn: 0`:

\`\`\`rust
let result = DeriveContextCmd {
    // ...
    tci_type,
    target_locality: pl0_pauser_locality,
    svn: 0,   // <--- BUG: should be measurement_log_entry.svn
}
\`\`\`

### ROM correctly stores SVN

`rom/dev/src/flow/cold_reset/fw_processor.rs` saves `stash_measurement.svn` into `MeasurementLogEntry.svn`.

### RT stash correctly passes SVN

`runtime/src/stash_measurement.rs` line 66 passes `svn` from the caller to `DeriveContextCmd`.

## Impact

1. **CDI derivation is wrong** — The derived CDI does not reflect the actual SVN of ROM-phase measured components.
2. **Certificates misreport SVN** — X.509 MultiTcbInfo shows SVN=0 for ROM-phase measurements regardless of actual value.
3. **Anti-rollback is weakened** — SVN-based anti-rollback cannot detect firmware rollback for ROM-phase measured components.
4. **Inconsistency** — The same STASH_MEASUREMENT gets different DPE treatment depending on whether it was sent before or after RT firmware is running.

## Reproduction

Send `STASH_MEASUREMENT` with `svn=42` via both ROM and RT paths, then compare the DPE leaf certs:

```
ROM-stash path TcbInfo entries:
  tci_type: TEST, svn: 0      <-- BUG: should be 42

RT-stash path TcbInfo entries:
  tci_type: TEST, svn: 42     <-- correct
```

## Proposed Fix

One-line change in `initialize_dpe()`:

\`\`\`rust
svn: measurement_log_entry.svn,   // was: svn: 0,
\`\`\`

raghuncstate:
## Summary

- Fix `initialize_dpe()` to use `measurement_log_entry.svn` instead of hardcoded `0` when replaying ROM measurement log entries into DPE
- Add regression test `test_svn_preserved_in_rom_stash_measurement` that verifies ROM-stash and RT-stash paths produce identical DPE certs with correct SVN

## The Bug

`runtime/src/drivers.rs` line 726 hardcodes `svn: 0` in `DeriveContextCmd` when replaying ROM measurements, even though `MeasurementLogEntry.svn` stores the correct value from the original `STASH_MEASUREMENT` call. The RT stash path (`stash_measurement.rs:66`) correctly passes SVN.

This causes ROM-phase measurements to get wrong SVN in DPE, producing incorrect CDI derivation and X.509 certificates (MultiTcbInfo shows SVN=0 instead of actual value).

## Fix

One-line change: `svn: 0` → `svn: measurement_log_entry.svn`

## Test plan

- [x] `test_svn_preserved_in_rom_stash_measurement` — sends `STASH_MEASUREMENT(svn=42)` via ROM and RT paths, parses MultiTcbInfo from DPE leaf certs, asserts both show SVN=42 and certs match
- [ ] Existing `test_all_measurement_apis` (uses `svn: 0`, should still pass)

Fixes #3500

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_certs.rs
+++ b/runtime/tests/runtime_integration_tests/test_certs.rs
@@ -869,9 +869,8 @@ fn test_svn_preserved_in_rom_stash_measurement() {
     .unwrap();
 
     //
-    // 1. ROM STASH MEASUREMENT path (SVN gets dropped — the bug)
+    // 1. ROM STASH MEASUREMENT path
     //    Send STASH_MEASUREMENT with svn=42 before FW upload, then boot to runtime.
-    //    initialize_dpe() replays the measurement with svn: 0.
     //
     let fuses = Fuses {
         fuse_pqc_key_type: pqc_key_type as u32,
@@ -904,7 +903,7 @@ fn test_svn_preserved_in_rom_stash_measurement() {
         .unwrap()
         .unwrap();
 
-    // Boot to runtime (triggers initialize_dpe which replays measurement with svn: 0)
+    // Boot to runtime (triggers initialize_dpe which replays ROM measurement into DPE)
     crate::common::test_upload_firmware(&mut hw, &fw_image, pqc_key_type);
     hw.step_until(|m| m.soc_ifc().cptra_flow_status().read().ready_for_runtime());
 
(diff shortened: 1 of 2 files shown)


runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -745,3 +745,262 @@ pub fn test_all_measurement_apis() {
...
+    }
+    eprintln!("RT-stash path TcbInfo entries:");
+    for (tci_type_bytes, svn_val) in &rt_tcb_entries {
+        eprintln!(
+            "  tci_type: {} ({:02x?}), svn: {:?}",
+            String::from_utf8_lossy(tci_type_bytes),
+            tci_type_bytes,
+            svn_val
+        );
+    }
```
zhalvorsen: Lets remove all of these debug prints

runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -745,3 +745,262 @@ pub fn test_all_measurement_apis() {
...
+    let rom_test_entry = rom_tcb_entries
+        .iter()
+        .find(|(t, _)| *t == test_tci_be)
+        .unwrap_or_else(|| {
+            // Also try raw bytes in case encoding differs
+            rom_tcb_entries
+                .iter()
+                .find(|(t, _)| *t == tci_type)
+                .expect("ROM-stash cert should have a TEST TcbInfo entry")
+        });
```
zhalvorsen: There should only be one endianness option. We shouldn't need to search for both

runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -745,3 +745,262 @@ pub fn test_all_measurement_apis() {
...
+                .find(|(t, _)| *t == tci_type)
+                .expect("RT-stash cert should have a TEST TcbInfo entry")
+        });
+
+    let rom_svn = rom_test_entry.1.unwrap_or(0);
+    let rt_svn = rt_test_entry.1.unwrap_or(0);
+
+    eprintln!("\n=== SVN Values for 'TEST' measurement ===");
+    eprintln!("  ROM-stash SVN: {} (expected: 42)", rom_svn);
+    eprintln!("  RT-stash  SVN: {} (expected: 42)", rt_svn);
```
zhalvorsen: We can drop these prints as well. If it fails, it will be shown in the `assert_eq` calls below

runtime/tests/runtime_integration_tests/test_certs.rs
```diff
@@ -745,3 +745,262 @@ pub fn test_all_measurement_apis() {
...
+        image_options,
+    )
+    .unwrap()
+    .to_bytes()
+    .unwrap();
+
+    //
+    // 1. ROM STASH MEASUREMENT path (SVN gets dropped — the bug)
+    //    Send STASH_MEASUREMENT with svn=42 before FW upload, then boot to runtime.
+    //    initialize_dpe() replays the measurement with svn: 0.
```
zhalvorsen: Looks like this comment doesn't apply anymore now that the SVNs match

zhalvorsen (dismissed):

zhalvorsen (dismissed):

PR #3506, opened 2026-03-18
2.0 Warm reset debug unlocked workaround

nquarton:
Re-derive dummy FMC key pairs on warm reset in debug unlocked mode. This is required because the key vault is reset to debug values on the warm reset but FMC key pairs are still needed to sign RT alias. These keys will not match the FMC cert generated by ROM on the cold boot but allows the boot to proceed.


nquarton:
Consolidating cherry-picks to a single PR

PR #3507, opened 2026-03-18
rom: Seed ABR_ENTROPY on startup for SCA

swenson:
ROM needs to seed these registers with random values for SCA protection.


zhalvorsen (approved):

swenson:
Thanks!

PR #3508, opened 2026-03-18
rom: Seed MLDSA_ENTROPY on startup for SCA (2.0)

swenson:
ROM needs to seed these registers with random values for SCA protection.


swenson:
Thanks!

PR #3509, opened 2026-03-18
add dgst field to GET_IMAGE_INFO response

parvathib:


runtime/src/get_image_info.rs
```diff
@@ -47,6 +47,7 @@ impl GetImageInfoCmd {
             resp.image_load_address_low = metadata.image_load_address.lo;
             resp.image_staging_address_high = metadata.image_staging_address.hi;
             resp.image_staging_address_low = metadata.image_staging_address.lo;
+            resp.digest = transmute!(metadata.digest);
```
swenson: question: is this in "HW" format, with dwords with swapped byte order? And do we want to return it in HW format or in standard format?

mlvisaya:
can you also add a unit test in test_get_image_info

swenson:
(I think we also need update the runtime/README.md docs)

runtime/src/get_image_info.rs
swenson: After discussion, this is probably correct, as we should match the auth manifest format, as much as I don't like that we don't follow the SHA2 standard format.

Issue #3027, opened 2025-12-17
[1.x Backport] Wrong RT TCI update during update-reset


nquarton:
Duplicate of https://github.com/chipsalliance/caliptra-sw/issues/2775 for 1.x

nquarton:
Completed in #3221

PR #3510, opened 2026-03-18
[dpe] Use both RT PCRs in RT DPE context

### LINKED ISSUE #2775: Wrong RT TCI update during update-reset
* Caliptra Version : 1.x, 2.x  latest

During Caliptra update-reset, it seems to hand over wrong value to the DPE(RT TCI)
For the tci_current, **RT_FW_CURRENT_PCR** must be assigned, but **RT_FW_JOURNEY_PCR** is being assigned.
Please check if this is intended.

Corresponding codes: 
        let latest_pcr = <[u8; 48]>::from(drivers.pcr_bank.read_pcr(RT_FW_JOURNEY_PCR));
        dpe.contexts[root_idx].tci.tci_current = TciMeasurement(latest_pcr);
        dpe.contexts[root_idx].tci.tci_cumulative = TciMeasurement(latest_pcr);

1.x:
https://github.com/chipsalliance/caliptra-sw/blob/f25760f542410e83db5814d956d31d954774d735/runtime/src/drivers.rs#L282-L284

2.x:
https://github.com/chipsalliance/caliptra-sw/blob/962e12b152f3edd083a009ab6c1e304a04c6c698/runtime/src/drivers.rs#L341-L343

nquarton:
Fixes #2775

The original design for DPE only considered only having one measurement per context. Later a design change added another to accomodate current and journey measurements. This was not reflected in the RT DPE context so journey was used for both measurements. This was partially due to the fact that the RT current measurement is also already reported in the RT alias certificate.

This changes 2.x to report both current and journey measurements in its DPE context. A similar change will be added to 1.x as tracked in #3027.

### DIFF
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -1558,6 +1558,11 @@ impl CaliptraError {
             0x000E0073,
             "Runtime Error: Reallocate DPE context requested fewer PL1 contexts than are used currently"
         ),
+        (
+            RUNTIME_RT_CURRENT_PCR_VALIDATION_FAILED,
+            0x000E0074,
+            "Runtime Error: RT current PCR validation failed"
+        ),
         // FMC Errors
         (FMC_GLOBAL_NMI, 0x000F0001, "FMC Error: Global NMI"),
         (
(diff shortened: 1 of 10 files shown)


nquarton:
Consolidating cherry-picks to a single PR

PR #3521, opened 2026-03-20
Revert "rom: Seed MLDSA_ENTROPY on startup for SCA (#3508)"

swenson:
This is already done as part of the normal MLDSA87 operations.
    
We also add some tests in the emulator to ensure that entropy is seeded
before keygen or signing and removed an unnecessary byte swap
for mldsa entropy.
    
This reverts commit 3b9102c3f9e891a831b979027f366b511cbdc748.


swenson:
Thanks!

PR #3522, opened 2026-03-20
Add emulator check for ABR_ENTROPY

swenson:
This also removes an unnecessary byte swap in the mldsa87 driver and clarifies the names and comments around the entropy.

This is a cherry pick of the relevant changes from the 2.0 PR #3521


swenson:
Thanks!

PR #3524, opened 2026-03-23
ci: set FIPS_TEST_ROM_EXP_VERSION for frozen ROM 2.1 tests

swenson:
When running tests against the frozen ROM 2.1 binary (version 0x1040), the FIPS test suite defaults to ROM_EXP_CURRENT (0x1041) since the ROM version was bumped to 2.1.1. This causes services::execute_all_services_rom, services::execute_all_services_rt, and services::version_info_update to fail with version assertion mismatches in all rom-2.1 CI variants.


swenson:
Thanks!

PR #3554, opened 2026-03-27
[CI] refactor size-history to be more generic

leongross:
The idea is to refactor size-history to be less tightly coupled to caliptra-sw, so we can move it into [calitpra-infa](https://github.com/leongross/caliptra-infra/tree/leongross/size-history-refactor) and also re-use the code for [caliptra-mcu-sw](https://github.com/chipsalliance/caliptra-mcu-sw), while prioritizing code as config patterns here. Each repo can then implement the necessary traits and implement the size history tracking with less code duplication. 

Manual [CI run](https://github.com/chipsalliance/caliptra-sw/actions/runs/23896440474)


ci-tools/size-history/Cargo.toml
```diff
@@ -1,17 +1,18 @@
...
 [dependencies]
 caliptra-builder.workspace = true
-ghac = "0.2.0"
-prost = "=0.13.2"
-prost-derive = "=0.13.2"
-serde = { workspace = true, features = ["derive"] }
-serde_json.workspace = true
-tinytemplate.workspace = true
+
+size-history = { git = "https://github.com/chipsalliance/caliptra-infra.git", package = "caliptra-size-history", branch = "main" }
```
clundin25: Let's pin this to a git commit

clundin25 (dismissed):
@leongross looks great! I think we can collapse this into xtask but feel free to do in a follow up (discussed offline as well)

ci-tools/size-history/Cargo.toml
leongross: I pinned it to the last commit of infra.

clundin25 (dismissed):

clundin25 (approved):

PR #3594, opened 2026-04-07
[doc] Document byte order of cryptographic fields in mailbox commands

mhatrevi:


runtime/README.md
```diff
@@ -314,6 +314,67 @@ Relevant registers:
 Mailbox user 0xFFFF_FFFF is reserved for Caliptra internal use. All mailbox
 commands from that user will fail.
 
+### Byte order of cryptographic fields
+
+Several mailbox commands include ECC P-384 and ML-DSA-87 public keys,
+signatures, and SHA digest values stored as `[u32]` arrays. These `u32` words
+are stored on the mailbox in Caliptra's native **little-endian** byte order.
```
swenson: I would just delete this sentence, as we directly contradict it below, as the ECDSA public keys and SHA digest values are in big-endian order.

```suggestion
signatures, and SHA digest values stored as `[u32]` arrays.
```

runtime/README.md
```diff
@@ -314,6 +314,67 @@ Relevant registers:
...
+The "big endian" / "little endian" labels in individual command tables describe
+how the **standard (big-endian) byte representation** produced by cryptographic
+tools maps to those `u32` words.
```
swenson: Standard byte representation has nothing to do with endianness -- there is a standard byte order that is specified by FIPS, etc.

```suggestion
how the **standard byte representation** produced by cryptographic
tools maps to those `u32` words.
```

runtime/README.md
```diff
@@ -314,6 +314,67 @@ Relevant registers:
...
+
```
swenson, line 324: More accurately, it might be good to say
```suggestion
Generally, any fields specified as **little-endian** byte representation will match the standard byte representation of the cryptographic protocols, since the RISC-V processor is a little-endian processor.
```

swenson (commented):
Thanks for doing this. I made some suggestions to clarify.

I tried to be explicit with the cryptographic mailbox commands in each command, but I don't know if it is worth mentioning that the cryptographic mailbox commands all use the byte order as specified in cryptographic standards, even for hashes and ECDSA public keys.

runtime/README.md
mhatrevi: Updated.

runtime/README.md
mhatrevi: Added

runtime/src/pcr.rs
```diff
@@ -77,7 +77,8 @@ impl GetPcrQuoteCmd {
                 }
                 resp.reset_ctrs = drivers.persistent_data.get().fw.pcr_reset.all_counters();
 
-                // Change the word endianness for ECC verification. Take lower 48 bytes.
+                // ECC uses big-endian word order; take lower 48 bytes of the SHA2-512/384 digest.
```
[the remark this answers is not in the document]
mhatrevi: Updated.

runtime/README.md
mhatrevi: Updated.

swenson (approved):
LGTM

PR #3602, opened 2026-04-10
[update] Reject  prod debug unlock request if pk hash fuse is zeroized or uinitialized

mhatrevi:


runtime/tests/runtime_integration_tests/test_debug_unlock.rs
```diff
@@ -1042,3 +1042,131 @@ fn test_dbg_unlock_prod_unlock_levels_success() {
         assert!(soc_debug_level == unlock_level);
     }
 }
+
+/// Test that production debug unlock fails when the PK hash fuse is all 1s
+/// (unprogrammed/erased OTP fuse). This simulates a device where the debug
+/// unlock PK hash was never provisioned or disabled.
+#[test]
+#[cfg(not(any(feature = "fpga_realtime", feature = "fpga_subsystem")))]
```
swenson, line 1050: We should be able to test this on FPGA I thought?

common/src/debug_unlock.rs
```diff
@@ -67,6 +68,18 @@ pub fn create_debug_unlock_challenge(
...
     }
 
+    // Reject request if the PK hash fuse for this level is zeroized (i.e. all 1s).
+    let debug_auth_pk_offset =
+        soc_ifc.debug_unlock_pk_hash_offset(request.unlock_level as u32)? as u64;
+    let mci_base: AxiAddr = soc_ifc.mci_base_addr().into();
+    let debug_auth_pk_hash_base = mci_base + debug_auth_pk_offset;
+    let mut fuse_digest: [u32; 12] = [0; 12];
+    dma.read_buffer(debug_auth_pk_hash_base, &mut fuse_digest);
+    if Array4x12::from(fuse_digest) == Array4x12::from([0xFFFF_FFFFu32; 12]) {
```
swenson: Maybe we should check if it is all 0 or all 1?

common/src/debug_unlock.rs
```diff
@@ -67,6 +68,18 @@ pub fn create_debug_unlock_challenge(
...
+        crate::cprintln!("Prod debug unlock disabled: PK hash is all 1s (unprogrammed)");
```
zhalvorsen: I understand it can depend on the fusing strategy, but from what I understand we have been assuming all zeros is unprogrammed and all ones is "erased" (zeroized). Are you intending this check to be for zeroized or unprogrammed?

In my opinion we should check for both all zeros (unprogrammed) and and all ones (zeroized)

mhatrevi: This check was intentionally added for the debug-unlock-disabled scenario (i.e., zeroized state). While both the unprogrammed and zeroized states would naturally fail the hash check, the explicit check is required to meet FIPS requirements.

zhalvorsen: Okay, I guess what threw me off is there are a bunch of "unprogrammed" comments in both the firmware and tests.

common/src/debug_unlock.rs
mhatrevi: Zach had a similar comment. This was my reply "This check was intentionally added for the debug-unlock-disabled scenario (i.e., zeroized state). While both the unprogrammed and zeroized states would naturally fail the hash check, the explicit check is required to meet FIPS requirements." If we feel strongly about the 0s check, I can go add that as well.

swenson: Mostly, I'm not 100% sure what fuse values will be returned by the fuse controller once it has been "zeroized" -- the fuses will be set to all 1s, but I don't know if our fuse controller (or any other fuse controller) will detect that it has been zeroized and return 0s instead?

Seems safer to check for either case, if we're being extra careful for FIPS.

mhatrevi: Added.

common/src/debug_unlock.rs
mhatrevi: Added the check.

runtime/tests/runtime_integration_tests/test_debug_unlock.rs
mhatrevi: Most of the test_dbg_unlock_* tests in runtime integration tests have this filtering so I kept the same filter.

swenson (approved):
LGTM

zhalvorsen (approved):

runtime/tests/runtime_integration_tests/test_debug_unlock.rs
mhatrevi: Discussed offline.

timothytrippel: we will definitely want to test this on FPGA. but for Subsystem we probably want to test this E2E over JTAG , which lives in https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/tests/integration/src/jtag/test_prod_debug_unlock.rs

common/src/debug_unlock.rs
```diff
@@ -67,6 +68,20 @@ pub fn create_debug_unlock_challenge(
...
+    // Check both all-0s (unprogrammed) and all-1s (zeroized).
+    let debug_auth_pk_offset =
+        soc_ifc.debug_unlock_pk_hash_offset(request.unlock_level as u32)? as u64;
+    let mci_base: AxiAddr = soc_ifc.mci_base_addr().into();
+    let debug_auth_pk_hash_base = mci_base + debug_auth_pk_offset;
+    let mut fuse_digest: [u32; 12] = [0; 12];
+    dma.read_buffer(debug_auth_pk_hash_base, &mut fuse_digest);
+    let digest = Array4x12::from(fuse_digest);
+    if digest == Array4x12::from([0xFFFF_FFFFu32; 12]) || digest == Array4x12::from([0u32; 12]) {
+        crate::cprintln!("Prod debug unlock disabled: PK hash fuse is zeroized/unprogrammed");
```
ArthurHeymans, line 81: Would cfi_assert be needed here

PR #3603, opened 2026-04-10
DPE core upgrades: SVN, State, Crypto API, PCRs, env unification

parvathib:
Squashed cherry-picks from main:
- #3013: Update to just before adding SVN (DPE sub → b1c0e99)
- #3018: Support SVN in DeriveContext (DPE sub → c9b67f5)
- #3022: Use dpe::State for shared state (DPE sub → 57690bd)
- #3029: New Crypto API (DPE sub → 81aea72)
- #3038: Use both RT PCRs in RT DPE context
- #3164: Unify DPE environment creation

Note: CertifyKeyCsr DPE verification test is temporarily disabled. It requires DPE commit 4986ac5 ("verification: Add ECA EKU and fix CSR signature verification (#510)") which will be picked up in a later phase when the DPE submodule is updated further.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -998,6 +998,7 @@ dependencies = [
  "caliptra-runtime",
  "caliptra-test-harness",
  "cfg-if",
+ "dpe",
  "ufmt",
  "zerocopy",
 ]
@@ -1379,6 +1380,7 @@ dependencies = [
  "arrayvec",
  "caliptra-cfi-derive-git",
  "caliptra-cfi-lib-git",
+ "zerocopy",
  "zeroize",
 ]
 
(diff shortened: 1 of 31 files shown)


hw/latest/caliptra-ss
zhalvorsen, line 1: Should these submodules be updated as part of this change? I'm worried we might be updating them to a 2.1 commit

zhalvorsen (dismissed):

hw/latest/caliptra-ss
zhalvorsen: The DPE submodule definitely should, but I'm not sure about the `hw/*` submodules

parvathib: You are right! I reverted the hw/* submodule changes. Only DPE submodule changed

zhalvorsen (dismissed):

zhalvorsen (approved):

swenson (approved):

PR #3619, opened 2026-04-14
Fix the issue in ACTIVATE_FIRMWARE that uses fw_id instead of exec_bit

### LINKED ISSUE #3605: runtime ACTIVATE_FIRMWARE uses fw_id instead of exec_bit for SS_GENERIC_FW_EXEC_CTRL bitmap
* Caliptra Version: RTL version 2.1.2, SW version: rom-2.1.1
* Caliptra Mode: Subsystem Mode

In runtime/src/activate_firmware.rs, the ACTIVATE_FIRMWARE command handler uses fw_id instead of exec_bit as the bit index when building the images_to_activate_bitmap that drives SS_GENERIC_FW_EXEC_CTRL. This causes no bit at all to be set in SS_GENERIC_FW_EXEC_CTRL when fw_id is larger than the size of images_to_activate_bitmap (128). 

Bug location:
https://github.com/chipsalliance/caliptra-sw/blob/6b4ce093469cc43586bce3d665b98d4627c10e71/runtime/src/activate_firmware.rs#L111

wlibo-mrvl:
Fixed https://github.com/chipsalliance/caliptra-sw/issues/3605

### DIFF
--- a/runtime/src/activate_firmware.rs
+++ b/runtime/src/activate_firmware.rs
@@ -108,7 +108,7 @@ impl ActivateFirmwareCmd {
             if exec_bit == 0 || exec_bit == 1 || exec_bit > MAX_EXEC_GO_BIT_INDEX {
                 return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
             }
-            Self::set_bit(&mut images_to_activate_bitmap, fw_id as usize);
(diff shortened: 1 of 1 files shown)


mlvisaya (approved):
Looks good to me

zhalvorsen (approved):

PR #3621, opened 2026-04-15
[dpe] Update to non-generic crypto API

zhalvorsen:
Having generics in the API was duplicating much of the complicated DPE code (certificate encoding) which bloated the code size unnecessarily. DPE was updated to remove the generics so this gains some codespace back.


fdamato (approved):
LGTM! Thanks Zach !

drivers/src/sha2_512_384.rs
```diff
@@ -628,6 +628,68 @@ pub struct Sha384;
 /// SHA-512 variant implementation
 pub struct Sha512;
 
+#[cfg(feature = "runtime")]
+pub struct DpeHasher<'a> {
+    op: Sha2DigestOp<'a, Sha384>,
+    active: Option<()>,
```
clundin25, line 634: Does a `state` Enum make sense?

clundin25: Alternatively, could you use an `Option<Op>` to imply it's active?

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -1011,7 +1011,7 @@ fn test_authorize_and_stash_after_update_reset_multiple_set_manifest() {
         fw_id: FW_ID_0,
         measurement: IMAGE_DIGEST1,
         source: ImageHashSource::InRequest as u32,
-        flags: 0, // Don't skip stash
```
clundin25, line 1014: Why change all of these tests?

clundin25 (approved):
Nice code space gains !

drivers/src/sha2_512_384.rs
zhalvorsen: `Option<Op>` has lifetime issues I haven't been able to untie. I thought about using a separate enum, but I like the idea of using `take` logic and the enum would effectively only be `Active` or `Inactive`

timothytrippel (approved):

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
zhalvorsen: DPE can no longer have two contexts with the same `tci_type` so these tests were failing

clundin25: Oh I see, do we need to go back and fix this? Worried that this will break integrations

drivers/src/sha2_512_384.rs
clundin25: Maybe the trait could be implemented on the enum itself to address the lifetime issues? Feel free to ignore, just fun to think about

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
zhalvorsen: I created https://github.com/chipsalliance/caliptra-sw/issues/3660 to track this before the next release.

drivers/src/sha2_512_384.rs
zhalvorsen: Potentially, keeping a handle the driver and the SHA operation were the main pain points. I'll think about it some more and maybe improve this in later changes

PR #3622, opened 2026-04-15
Add DPE ML-DSA support to Caliptra 2.0

ArthurHeymans:
Squashed backport of:
 - [dpe] Enable hybrid DPE feature (#3243)
 - [dpe] Add ML-DSA DPE Crypto trait (#3315)
 - [dpe] Add support for 64 DPE contexts (#3246)
 - [dpe] Add ML-DSA DPE command (#3326)
 - [dpe] Test ML-DSA CertifyKey (#3357)
 - [dpe] Test using external mu with ML-DSA profile (#3371)
 - [dpe] reduce nesting of InvokeDpeCmd::execute (#3386)
 - [dpe] Add optional ML-DSA response over DMA (#3391)
 - [dpe] Add HW model support for large DPE responses (#3403)
 - [dpe] Support both DPE profiles in more tests (#3407)
 - [dpe] Add a test for worst case scenario certs and CSRs (#3417)
 - [dpe] Fail early for response sizing (#3415)
 - [dpe] Add ML-DSA support to CertifyKeyExtended (#3426)
 - [dpe] Test ML-DSA profile with golang verification (#3454)
 - [dpe] Add comment that initialization is profile agnostic (#3460)

 DPE rev: a26db5b — last rev bumped by PRs in this list (#3371).

 caliptra-2.0 adaptations:
 - DpeMldsaCrypto uses MldsaReg (2.0 has dedicated ML-DSA HW register;
   main uses AbrReg shared Adams Bridge register for ML-DSA and ML-KEM)
 - SignData::Mu returns CryptoError::NotImplemented (no external-mu on 2.0)
 - Removed ML-KEM commands (no ML-KEM HW on 2.0)
 - Removed main-only commands (ocp_lock, shake256)
 - Flat PersistentData paths (2.0 uses single struct, not rom/fw split)
 - Kept 2.0 caliptra-cfi-*-git package names to avoid ROM recompilation
 - STACK_SIZE=136KB (main uses 154KB; 2.0 has ~30KB more PersistentData)
 - RUNTIME_SIZE=176KB (main uses 210KB; 2.0 has fewer runtime features)

 Additional fixes on top:
 - Fix stack overflow in initialize_dpe by using execute_serialized
 - Split InvokeDpe and CertifyKeyExtended into per-profile dispatch
 - Defer common response buffer allocation to reduce stack usage
 - Remove external AXI response support (unnecessary in 2.0)
 - Remove subsystem mailbox size limit handling (mailbox not smaller in 2.0)
 - Reduce MAX_HANDLES to preserve ROM layout
 - Derive DPE context thresholds from MAX_HANDLES
 - Fix tests for 32/64 context configurations


runtime/src/invoke_dpe.rs
```diff
@@ -12,148 +12,264 @@ Abstract:
...
+        // Execute the DPE command and get the length of the response that was written
+        let len = Self::execute(drivers, dpe_cmd_buf, mbox_resp, CaliptraDpeProfile::Mldsa87)?;
+
+        // We are done if the response is going over the mailbox
+        let respond_to_mailbox = !cmd.flags.external_axi_response();
+        if respond_to_mailbox {
+            return Ok(len);
+        }
+
+        // Populate the checksum so the full response can be checked at the destination
```
zhalvorsen: I think we should remove external response support for 2.0. The mailboxes are always big enough to support the full response and this just adds unnecessary complexity.

runtime/src/certify_key_extended.rs
```diff
@@ -12,30 +12,135 @@ Abstract:
...
+            mbox_resp,
+        )?;
+
+        // We are done if the response is going over the mailbox
+        let respond_to_mailbox = !cmd.flags.external_axi_response();
+        if respond_to_mailbox {
+            return Ok(len);
+        }
+
+        // Populate the checksum so the full response can be checked at the destination
```
zhalvorsen: If we remove it from `invoke_dpe.rs` we should probably remove it from here as well

zhalvorsen (dismissed):

runtime/src/invoke_dpe.rs
parvathib: done

runtime/src/certify_key_extended.rs
parvathib: removed external axi response here too

runtime/src/invoke_dpe.rs
```diff
@@ -12,148 +12,206 @@ Abstract:
...
+        // handle rotation so the caller doesn't lose the handle.
+        let mbox_resp = if drivers.soc_ifc.subsystem_mode() {
+            // The mailbox size is smaller when subsystem is enabled
+            let len = usize::min(mbox_resp.len(), SUBSYSTEM_MAILBOX_SIZE_LIMIT);
+            mbox_resp
+                .get_mut(..len)
+                .ok_or(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY)?
+        } else {
+            mbox_resp
+        };
```
zhalvorsen: In 2.0 hardware the mailbox isn't smaller in subsystem mode, so you can remove this special handling here and in certify_key_extended.


```suggestion
```

parvathib: added suggestion and removed the constant wrongly introduced too.

zhalvorsen (approved):

PR #3625, opened 2026-04-16
[feat] Add Stable Owner Key derivation from HEK seed

mhatrevi:
ROM derives a Stable Owner Root Key from HEK seed via HKDF into KV15 during cold boot. CM_DERIVE_STABLE_KEY gains OwnerKey (0x3) which uses CMAC-KDF + HMAC-KDF with label "Stable Owner Key" to produce child keys as encrypted CMKs.

Both ROM firmware processor and runtime handle the new key type with identical derivation logic. Adds integration tests for all three key types (IDevId, LDevId, OwnerKey) and a separate test verifying different personalization seeds produce different keys.


mhatrevi:
Implements https://github.com/chipsalliance/Caliptra/issues/657

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -170,6 +176,86 @@ impl InitDevIdLayer {
...
+                KeyUsage::default().set_hmac_key_en(),
+            )),
+            HmacMode::Hmac512,
+        )?;
+
+        // Step 2: HKDF-Expand — HMAC-KDF(key=PRK, label) → stable owner root key
+        hmac_kdf(
+            &mut env.hmac,
+            HmacKey::Key(KeyReadArgs::new(hek_seed_temp)),
+            b"stable_owner_root_key",
```
swenson: nit: Maybe pull this out into a constant?

swenson (dismissed):
LGTM, one nit

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -170,6 +179,87 @@ impl InitDevIdLayer {
...
+    /// Decrypt HEK Seed
+    ///
+    /// # Arguments
+    ///
+    /// * `env` - ROM Environment
+    /// * `key_id` - Key Vault slot to store the decrypted HEK seed in
+    #[cfg_attr(feature = "cfi", cfi_impl_fn)]
+    fn decrypt_hek_seed(env: &mut RomEnvFips, key_id: KeyId) -> CaliptraResult<()> {
+        env.doe.decrypt_hek_seed(&DOE_IV, key_id)?;
+        report_boot_status(IDevIdDecryptHekSeedComplete.into());
```
clundin25, line 199: Posting this so I don't forget.

This should be gated behind an OCP LOCK feature check: https://github.com/chipsalliance/caliptra-sw/blob/main/rom/dev/src/flow/cold_reset/ocp_lock.rs#L32-L36 like this

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -170,6 +179,87 @@ impl InitDevIdLayer {
...
+            &mut env.trng,
+            HmacTag::Key(KeyWriteArgs::new(
+                stable_owner,
+                KeyUsage::default().set_aes_key_en(),
+            )),
+            HmacMode::Hmac512,
+        )?;
+
+        // Erase the temporary HEK seed slot
+        env.key_vault.erase_key(hek_seed_temp)?;
```
clundin25: This won't be erased if line 254 errors. Maybe not a big deal, just wanted to point that out

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2586,21 +2586,28 @@ impl Commands {
         let aes_key = match key_type {
             CmStableKeyType::IDevId => AesKey::KV(KeyReadArgs::new(KEY_ID_STABLE_IDEV)),
             CmStableKeyType::LDevId => AesKey::KV(KeyReadArgs::new(KEY_ID_STABLE_LDEV)),
+            CmStableKeyType::OwnerKey => AesKey::KV(KeyReadArgs::new(KEY_ID_STABLE_OWNER)),
```
clundin25, line 2596: Should return error if OCP LOCK is enabled? https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/src/ocp_lock/mod.rs#L1023-L1025

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Updated.

runtime/src/cryptographic_mailbox.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -76,12 +79,26 @@ impl InitDevIdLayer {
...
+
         // Clear Deobfuscation Engine Secrets
         Self::clear_doe_secrets(env)?;
 
         // Derive the DICE CDI from decrypted UDS
         Self::derive_cdi(env, KEY_ID_UDS, KEY_ID_ROM_FMC_CDI)?;
 
+        if derive_stable_owner {
+            // Derive stable owner root key from HEK seed (HKDF-Extract + Expand)
+            Self::derive_stable_owner_root_key(env, KEY_ID_HEK_SEED, KEY_ID_STABLE_OWNER)?;
```
bluegate010, line 99: Nit: why not have this line appear right after `Self::decrypt_hek_seed(env, KEY_ID_HEK_SEED)` above.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2583,24 +2583,38 @@ impl Commands {
 
         let key_type: CmStableKeyType = request.key_type.into();
 
+        // The Stable Owner Key feature is mutually exclusive with OCP LOCK:
+        // when OCP LOCK is enabled the HEK seed is reserved for OCP LOCK use
+        // and KEY_ID_STABLE_OWNER is not populated by ROM.
+        if key_type == CmStableKeyType::OwnerKey && drivers.soc_ifc.ocp_lock_enabled() {
+            Err(CaliptraError::CMB_STABLE_OWNER_KEY_NOT_AVAILABLE)?;
+        }
```
bluegate010, line 2591: After the mutual exclusivity is addressed, I think this conditional will need to change to be, "not only is LOCK enabled, but we used the HEK seed to derive the stable owner key", since in 2.2 we would use something else besides the HEK seed to make the stable owner key.

Since that's a strict relaxation of the constraint here, I think that could come in a 2.2 ROM/RT release, so just noting here for posterity. Perhaps could amend the comment above with a TODO.

rom/dev/src/flow/cold_reset/idev_id.rs
```diff
@@ -76,12 +79,26 @@ impl InitDevIdLayer {
         // Decrypt the Field Entropy
         Self::decrypt_field_entropy(env, KEY_ID_FE)?;
 
+        // Stable Owner Key and OCP LOCK are mutually exclusive consumers of
+        // `fuse_hek_seed`. OCP LOCK reads it directly from soc_ifc, so when
+        // it is enabled we skip the HEK seed DOE decrypt and the Stable
+        // Owner Root Key derivation.
+        let derive_stable_owner = !env.soc_ifc.ocp_lock_enabled();
```
clundin25: Is it enough to check if OCP LOCK is enabled in hardware or should we check the firmware feature flag too?

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: Updated.

runtime/src/cryptographic_mailbox.rs
mhatrevi: Updated.

rom/dev/src/flow/cold_reset/idev_id.rs
mhatrevi: I think hardware check should be enough.

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2583,29 +2583,53 @@ impl Commands {
...
+                    &mut drivers.hmac,
+                    (&Array4x16::from(k0)).into(),
+                    &data,
+                    None,
+                    &mut drivers.trng,
+                    (&mut tag).into(),
+                    HmacMode::Hmac512,
+                )?;
+            }
+            _ => {
```
mlvisaya: qq: looks like this catch all case will also match CmStableKeyType::Reserved, is that ok? 
or should it be:

CmStableKeyType::IDevId | CmStableKeyType::LDevId => {
   // DOT_PREFIX code here
}
_ => {
  // some error
}

mhatrevi: The reserved type fails earlier, but I agree, this is a good sanity check. Updated.

drivers/src/soc_ifc.rs
```diff
@@ -644,6 +644,24 @@ impl SocIfc {
         )
     }
 
+    /// Check if the Stable Owner Key feature is enabled via SS_STRAP_GENERIC[3] bit 0.
+    pub fn stable_owner_key_enabled(&self) -> bool {
+        self.soc_ifc.regs().ss_strap_generic().at(3).read() & 1 != 0
```
swenson, line 649: I think we need to document this strap's value in the ROM documentation, and also update the runtime docs for the stable owner key CM method to indicate that this strap is required to be able to use the stable owner key.

swenson (dismissed):
LGTM

drivers/src/soc_ifc.rs
mhatrevi: Added.

swenson (approved):
LGTM

PR #3626, opened 2026-04-16
Add release history tables and ci-tools/release/release_info.py

swenson:
Add tables to README.md listing every official ROM and FMC+Runtime firmware release with version, git commit, SVN, and hashes:

- ROM rows include the SHA256 embedded in the RomInfo struct and the SHA384 over the full ROM binary, for both the no-log and with-log variants.
- FW rows include the FMC and Runtime SHA384 digests extracted from the firmware image manifest's TOC entries (not hashes of the bundle file itself).

Add ci-tools/release/release_info.py to (re)generate these tables. The script downloads pre-built release assets from GitHub, parses the 1.x and 2.x firmware manifest formats, and verifies each ROM's embedded SHA256 by recomputing it with the builder's word-reversed hashing convention. Releases without pre-built binaries (rom-1.0.3) can be built from source via --build, which checks out a git worktree at the tag and invokes caliptra-builder.


ci-tools/release/release_info.py
```diff
@@ -0,0 +1,524 @@
+#!/usr/bin/env python3
```
clundin25: I think it makes sense to just make this a subcommand of `cargo xtask release`?

swenson: Sure. I think it would be straightforward enough to port this to Rust.

swenson: Done.

xtask/src/main.rs
```diff
@@ -68,6 +69,24 @@ pub enum ReleaseCommands {
...
+    ///
+    /// Shells out to `gh` to download release assets and verifies embedded
+    /// ROM SHA256 digests. Use `--build` to also build releases that lack
+    /// pre-built binary artifacts (e.g. rom-1.0.3).
+    Info {
+        /// Release name (e.g. 'rom-2.1.1', 'fw-2.0.1'). Omit when using --all.
+        release_name: Option<String>,
+        /// Process every release in the built-in list.
+        #[arg(long)]
+        all: bool,
```
clundin25, line 82: nit: I think `release_name: None` implies this anyway

xtask/src/release_info.rs
```diff
@@ -0,0 +1,810 @@
...
+    pub rt_digest: Option<String>,
+
+    pub svn: u32,
+}
+
+fn run_gh(args: &[&str]) -> Result<String> {
+    let out = Command::new("gh")
+        .args(args)
+        .output()
+        .context("failed to run gh")?;
```
clundin25, line 240: nit: Could alternatively use Octocrab and a zip crate to remove CLI dependencies, though I think they will almost always be there anyway.

clundin25 (dismissed):

xtask/src/release_info.rs
swenson: Yeah, I went the `gh` path in the Python script to avoid having to add Python dependencies. I can see if it makes sense to use octocrab. Zips shouldn't be a problem.

clundin25 (approved):
Awesome!

mhatrevi (approved):

swenson:
Thanks!

PR #3637, opened 2026-04-20
emulator: Allow writing to prod debug unlock num / offset registers

swenson:
These registers are allowed to be written by SoC / MCU, but not by Caliptra. We don't have a good way to enforce such a separation in the emulator at the moment, so we mark them as RW.


timothytrippel (approved):

zhalvorsen (approved):

amit-kumarh (approved):

swenson:
Thanks!

PR #3638, opened 2026-04-20
[fw_info] Add more information about how firmware was verified (2.0)

zhalvorsen:
This adds the following fields to `FwInfoResp`:

* vendor_pub_key_hash
* image_manifest_pqc_type
* vendor_ecc384_pub_key_index
* vendor_pqc_pub_key_index

These will be used by MCU key revocation APIs to make sure it is safe to revoke a key.


swenson (approved):
LGTM

Issue #3362, opened 2026-02-20
[OCP-LOCK]: Enable Derive MEK tests on subsystem FPGA


clundin25:
https://github.com/chipsalliance/caliptra-sw/pull/3358 integrates with the Encryption Engine block.

We still need to write an integration that is compatible with subsystem.

clundin25:
@GwangbaeCHOI to work on this item.

PR #3640, opened 2026-04-21
[OCP-LOCK] Enable encryption engine interacting tests on subsystem FPGA

GwangbaeCHOI:
- Enable DERIVE_MEK tests (#3362)
- Enable CLEAR_KEY_CACHE tests
- Enable LOAD_MEK tests
- Enable UNLOAD_MEK tests

### DIFF
--- a/drivers/src/dma.rs
+++ b/drivers/src/dma.rs
@@ -1176,11 +1176,19 @@ bitfield! {
     command_code, set_command_code: 5, 2;
     error_code, _: 19, 16;
     ready_bit, _: 31;
+
+    #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
+    _, set_ready_bit: 31;
 }
 impl EncryptionEngineCtrl {
     fn clear() -> Self {
         let mut ctrl = Self(0);
         ctrl.set_done_bit(true);
+
+        // To maintain the RDY bit to be set
+        #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
+        ctrl.set_ready_bit(true);
+
         ctrl
     }
 
@@ -1188,6 +1196,11 @@ impl EncryptionEngineCtrl {
         let mut ctrl = Self(0);
         ctrl.set_command_code(cmd.into());
         ctrl.set_execute_bit(true);
+
+        // To maintain the RDY bit to be set
+        #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
+        ctrl.set_ready_bit(true);
+
         ctrl
     }
 }
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -6,8 +6,8 @@
 use crate::api_types::{DeviceLifecycle, Fuses};
 use crate::bmc::Bmc;
 use crate::fpga_regs::{
-    Control, FifoData, FifoRegs, FifoStatus, FlashControl, FlashCtrlRegs, FlashOpStatus,
-    ItrngFifoStatus, WrapperRegs,
+    Control, EncryptionEngineControl, FifoData, FifoRegs, FifoStatus, FlashControl, FlashCtrlRegs,
+    FlashOpStatus, ItrngFifoStatus, WrapperRegs,
 };
 use crate::keys::{DEFAULT_LIFECYCLE_RAW_TOKENS, DEFAULT_MANUF_DEBUG_UNLOCK_RAW_TOKEN};
 use crate::mcu_boot_status::McuBootMilestones;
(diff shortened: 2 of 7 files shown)


drivers/src/dma.rs
```diff
@@ -1176,18 +1176,31 @@ bitfield! {
     command_code, set_command_code: 5, 2;
     error_code, _: 19, 16;
     ready_bit, _: 31;
+
+    #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
```
GwangbaeCHOI, line 1180: These codes are not really needed for `fpga_realtime`. But I found that the current fpga-subsystem tests on MCU uses fpga-realtime ROM of Caliptra, and that's why I added `fpga_realtime` feature here.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -902,6 +912,118 @@ impl ModelFpgaSubsystem {
         }
     }
 
+    fn realtime_encryption_engine_register_handler(
```
GwangbaeCHOI, line 915: This thread simulates the encryption engine similar to the thread for internal TRNG. Since we don't have exactly same RW access to registers, especially the `RDY` bit of `CTRL` register, There are additional codes to retain the `RDY` bit.

drivers/src/dma.rs
```diff
@@ -1176,18 +1176,31 @@ bitfield! {
...
+    _, set_ready_bit: 31;
 }
 impl EncryptionEngineCtrl {
     fn clear() -> Self {
         let mut ctrl = Self(0);
         ctrl.set_done_bit(true);
+
+        // To maintain the RDY bit to be set
+        #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
+        ctrl.set_ready_bit(true);
```
GwangbaeCHOI, line 1190: `RDY` bit will remain `1` after clearing the register.

drivers/src/dma.rs
```diff
@@ -1176,18 +1176,31 @@ bitfield! {
...
     }
 
     fn execute(cmd: EncryptionEngineCommandCode) -> Self {
         let mut ctrl = Self(0);
         ctrl.set_command_code(cmd.into());
         ctrl.set_execute_bit(true);
+
+        // To maintain the RDY bit to be set
+        #[cfg(any(feature = "fpga_realtime", feature = "fpga_subsystem"))]
+        ctrl.set_ready_bit(true);
```
GwangbaeCHOI, line 1202: To keep `RDY` bit to `1` after executing a command

drivers/src/dma.rs
clundin25: Confusingly, `fpga_realtime` in MCU == `fpga_subsystem` in Caliptra.

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -469,6 +474,11 @@ pub struct ModelFpgaSubsystem {
     saved_ocp_lock_en: bool,
     saved_security_state: SecurityState,
     saved_lc_state: Option<LifecycleControllerState>,
+
+    // OCP LOCK encryption engine FPGA emulator
+    pub realtime_encryption_engine: Option<thread::JoinHandle<()>>,
+    pub realtime_encryption_engine_exit_flag: Arc<AtomicBool>,
+    pub realtime_encryption_engine_paused: Arc<AtomicBool>,
```
clundin25, line 481: nit: I think you can skip using `Arc` to wrap the `AtomicBool`? IIRC it should be safe to copy and share the `AtomicBool`s across threads

hw-model/src/model_fpga_subsystem.rs
clundin25: Should we make a change to the FPGA wrapper RTL?

drivers/src/dma.rs
clundin25: Why does this bit need special treatment on the FPGA? What would it take to eliminate this feature flagged code and simplify the firmware?

hw-model/src/model_fpga_subsystem.rs
GwangbaeCHOI: If we can make RDY bit sticky, it would be better

drivers/src/dma.rs
GwangbaeCHOI: WIthin the FPGA wrapper, CTRL register is RW. So, Caliptra can overwrite RDY bit which supposed to be read-only. If we can make RDY bit to be sticky, we won't need this code.

clundin25: Thanks @GwangbaeCHOI let me see if we can make a change to the FPGA Wrapper

clundin25: @GwangbaeCHOI sounds like we can make that happen, want to wait for a new bitstream or move ahead ?

GwangbaeCHOI: I can wait. So that we won't need to update again.

PR #3641, opened 2026-04-21
Add ML-DSA-87 software implementation

zhalvorsen:
This adds an embedded friendly implementation of SHAKE and ML-DSA-87. The implementation prioritizes size over speed. It is able to build into Caliptra firmware using ~10KB of codespace for key-gen, sign, and verify. Other benchmarks still need to be taken.

Using the [sha3](https://crates.io/crates/sha3) crate instead of the local SHAKE implementation adds another ~2KB of codespace.


common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -0,0 +1,1326 @@
...
+    shake256.squeeze(&mut rho_prime);
+
+    let mut sign = Signature {
+        c_tilde: [0u8; 2 * LAMBDA_BYTES],
+        z: Vector7::default(),
+        h: Vector8::default(),
+    };
+    let mut w1 = Vector8::default();
+
+    // We use a union in C, in Rust we can just use separate variables or reuse if needed.
```
ternst-nv: We could also use an enum if we needed.  As it stands analysis puts us at about 80Kb stack usage to implement signing logic, with 85Kb reserved.  So any vectors we can save would be important for having sufficient stack space to handle the command overhead before calling into this functionality.

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -0,0 +1,1326 @@
...
+        let mut y = Vector7::default();
+        vector7_expand_mask(&mut y, &rho_prime, kappa);
+        vector7_add(&mut sign.z, &y, &cs1);
+
+        vector8_mul_scalar(&mut cs2, &priv_key.s2_ntt, &c_ntt);
+        vector8_inverse_ntt(&mut cs2);
+
+        let mut r0 = Vector8::default();
+        vector8_sub(&mut r0, &w, &cs2);
+        let r0_copy = r0;
```
ternst-nv: I did some examining here, we could likely use an `in_place` version of the `vector8_low_bits` to avoid having to make this copy.  That would save ~8Kb in stack space.

ternst-nv (dismissed):

common/crypto/mldsa/src/mldsa87.rs
zhalvorsen: There were some similar optimizations I added that should save ~31KB. Thanks for pointing this out!

common/crypto/mldsa/src/mldsa87.rs
zhalvorsen: Changed to an enum. Thanks!

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -877,13 +877,22 @@ fn sign_internal(
     };
     let mut w1 = Vector8::default();
 
-    // We use a union in C, in Rust we can just use separate variables or reuse if needed.
-    // Given stack space is usually not as tight in Rust unless specified, but let's be careful.
-    let mut cs1 = Vector7::default();
-    let mut cs2 = Vector8::default();
+    #[derive(Clone, Copy)]
```
ternst-nv: [nit] Might want to drop `Copy` here to prevent unexpected copies of the parameter.

zhalvorsen: Good point! Done

ternst-nv (dismissed):
LGTM

ternst-nv (approved):

PR #3642, opened 2026-04-21
emulator: Allow writing to prod debug unlock num / offset registers (2.1)

swenson:
These registers are allowed to be written by SoC / MCU, but not by Caliptra. We don't have a good way to enforce such a separation in the emulator at the moment, so we mark them as RW.

(cherry picked from commit edcdfbd6526d16f40296eee8b0ab3196f3da6514)


mhatrevi (approved):

swenson:
Thanks!

PR #3643, opened 2026-04-21
fpga: use strap UDS/FE for deterministic IDevID for tests

parvathib:
Switch `secrets_valid` to True so DOE reads UDS and FE from strap registers
 instead of OTP. This gives deterministic and identical IdevID on FPGA and emulator
 for SPDM attestation tests


swenson (approved):
LGTM

mhatrevi (approved):

PR #3644, opened 2026-04-22
CPU emulator interrupt robustness: deliver only active PIC IRQs

v-ramalleti:
Problem:
Unhandled interrupt failures were seen when deferred external interrupt events were replayed after interrupt state transitions.

**What changed:**

ExtInt handling now checks PIC pending state at delivery time.
Delivery uses the active highest-priority pending IRQ, not the stale queued IRQ.
Deferred events with no active pending source are ignored.
save=true deferral path remains unchanged.

**Behavioral impact:**

Valid interrupts are still delivered normally.
Spurious stale deferred events no longer trigger trap handling.
Reduces intermittent interrupt-related test flakiness.

Before this fix we are observing Unhandled interrupt 38
[MCI_TIMER_BEFORE_CPU_FIX.txt](https://github.com/user-attachments/files/26950512/MCI_TIMER_BEFORE_CPU_FIX.txt)

After this fix the test passed. 
[MCI_TIMER_AFTER_CPU_FIX.txt](https://github.com/user-attachments/files/26950517/MCI_TIMER_AFTER_CPU_FIX.txt)

Apart from cpu fix, I have added another register fix which says read/write register but code is written as readonly. 
image

### DIFF
--- a/sw-emulator/lib/cpu/src/cpu.rs
+++ b/sw-emulator/lib/cpu/src/cpu.rs
@@ -304,6 +304,9 @@ pub struct Cpu<TBus: Bus> {
     /// Machine External interrupt enabled
     ext_int_en: bool,
 
+    /// Tracks ExtInt events that were deferred while interrupts were masked.
+    deferred_ext_int: [bool; 256],
+
     /// Halted state
     halted: bool,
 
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM -- I ran some MCU integration tests with this change and it seems to work well there as well.

PR #3656, opened 2026-04-24
Add MRAC (Memory Region Access Control) emulation

### LINKED ISSUE #3601: MRAC emulation in CPU
It might be nice to have optional emulation of the VeeR MRAC RISC-V CSR: https://chipsalliance.github.io/Cores-VeeR-EL2/html/main/docs_rendered/html/memory-map.html#region-access-control-register-mrac

I think this would be best implemented as a RootBus wrapper, that takes an inner RootBus that it delegates certain reads and writes to.

At least as far as we should honor two aspects:

- If a region of the memory map is set to side effect, then we allow 32-bit level access to the correct word always (but disallow non-aligned access to 8-bit or 16-bit) that is passed through to the inner RootBus.
- If a region is set to cachable, then the outer RootBus accepts any 8, 16, or 32-bit access with any alignment, but always does 64-bit access to the inner RootBus (translated as two 32-bit accesses)

Notably, I would expect two consequences of this:

1. Non-aligned access to a memory should fail it is set to side-effect
2. If side effect is not turned on correctly for the mailbox, then its lock will probably not work correctly (since the user locking register is adjacent to another register).

However, I also fear that enabling MRAC emulation might break a lot of things, so maybe we should leave it off by default.

iyernaveenr:
## Add MRAC (Memory Region Access Control) emulation

Emulate the VeeR EL2 MRAC CSR (0x7C0) so the sw-emulator and hw-model exhibit the same per-region access constraints as real hardware. Reference: [VeeR EL2 spec](https://chipsalliance.github.io/Cores-VeeR-EL2/html/main/docs_rendered/html/memory-map.html#region-access-control-register-mrac).

### What this does

The 32-bit address space is split into 16 fixed 256 MB regions; each region has two MRAC bits (`cacheable`, `sideeffect`) that gate LSU behavior:

- **LSU data load:**
  - `sideeffect`: misaligned access is rejected; aligned access goes through with the requested size (no widening).
  - `cacheable`: read is widened to a 64-bit aligned dword fetch; write goes through with the requested size.
  - neither: passthrough.
- **`0b11` (both bits set):** illegal in RTL; sanitized to `0b10` (sideeffect-only) on CSR write.

IFU (instruction fetch) is handled at the cpu side rather than in `MracBus`: `Cpu::read_instr` issues word-aligned reads to the bus and slices the requested halfword/word out locally, with a second word read only when an `RVC` instruction straddles a 4-byte boundary. This naturally bypasses MracBus's LSU rules (which IFU does not have) without needing to thread any access-type information through the `Bus` trait.

### Why this matters

Without MRAC enforcement the emulator does not catch access-pattern bugs that would fail on real hardware. Concrete example: reading a mailbox `USER` register at offset 4 in a non-side-effect region triggers a widened 64-bit read that *also* reads the adjacent `LOCK` register at offset 0, silently acquiring the lock. The `sideeffect` flag prevents this by restricting reads to the exact requested address.

### Implementation

1. New `MracBus<TBus: Bus>` wrapper in `sw-emulator/lib/bus` that intercepts LSU reads and writes and enforces the rules above before delegating to the inner bus. The widening helper handles accesses that straddle the 8-byte dword boundary.
2. MRAC CSR (0x7C0) added to `CsrFile` with a shared `Rc<Cell<u32>>`; CSR writes update the cell that `MracBus` reads on every access. `CsrFile::reset()` re-syncs the cell so warm reset cannot leave `MracBus` enforcing stale region rules.
3. `Cpu::read_instr` rewritten to do its own word-aligned bus reads + local slicing for IFU. The `Bus::read` trait signature is **unchanged** from `main` — no breaking change for downstream consumers.
4. `ModelEmulated` and the `sw-emulator` app both wrap their root bus in `MracBus`. The CLI gains a `--mrac` flag; `ModelEmulated` enables MRAC by default so existing integration tests exercise the new path. This matches real hardware, where the boot code unconditionally writes CSR 0x7C0 at startup.

### Tests

- `mrac_bus` unit tests cover disabled passthrough, all four region encodings, alignment enforcement for side-effect, cacheable LSU widening including boundary-crossing accesses, the mailbox-lock spurious-acquire case that motivated this work, that cacheable writes are not widened, and shared-cell update propagation.
- `csr_file` unit tests cover MRAC CSR write sanitization across all 16 regions, preservation of legal combinations, and the reset-time cell resync.
- The full `caliptra-drivers` integration test binary passes with MRAC enabled.

Closes #3601

### DIFF
--- a/sw-emulator/lib/cpu/src/cpu.rs
+++ b/sw-emulator/lib/cpu/src/cpu.rs
@@ -22,6 +22,7 @@ use crate::Pic;
 use bit_vec::BitVec;
 use caliptra_emu_bus::{Bus, BusError, Clock, Event, TimerAction};
 use caliptra_emu_types::{RvAddr, RvData, RvException, RvSize};
+use std::cell::Cell;
 use std::fs::File;
 use std::io::Write;
 use std::rc::Rc;
@@ -374,10 +375,16 @@ pub enum StepAction {
 
 impl<TBus: Bus> Cpu<TBus> {
     /// Create a new RISCV CPU
-    pub fn new(bus: TBus, clock: Rc<Clock>, pic: Rc<Pic>, args: CpuArgs) -> Self {
+    pub fn new(
+        bus: TBus,
+        clock: Rc<Clock>,
+        pic: Rc<Pic>,
+        args: CpuArgs,
+        mrac: Rc<Cell<u32>>,
+    ) -> Self {
         Self {
             xregs: XRegFile::new(),
-            csrs: CsrFile::new(clock.clone(), pic.clone()),
+            csrs: CsrFile::new(clock.clone(), pic.clone(), mrac),
             pc: args.org.reset_vector,
             next_pc: args.org.reset_vector,
             bus,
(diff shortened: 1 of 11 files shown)


sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -769,7 +776,7 @@ impl<TBus: Bus> Cpu<TBus> {
 
         match size {
             RvSize::Byte => Err(RvException::instr_access_fault(addr)),
-            _ => match self.bus.read(size, addr) {
+            _ => match self.bus.read(size, addr, BusAccessType::InstrFetch) {
```
swenson: As far as I can tell, this is the only place that the `BusAccessType::InstrFetch` logic is ever invoked.

Would it be simpler, at least in terms of code differences, to make this have some special logic to bypass the `MracBus::read()` logic and always read `u32` and translate here to the `u8` or `u16` as needed?

Then we can eliminate all of the noise from having the `BusAccessType` added to every interface -- every access is `DataLoad` except here.

iyernaveenr: Thanks for your review! I agree -- Handling IFU at the call site is cleaner. Implemented your suggestion. Always-u64 turned out to break the existing instr tests -- the test harness sizes its ROM exactly to .text, so fetching the unused half of an 8-byte chunk hits an out-of-bounds read.

So the helper does: word-aligned u32 read, second u32 only on RVC straddle. Matches what you proposed verbatim. BusAccessType dropped across the tree.

sw-emulator/lib/periph/src/hash_sha512.rs
```diff
@@ -11,7 +11,6 @@ Abstract:
     File contains SHA512 peripheral implementation.
 
 --*/
-
```
swenson, line 14: nit: could you remove these unrelated, whitespace-only changes from the PR?

swenson (commented):
Code LGTM aside from nits. I'm running the caliptra-mcu-sw integration tests as well to see if there are any updates needed there.

sw-emulator/lib/periph/src/hash_sha512.rs
iyernaveenr: Done. Thanks!

swenson (approved):
LGTM

PR #3663, opened 2026-04-27
Feature/optimize zeroize impls

ternst-nv:
Implement custom zeroization logic for large data structures which relies on the byte slice zeroization implemented by the zeroization crate.  This reduces instruction usage, since a single zeroization method can be called rather than recursively calling each individual field's zeroize.

This is guaranteed to be safe, since the various structures implement ZeroCopy traits which guarantee their layout as well as all 0s being a safe value.

Add a unit test to verify the data is cleared.

### DIFF
--- a/hw-model/src/model_emulated.rs
+++ b/hw-model/src/model_emulated.rs
@@ -105,6 +105,16 @@ impl ModelEmulated {
     }
 }
 
+impl ModelEmulated {
+    /// Read a slice of the DCCM (Data Closely Coupled Memory) by absolute address.
+    /// The `addr` and `len` are in the CPU address space (starting at 0x5000_0000).
+    pub fn dccm_read(&self, addr: u32, len: usize) -> &[u8] {
+        const DCCM_BASE: u32 = 0x5000_0000;
+        let offset = (addr - DCCM_BASE) as usize;
+        &self.cpu.bus.bus.dccm.data()[offset..offset + len]
+    }
+}
+
 fn hash_slice(slice: &[u8]) -> u64 {
     let mut hasher = DefaultHasher::new();
     std::hash::Hash::hash_slice(slice, &mut hasher);
--- a/hw-model/src/model_emulated.rs
+++ b/hw-model/src/model_emulated.rs
@@ -105,16 +105,6 @@ impl ModelEmulated {
     }
 }
 
-impl ModelEmulated {
-    /// Read a slice of the DCCM (Data Closely Coupled Memory) by absolute address.
-    /// The `addr` and `len` are in the CPU address space (starting at 0x5000_0000).
-    pub fn dccm_read(&self, addr: u32, len: usize) -> &[u8] {
-        const DCCM_BASE: u32 = 0x5000_0000;
-        let offset = (addr - DCCM_BASE) as usize;
-        &self.cpu.bus.bus.dccm.data()[offset..offset + len]
-    }
-}
-
 fn hash_slice(slice: &[u8]) -> u64 {
     let mut hasher = DefaultHasher::new();
     std::hash::Hash::hash_slice(slice, &mut hasher);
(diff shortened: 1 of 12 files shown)


hw-model/src/model_emulated.rs
```diff
@@ -105,6 +105,16 @@ impl ModelEmulated {
     }
 }
 
+impl ModelEmulated {
+    /// Read a slice of the DCCM (Data Closely Coupled Memory) by absolute address.
+    /// The `addr` and `len` are in the CPU address space (starting at 0x5000_0000).
+    pub fn dccm_read(&self, addr: u32, len: usize) -> &[u8] {
```
zhalvorsen: You could add this function to the `HwModel` trait with a default implementation that is unsupported

zhalvorsen: This API might be a bit more intuitive with an offset parameter instead of an address. It could also help if we decide to implement it for other hardware models. What do you think?

image/types/src/lib.rs
```diff
@@ -23,9 +23,34 @@ use core::ops::Range;
...
+/// compiler from eliding the zeroing as a dead store.
+pub trait ZeroizeWithByteScrub: FromZeros + IntoBytes + Sized {
+    fn zeroize_scrub(&mut self) {
+        // SAFETY: `Self: IntoBytes` guarantees a contiguous byte representation
+        // with no uninitialized padding. `Self: FromZeros` guarantees all-zeros
+        // is a valid bit pattern. `self` is a valid mutable reference.
+        let bytes = unsafe {
+            core::slice::from_raw_parts_mut(self as *mut Self as *mut u8, size_of::<Self>())
+        };
+        bytes.zeroize();
```
zhalvorsen, line 50: You should be able to avoid the unsafe here using zerocopy:


```suggestion
        self.as_mut_bytes().zeroize();
```

hw-model/src/model_emulated.rs
ternst-nv: Good suggestion, will do.

image/types/src/lib.rs
ternst-nv: Unfortunately `as_mut_bytes` is only available for `FromBytes` and not `TryFromBytes` since it would be possible to put an invalid structure into the byte slice.

The `PersistentData` only implements `TryFromBytes`, as there are at least a couple enumerations which are not exhaustive.  I did run down the various elements of `PersistentData` though and they can be safely represented as 0s, so the `FromZeros` is a safe trait to decorate them with.

zhalvorsen: Oh, gotcha. I didn't realize `as_mut_bytes` had that constraint

(leaving this link for quick future reference https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#method.as_mut_bytes)

zhalvorsen (approved):

image/types/src/lib.rs
zhalvorsen: Could you use the `FromZeros` trait:

*self = Self::new_zeroed()

zhalvorsen: I wonder if that would blow up the stack though

ternst-nv: I thought about that, but my concern was the code being elided out if the compiler detects that the value is never used after the zeroization.  `Zeroize` has protections against that, but I don't think `new_zeroed` does.

zhalvorsen: You could wrap it in a `black_box`

ternst-nv: I'd be a little nervous about that, since the docs say it is only a hint and doesn't guarantee the compiler won't optimize things out: https://doc.rust-lang.org/beta/std/hint/fn.black_box.html.  Granted we now have a unit test so would detect if it wasn't doing it (for at least the emulator build), but my preference would be to rely on zeroize.

zhalvorsen: Okay, SGTM

PR #3673, opened 2026-04-30
[fix] Fix mailbox API size constants for subsystem mode

mhatrevi:


api/src/mailbox.rs
```diff
@@ -5745,12 +5749,10 @@ pub fn mbox_write_fifo(
     mbox: &mbox::RegisterBlock<impl MmioMut>,
     buf: &[u8],
 ) -> core::result::Result<(), CaliptraApiError> {
-    const MAILBOX_SIZE: u32 = 256 * 1024;
-
     let Ok(input_len) = u32::try_from(buf.len()) else {
         return Err(CaliptraApiError::BufferTooLargeForMailbox);
     };
-    if input_len > MAILBOX_SIZE {
+    if input_len > PASSIVE_MAILBOX_SIZE_LIMIT as u32 {
```
zhalvorsen: Should this check which configuration this is in so it doesn't overflow in subsystem mode?

rom/dev/tests/rom_integration_tests/test_cm_sha.rs
```diff
@@ -214,11 +211,7 @@ fn test_cm_sha_full_mailbox_all_0xff() {
 
     // Create a full mailbox payload filled with 0xff (12 bytes overhead)
 
-    let size = if cfg!(feature = "fpga_subsystem") {
-        (MBOX_SIZE_SUBSYSTEM - 12) as usize
-    } else {
-        MAX_CM_SHA_INPUT_SIZE
-    };
+    let size = MAX_CM_SHA_INPUT_SIZE;
```
zhalvorsen, line 215: All of the tests are only testing up to the Subsystem max size. Should we add a separate test run in passive mode for the full 256KB SHA command?

zhalvorsen (dismissed):

api/src/mailbox.rs
mhatrevi: Makes sense; I modified few things to address this.

rom/dev/tests/rom_integration_tests/test_cm_sha.rs
mhatrevi: Added.

zhalvorsen (approved):

PR #3679, opened 2026-05-01
Add support for `SIGN_WITH_EXPORTED_MLDSA`

### LINKED ISSUE #3681: Add ML-DSA support for signing with an exported CDI
This was missed while implementing ML-DSA for DPE.

zhalvorsen:
This adds similar functionality to `SIGN_WITH_EXPORTED_ECDSA`, but with ML-DSA-87. This adds similar tests to the legacy version as well.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -711,6 +711,7 @@ pub enum MailboxReq {
     VerifyAuthManifest(VerifyAuthManifestReq),
     AuthorizeAndStash(AuthorizeAndStashReq),
     SignWithExportedEcdsa(SignWithExportedEcdsaReq),
+    SignWithExportedMldsa(SignWithExportedMldsaReq),
     RevokeExportedCdiHandle(RevokeExportedCdiHandleReq),
     GetImageInfo(GetImageInfoReq),
     CmStatus(MailboxReqHeader),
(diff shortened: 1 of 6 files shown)


api/src/mailbox.rs
```diff
@@ -2371,28 +2371,57 @@ pub struct RevokeExportedCdiHandleResp {
...
+}
+
 // SIGN_WITH_EXPORTED_MLDSA
 #[repr(C)]
 #[derive(Debug, IntoBytes, FromBytes, KnownLayout, Immutable, PartialEq, Eq)]
 pub struct SignWithExportedMldsaReq {
     pub hdr: MailboxReqHeader,
     pub exported_cdi_handle: [u8; Self::EXPORTED_CDI_MAX_SIZE],
-    pub tbs: [u8; Self::MAX_DIGEST_SIZE],
+    pub sign_type: u32,
```
clundin25, line 2519: Should be `MldsaSignType`?

runtime/src/sign_with_exported_mldsa.rs
```diff
@@ -1,22 +1,126 @@
...
+    /// * `env` - DPE environment containing Crypto and Platform implementations
+    /// * `data` - The data to be signed
+    /// * `exported_cdi_handle` - A handle from DPE that is exchanged for a CDI.
+    #[cfg_attr(feature = "cfi", cfi_impl_fn)]
+    fn mldsa_sign(
+        env: &mut DpeCrypto,
+        data: &SignData,
+        exported_cdi_handle: &[u8; MAX_EXPORTED_CDI_SIZE],
+    ) -> CaliptraResult<(Signature, PubKey)> {
+        let profile_desc = b"Exported ML-DSA";
```
clundin25: Can we elevate this to a constant? Makes it easier to read IMO

clundin25 (dismissed):

api/src/mailbox.rs
zhalvorsen: From what I've seen, we don't use enums directly in mailbox commands/responses. We usually use u32's and then convert them in the handler. Are we already using enums somewhere?

runtime/src/sign_with_exported_mldsa.rs
zhalvorsen: Good call, done!

api/src/mailbox.rs
clundin25: Hmm that's a fair point, using it directly may also require `TryFromBytes` instead of `FromBytes` so I think I am wrong

clundin25 (dismissed):

clundin25 (approved):

PR #3682, opened 2026-05-04
fix(emulator): correct flow status handling in soc_reg

v-ramalleti:
The ready_for_fuses bit is driven by hardware (Boot FSM) and should not be overwritten
by software writes in the emulator.

Previously, writes to the SOC register would overwrite all bits, including
ready_for_fuses. This resulted in incorrect behavior where firmware could
observe the bit being cleared unintentionally.

This change preserves the ready_for_fuses bit during register writes by:
- retaining the existing value of the bit from the current register state
- applying all other bits from the incoming write value

This aligns emulator behavior with RTL, where ready_for_fuses is hardware-driven.


zhalvorsen (approved):

swenson (commented):
LGTM

PR #3685, opened 2026-05-05
dpe: Integrate size/encode optimization

ternst-nv:
Integrate the DPE patch to unify the get_size/encode functionality resulting in a size savings of of ~1.2Kb of imem.


zhalvorsen (commented):
I don't think anything has significantly changed in the golang verifier, but can you run the xtask flow to keep them in sync?

```
cargo xtask update-dpe
```

zhalvorsen (approved):

PR #3701, opened 2026-05-07
[OCP-LOCK]: Update OCP LOCK key ladder to use DOE

clundin25:
- Updated ROM to use DOE for HEK seed deobfuscation to comply with HW spec.
- Implemented two-step HKDF for HEK derivation because DOE secrets can only be used as data.
- Updated tests in `derive.rs` to match new ladder.

Resolves chipsalliance/caliptra-sw#3584

NOTE:

- The OCP LOCK flow from earlier IS STILL SECURE.
- This ROM creates a different HEK to previous versions.

### DIFF
--- a/rom/dev/src/flow/cold_reset/ocp_lock.rs
+++ b/rom/dev/src/flow/cold_reset/ocp_lock.rs
@@ -11,17 +11,15 @@ use caliptra_common::{
     crypto::Crypto,
     keyids::{
         ocp_lock::{KEY_ID_HEK, KEY_ID_MDK},
-        KEY_ID_ROM_FMC_CDI,
+        KEY_ID_HEK_SEED, KEY_ID_ROM_FMC_CDI, KEY_ID_TMP,
     },
 };
 use caliptra_drivers::{
-    Array4x8, CaliptraResult, HekSeedState, HmacMode, KeyUsage, KeyVault, Lifecycle,
-    PersistentData, SocIfc,
+    Array4x8, CaliptraResult, HekSeedState, HmacData, HmacKey, HmacMode, KeyReadArgs, KeyUsage,
+    KeyVault, Lifecycle, PersistentData, SocIfc,
 };
 
 use crate::rom_env::RomEnv;
-
-use zerocopy::IntoBytes;
 
 /// Run the OCP LOCK Cold Reset flow
 ///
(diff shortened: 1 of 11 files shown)


clundin25:
NOTE: We should check this on FPGA after https://github.com/chipsalliance/caliptra-sw/pull/3640, before we cut a release (or merge)

rom/dev/src/flow/cold_reset/ocp_lock.rs
```diff
@@ -48,17 +47,44 @@ pub fn ocp_lock_cold_reset_flow(env: &mut RomEnv) -> CaliptraResult<()> {
...
-        b"ocp_lock_hek",
-        Some(hek_seed.as_bytes()),
-        KEY_ID_HEK,
-        HmacMode::Hmac512,
-        KeyUsage::default().set_hmac_key_en(),
-    )?;
+    // Decrypt HEK seed via DOE
+    env.doe.decrypt_hek_seed(&DOE_IV, KEY_ID_HEK_SEED)?;
+
+    // HKDF-Extract — HMAC(key=CDI, data=HEK_seed_in_KV)
```
bluegate010, line 49: One wrinkle here is that ROM already has a [flow](https://github.com/chipsalliance/caliptra-sw/blob/d32df9f15ff53c9562fb8a9fdc128e08ecbb7a08/rom/dev/src/flow/cold_reset/ldev_id.rs#L132) where it needs to mix two KV values together. In that flow it calculates: LDevID_CDI = HMAC(HMAC(IDevID_CDI, "ldevid_cdi"), field_entropy)).

I believe we did the math and figured out how to justify this from a FIPS perspective. We should probably be consistent.

clundin25: I mirrored the implementation for stable owner keys, which has this statement ["The DOE hardware restricts the HEK seed to HMAC_BLOCK only"](https://github.com/chipsalliance/caliptra-sw/blob/main/rom/dev/src/flow/cold_reset/idev_id.rs#L208-L209)

clundin25: Here is the HW rule: 

* https://github.com/chipsalliance/caliptra-rtl/blob/a687e263ab4550b40ab428dee1494a07a9add7d5/src/keyvault/rtl/kv_defines_pkg.sv#L64
* https://github.com/chipsalliance/caliptra-rtl/blob/a687e263ab4550b40ab428dee1494a07a9add7d5/src/doe/rtl/doe_fsm.sv#L268

bluegate010: Ah very good then.

bluegate010 (dismissed):

zhalvorsen (dismissed):

RaunakGu:
hi @clundin25,

Can you please merge this PR, if no pending activity

clundin25:
@RaunakGu this is ready, just needs to be approved.

zhalvorsen (approved):

PR #3702, opened 2026-05-08
ROM: verify firmware after loading to ICCM

ArthurHeymans:
Add a post-copy digest check for firmware loaded into ICCM. Cold boot now
verifies the loaded FMC and runtime images against the firmware manifest
stored in DCCM before completing firmware processing. Update reset verifies
the loaded runtime image before reporting success.

This is a bug fix and defense-in-depth improvement: the image was already
verified before copying, but this also detects corruption during transfer
into ICCM.

Related: chipsalliance/caliptra-sw#817

### DIFF
--- a/rom/dev/src/flow/update_reset.rs
+++ b/rom/dev/src/flow/update_reset.rs
@@ -172,6 +172,9 @@ impl UpdateResetFlow {
                 &mut env.dma,
                 staging_addr,
             )?;
+            crate::flow::loaded_image::verify_runtime(&manifest, &mut env.sha2_512_384)?;
+            // Complete only after the loaded ICCM contents have been verified.
+            recv_txn.complete(true)?;
             Ok(())
         };
         if let Err(e) = process_txn() {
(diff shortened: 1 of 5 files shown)


rom/dev/src/flow/update_reset.rs
```diff
@@ -343,6 +344,52 @@ impl UpdateResetFlow {
...
     }
 
+    /// Verify the runtime image after it has been loaded into ICCM.
+    ///
+    /// # Arguments
+    ///
+    /// * `manifest`       - Manifest
+    /// * `sha2_512_384`   - SHA2-512/384 engine
+    #[cfg_attr(feature = "cfi", cfi_impl_fn)]
+    fn verify_loaded_runtime(
```
zhalvorsen: Can we put these functions somewhere common so they aren't copy pasted in both?

zhalvorsen (dismissed):

zhalvorsen (dismissed):

swenson (approved):
LGTM

mhatrevi (approved):

PR #3711, opened 2026-05-12
hw-model: fix UTF-8 search panic and expose setters for cold_reset state

swenson:
Two small fixes for tests:

* output.rs: When the buffered UART output contains multi-byte UTF-8 characters, `process_new_data` could compute `search_pos = output.len() - term.len()` and land inside a multi-byte codepoint. The subsequent `self.output[self.search_pos..]` slice would panic. Floor the index down to a UTF-8 character boundary before storing it. Added regression tests.

* model_fpga_subsystem.rs: With #3413, `cold_reset()` now does a full AXI reset and re-provisions OTP from `saved_lc_state` and `otp_init`. Add two pub setters so tests can adjust the state cold_reset will restore: `set_saved_lc_state` (for tests that transition LC over JTAG and then cold-reset) and `set_otp_init` (for tests that need particular OTP contents to survive a cold reset).

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -1705,6 +1705,22 @@ impl ModelFpgaSubsystem {
         unsafe { core::slice::from_raw_parts_mut(self.otp_mem_backdoor, OTP_SIZE) }
     }
 
+    /// Override the lifecycle controller state that will be provisioned into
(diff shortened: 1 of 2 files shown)


mhatrevi (approved):

zhalvorsen (approved):

swenson:
Thanks!

PR #3720, opened 2026-05-13
runtime: add ACTIVATE_FIRMWARE INITIAL_ACTIVATE flag

swenson:
Lets MCU ROM finalize the encrypted-boot flow: after MCU ROM decrypts firmware in MCU SRAM via CM_AES_GCM_DECRYPT_DMA, it sends ACTIVATE_FIRMWARE with this flag so Caliptra publishes FW_EXEC_CTRL[MCU] and sets RESET_REASON.FwBootUpdReset, without performing the hitless-update reset/reload/verify dance. Without this MCI's BOOT_RST_MCU state holds MCU in reset forever after the post-decrypt warm reset.

Gated on BootMode::EncryptedFirmware (set by ROM only on RI_DOWNLOAD_ENCRYPTED_FIRMWARE) and FW_EXEC_CTRL[MCU] == 0 to prevent a real hitless update from masquerading as initial activation.

The new field is appended; older callers that omit it are treated as flags=0 (existing hitless-update behavior).

Tested locally with an appropriate MCU ROM change on FPGA to ensure that this new flow works. (Will send a separate PR to update the MCU commit and re-enable he encrypted firmware test when MCU ROM adds this new flag.)

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -1185,6 +1185,8 @@ pub struct ActivateFirmwareReq {
     pub fw_id_count: u32,
     pub fw_ids: [u32; ActivateFirmwareReq::MAX_FW_ID_COUNT],
     pub mcu_fw_image_size: u32,
+    /// Optional flags. See [`ActivateFirmwareFlags`].
+    pub flags: u32,
 }
 impl Request for ActivateFirmwareReq {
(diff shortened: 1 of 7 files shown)


mhatrevi (approved):

swenson:
Thanks!

PR #3727, opened 2026-05-15
[fix] Run ML-KEM KATs before driver use

mhatrevi:
Make MlKem1024 construction fallible so the ML-KEM KAT runs before any ML-KEM operation, and propagate the result through HPKE, runtime mailbox, and test firmware call sites.

### DIFF
--- a/drivers/src/hpke/kem/hybrid.rs
+++ b/drivers/src/hpke/kem/hybrid.rs
@@ -215,7 +215,7 @@ impl
 
         let (pq_enc, pq_shared_secret) = {
             let (pq_seed, _) = Self::expand_seed(ctx.sha, &self.seed)?;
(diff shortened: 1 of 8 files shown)


drivers/src/hpke/kem/hybrid.rs
```diff
@@ -215,7 +215,7 @@ impl
 
         let (pq_enc, pq_shared_secret) = {
             let (pq_seed, _) = Self::expand_seed(ctx.sha, &self.seed)?;
-            let mut ml_kem_driver = MlKem1024::new(ctx.abr_reg);
+            let mut ml_kem_driver = MlKem1024::new(ctx.abr_reg)?;
```
swenson, line 218: Won't this run them before *every* use?

PR #3728, opened 2026-05-15
Run AES KATs from central harness

mhatrevi:
Add Aes::run_kats() for non-GCM AES KATs and keep ROM GCM/CMAC-KDF KAT execution in the central KAT harness.

Make AesGcm::new() only seed AES entropy so ROM startup avoids duplicate GCM and CMAC-KDF KAT execution.


zhalvorsen (approved):

mhatrevi:
No longer needed.

PR #3729, opened 2026-05-15
runtime: Add MCU RT to DPE with SVN for attestation

parvathib:
- Store SoC manifest SVN in PersistentData during set_auth_manifest
- Pass authenticated SVN in AuthorizeAndStashReq during recovery boot
- Add UPDATE_EXISTING flag to AUTHORIZE_AND_STASH for hitless update
- Add update_measurement() using RECURSIVE DeriveContext to extend MCU RT DPE context TCI in place without allocating new context slots
- MCU RT hitless update (activate_firmware) now uses UPDATE_EXISTING instead of SKIP_STASH to keep DPE attestation current
- Add integration tests verifying DPE state is correctly updated

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -57,6 +57,8 @@ pub const CMB_AES_KEY_SHARE_SIZE: u32 = 32;
 pub const DOT_OWNER_PK_HASH_SIZE: u32 = 13 * 4;
 pub const CLEARED_NON_FATAL_FW_ERROR_SIZE: u32 = 4;
 pub const MCU_FIRMWARE_LOADED_SIZE: u32 = 4;
+#[cfg(feature = "runtime")]
+pub const SOC_MANIFEST_SVN_SIZE: u32 = 4;
 pub const _DPE_PL_CONTEXT_LIMITS_WITH_PAD_SIZE: u32 = 4; // u8 + u8 + 2 bytes padding
 
 #[cfg(feature = "runtime")]
@@ -94,6 +96,15 @@ pub type StashMeasurementArray = [MeasurementLogEntry; MEASUREMENT_MAX_COUNT];
 #[cfg(feature = "runtime")]
 pub type AuthManifestImageMetadataList =
     [AuthManifestImageMetadata; AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT];
+
+#[cfg(feature = "runtime")]
+#[derive(Clone, Copy, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize, Default)]
+#[repr(C)]
+pub struct CaliptraManagedDpeContextIndices {
+    pub cciv: u8,
+    pub mcu_rt: u8,
+    pub reserved: [u8; 2],
+}
 
 #[derive(Clone, Immutable, IntoBytes, KnownLayout, TryFromBytes, Zeroize)]
 #[repr(C)]
@@ -381,6 +392,14 @@ pub struct PersistentData {
     /// changing the frozen ROM binary. FMC writes this during handoff,
     /// runtime reads it for MLDSA signing operations.
     pub rt_mldsa_keypair_seed_kv_hdl: HandOffDataHandle,
+
+    /// SoC manifest SVN, stored after auth manifest validation.
+    /// Used as the MCU RT current_svn when creating the MCU RT DPE context.
+    #[cfg(feature = "runtime")]
+    pub soc_manifest_svn: u32,
+
+    #[cfg(feature = "runtime")]
+    pub caliptra_managed_dpe_context_indices: CaliptraManagedDpeContextIndices,
 }
 
 impl PersistentData {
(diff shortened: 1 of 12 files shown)


mhatrevi (changes requested):
This PR needs to be split into two parts:
1. Adding "update DPE context" feature (Existing draft PR: https://github.com/chipsalliance/caliptra-sw/pull/3587)
2. Adding MCU RT measurement to DPE.

drivers/src/persistent.rs
```diff
@@ -373,6 +374,10 @@ pub struct PersistentData {
 
     pub mcu_firmware_loaded: u32,
 
+    /// SoC manifest SVN, stored after auth manifest validation.
+    /// Used as the MCU RT current_svn when creating the MCU RT DPE context.
+    pub soc_manifest_svn: u32,
```
zhalvorsen: It would probably be good to append this at the end of the struct to make hitless updates compatible with this version

runtime/src/drivers.rs
```diff
@@ -287,49 +290,72 @@ impl Drivers {
...
-    /// # Returns
-    ///
-    /// * `usize` - Index containing the CCIV DPE context
     #[inline(always)]
-    pub fn get_dpe_cciv_context_idx(dpe: &caliptra_dpe::State) -> CaliptraResult<usize> {
-        // Find the root context index using your existing helper
-        let root_idx = Self::get_dpe_root_context_idx(dpe)? as u8;
-
-        let cciv_type = u32::from_be_bytes(*b"CCIV");
+    pub fn get_dpe_context_idx_by_tci_type(
```
zhalvorsen, line 298: I'm worried about the non-deterministic nature of looking up by TCI type. I'm worried about a pathway that could block hitless updates by creating a context with specific features.

What if we looked up the index at creation time instead? We could pretty easily just look for the index of the default handle then cache the indices in persistent data. Then use the index directly when we go to extend the journey measurement.

runtime/tests/runtime_integration_tests/test_recovery_flow.rs
```diff
@@ -106,3 +106,54 @@ fn test_mcu_fw_bad_signature() {
         30_000_000,
     );
 }
+
+#[cfg_attr(
+    any(
+        feature = "verilator",
+        feature = "fpga_realtime",
+        feature = "fpga_subsystem"
```
zhalvorsen: Is there a reason this wouldn't work on `fpga_subsystem`?

runtime/tests/runtime_integration_tests/test_recovery_flow.rs
```diff
@@ -106,3 +106,54 @@ fn test_mcu_fw_bad_signature() {
...
+        ..Default::default()
+    });
+    args.soc_manifest = Some(soc_manifest);
+    args.mcu_fw_image = Some(&mcu_fw);
+    args.soc_manifest_svn = Some(5);
+    args.soc_manifest_max_svn = Some(127);
+    let mut model = run_rt_test(args);
+    model.step_until_boot_status(RT_READY_FOR_COMMANDS, true);
+    // Recovery boot completed successfully with SVN=5 in the SoC manifest.
+    // The MCU RT DPE context was created with the SoC manifest SVN
```
zhalvorsen: Can we pull the DPE certificate to make sure the SVN in the certificate also says 5?

runtime/src/drivers.rs
parvathib: Done. Added a cached index struct in  `PersistentData`  for Caliptra-managed contexts. `CCIV` and `MCFW` indices are cached when created, repopulated on update reset after DPE validation, and the backdoor update path now uses the cached index instead of doing a TCI-type lookup

runtime/tests/runtime_integration_tests/test_recovery_flow.rs
parvathib: no specific reason. I removed the `fpga_subsystem` ignore so the test can run there too.

drivers/src/persistent.rs
parvathib: Done. I kept the existing fields intact and appended the new SoC manifest SVN and Caliptra-managed DPE context index state at the end of `PersistentData`.

For hitless update compatibility, the new context-index cache is repopulated after DPE validation on update reset, so it does not rely on older runtime images having initialized the appended fields.

runtime/tests/runtime_integration_tests/test_recovery_flow.rs
parvathib: Done. The test now pulls a DPE certificate, parses the MultiTcbInfo extension, finds the MCFW  TCB info, and verifies its SVN is 5 .

drivers/src/persistent.rs
```diff
@@ -381,6 +390,12 @@ pub struct PersistentData {
     /// changing the frozen ROM binary. FMC writes this during handoff,
     /// runtime reads it for MLDSA signing operations.
     pub rt_mldsa_keypair_seed_kv_hdl: HandOffDataHandle,
+
+    /// SoC manifest SVN, stored after auth manifest validation.
+    /// Used as the MCU RT current_svn when creating the MCU RT DPE context.
+    pub soc_manifest_svn: u32,
```
mhatrevi, line 399: Both these fields are Runtime only so should be under #[cfg(feature = "runtime")]

mhatrevi (changes requested):

runtime/src/drivers.rs
```diff
@@ -287,49 +294,139 @@ impl Drivers {
...
-                    return Err(CaliptraError::RUNTIME_CCIV_CONTEXT_NOT_FOUND);
-                }
-                Ok(idx)
-            }
-            None => Err(CaliptraError::RUNTIME_CCIV_CONTEXT_NOT_FOUND),
+        found_idx.ok_or(CaliptraError::RUNTIME_DPE_CONTEXT_NOT_FOUND)
+    }
+
+    #[inline(always)]
+    fn get_caliptra_managed_dpe_context_index(&self, tci_type: u32) -> CaliptraResult<usize> {
```
zhalvorsen, line 324: What do you think having an enum here instead of the u32 tci_type? It should make the caller more intuitive and it wouldn't need any failure logic.

```Rust
enum CaliptraManagedContexts {
    CaliptraConfiguredInitializationValues,
    McuRt,
}

   fn get_caliptra_managed_dpe_context_index(&self, context: CaliptraManagedContexts) -> CaliptraResult<usize> {
        let indices = self
            .persistent_data
            .get()
            .caliptra_managed_dpe_context_indices;
        match tci_type {
            CaliptraManagedContexts::CaliptraConfiguredInitializationValues => indices.cciv,
            CaliptraManagedContexts::McuRt => indices.mcu_rt,
        }
    }
```

runtime/src/drivers.rs
zhalvorsen, line 232: Could you set all of the indices (including the reserved indices) in persistent data to `0xff` at the beginning of the cold reset flow so if by some chance something is not initialized correctly or we hitlessly update to a firmware that adds another Caliptra managed context we don't accidentally wield the 0th index context

runtime/src/drivers.rs
```diff
@@ -412,29 +509,34 @@ impl Drivers {
...
+        if cciv_ctx.state == ContextState::Inactive
+            || cciv_ctx.context_type != ContextType::Normal
+            || cciv_ctx.tci.tci_type != Self::CCIV_TCI_TYPE
+        {
+            return Err(CaliptraError::RUNTIME_DPE_CONTEXT_NOT_FOUND);
+        }
+        let prev_current: [u8; 48] = cciv_ctx.tci.tci_current.0;
         if prev_current == initialization_values_hash {
             return Ok(());
         }
```
zhalvorsen, line 533: It looks like if we did the check for measurement change inside `update_caliptra_managed_measurement` we could remove all of this logic. Is that true? It looks like that function is already getting the index, and checking the context configuration internally. If we can, then we also wouldn't need to duplicate this for other measurements. I imagine, we would also want to do this check for the MCU RT?

drivers/src/persistent.rs
parvathib: done

mhatrevi (approved):

PR #3730, opened 2026-05-15
Add funcitons and tests to run MLDSA-87 NIST ACVP vectors

ravithakurAMD:
Added MLDSA ACVP test code


drivers/src/mldsa87.rs
```diff
@@ -157,6 +157,29 @@ impl Mldsa87 {
...
+    /// # Arguments
+    ///
+    /// * `seed` - Either an array of 4x8 bytes or a key vault key to use as seed.
+    /// * `trng` - TRNG driver instance.
+    /// * `priv_key_out` - Optional output parameter to store the private key.
+    ///
+    /// # Returns
+    ///
+    /// * `Mldsa87PubKey` - Generated MLDSA-87 Public Key
+    pub fn key_pair_no_pct(
```
zhalvorsen: We should add it to this function as well

drivers/src/mldsa87.rs
zhalvorsen, line 1: Is there a reason you changed the file permissions on these?

ravithakurAMD: That was not intended :( I'll address it

drivers/test-fw/src/bin/mldsa87_tests.rs
```diff
@@ -936,6 +936,64 @@ fn test_verify_failure() {
...
+            &msg,
+            &sign_rnd,
+            &mut trng,
+        )
+        .unwrap();
+
+    assert_eq!(signature, Mldsa87Signature::from(MLDSA_SIGN));
+}
+
+//// Diagnostic test: exercises key_pair_no_pct() with a direct Array4x8 seed against
```
zhalvorsen: Do we want to keep this commented out?

zhalvorsen (dismissed):

drivers/test-fw/src/bin/mldsa87_tests.rs
ravithakurAMD: Yeah it passes on its own, but for some reason when test_sign_caller_provided_private_key_var_msg or test_keygen_caller_provided_seed_var_msg are included in the test suite, this test fails. I will debug it, but I did not want that to hold this PR back

nquarton:
Increasing the stack size for these driver tests should allow you to re-enable that commented out test. It actually looks like it was increased in main but not 2.0 already. I would just make it [0x10000 as it is in main](https://github.com/chipsalliance/caliptra-sw/blob/d89f997d833faf84f30e1a0f4f525a3dcd4fe6df/test-harness/scripts/rom.ld#L29)

nquarton (approved):

zhalvorsen (approved):

PR #3749, opened 2026-05-20
compiler: Update to rust 1.95

ternst-nv:
Update the compiler version to 1.95 to support a unified CFI and DPE library with the 2.x Caliptra FW.  This will enable us to utilize existing MLDSA DPE and Cert implementations within Caliptra 1.x.

This required a number of fixes:
1. Unroll a variable initialization to avoid memcpy prior to CFI init
2. Reduce ROM with log size to fit within the 64Kb (deployable ROM without logs remains significantly under the 64Kb budget)
3. Manually implement mutable Packet to avoid duplicate 14Kb copies of the size.  This codified a stack return-value optimization that was previously done by the 1.70 compiler
4. Patch some undefined behavior which did not compile
5. Fixed formatting issues
6. Fixed clippy issues.


drivers/src/trng_ext.rs
```diff
@@ -26,8 +27,30 @@ impl TrngExt {
                 return Err(CaliptraError::DRIVER_TRNG_EXT_TIMEOUT);
             }
         }
-        let result = Array4x12::read_from_reg(regs.cptra_trng_data());
+
+        // Manually unroll the generation of rng data to avoid a memcpy.
```
swenson, line 31: suggestion: should we change the implementation of `Array4x12::read_from_reg` rather than adding this special case?

And since these are registers, I think we should be using read_volatile and write_volatile to add extra protections against memcpy (which might result in the wrong size read/writes, which might matter).

swenson (dismissed):
LGTM, one suggestion

drivers/src/trng_ext.rs
ternst-nv: Good call on the `volitile_*` I'll make that change.  In regard to changing `read_from_reg`, I think that is more a question of whether we want to disable memcpy in general? It looks like this is the only place it is getting invoked prior to cfi init, so I'm inclined to leave as is, unless we think it is a general concern.

swenson: `memcpy` in general is okay I think, though I'd be open to exploring disabling it altogether if we could, since I don't know that it buys us much on our relatively simple RISC-V core.

swenson (dismissed):
LGTM

zhalvorsen (approved):

PR #3754, opened 2026-05-21
[fix] Nightly OCP LOCK tests for 2.1.0 ROM

zhalvorsen:
There are tests that do the expected derivations and check the same inputs to Caliptra result in the expected MEKs. This ensures we are deriving them the way we have planned.

In 4de7892 HEK derivation was changed to use the DOE. This affected the test MEK deriver. The nightly tests have started failing because the derivations using the 2.1 ROM don't match the new derivation scheme. This change checks to see which way it is supposed to derive them depending on which ROM it is using so the nightly tests will pass with both ROM versions.

### DIFF
--- a/test/src/derive.rs
+++ b/test/src/derive.rs
@@ -322,15 +322,29 @@ impl OcpLockKeyLadderBuilder {
     }
 
     pub fn add_hek(self) -> Self {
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM

mhatrevi (approved):

PR #3764, opened 2026-05-21
mldsa sign without verify

ravithakurAMD:
Added function to do MLDSA sign without verify, for acvp test


drivers/src/mldsa87.rs
```diff
@@ -470,6 +470,61 @@ impl<'a> Mldsa87<'a> {
         }
     }
 
+    /// Sign a variable-length message with the specified private key, skipping the
+    /// post-sign verification step. Intended for ACVP testing only, where the public
+    /// key is not available and the anti-glitch check cannot be performed.
+    pub fn sign_var_no_verify(
```
zhalvorsen, line 477: Can we put this behind a feature flag so it can't be used in regular firmware by accident?

ravithakurAMD: Yes, I'll make the edit now.

zhalvorsen (approved):

nquarton (approved):

PR #3767, opened 2026-05-21
Proposal: WIP

mhatrevi:


DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,185 @@
...
+Active remains Active
+Retired remains Retired
+Inactive contexts are never updated
+```
+
+## Effect on Future CDI / CertifyKey
+
+No descendant recomputation is required.
+
+DPE computes the measurement hash later when a caller requests operations such as `CertifyKey` or `Sign`, by walking the context ancestry.
```
zhalvorsen, line 183: nit: CDI here instead of hash?

```suggestion
DPE computes the context CDI later when a caller requests operations such as `CertifyKey` or `Sign`, by walking the context ancestry.
```

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,185 @@
+# DPE Context Update Proposal
+
+## Summary
+
+This proposal introduces three related changes:
```
zhalvorsen, line 5: There are two contexts I think we should handle differently.

* `RTMR` that keeps track of the Caliptra RT measurement
* `MCFW` that keeps track of the MCU FW measurement

Because both of these are owned by Caliptra Core, they should only be updated by it. We should fail if either of those measurements are being updated through `AUTHORIZE_AND_STASH`. During cold boot we should also cache the index for these in the context array so Caliptra can easily find them without needing to do a lookup by measurement ID with possible clashes

zhalvorsen (approved):
This looks really good. I think you covered everything we talked about yesterday really well. I had a one thought while reading this and then a small nit.

PR #3778, opened 2026-05-22
cfi: unified cfi

ternst-nv:
Utilize the CFI library for all CFI operations instead of including both a git and local copy.  This prepares for landing a unified DPE environment and saves ~2Kb of duplicated instructions.


Cargo.toml
zhalvorsen: We should turn CFI on for these as applicable as well

ternst-nv: Ahh, good catch.  Will do.

ternst-nv: Pushed a fix with this, and took a look at main, we'll likely want to enable there as well.

zhalvorsen: Oh dang, I guess that similarly slipped through the cracks

zhalvorsen (dismissed):

zhalvorsen (approved):

PR #3784, opened 2026-05-26
pqc: mldsa x509

ternst-nv:
Implement support for MLDSA-87 x509 creation.  This includes a number of steps, including renaming the existing solution to support both, supporting a new openssl library for testing purposes, as well as altering the x509 generator to handle both ECDSA and MLDSA-87.  

The bulk of the generator and MLDSA-87 certificate changes were ported from the main branch supporting MLDSA-87 in hardware.

### DIFF
--- a/x509/build/build.rs
+++ b/x509/build/build.rs
@@ -56,7 +56,7 @@ fn gen_init_devid_csr(out_dir: &str) {
         .add_key_usage_ext(usage)
         .add_ueid_ext(&[0xFF; 17]);
     let template = bldr.tbs_template("Caliptra 1.0 IDevID");
-    CodeGen::gen_code("InitDevIdCsrTbs", template, out_dir);
+    CodeGen::gen_code("InitDevIdCsrTbsEcc384", template, out_dir);
 }
 
 #[cfg(feature = "generate_templates")]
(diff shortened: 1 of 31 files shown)


x509/build/build.rs
```diff
@@ -51,12 +51,20 @@ fn main() {
...
         .add_ueid_ext(&[0xFF; 17]);
     let template = bldr.tbs_template("Caliptra 1.0 IDevID");
-    CodeGen::gen_code("InitDevIdCsrTbs", template, out_dir);
+    CodeGen::gen_code("InitDevIdCsrTbsEcc384", template, out_dir);
+
+    let bldr = csr::CsrTemplateBuilder::<MlDsa87Algo>::new()
+        .add_basic_constraints_ext(true, 5)
+        .add_key_usage_ext(usage)
+        .add_ueid_ext(&[0xFF; 17]);
+    let template = bldr.tbs_template("Caliptra 1.0 MlDsa87 IDevID");
```
swenson: Question: are we actually adding an MLDSA87 IDevID cert as part of the upcoming work?

It feels like that IDevID is composed of a hardware Caliptra core DICE secret, which 1.x can't provide. I'm just not sure if it is disingenuous to call something not based on a core HW secret an "IDevID"?

swenson (dismissed):
LGTM, one question

x509/build/build.rs
zhalvorsen: What would you recommend calling it in that case? The SoC manager will need to provide the base entropy used to derive the ML-DSA identity keys via the mailbox. That entropy should be rooted in hardware ideally some form of secret fuses. I've seen some people refer to this as the PQDevID

zhalvorsen (dismissed):

x509/build/build.rs
ternst-nv: Makes sense, taking a look the RFC calls it the PQDevID, as Zach suggests. Would that name work for you @swenson?

swenson (approved):
LGTM

PR #3788, opened 2026-05-26
rom: set CSRNG entropy_src CONF.THRESHOLD_SCOPE to false

### LINKED ISSUE #3771: CSRNG needs to set THRESHOLD_SCOPE to false
So that health tests are triggered for individual lanes.

May require tweaking the tests since they are highly tuned for specific failures.

swenson:
Score the entropy_src Adaptive Proportion (and Markov) health checks
per RNG bus lane instead of across the aggregate sum, so a stuck or
biased lane is detected even when the other lanes compensate.

We also bump the default FIPS health-test window to 1024 clock cycles
(1024 bits per lane, 4096 bits aggregate) and derive the default ADAPTP
thresholds from that window: HI = 3*W/4, LO = W/4 (so 768 / 256 at the
new default, and they scale automatically with any SoC override).

The emulator's entropy_src model is updated to faithfully implement
THRESHOLD_SCOPE per entropy_src_adaptp_ht.sv, and to honor SoC writes
to the HEALTH_TEST_WINDOWS register (previously read-only) so the
boot-time tester, seed conditioner, and continuous-test simulator all
agree on the active window size.

We needed to regenerate the test data, so rather than adding another
test data file with unknown provenance, wrote a simple Rust binary to
regenerate these files automatically.

Hardcoded generate12 expected values that depend on the conditioned
seed are updated to reflect the larger window.

Fixes #3771


drivers/src/csrng.rs
```diff
@@ -449,19 +451,20 @@ fn read_entropy_configuration(
...
-    // This window value of 2048 comes from the OpenTitan documentation, since two noise
-    // channels are used. https://opentitan.org/book/hw/ip/entropy_src/index.html#description
-    const ADAPTP_WINDOW_SIZE_BITS: u32 = 2048;
-    const ADAPTP_DEFAULT_HI: u32 = 3 * (ADAPTP_WINDOW_SIZE_BITS / 4);
-    const ADAPTP_DEFAULT_LO: u32 = ADAPTP_WINDOW_SIZE_BITS / 4;
+    // Use 75% and 25% of the per-lane FIPS window size for the default HI and LO
+    // thresholds respectively. The OpenTitan-derived FIPS window of 2048 bits is
+    // distributed across 4 RNG lanes, so each lane is scored over 512 bits.
+    // https://opentitan.org/book/hw/ip/entropy_src/index.html#description
+    const ADAPTP_PER_LANE_WINDOW_BITS: u32 = 2048 / 4;
```
moidx: Should this take into account the `health_test_window` value as opposed to a hardcoded value?

swenson: Good call. I'll update this.

drivers/tests/csrng_test_data_gen/src/lib.rs
```diff
@@ -0,0 +1,212 @@
+// Licensed under the Apache-2.0 license
+
+//! Deterministic generator for the CSRNG ADAPTP integration-test windows.
+//!
+//! With `CONF.THRESHOLD_SCOPE = FALSE` the entropy_src ADAPTP health test
+//! scores each of the 4 RNG bus lanes independently. The integration tests
+//! at `drivers/tests/drivers_integration_tests/main.rs` use 8 hand-crafted
+//! 512-nibble windows where each lane carries a specific number of ones.
```
mhatrevi: This should be 1024.

mhatrevi (dismissed):

drivers/tests/csrng_test_data_gen/src/lib.rs
swenson: Oops!

drivers/src/csrng.rs
```diff
@@ -449,39 +455,43 @@ fn read_entropy_configuration(
...
-    // This window value of 2048 comes from the OpenTitan documentation, since two noise
-    // channels are used. https://opentitan.org/book/hw/ip/entropy_src/index.html#description
-    const ADAPTP_WINDOW_SIZE_BITS: u32 = 2048;
-    const ADAPTP_DEFAULT_HI: u32 = 3 * (ADAPTP_WINDOW_SIZE_BITS / 4);
-    const ADAPTP_DEFAULT_LO: u32 = ADAPTP_WINDOW_SIZE_BITS / 4;
+    // Each clock samples one bit on each of the 4 RNG lanes, so the per-lane
+    // window in bits is exactly `health_test_window` (the FIPS_WINDOW value
+    // written to the HEALTH_TEST_WINDOWS register). The defaults therefore
+    // scale automatically when the SoC overrides the window size.
+    // https://opentitan.org/book/hw/ip/entropy_src/index.html#description
```
vogelpi: @swenson , I recommend putting this link here instead https://opentitan.org/earlgrey_1.0.0/book/hw/ip/entropy_src/index.html#description

Because this version of the documentation corresponds to the version of the hardware IP included in Caliptra. The other link points to master which may change over time.

drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -931,10 +931,10 @@ fn test_csrng_repetition_count() {
         let rom = caliptra_builder::build_firmware_rom(test_fwid).unwrap();
 
         let itrng_nibbles = Box::new({
-            // The boot-time health testing requires two consecutive windows of 2048-bits to
-            // pass or fail health tests. So let's set up our windows to begin with the
-            // repeated bits followed by known good entropy bits.
-            const NUM_TEST_WINDOW_NIBBLES: usize = 2048 / 4;
+            // The boot-time health testing requires two consecutive 4096-bit
```
vogelpi: This should be "startup health testing" (rather than boot-time health testing). See e.g. https://opentitan.org/earlgrey_1.0.0/book/hw/ip/entropy_src/doc/theory_of_operation.html#startup-health-testing

The boot-time / bypass mode is something completely different. It can be used to generate an initial seed without conditioning and by just using 384 bits. This is not FIPS compliant. And AFAIU, you don't use this Caliptra (at least in csrng.rs, you configure the FIPS_ENABLE setting to TRUE).

drivers/src/csrng.rs
```diff
@@ -449,39 +455,43 @@ fn read_entropy_configuration(
...
-    // Use 75% and 25% of the 2048 bit FIPS window size for the default HI and LO thresholds
-    // respectively.
+    //  * In any window any single RNG lane has more than the HI threshold of 1's; or,
+    //  * Any RNG lane has less than the LO threshold of 1's.
+    // (Each of the 4 RNG bus lanes is scored individually because the ROM configures
+    // entropy_src with CONF.THRESHOLD_SCOPE = FALSE.)
+    // See section 4.4.2 of NIST.SP.800-90B for more information about this test.
+
+    // If the SoC doesn't override the ADAPTP thresholds, use 75% and 25% of the
+    // per-lane FIPS window size for the default HI and LO thresholds respectively.
```
vogelpi, line 469: You may want to say here that this should lead to a min-entropy value of roughly 0.6 (see Table 2 in Section 4.4.2 / Page 27 of NIST SP 800-90B).

drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -1017,61 +1017,64 @@ fn test_csrng_repetition_count() {
...
 
-    // 512 ones; 1536 zeros - should pass inclusive LO threshold.
+    // 256 ones per lane (1024 ones aggregate) - should pass inclusive LO threshold.
     test_csrng_with_nibbles(
         PASS,
         Box::new({
-            const WINDOW: [u8; 512] = *include_bytes!("test_data/csrng/512_ones_1536_zeros");
-            // Boot-time health checks require testing two 2048 bit windows.
+            const WINDOW: [u8; 1024] = *include_bytes!("test_data/csrng/1024_ones_3072_zeros");
+            // Boot-time health checks require testing two 4096 bit windows.
```
vogelpi: Boot-time -> startup

drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -1197,14 +1200,14 @@ fn test_csrng_runtime_health_failure() {
     )
     .unwrap();
 
-    // Provide just enough good entropy for boot health testing (two 2048-bit windows),
+    // Provide just enough good entropy for boot health testing (two 4096-bit windows),
     // then switch to bad entropy. The bad entropy (stuck at all 1s) will cause
     // the repetition count test to fail during subsequent generate operations.
     //
-    // Boot health testing: 2 windows × 512 nibbles = 1024 nibbles
+    // Boot health testing: 2 windows × 1024 nibbles = 2048 nibbles
```
vogelpi: Boot -> Startup

sw-emulator/lib/periph/src/csrng/health_test.rs
```diff
@@ -143,8 +166,14 @@ pub struct AdaptiveProportionTester {
     hi_threshold: u32,
     lo_failures: u32,
     hi_failures: u32,
-    num_ones_seen: u32,
+    // When true, the test scores the sum of ones across all RNG lanes
+    // against hi/lo thresholds (mubi4 strict-true value). When false,
+    // the test scores each lane individually: a failure is counted if
+    // any lane's count exceeds hi or falls below lo.
+    threshold_scope: bool,
```
vogelpi, line 171: My understanding is that you put the hardware reset value here but that software will configure the value false when configuring the hardware, correct?

vogelpi (approved):
Thanks @swenson for the PR. This looks good. I have some minor comments.

As I wrote in my email, some more things might need changing which are currently not yet touched by this PR. Those could be done in a follow-up.

sw-emulator/lib/periph/src/csrng/health_test.rs
swenson: Correct.

mhatrevi (approved):

PR #3790, opened 2026-05-27
drivers: Fix AES-GCM streaming GHASH save/restore bug

swenson:
The GHASH accumulator was incorrectly discarded on the first streaming update call when no AAD had been provided, causing the final tag to omit the authentication contribution of the first batch of data. Always save the GHASH state after processing blocks in aes_256_gcm_update.

Adds integration tests covering streaming AES-GCM with empty AAD across a range of plaintext sizes and chunk splits.


zhalvorsen (dismissed):

zhalvorsen (approved):

swenson:
Thanks!

PR #3791, opened 2026-05-27
Fix build errors for external consumers using hw-model as a dependency

chris-hinson:
Projects that source Caliptra's hw-model as an external dependency are currently failing to build. The upstream caliptra-sw workspace is unaffected because its checked-in Cargo.lock pins ml-dsa to exactly rc.7, and Cargo honors that lockfile when building the workspace as the root. 
 
 However, Cargo.lock is not propagated to library consumers when an external project depends on hw-model, it performs its own dependency resolution and can resolve ml-dsa to a newer version (e.g. rc.9) where SigningKey::from_expanded() no longer exists, producing a hard compile error.
 
 This PR fixes the breakage by bumping the ml-dsa pin to =0.1.0-rc.9 and migrating the one affected call site in sw-emulator/lib/periph/src/abr.rs to the replacement API: ExpandedSigningKey::from_expanded().   rc.9 introduces a dedicated ExpandedSigningKey\<P\> type for keys already in expanded form (the format stored in Caliptra's key vault), separating it from the compact-seed SigningKey\<P\>.  Signing behavior over ML-DSA-87 is identical this is purely an API migration to unblock downstream consumers.

  The version bump also transitively stabilizes several cryptographic dependencies (keccak, sha3, pkcs8, spki all graduate from RC to stable). Two additional transitive dependencies (cmov, ctutils) are pulled in by module-lattice 0.2.2.    
  
  Changes:
  - Cargo.toml: bump ml-dsa exact pin from =0.1.0-rc.7 to =0.1.0-rc.9
  - Cargo.lock: update transitive dependency tree accordingly                                                                                                                                                                                     - sw-emulator/lib/periph/src/abr.rs: replace deprecated SigningKey::from_expanded() with ExpandedSigningKey::from_expanded(), removing the #[allow(deprecated)] suppression


zhalvorsen (dismissed):

sw-emulator/lib/periph/src/abr.rs
```diff
@@ -765,14 +765,9 @@ impl Abr {
     }
 
     fn mldsa_sign_mu(&mut self, sk_bytes: [u8; SK_LEN]) -> Vec<u8> {
-        //note: we use the deprecated ::from_expanded function here because the ::decode function it replaces
-        //was not given a direct replacement, and the ::from_seed function reccomended as a replacement does
-        //not take the raw key bytes as an input.  If this function is removed in a future release, thought
-        //may need to be put into how to best replace it as it seems like creating a signing key from raw bytes
-        //is not something the maintainers are particularly interested in supporting.
         #[allow(deprecated)]
```
zhalvorsen, line 774: Does the clippy `allow(deprecrated)` still need to be here?

chris-hinson: unfortunately it appears so, from_expanded is still deprecated in rc.9, and the previous failure in the checks action was due to clippy failing on this function being deprecated.  moving to the suggested from_seed function is out of my range of expertise, as it would require mucking with the actual keys logic, so I opted to continue using the deprecated function to minimize logic changes.

zhalvorsen: Can we keep the comment then?

chris-hinson: agreed, rewrote and re-added comment

zhalvorsen (dismissed):

zhalvorsen (dismissed):

chris-hinson:
hey @zhalvorsen gentle ping here - had to update this pr to resolve a Cargo.lock conflict, could you re-aprove and release the CI workflows? Not sure why they need manual approval to run now.....

zhalvorsen (dismissed):

clundin25:
@chris-hinson looks like there is an issue with the Cargo.lock changes. Do you mind fixing them?

chris-hinson:
@clundin25 Cargo.lock error should be fixed now. build_and_test ran successfully in the actions for my fork.

clundin25 (approved):

PR #3795, opened 2026-05-29
[rom] Add MLDSA-87 SigVer KAT

### LINKED ISSUE #3772: MLDSA signature verification KAT missing
We need to have an explicit signature verification KAT that takes a known, fixed public key and signature and verifies them. (Today, we generate the signature and public key, which is not what the test requires.)

swenson:
FIPS 140-3 IG 10.3.A requires a CAST for the signature verification algorithm using fixed inputs and a pre-computed expected value. The existing MLDSA-87 KAT only exercised KeyGen and SigGen against freshly generated outputs.

Reshape Mldsa87Kat into three CASTs (kat_keygen / kat_sign / kat_verify) that compare directly against newly added KAT_PUB_KEY and KAT_SIGNATURE constants generated with fips204 0.4.6 from the existing SEED and KAT_MESSAGE. Add CaliptraError::KAT_MLDSA87_VERIFY_FAILURE for the new SigVer failure path.

Also, remove KATs from fake ROM.

Fixes #3543.
Fixes #3772.


drivers/src/aes.rs
```diff
@@ -1485,6 +1485,13 @@ impl AesGcm {
         Ok(Self { aes })
     }
 
+    /// Construct an `AesGcm` driver without running KATs.
+    pub fn new_skip_kats(aes: AesReg, aes_clp: AesClpReg) -> Self {
```
zhalvorsen: Can we put this behind a feature flag so we don't accidentally use it in the production firmware in the future?

rom/dev/src/main.rs
```diff
@@ -157,58 +159,64 @@ pub extern "C" fn rom_entry() -> ! {
     // Start the watchdog timer
     wdt::start_wdt(&mut env.soc_ifc);
 
-    let initialized = if cfg!(feature = "fake-rom") {
-        InitializedDrivers {
-            sha1: Sha1::new().unwrap(),
+    let initialized = {
+        #[cfg(feature = "fake-rom")]
```
zhalvorsen: nit: can you still use an `if cfg!` so we can just use an else below? I think that makes it a little more readable

rom/dev/src/main.rs
```diff
@@ -251,6 +259,12 @@ pub extern "C" fn rom_entry() -> ! {
     caliptra_drivers::ExitCtrl::exit(0);
 }
 
+#[cfg(feature = "fake-rom")]
+fn run_fips_tests(_env: &mut KatsEnv<'_, '_>) -> CaliptraResult<InitializedDrivers> {
```
zhalvorsen: It doesn't look like this is used on the `fake-rom` path?

zhalvorsen (dismissed):

mhatrevi (dismissed):

rom/dev/src/main.rs
swenson: oh sure. I think I generally have been using this pattern because that's what we do in MCU runtime to work around a Rust compiler bug where the stack space is still reserved if you use `if cfg!` with an async function in RISC-V. But I can fix it here to be more readable.

drivers/src/aes.rs
clundin25: Additionally, it should be an `unsafe` API?

swenson:
This is now blocked on runtime size after #3799 / #3775.

swenson:
Turns out we don't need the full signature, etc., and can still use hashes. No longer blocked.

kat/src/mldsa87_kat.rs
```diff
@@ -124,6 +117,26 @@ impl Mldsa87Kat {
...
+    fn kat_verify(
+        mldsa87: &mut Mldsa87,
+        pub_key: &Mldsa87PubKey,
+        signature: &Mldsa87Signature,
+    ) -> CaliptraResult<()> {
+        let result = mldsa87
+            .verify_var(pub_key, KAT_MESSAGE.as_bytes(), signature)
+            .map_err(|_| CaliptraError::KAT_MLDSA87_VERIFY_FAILURE)?;
+
+        if result != Mldsa87Result::Success {
```
ternst-nv, line 133: [nit] Do we want a negative case here as well?

ternst-nv (dismissed):
Code LGTM.  It would be good to have a reference doc, guideline for doing the signature generation and checking it instead of requiring a hard-coded value.

kat/src/mldsa87_kat.rs
swenson: I don't think it is strictly necessary.

swenson:
@ternst-nv I can add some references in a follow-up. I sent an email asking if they had some good things we could point to.

kat/src/mldsa87_kat.rs
swenson: (And we generally don't have negative tests for all our KATs, and we're always concerned with startup time.)

drivers/src/memory_layout.rs
```diff
@@ -79,7 +79,7 @@ pub const DCCM_SIZE: u32 = 256 * 1024;
 pub const ROM_DATA_SIZE: u32 = 1024;
 pub const ROM_DATA_RESERVED_SIZE: u32 = 996;
 pub const STACK_SIZE: u32 = 154 * 1024;
-pub const ROM_STACK_SIZE: u32 = 110 * 1024;
+pub const ROM_STACK_SIZE: u32 = 120 * 1024;
```
mhatrevi, line 82: Curious why this was needed.

mhatrevi (approved):

drivers/src/memory_layout.rs
swenson: SigVer added more stack state to check private key, public key, and signature.

PR #3796, opened 2026-05-29
tools: track stack usage

ternst-nv:
Add tests and tooling to track the stack usage for each command.


ci-tools/stack-size/stack-size.sh
```diff
@@ -0,0 +1,12 @@
...
+#!/usr/bin/env bash
+# Run the per-command runtime stack-usage measurement test and print only the
+# resulting usage table.
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+cargo test -p caliptra-runtime --test runtime_integration_tests \
+    test_stack_usage::measure_runtime_command_stack_usage -- --nocapture 2>&1 |
+    awk '/^Runtime command peak stack usage/{p=1} /^test /{p=0} p'
```
pjhades, line 12: I wonder if this script can be avoided at all. The script doesn't do much but is bound to the test output and can potentially go out of sync with it.

pjhades (dismissed):
LGTM in general, just one comment: based on my experience so far, the test could crash if the stack overflow triggered some other issue like NMI, so it might not print the whole table as expected.

sw-emulator/lib/cpu/src/cpu.rs
```diff
@@ -93,6 +93,10 @@ pub struct StackInfo {
     images: Vec<ImageInfo>,
     max_stack_overflow: u32,
     has_overflowed: bool,
+    /// Lowest stack pointer observed while executing within a tracked image.
+    /// Stacks grow downward, so this is the high-water mark of stack usage.
+    /// `None` means no stack-pointer activity has been observed yet.
+    min_sp: Option<u32>,
```
clundin25: nit: This mixes terms. Perhaps `stack_watermark` is sufficient, to avoid leaking the direction of the stack?

ci-tools/stack-size/stack-size.sh
clundin25, line 1: nit: Suggest moving this into a subfolder to keep the root folder cleaner

ci-tools/stack-size/stack-size.sh
ternst-nv: Yeah, it could just be invoked by the cargo test command.  I think it would be useful to keep as script because it also removes the test output fluff from a standard run.

Are you okay with giving it a bit and if it becomes an integration burden I'm happy to remove?

clundin25 (approved):

ci-tools/stack-size/stack-size.sh
pjhades: Yeah I'm okay, and I like @clundin25's suggestion of moving it to a subfolder.

swenson (approved):
LGTM

Issue #3774, opened 2026-05-22
Missing KATs for KDF, AES, MLKEM, ECDH in runtime startup


swenson:
We need to verify that all KATs are run by the time that runtime has finished startup. A few KATs are definitely not run after we refactored to have some KATs skipped in ROM.

PR #3799, opened 2026-06-01
fips: Add mlkem invocation from start up (#3774)

ternst-nv:
Add the runtime invociation for the mlkem FIPS KAT test.


swenson (approved):
LGTM

PR #3803, opened 2026-06-01
[fix] Mark hash-based ECDSA/LMS verify as FIPS non-approved

### LINKED ISSUE #3773: ECDSA and LMS signature verify commands that accept hashes as inputs need to be marked as FIPS non-approved
We have ECDSA384_SIGNATURE_VERIFY and LMS_SIGNATURE_VERIFY commands (ECDSA in ROM, both in runtime) that accept hashes as inputs. To meet FIPS requirements, they would have to accept the raw message, and not a hash, to be approved. MLDSA87_SIGNATURE_VERIFY and CM_ECDSA_VERIFY are *not* affected, as they use the raw message.

As such, as need to mark the status as non-FIPS approved.

We should add two additional commands to verify ECDSA and LMS signatures that accept the raw message as input, similar to the MLDSA signature verify.

mhatrevi:
Update successful ECDSA384_SIGNATURE_VERIFY and LMS_SIGNATURE_VERIFY responses to report FIPS_NOT_APPROVED because those commands accept a caller-provided digest rather than the raw message. Keep signature failures as command failures, and update tests/docs for the new status.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -737,7 +737,7 @@ Command Code: `0x4543_5632` ("ECV2")
 | **Name**      | **Type** | **Description**
 | --------      | -------- | ---------------
 | chksum        | u32      | Checksum over other output arguments, computed by Caliptra. Little endian.
-| fips\_status  | u32      | `FIPS_NOT_APPROVED`, because the caller supplies the digest.
+| fips\_status  | u32      | `FIPS_NOT_APPROVED_USER_SUPPLIED_DIGEST`, because the caller supplies the digest.
 
 ### LMS\_SIGNATURE\_VERIFY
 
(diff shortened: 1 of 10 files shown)


runtime/README.md
```diff
@@ -3157,15 +3158,15 @@ chksum field.
...
-currently no use case for any other responses or error values.
+For successful commands, the firmware responds with a FIPS status indicating whether the service is FIPS approved. A non-approved status is not a command error.
 
 *Table: FIPS status codes*
 
 | **Name**        | **Value**                   | Description                                         |
 | --------------- | --------------------------- | --------------------------------------------------- |
 | `FIPS_APPROVED` | `0x0000_0000`               | Status of command is FIPS approved                  |
-| `RESERVED`      | `0x0000_0001 - 0xFFFF_FFFF` | Other values reserved, will not be sent by Caliptra |
+| `FIPS_NOT_APPROVED` | `0x0000_0001`           | Status of command is not FIPS approved              |
```
swenson: I think we should have a specific FIPS non-approved because user-supplied digest error. We have a similar specific FIPS non-approval error due to non-zeroizable key, for example.

swenson: (in 2.0 we have the specific FIPS non-approved for non-zeroizable key)

mhatrevi: Added FIPS_NOT_APPROVED_USER_SUPPLIED_DIGEST

mhatrevi: Updated.

swenson (approved):
LGTM

PR #3804, opened 2026-06-01
[fix] Disallow UDS programming when debug intent is set

### LINKED ISSUE #3800: Don't allow UDS programming if debug intent is set
The CSRNG uses a fixed seed if debug intent is set, which means a predictable UDS would be written.

I don't know why anyone would ever want attempt to write UDS if debug_intent is set, but we should disallow it.

mhatrevi:
Reject UDS programming when SS_DEBUG_INTENT.DEBUG_INTENT is set to avoid programming UDS from deterministic/fake RNG output in debug-intent flows.

Adds a ROM error code and regression coverage for the debug-intent case.


swenson (dismissed):
LGTM

swenson (approved):
LGTM

PR #3805, opened 2026-06-01
pqc: mldsa driver

ternst-nv:
Add a simple MLDSA driver which invokes the recently added MLDSA library.  This includes support for some additional FIPS ACVP tests as part of validating the PQC implementation.


drivers/src/mldsa87.rs
```diff
@@ -0,0 +1,175 @@
...
+};
+
+/// ML-DSA-87 deterministic key-generation / signing seed (32 bytes).
+pub type Mldsa87Seed = [u8; MLDSA87_PRIVATE_SEED_BYTES];
+
+/// ML-DSA-87 encoded public key (2,592 bytes).
+pub type Mldsa87PubKey = [u8; MLDSA87_PUBLIC_KEY_BYTES];
+
+/// ML-DSA-87 encoded signature (4,627 bytes).
+pub type Mldsa87Signature = [u8; MLDSA87_SIGNATURE_BYTES];
```
swenson, line 35: I'd recommend using the `LEArray4xN` types, as they are more efficient in code space. (Especially for this one, which isn't divisible by 4, you'll get a lot of edge cases generated by the compiler.)

common/crypto/mldsa/src/acvp/extract.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+# Licensed under the Apache-2.0 license
```
swenson, line 2: I think we might need to add a note somewhere that the data files are under a different license, and we need to reproduce the license somehow (maybe in a README.md in this directory): https://github.com/usnistgov/ACVP-Server#license

It's an MIT-like license, which is fine, but we have to reproduce the text exactly and we can't relicense to Apache.

It's debatable if the test vectors themselves count as "software", but better safe than sorry.

drivers/src/mldsa87.rs
ternst-nv: Sounds good.  The `LEArray4xN` is currently not on this branch, so I'll file a follow PR which includes both the type, and switches these instances.

swenson (approved):
LGTM

swenson (approved):

PR #3808, opened 2026-06-02
[update] Zeroize AES, MLDSA, MLKEM, and SHA3/SHAKE during runtime FIPS shutdown before completing the command.

### LINKED ISSUE #3882: [FIPS-140-3] SHUTDOWN mailbox command — Runtime incomplete zeroize
Runtime SHUTDOWN is missing engine zeroizes — needed for FIPS L2+
runtime/src/fips.rs:35-58 (FipsModule::zeroize) zeroizes ECC, HMAC, SHA-256, SHA-512/384, SHA-Acc, and the Key Vault — but not the following, all of which retain plaintext CSP-derived state after operation:

ISO/IEC 19790:2025 §7.9.7.1 AS09.29 requires methods to zeroize all plaintext CSPs; §7.9.7.3 AS09.32 makes the zeroization a positive obligation at L2+

Image

mhatrevi:
This brings runtime shutdown zeroization coverage in line with the crypto engines that may retain sensitive state.

### DIFF
--- a/runtime/src/fips.rs
+++ b/runtime/src/fips.rs
@@ -12,14 +12,18 @@ Abstract:
 
 --*/
 use caliptra_cfi_derive::{cfi_impl_fn, cfi_mod_fn};
+use caliptra_drivers::Aes;
 use caliptra_drivers::CaliptraError;
 use caliptra_drivers::CaliptraResult;
 use caliptra_drivers::Ecc384;
 use caliptra_drivers::Hmac;
 use caliptra_drivers::KeyVault;
+use caliptra_drivers::MlKem1024;
+use caliptra_drivers::Mldsa87;
 use caliptra_drivers::Sha256;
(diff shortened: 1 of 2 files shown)


runtime/src/fips.rs
```diff
@@ -12,14 +12,19 @@ Abstract:
 
 --*/
 use caliptra_cfi_derive::{cfi_impl_fn, cfi_mod_fn};
+use caliptra_common::cprintln;
```
zhalvorsen: nit: it doesn't look like this is being used

zhalvorsen (dismissed):

runtime/src/fips.rs
mhatrevi: Fixed.

zhalvorsen (dismissed):

nquarton (approved):

zhalvorsen (approved):

PR #3809, opened 2026-06-02
[update] Use SS_STRAP_GENERIC[2] to configure entropy_src single-bit mode during CSRNG initialization.

### LINKED ISSUE #3266: rom: Allow setting entropy_src RNG single-bit mode
In single-bit mode, we need to set entropy_src THRESHOLD_SCOPE to false, RNG_BIT_ENABLE to true, and RNG_BIT_SEL configurable by the SoC.

This should be configurable based on fuses / straps.

We will we need at least ~3 bits to enable this: 1 bit to set that this mode is going to be used, and then 2 bits to select which bit of the RNG to use.

See the conversation https://github.com/chipsalliance/caliptra-sw/pull/3256#discussion_r2767304323 for context.

Targetted for 2.0.2+ and 2.1.1+ ROM

mhatrevi:
Cache the single-bit enable and RNG bit select values in ROM persistent entropy configuration so warm reset reuses the cold-boot settings. Update the warm-reset entropy config test and ROM documentation for the new strap bits.


drivers/src/persistent.rs
```diff
@@ -366,13 +366,18 @@ pub struct EntropyConfiguration {
...
-    pub rng_bit_sel: u32,
     pub repcnt_threshold: u32,
     pub adaptp_hi_threshold: u32,
     pub adaptp_lo_threshold: u32,
 }
 
+#[derive(Clone, Default, FromZeros, IntoBytes, KnownLayout, Zeroize)]
+#[repr(C)]
+pub struct EntropyConfigurationExtension {
+    pub rng_bit_enable: u32,
```
zhalvorsen: nit: You could use `U8Bool` here to save some space or start a `bitflags` field

zhalvorsen (dismissed):

drivers/src/persistent.rs
swenson: I went with the bitfield approach, as I like it a bit more.

swenson (dismissed):
LGTM

drivers/src/persistent.rs
zhalvorsen: Sounds good, we have a lot of wasted bits in the current implementation. If we would have done this from the beginning I think we'd have quite a bit of space.

zhalvorsen (dismissed):

zhalvorsen (dismissed):

swenson:
I went ahead and updated the emulator to properly emulate the single bit mode the same as the RTL; this revealed that the failure was due to not adjusting the window size appropriately for this mode. So, I updated the defaults to adjust by a factor of 4 by default (but does not modify SoC-provided overrides from fuses).

zhalvorsen (approved):

swenson:
Thanks!

PR #3813, opened 2026-06-02
[nightly]: Update DPE cert retrieval in tests

clundin25:
The nightly release was failing due to compilation errors in `runtime_integration_tests`. A previous commit updated `get_dpe_leaf_cert` to return `CertifyKeyP384Resp` instead of `CertifyKeyResp`, but did not update all callers. This PR fixes the compilation by accessing the `cert` field directly on `CertifyKeyP384Resp` and slicing it with `cert_size`, instead of calling the non-existent `cert()` method.


swenson (approved):
LGTM

mhatrevi (approved):

PR #3816, opened 2026-06-02
runtime: small code-size reductions

swenson:
Reclaim ~550 B of runtime .text:

- abr: mark with_mldsa87/with_ml_kem #[inline(always)] so the five generic monomorphizations of the closure wrappers collapse into their call sites instead of emitting separate function bodies.
- lib: have human_readable_command return &str ("?" sentinel) instead of Option<&str>, collapsing three cprintln! if/else pairs into single calls.

No functional change.


ternst-nv (dismissed):
LGTM

mhatrevi (dismissed):

ArthurHeymans (approved):

mhatrevi (approved):

swenson:
Thanks!

PR #3821, opened 2026-06-03
Adding hashing step to ECDSA KAT

### LINKED ISSUE #3770: ECDSA KAT needs to hash message before signing
Per the latest CMVP guidance, we need to hash the message that ECDSA signs in the KAT, rather than just doing the asymmetric operation (as we do today.

nquarton:
- KAT flow performs SHA384 operation and provides digest to key gen PCT
- Added generation function in test/src/crypto.rs to output expected KAT key pair


mhatrevi (dismissed):

swenson (dismissed):
LGTM

swenson (approved):
LGTM

PR #3824, opened 2026-06-03
[feat] Add owner authorization manifest support

mhatrevi:
Add owner-only authorization manifest types and runtime handling for SET_OWNER_AUTH_MANIFEST, including verification of owner key endorsements, owner IMC signatures, update-in-place/replace behavior, and owner-only authorization fallback in AUTHORIZE_AND_STASH.

Add owner authorization manifest generation to the auth-manifest tooling and expose it through create-owner-auth-man. Document the owner manifest wire format and update integration tests to exercise the shared generator and owner-only authorization flow.

### DIFF
--- a/auth-manifest/app/src/main.rs
+++ b/auth-manifest/app/src/main.rs
@@ -13,7 +13,9 @@ Abstract:
 --*/
 
 use anyhow::Context;
-use caliptra_auth_man_gen::{AuthManifestGenerator, AuthManifestGeneratorConfig};
+use caliptra_auth_man_gen::{
+    AuthManifestGenerator, AuthManifestGeneratorConfig, OwnerAuthManifestGeneratorConfig,
+};
 use caliptra_auth_man_types::AuthManifestFlags;
 #[cfg(feature = "openssl")]
 use caliptra_image_crypto::OsslCrypto as Crypto;
@@ -28,10 +30,13 @@ use zerocopy::IntoBytes;
 
 mod config;
 
-/// Entry point
-fn main() {
-    let sub_cmds = vec![Command::new("create-auth-man")
-        .about("Create a new authorization manifest")
+fn manifest_command(
+    name: &'static str,
+    about: &'static str,
+    flags_help: Option<&'static str>,
+) -> Command {
+    let command = Command::new(name)
+        .about(about)
         .arg(
             arg!(--"version" <U32> "Manifest Version Number")
                 .required(true)
@@ -39,11 +44,6 @@ fn main() {
         )
         .arg(
             arg!(--"svn" <U32> "Manifest Security Version Number")
-                .required(true)
-                .value_parser(value_parser!(u32)),
-        )
-        .arg(
-            arg!(--"flags" <U32> "Manifest Flags")
                 .required(true)
                 .value_parser(value_parser!(u32)),
         )
@@ -66,7 +66,34 @@ fn main() {
             arg!(--"out" <FILE> "Output file")
                 .required(true)
                 .value_parser(value_parser!(PathBuf)),
-        )];
+        );
+
+    if let Some(flags_help) = flags_help {
+        command.arg(
+            arg!(--"flags" <U32>)
+                .help(flags_help)
+                .required(true)
+                .value_parser(value_parser!(u32)),
+        )
+    } else {
+        command
+    }
+}
+
+/// Entry point
+fn main() {
+    let sub_cmds = vec![
+        manifest_command(
+            "create-auth-man",
+            "Create a new authorization manifest",
+            Some("Manifest Flags"),
+        ),
+        manifest_command(
+            "create-owner-auth-man",
+            "Create a new owner authorization manifest",
+            None,
+        ),
+    ];
 
     let cmd = Command::new("caliptra-auth-man-app")
         .arg_required_else_help(true)
@@ -76,6 +103,7 @@ fn main() {
 
     let result = match cmd.subcommand().unwrap() {
         ("create-auth-man", args) => run_auth_man_cmd(args),
+        ("create-owner-auth-man", args) => run_owner_auth_man_cmd(args),
         (_, _) => unreachable!(),
     };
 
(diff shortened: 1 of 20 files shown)


auth-manifest/app/src/main.rs
```diff
@@ -178,3 +206,78 @@ pub(crate) fn run_auth_man_cmd(args: &ArgMatches) -> anyhow::Result<()> {
...
+pub(crate) fn run_owner_auth_man_cmd(args: &ArgMatches) -> anyhow::Result<()> {
+    let version: &u32 = args
+        .get_one::<u32>("version")
+        .with_context(|| "version arg not specified")?;
+
+    let svn: &u32 = args
+        .get_one::<u32>("svn")
+        .with_context(|| "svn arg not specified")?;
+
+    if *svn > 255 {
```
zhalvorsen, line 219: There are only 8 strap bits allocated for the min SVN. Is MCU expected to translate 256 bits of fuses to the integer value? In other words, how many bits of fuses should be allocated for this field?

runtime/src/authorize_and_stash.rs
```diff
@@ -79,24 +95,36 @@ impl AuthorizeAndStashCmd {
...
+        let auth_and_stash_flags: AuthAndStashFlags = cmd.flags.into();
+        let (metadata_entry, success_code) =
+            if let Some(entry) = find_metadata_entry(auth_manifest_image_metadata_col, cmd_fw_id) {
+                (Some(entry), IMAGE_AUTHORIZED_VENDOR_OWNER)
+            } else if let Some(entry) =
+                find_owner_metadata_entry(owner_auth_manifest_image_metadata_col, cmd_fw_id)
+            {
+                (Some(entry), IMAGE_AUTHORIZED_OWNER_ONLY)
+            } else {
+                (None, IMAGE_NOT_AUTHORIZED)
```
zhalvorsen: Would the logic below be cleaner, if it just returned here?

e.g.

```suggestion
                return Ok(IMAGE_NOT_AUTHORIZED)
```

drivers/src/persistent.rs
```diff
@@ -985,11 +1013,12 @@ mod tests {
             (offset_of!(FwPersistentData, rt_dice_mldsa_sign), 15364, "rt_dice_mldsa_sign"),
             (offset_of!(FwPersistentData, pcr_reset), 19992, "pcr_reset"),
             (offset_of!(FwPersistentData, auth_manifest_image_metadata_col), 21016, "auth_manifest_image_metadata_col"),
-            (offset_of!(FwPersistentData, fmc_alias_csr), 31256, "fmc_alias_csr"),
-            (offset_of!(FwPersistentData, mcu_firmware_loaded), 40472, "mcu_firmware_loaded"),
-            (offset_of!(FwPersistentData, ocp_lock_metadata), 40476, "ocp_lock_metadata"),
-            (offset_of!(FwPersistentData, version), 40480, "version"),
-            (offset_of!(FwPersistentData, marker), 40484, "marker"),
+            (offset_of!(FwPersistentData, owner_auth_manifest_image_metadata_col), 27672, "owner_auth_manifest_image_metadata_col"),
+            (offset_of!(FwPersistentData, fmc_alias_csr), 30744, "fmc_alias_csr"),
```
zhalvorsen: Can we fit the `owner_auth_manifest_image_metadata_col` inside the previous `auth_manifest_image_metadata_col` such that the other field offsets don't change to maintain backwards compatibility?

runtime/src/set_owner_auth_manifest.rs
```diff
@@ -0,0 +1,480 @@
...
+        let buf = cmd_buf
+            .get(..imc_size)
+            .ok_or(CaliptraError::RUNTIME_OWNER_AUTH_MANIFEST_IMC_INVALID_SIZE)?;
+
+        // SAFETY: `OwnerAuthManifestImageMetadataCollection` is
+        // `#[repr(C)]` + `FromBytes`, alignment is u32 == 4 which the
+        // mailbox buffer satisfies. We only mutate via the sort routine
+        // before copying into persistent storage.
+        let imc_mailbox =
+            unsafe { &mut *(buf.as_ptr() as *mut OwnerAuthManifestImageMetadataCollection) };
```
zhalvorsen: Why can't we use OwnerAuthManifestImageMetadataCollection::[mut_from_prefix](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#method.mut_from_prefix) here?

runtime/src/authorize_and_stash.rs
mhatrevi: Updated.

runtime/src/set_owner_auth_manifest.rs
mhatrevi: Updated.

drivers/src/persistent.rs
mhatrevi: Updated.

auth-manifest/app/src/main.rs
mhatrevi: Yes. The 8-bit strap is the binary transport value, not the fuse bitmap itself. MCU firmware is expected to convert the platform’s monotonic SVN storage into a value from 0 through 255 and program `SS_STRAP_GENERIC[3][15:8]` before Caliptra starts.

This PR does not define a new Caliptra fuse field. If the platform uses the same thermometer encoding as the existing SVN fuses, it should reserve a 256-bit-aligned field, use bits 0 through 254 to encode SVN values 1 through 255, and reserve bit 255. An all-zero field represents SVN 0.

I clarified this contract in the `SET_OWNER_AUTH_MANIFEST` section of README.md and linked to it from README.md.

drivers/src/persistent.rs
zhalvorsen: I'd rather not let this list diverge if we can avoid it, could something like this work?
```suggestion
        let actual_expected = [
            (offset_of!(FwPersistentData, dpe), 0, "dpe"),
            (offset_of!(FwPersistentData, ecc_rtalias_tbs), 10240, "ecc_rtalias_tbs"),
            (offset_of!(FwPersistentData, mldsa_rtalias_tbs), 11264, "mldsa_rtalias_tbs"),
            (offset_of!(FwPersistentData, rtalias_mldsa_tbs_size), 15360, "rtalias_mldsa_tbs_size"),
            (offset_of!(FwPersistentData, rt_dice_mldsa_sign), 15364, "rt_dice_mldsa_sign"),
            (offset_of!(FwPersistentData, pcr_reset), 19992, "pcr_reset"),
            (offset_of!(FwPersistentData, auth_manifest_image_metadata_col), 21016, "auth_manifest_image_metadata_col"),
            #[cfg(feature = "runtime")]
            (offset_of!(FwPersistentData, owner_auth_manifest_image_metadata_col), 27672, "owner_auth_manifest_image_metadata_col"),
            (offset_of!(FwPersistentData, fmc_alias_csr), 31256, "fmc_alias_csr"),
            (offset_of!(FwPersistentData, mcu_firmware_loaded), 40472, "mcu_firmware_loaded"),
            (offset_of!(FwPersistentData, ocp_lock_metadata), 40476, "ocp_lock_metadata"),
            (offset_of!(FwPersistentData, version), 40480, "version"),
            (offset_of!(FwPersistentData, marker), 40484, "marker"),
        ];
```

runtime/src/set_owner_auth_manifest.rs
```diff
@@ -0,0 +1,480 @@
...
+            .get(..auth_metadata_persistent.entry_count as usize)
+            .ok_or(CaliptraError::RUNTIME_INTERNAL)?;
+        if sorted_metadata_lists_overlap(new_slice, auth_slice) {
+            Err(CaliptraError::RUNTIME_OWNER_AUTH_MANIFEST_IMC_DUPLICATE_FW_ID)?;
+        }
+
+        persistent.zeroize();
+        persistent
+            .as_mut_bytes()
+            .copy_from_slice(imc_candidate.as_bytes());
```
zhalvorsen: nit: to maintain type safety you should be able to do something like this:

```suggestion
        *persistent = imc_candidate.clone();
```

drivers/src/persistent.rs
mhatrevi: Updated.

runtime/src/set_owner_auth_manifest.rs
mhatrevi: Updated.

zhalvorsen (approved):

clundin25 (approved):

daeyoungkim-87:
Hello, @mhatrevi  @zhalvorsen  @clundin25 

Due to the SoC_Manifest format changes in this MR, previously built firmware is no longer compatible. 

I have created an issue to track this.
https://github.com/chipsalliance/caliptra-sw/issues/4095

Can we review the possibility of reverting the SoC_Manifest format to maintain backward compatibility?

Thank you.

PR #3826, opened 2026-06-04
[fix] FIPS fixes

### LINKED ISSUE #3770: ECDSA KAT needs to hash message before signing

### LINKED ISSUE #3773: ECDSA and LMS signature verify commands that accept hashes as inputs need to be marked as FIPS non-approved

mhatrevi:
This is a combo PR for the following Prs:

1. https://github.com/chipsalliance/caliptra-sw/pull/3821
2. https://github.com/chipsalliance/caliptra-sw/pull/3803
3. https://github.com/chipsalliance/caliptra-sw/pull/3809

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -55,7 +55,7 @@ pub const MEASUREMENT_MAX_COUNT: usize = 8;
 pub const CMB_AES_KEY_SHARE_SIZE: u32 = 32;
 pub const DOT_OWNER_PK_HASH_SIZE: u32 = 13 * 4;
 pub const ROM_OCP_LOCK_METADATA_SIZE: u32 = 8;
-pub const ROM_ENTROPY_CFG_SIZE: u32 = 24;
+pub const ROM_ENTROPY_CFG_SIZE: u32 = 32;
 pub const CLEARED_NON_FATAL_FW_ERROR_SIZE: u32 = 4;
 pub const BOOT_MODE_SIZE: u32 = 4;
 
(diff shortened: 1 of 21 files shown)


swenson (approved):
LGTM; I'll modify my threshold scope PR, as we want to set it false always, but it doesn't affect this PR.

drivers/src/persistent.rs
```diff
@@ -903,10 +905,10 @@ mod tests {
             (offset_of!(RomPersistentData, cleared_non_fatal_fw_error), 60532, "cleared_non_fatal_fw_error"),
             (offset_of!(RomPersistentData, ocp_lock_metadata), 60536, "ocp_lock_metadata"),
             (offset_of!(RomPersistentData, entropy_cfg), 60544, "entropy_cfg"),
-            (offset_of!(RomPersistentData, boot_mode), 60568, "boot_mode"),
-            (offset_of!(RomPersistentData, major_version), 60572, "major_version"),
-            (offset_of!(RomPersistentData, minor_version), 60574, "minor_version"),
-            (offset_of!(RomPersistentData, marker), 60576, "marker"),
+            (offset_of!(RomPersistentData, boot_mode), 60576, "boot_mode"),
```
zhalvorsen, line 908: How are firmwares going to be able to know where these values are from old ROMs? I think this will fail the nightlies run against old ROMs

These will need to be pulled out of the of EntropyConfiguration struct and appended to the RomPersistentData struct so it doesn't affect any other offsets. We should also bump the version of the struct. If runtime needs those values it will need to check the version of the struct to see if they were populated

mhatrevi:
Abandoning this PR and going back to the individual PRs.

PR #3827, opened 2026-06-04
noop; Update to the latest RTL submodules

swenson:


helloxiling (approved):
Looks good to me.

hw/latest/registers/src/i3ccsr.rs
```diff
@@ -136,21 +136,21 @@ impl<TMmio: caliptra_ureg::Mmio> RegisterBlock<TMmio> {
     #[inline(always)]
     pub fn tti(&self) -> TtiBlock<&TMmio> {
         TtiBlock {
-            ptr: unsafe { self.ptr.add(0x1c0 / core::mem::size_of::<u32>()) },
+            ptr: unsafe { self.ptr.add(0x200 / core::mem::size_of::<u32>()) },
```
zhalvorsen, line 139: Is this a no-op if the register offsets have changed?

zhalvorsen (approved):

hw/latest/registers/src/i3ccsr.rs
swenson: Yes; these registers are not used by caliptra core (we only use the recovery block, not the TTI, etc. blocks)

zhalvorsen: Gotcha, thanks!

zhalvorsen (approved):

swenson:
Thanks!

mhatrevi (approved):

Issue #3700, opened 2026-05-07
[RFC] Caliptra 1.x: Add MLDSA87_SIGNATURE_VERIFY Runtime Mailbox Command


sameer-katti:
# Caliptra 1.x: Add MLDSA87_SIGNATURE_VERIFY Runtime Mailbox Command

Add a firmware-only Caliptra Runtime mailbox command, `MLDSA87_SIGNATURE_VERIFY` (`0x4D44_5356`, `"MDSV"`), allowing SoC firmware to verify ML-DSA-87 signatures through Caliptra. The command wraps the ML-DSA-87 software implementation from PR#3641 and is additive: no ROM, hardware, firmware image verification, auth manifest, DPE, ECDSA, or LMS behavior changes are proposed.

## Scope

This proposal affects Caliptra Runtime mailbox API surface, runtime dispatch, error reporting, tests, and documentation.

* Specification/documentation changes:
  * Document a new `MLDSA87_SIGNATURE_VERIFY` command in the Caliptra runtime specification / `runtime/README.md`.
  * Add command ID `0x4D44_5356` (`"MDSV"`).
  * Define request format:
    * `chksum: u32`
    * `pub_key: u8[2592]`
    * `signature: u8[4627]`
    * `_sig_pad: u8`, reserved and must be zero
    * `message: u8[64]`, intended for SHA-256/SHA-384/SHA-512 digests or other fixed-size messages
  * Define response as `MailboxRespHeader` with the existing checksum and `fips_status` pattern used by ECDSA/LMS verification commands.
  * Expected files: `api/src/mailbox.rs`, `runtime/src/verify.rs`, `runtime/src/lib.rs`, `runtime/Cargo.toml`, `error/src/lib.rs`, `libcaliptra/src/caliptra_mbox.h`, related libcaliptra API/types if exposed, runtime docs, and integration tests.

* FIPS 140-3 security posture:
  * Adds a new ML-DSA-87 / FIPS 204 signature verification service to the runtime mailbox interface.
  * The implementation should be assessed as an approved-service addition, including service table updates, algorithm validation evidence, KAT/ACVP coverage as applicable, and `fips_status` behavior.
  * ML-DSA-87 provides stateless post-quantum signature verification and avoids LMS signer-state and key-exhaustion risks.

* Memory consumption:
  * Incremental mailbox plumbing is expected to be small.
  * The command relies on the existing ML-DSA-87 / PQC implementation proposed in PR#3641.
[...]

PR #3828, opened 2026-06-04
runtime: add MLDSA87_SIGNATURE_VERIFY mailbox command (RFC #3700)

elunogv:
Add a new runtime mailbox command that verifies an ML-DSA-87 (FIPS 204) signature over a fixed 64-byte message. Unlike the ECDSA / LMS verify commands, the message is supplied directly in the request rather than read from the SHA accelerator; callers are expected to pre-hash large payloads.
The command (and its supporting plumbing) is gated behind a new mldsa_attestation cargo feature on caliptra-runtime, which is OFF by default. With the feature off, builds and binary size are unchanged. 

Changes:
- api: new CommandId::MLDSA87_SIGNATURE_VERIFY ("MDSV") + Mldsa87VerifyReq.
- error: new RUNTIME_MLDSA87_VERIFY_FAILED (0xE0062) and RUNTIME_MLDSA87_VERIFY_INVALID_PADDING (0xE0063).
- runtime: dispatcher arm + Mldsa87VerifyCmd handler in verify.rs, feature-gated. Validates the reserved _sig_pad byte before calling caliptra-drivers Mldsa87::verify.
- runtime/tests: 8 integration tests under test_mldsa_verify.rs covering happy path (2 seeds x 2 messages), crypto-negative cases (tampered signature, wrong message, wrong pub key, all-zero), and malformed requests (bad chksum, invalid padding, truncated request).
- libcaliptra: C bindings (caliptra_mldsa87_signature_verify_req, caliptra_mldsa87_signature_verify) + smoke test in examples/generic.
- docs: README mailbox-command reference + test-coverage.md entries. Two prerequisites are tracked separately and must land before mldsa_attestation can be enabled in a shipping build: (1) panic-freedom in caliptra-mldsa / caliptra-shake (for test_panic_missing), and (2) stack-budget reduction in caliptra-mldsa::verify_internal.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -21,6 +21,7 @@ impl CommandId {
     pub const GET_RT_ALIAS_CERT: Self = Self(0x43455252); // "CERR"
     pub const ECDSA384_VERIFY: Self = Self(0x53494756); // "SIGV"
     pub const LMS_VERIFY: Self = Self(0x4C4D5356); // "LMSV"
+    pub const MLDSA87_SIGNATURE_VERIFY: Self = Self(0x4D445356); // "MDSV"
     pub const STASH_MEASUREMENT: Self = Self(0x4D454153); // "MEAS"
     pub const INVOKE_DPE: Self = Self(0x44504543); // "DPEC"
     pub const DISABLE_ATTESTATION: Self = Self(0x4453424C); // "DSBL"
@@ -265,6 +266,7 @@ impl Default for MailboxResp {
 pub enum MailboxReq {
     EcdsaVerify(EcdsaVerifyReq),
     LmsVerify(LmsVerifyReq),
+    Mldsa87Verify(Mldsa87VerifyReq),
     GetLdevCert(GetLdevCertReq),
     StashMeasurement(StashMeasurementReq),
     InvokeDpeCommand(InvokeDpeReq),
@@ -294,6 +296,7 @@ impl MailboxReq {
         match self {
             MailboxReq::EcdsaVerify(req) => Ok(req.as_bytes()),
             MailboxReq::LmsVerify(req) => Ok(req.as_bytes()),
+            MailboxReq::Mldsa87Verify(req) => Ok(req.as_bytes()),
             MailboxReq::StashMeasurement(req) => Ok(req.as_bytes()),
             MailboxReq::InvokeDpeCommand(req) => req.as_bytes_partial(),
             MailboxReq::FipsVersion(req) => Ok(req.as_bytes()),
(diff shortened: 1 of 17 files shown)


api/src/mailbox.rs
```diff
@@ -605,6 +610,37 @@ impl Request for LmsVerifyReq {
...
+// The message digest to verify against is taken from Caliptra's SHA
+// accelerator (matching ECDSA384_VERIFY and LMS_VERIFY). Callers must
+// stream the message through the SHA accelerator before issuing this
+// command. Carrying the raw message in-band would put hashing outside
+// the FIPS module boundary and is therefore disallowed.
+#[repr(C)]
+#[derive(Debug, IntoBytes, FromBytes, Immutable, KnownLayout, PartialEq, Eq)]
+pub struct Mldsa87VerifyReq {
+    pub hdr: MailboxReqHeader,
+    pub pub_key: [u8; 2592],
```
ternst-nv: [nit] Could we use constants from the MLDSA library here?

api/src/mailbox.rs
```diff
@@ -605,6 +610,37 @@ impl Request for LmsVerifyReq {
...
+}
+impl Request for Mldsa87VerifyReq {
+    const ID: CommandId = CommandId::MLDSA87_SIGNATURE_VERIFY;
+    type Resp = MailboxRespHeader;
+}
+impl Default for Mldsa87VerifyReq {
+    fn default() -> Self {
+        Self {
+            hdr: MailboxReqHeader::default(),
+            pub_key: [0u8; 2592],
```
ternst-nv: [nit] Could we use the constants defined in the Driver/mldsa crate for these values?

libcaliptra/inc/caliptra_api.h
```diff
@@ -179,6 +179,9 @@ int caliptra_ecdsa384_verify(struct caliptra_ecdsa_verify_req *req, bool async);
 // LMS Verify
 int caliptra_lms_verify(struct caliptra_lms_verify_req *req, bool async);
 
+// ML-DSA-87 Signature Verify (RFC #3700)
```
ternst-nv: [nit] We can remove the RFC in this and the other comments.

runtime/src/verify.rs
```diff
@@ -133,3 +137,36 @@ impl LmsVerifyCmd {
...
     }
 }
+
+#[cfg(feature = "mldsa_attestation")]
+pub struct Mldsa87VerifyCmd;
+#[cfg(feature = "mldsa_attestation")]
+impl Mldsa87VerifyCmd {
+    #[cfg_attr(feature = "cfi", cfi_impl_fn)]
+    #[inline(never)]
+    pub(crate) fn execute(drivers: &mut Drivers, cmd_args: &[u8]) -> CaliptraResult<MailboxResp> {
```
ternst-nv: This command will need to be updated to the new format in this PR: https://github.com/chipsalliance/caliptra-sw/pull/3875, to receive a large stack savings.

ternst-nv (commented):
Changes look good, please address the command structure update and should be good to go.  Thanks!

api/src/mailbox.rs
elunogv: Pushed a new commit to address this

libcaliptra/inc/caliptra_api.h
elunogv: Pushed a new commit to address this

runtime/src/verify.rs
elunogv: Pushed a new commit to address this

ternst-nv (approved):
LGTM

Issue #3741, opened 2026-05-18
[PQC] Upgrade DPE to hybrid-capable caliptra-dpe


ternst-nv:
* Caliptra Version(s): 1.1
* Caliptra Mode: Core Mode

### Summary

Upgrade Runtime from the in-tree DPE implementation to external `caliptra-dpe` at rev `3e5b97094e71a5d5c6f6a5bd1882f4006830f2cc`, enabling `hybrid` and `arbitrary_max_handles` while preserving existing ECDSA behavior and hitless-update compatibility.

### Scope

- [x] Update DPE dependencies in workspace `Cargo.toml`.
- [x] Enable `hybrid` and `arbitrary_max_handles`.
- [ ] Verify `MAX_HANDLES` remains 24.
- [x] Keep `DPE_SIZE = 5 * 1024`.
- [x] Adapt `runtime/src/dpe_crypto.rs` to the new generic `Crypto` trait API.
- [x] Adapt `runtime/src/invoke_dpe.rs` to the new command execution API.
- [x] Adapt `runtime/src/dpe_platform.rs` to the new `Platform` trait.
- [x] Adapt `runtime/src/drivers.rs` DPE initialization.
- [x] Update other DPE type imports and tests.
- [x] Add stale/incompatible DPE-state handling based on the new marker/version behavior.

### Acceptance Criteria

- [ ] Workspace dependencies use external `caliptra-dpe` rev `3e5b97094e71a5d5c6f6a5bd1882f4006830f2cc` with `hybrid` and `arbitrary_max_handles`.
- [ ] Runtime binary size impact is reported.
- [x] `DPE_DCCM_STORAGE < DPE_SIZE` static assertion passes.
- [x] `PersistentData::assert_matches_layout()` passes.
- [ ] A guard verifies `MAX_HANDLES == 24`.
- [x] Incompatible old DPE state is detected via marker/version and re-initialized safely.
- [x] Existing ECDSA DPE command paths are migrated to the new APIs without intentional behavior changes.

PR #3831, opened 2026-06-04
Upgrade DPE to caliptra-dpe for 1.x PQC retrofit

pjhades:
This PR upgrades the DPE dependency to caliptra-dpe revision [a174f337998359080d5396fc84a5ee2ada4ae83a](https://github.com/chipsalliance/caliptra-dpe/tree/a174f337998359080d5396fc84a5ee2ada4ae83a), and adapt the codebase to the updated types, imports, and interfaces.

Issue: #3741


runtime/src/drivers.rs
```diff
@@ -479,62 +483,72 @@ impl Drivers {
...
             flags: DeriveContextFlags::MAKE_DEFAULT
                 | DeriveContextFlags::CHANGE_LOCALITY
-                | DeriveContextFlags::INPUT_ALLOW_CA
-                | DeriveContextFlags::INPUT_ALLOW_X509,
+                | DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT
+                | DeriveContextFlags::INPUT_ALLOW_X509
+                | DeriveContextFlags::ALLOW_RECURSIVE,
             tci_type: u32::from_be_bytes(*b"CCIV"),
             target_locality: pl0_pauser_locality,
+            svn: 0,
```
pjhades, line 549: @zhalvorsen Could you please take a look at this, and other DPE invocations? I'm not familiar enough with DPE and I see the flags and some command fields have been updated. I made the changes following `caliptra-sw/main` and cargo tests passed, which seems good. But it'd be nice to have more eyes on it.

drivers/src/persistent.rs
```diff
@@ -89,7 +89,7 @@ impl Zeroize for ExportedCdiHandles {
 }
 
 #[cfg(feature = "runtime")]
-const DPE_DCCM_STORAGE: usize = size_of::<DpeInstance>()
+const DPE_DCCM_STORAGE: usize = size_of::<State>()
```
zhalvorsen, line 92: This version of DPE also assumes 64 contexts. This can be set back to 32 contexts with a features flag and environment variable. It looks like you have the feature flag set, but you'll need the environment variable set as well. Caliptra 2.0 has it set like this https://github.com/chipsalliance/caliptra-sw/blob/88512eb6b83b7503625cd319668d30dc3bff6012/.cargo/config.toml#L4. That's how I would recommend doing it as well.

drivers/src/sha384.rs
```diff
@@ -34,6 +34,55 @@ pub struct Sha384 {
...
+#[cfg(feature = "runtime")]
+pub struct DpeHasher<'a> {
+    op: Sha384DigestOp<'a>,
+}
+
+#[cfg(feature = "runtime")]
+impl<'a> DpeHasher<'a> {
+    pub fn new(sha: &'a mut Sha384) -> Result<Self, CaliptraError> {
+        Ok(Self {
+            op: sha.digest_init()?,
```
zhalvorsen, line 47: The version in `main` also has a field to keep track of the state to make sure it is used properly (can't reuse the state after finalization). It would be good to add that here as well.

https://github.com/pjhades/caliptra-sw/blob/a38c99b60c97df406201e098df1b89e9301a26d1/drivers/src/sha2_512_384.rs#L634

runtime/src/drivers.rs
zhalvorsen: This looks good to me!

runtime/src/invoke_dpe.rs
```diff
@@ -57,91 +70,57 @@ impl InvokeDpeCmd {
...
                 }
-                Command::Sign(cmd) => cmd.execute(dpe, &mut env, locality),
-                Command::RotateCtx(cmd) => cmd.execute(dpe, &mut env, locality),
-                Command::GetCertificateChain(cmd) => cmd.execute(dpe, &mut env, locality),
-            };
+                _ => (),
+            }
+
+            let ueid = Some(drivers.soc_ifc.fuse_bank().ueid());
+            let resp = invoke_dpe_cmd(drivers, &command, None, ueid, None);
```
zhalvorsen: Is there are specific reason we wouldn't use the serialized version here?

runtime/src/stash_measurement.rs
```diff
@@ -49,48 +50,23 @@ impl StashMeasurementCmd {
...
-            let derive_context_resp = DeriveContextCmd {
+            let cmd = DeriveContextCmd {
                 handle: ContextHandle::default(),
-                data: *measurement,
+                data: TciMeasurement(*measurement),
                 flags: DeriveContextFlags::MAKE_DEFAULT
                     | DeriveContextFlags::CHANGE_LOCALITY
-                    | DeriveContextFlags::INPUT_ALLOW_CA
+                    | DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT
                     | DeriveContextFlags::INPUT_ALLOW_X509,
```
zhalvorsen: We should add the ALLOW_RECURSIVE flag here as well

runtime/tests/runtime_integration_tests/test_tagging.rs
```diff
@@ -213,13 +212,15 @@ fn test_tagging_retired_context() {
...
 
     // retire context via DeriveContext
-    let derive_context_cmd = DeriveContextCmd {
-        handle: ContextHandle::default(),
-        data: [0u8; DPE_PROFILE.get_hash_size()],
-        flags: DeriveContextFlags::empty(),
-        tci_type: 0,
-        target_locality: 0,
-    };
+    //let derive_context_cmd = DeriveContextCmd {
```
zhalvorsen: Lets remove the commented code here

drivers/src/sha384.rs
pjhades: Updated in [aedb308](https://github.com/chipsalliance/caliptra-sw/pull/3831/commits/aedb30891a653f32cca5896cb26b0b29edad7f85).

runtime/src/invoke_dpe.rs
pjhades: Good catch! I had an optimization patch on top of this where only the serialized version was called. But I extracted it for a later PR and forgot to change this. Updated in [5daacf5](https://github.com/chipsalliance/caliptra-sw/pull/3831/commits/5daacf529c66f597e0f5840decb65bd99ddad882).

runtime/src/stash_measurement.rs
pjhades: Updated in [bd50c9f](https://github.com/chipsalliance/caliptra-sw/pull/3831/commits/bd50c9f6c35d06c58a33fb34a5114de159072d73).

drivers/src/persistent.rs
pjhades: Yeah I saw that, and it's currently [defined as 32](https://github.com/chipsalliance/caliptra-sw/blob/caliptra-1.x/.cargo/config#L2).

@ternst-nv Could you please comment on this? #3741 asks to define it as 24. Is that still what is needed?

ternst-nv: Hmm, it looks like the previous limit is 24, and just based on stack usage I think it might be a good to keep a smaller value.  @zhalvorsen , is there any concern with sticking with 24?

libcaliptra/inc/caliptra_types.h
```diff
@@ -250,7 +250,8 @@ struct caliptra_certify_key_extended_req
 struct caliptra_certify_key_extended_resp
 {
     struct caliptra_resp_header hdr;
-    uint8_t certify_key_resp[6272];
+    uint32_t size;
+    uint8_t certify_key_resp[12672];
```
ternst-nv: [nit] Can we use the constant here?

ternst-nv (dismissed):
LGTM!

drivers/src/persistent.rs
zhalvorsen: Oh okay cool. I didn't realize that environment variable was already set. I'd be surprised if we wanted to lower that back to 24. We raised it to 32 (from 24) in 1.x about a year ago if I remember correctly. Where are you seeing it set to 24?

pjhades:
The runtime binary size increased by ~15 KiB. (209876 bytes, compared to 193516 bytes in 1.x)

drivers/src/persistent.rs
ternst-nv: This is where I saw it: https://github.com/chipsalliance/caliptra-dpe/blob/08cbee89763acfc0d9bd55edd82532f4d9c17175/dpe/src/lib.rs#L33.  Ahh, looks like there it is replaced by the MAX_HANDLES setting.  Ok, yeah lets stay with 32 then.

zhalvorsen: Sounds good to me. While we are at it, it would be good to get this change in https://github.com/chipsalliance/caliptra-dpe/issues/575. It should help with the stack usage, especially when we get to ML-DSA

zhalvorsen: Let me check with a contractor to see if they can do the change

libcaliptra/inc/caliptra_types.h
pjhades: Updated in [830dd1f](https://github.com/chipsalliance/caliptra-sw/pull/3831/commits/830dd1f4b40935bfc2367c42767d4814fef1f5aa).

drivers/src/persistent.rs
pjhades: I think we can merge this PR for now so that other PQC work can happen. When the change from https://github.com/chipsalliance/caliptra-dpe/issues/575 is merged we can bump dpe again to include it.

pjhades: Regarding response size, currently it's also [hardcoded](https://github.com/chipsalliance/caliptra-sw/blob/main/api/src/mailbox.rs#L1663). I think we can define the response size as an associated const from caliptra-dpe instead.

But types in libcaliptra need some care to avoid hardcoding. Maybe the types can be generated with cbindgen.

ternst-nv (approved):

drivers/src/persistent.rs
zhalvorsen: I would prefer that as well, I have a draft of what that could look like in DPE https://github.com/chipsalliance/caliptra-dpe/pull/574

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -476,7 +476,7 @@ fn test_authorize_and_stash_after_update_reset() {
         fw_id: FW_ID_0,
         measurement: IMAGE_DIGEST1,
         source: ImageHashSource::InRequest as u32,
-        flags: 0, // Don't skip stash
+        flags: 1, // Skip stash
```
zhalvorsen, line 479: FYI: if you update the DPE reference again, you can undo the changes in this file, runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs, and runtime/tests/runtime_integration_tests/test_reallocate_dpe_context_limits.rs,

zhalvorsen (approved):

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
pjhades: Yeah I noticed your recent PR reverting the TCI type check. Thanks for the reminder!

PR #3833, opened 2026-06-04
[runtime] Reduce redundant logging

zhalvorsen:
This removes some unnecessary logging to reduce runtime codespace. It adds a few new error codes to handle conditions where duplicate error codes were used but distinguished using print statements. This also removes some print statements that simply stated the error code description.

### DIFF
--- a/runtime/src/recovery_flow.rs
+++ b/runtime/src/recovery_flow.rs
@@ -90,10 +90,8 @@ impl RecoveryFlow {
             )?;
             dma_recovery.wait_for_payload_not_available();
 
-            cprintln!("[rt] Uploading MCU firmware");
             let mcu_size_bytes =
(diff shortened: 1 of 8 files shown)


runtime/src/recovery_flow.rs
```diff
@@ -112,7 +107,6 @@ impl RecoveryFlow {
             .pl0_pauser;
 
         let digest: [u8; 48] = digest.into();
-        cprintln!("[rt] Verifying MCU digest: {}", HexBytes(&digest));
```
swenson, line 115: This one might be good to keep?

swenson (dismissed):
LGTM, one I would keep personally

runtime/src/recovery_flow.rs
zhalvorsen: Good call, I added it back

swenson (approved):
LGTM

mhatrevi (approved):

clundin25 (approved):

PR #3836, opened 2026-06-05
Disable entropy_src repcnts health test in ROM

swenson:
The Repetition Count Symbol (repcnts) health test is recommended to be disabled for FIPS, so stop programming its threshold and leave it at the reset default, which disables it. A repeating 4-bit symbol also makes every individual RNG wire
constant, so the per-wire repcnt test still catches these streams.


ekarabu (approved):

clundin25 (approved):

swenson:
Thanks!

PR #3838, opened 2026-06-05
mldsa: sig kat optimization

ternst-nv:
Following FIPS guidance from the 2.x MLDSA KAT, optimize space usage by generating the signatures and keys as part of the KAT itself.  This saves ~7Kb in instruction memory space.

This requires the ability to export the Secret Key created during key generation to satisfy the Key Gen KAT requirements.


common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -814,14 +861,62 @@ fn generate_key_internal(
...
+/// transform), matching `BitPack(w, sub-?, sub)` for the appropriate range.
+fn pack_scalar_bits(out: &mut [u8], s: &Scalar, bits: u32, sub: u32) {
+    let mut acc: u64 = 0;
+    let mut nbits: u32 = 0;
+    let mut oi = 0;
+    for i in 0..K_DEGREE {
+        acc |= (mod_sub(sub, s.c[i]) as u64) << nbits;
+        nbits += bits;
+        while nbits >= 8 {
+            out[oi] = (acc & 0xff) as u8;
```
pjhades, line 881: I'm not familiar with the meaning of the variables passed into this. Maybe others can comment.

Can this panic?

Also, would it be a concern that `nbits` is not a multiple of 8? In that case there would be some remaining bits not written to the output.

pjhades (approved):
LGTM, left comment on a helper function.

common/crypto/mldsa/src/mldsa87.rs
swenson: It should not be possible for this to panic, and I believe that nbits will always be correctly sized -- we're not calling this with arbitrary values, but with certain fixed values.

If code size were not the limiting factor, we'd probably use macros or unroll these loops for performance, honestly.

swenson (approved):
LGTM

zhalvorsen (approved):

PR #3842, opened 2026-06-05
[fix] Report firmware verification recovery reason

### LINKED ISSUE #3840: rom: Set I3C device_status on fatal error
Use the vendor-specific code to register fatal errors in ROM in the I3C device status, in particular, to help BMCs and others know about things like particular authorization failures, that can help them take action.

We only have 127 or so codes available, so we should be prudent.

mhatrevi:
Set DEVICE_STATUS_0 to Boot Failure instead of Fatal Error when ROM firmware verification fails in subsystem recovery flow. Populate REC_REASON_CODE with a coarse verification failure reason so the recovery agent can distinguish key, signature, digest, layout, rollback, and PQC configuration failures.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor/mod.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor/mod.rs
@@ -103,6 +103,122 @@ pub struct FwProcInfo {
 pub struct FirmwareProcessor {}
 
 impl FirmwareProcessor {
+    fn recovery_reason_from_verification_error(err: CaliptraError) -> u32 {
+        let err = u32::from(err);
+
+        macro_rules! err_is {
+            ($($candidate:expr),+ $(,)?) => {
+                $(err == u32::from($candidate))||+
+            };
+        }
+
+        if err_is!(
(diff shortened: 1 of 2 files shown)


rom/dev/src/flow/cold_reset/fw_processor/mod.rs
```diff
@@ -103,6 +103,122 @@ pub struct FwProcInfo {
...
+    fn recovery_reason_from_verification_error(err: CaliptraError) -> u32 {
+        let err = u32::from(err);
+
+        macro_rules! err_is {
+            ($($candidate:expr),+ $(,)?) => {
+                $(err == u32::from($candidate))||+
+            };
+        }
+
+        if err_is!(
```
swenson, line 115: I feel like this could be replaced with a `match`?

swenson (approved):
LGTM, one suggestion/question

rom/dev/src/flow/cold_reset/fw_processor/mod.rs
mhatrevi: Since the PR is already queued, I'll fix this in an upcoming PR.

PR #3845, opened 2026-06-08
[ml-dsa] Add external-mu APIs

zhalvorsen:
This adds APIs to sign and verify using external-mu. This will be necessary for DPE integration as part of the 1.x retrofit.


ternst-nv (approved):
LGTM

PR #3847, opened 2026-06-08
Fix OTP FuseController DMA emulator for field entropy programming (#3753)

mlvisaya:
- Expand FUSE_BANK_SIZE_BYTES from 64 to 136 to accommodate UDS (64 bytes), UDS digest (8 bytes), and 4 FE partitions (4 x 16 bytes = 64 bytes)
- Read ss_uds_seed_base_addr_l dynamically from SoC registers instead of hardcoding FUSE_BANK_OFFSET=0x800, since firmware sets this register to the actual fuse bank base address
- Scope start_digest() to hash only the UDS portion (first 64 bytes) of the fuse bank, matching the hardware behavior
- Inline the 0x800 constant in soc_reg.rs since the FuseController no longer exposes FUSE_BANK_OFFSET

### DIFF
--- a/sw-emulator/lib/periph/src/dma/otp_fc.rs
+++ b/sw-emulator/lib/periph/src/dma/otp_fc.rs
@@ -13,7 +13,7 @@ Abstract:
 --*/
 
 use bitfield::size_of;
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

mlvisaya:
Thanks!

PR #3848, opened 2026-06-09
[fix] Fix TAP mailbox availability after debug unlock

### LINKED ISSUE #3834: Caliptra TAP holds mailbox lock indefinitely post MANUF debug unlock process
# Please include the following information

* Caliptra Version 2.0.2
* Caliptra Mode :Subsystem Mode

During the MANUF debug unlock process on either FPGA ENV/ DV simulation, the Caliptra TAP acquires the mailbox lock and fails to release post successful Debug unlock and the Mailbox lock is held indefinitely by the TAP.
Hence preventing MCU ROM from acquiring the lock for `RI_DOWNLOAD_FIRMWARE` post a successful Debug unlock process.

## 

After digging deep with DV simulation it was root caused that:
1. SS_DBG_MANUF_SERVICE_REG_RSP.TAP_MAILBOX_AVAILABLE is set to 0, immediately after a successful authentication by Caliptra ROM,
2.  This would prevent/gate release of Mailbox by TAP as the MBOX_EXECUTE (CLEAR) operation would be rejected by the Hardware.

mhatrevi:
Fixes the debug-unlock mailbox handoff so Caliptra does not clear TAP_MAILBOX_AVAILABLE when clearing the debug-unlock in-progress bit.

Previously, set_ss_dbg_unlock_in_progress(false) also cleared TAP_MAILBOX_AVAILABLE. After a successful MANUF debug unlock, this could prevent the TAP from clearing MBOX_EXECUTE, leaving the mailbox lock held indefinitely and blocking the MCU ROM from acquiring the mailbox for RI_DOWNLOAD_FIRMWARE.


rom/dev/src/flow/debug_unlock.rs
```diff
@@ -55,11 +53,28 @@ pub fn debug_unlock(env: &mut RomEnv) -> CaliptraResult<()> {
...
+    env.soc_ifc.set_ss_dbg_unlock_in_progress(true);
+}
+
+fn finish_debug_unlock(env: &mut RomEnv, response_sent: bool) {
+    env.soc_ifc.set_ss_dbg_unlock_in_progress(false);
+    if response_sent {
+        // Real TAP can clear MBOX_EXECUTE while ROM polls for mailbox idle. In
+        // the emulator, the requester clears MBOX_EXECUTE only after firmware
+        // returns control to the model, so polling here would deadlock tests.
+        #[cfg(not(feature = "emu"))]
```
swenson: I think this defeats the purpose of this PR. The "real" ROM uses the `emu` feature flag. https://github.com/chipsalliance/caliptra-sw/blob/main/builder/src/firmware.rs#L57

I'd rather fix the emulator than try to work around some difference.

runtime/src/lib.rs
```diff
@@ -279,6 +279,23 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
...
+    matches!(cmd_id, CommandId::PRODUCTION_AUTH_DEBUG_UNLOCK_TOKEN)
+        || (matches!(cmd_id, CommandId::PRODUCTION_AUTH_DEBUG_UNLOCK_REQ) && command_failed)
+}
+
+fn finish_debug_unlock_mailbox_access(drivers: &mut Drivers) {
+    drivers.soc_ifc.set_ss_dbg_unlock_in_progress(false);
+    // Real TAP can clear MBOX_EXECUTE while runtime polls for mailbox idle. In
+    // the emulator, the requester clears MBOX_EXECUTE only after firmware
+    // returns control to the model, so polling here would deadlock tests.
+    #[cfg(not(feature = "emu"))]
```
swenson: ditto here

rom/dev/src/flow/debug_unlock.rs
mhatrevi: I handled the debug unlock tests as a special case. We can address fixing the emulator as a follow-up task.

runtime/src/lib.rs
mhatrevi: I handled the debug unlock tests as a special case. We can address fixing the emulator as a follow-up task.

rom/dev/src/flow/debug_unlock.rs
```diff
@@ -115,14 +126,16 @@ fn handle_manufacturing(env: &mut RomEnv) -> CaliptraResult<()> {
...
+    drop(txn);
+    finish_debug_unlock(env, true);
 
     cprintln!("[dbg_manuf] --");
     result
 }
 
 fn handle_auth_debug_unlock_request(
     env: &mut RomEnv,
+    response_sent: &mut bool,
```
swenson: This seems like an anti-pattern in Rust -- we don't generally use mutable parameters to return values unless there is a very good reason. Why can't this just be something that is returned?

rom/dev/tests/rom_integration_tests/test_debug_unlock.rs
```diff
@@ -204,7 +206,6 @@ fn test_dbg_unlock_manuf_wrong_cmd() {
     });
 }
 
-//TODO: https://github.com/chipsalliance/caliptra-sw/issues/2070
```
swenson, line 207: Not sure why these are deleted -- these tests still don't work on FPGA (realtime).

rom/dev/src/flow/debug_unlock.rs
mhatrevi: Updated.

rom/dev/tests/rom_integration_tests/test_debug_unlock.rs
mhatrevi: Discussed offline; these debugs unlock tests are only available in SS mode.

swenson: Discussed, and these are probably okay to delete.

swenson (dismissed):
LGTM

swenson (approved):

Issue #3822, opened 2026-06-03
Explore stack optimizations in 1.x


pjhades:
We encountered stack overflows during the ongoing work of PQC retrofit to Caliptra 1.x. The PQC work depends on the caliptra-dpe crate with ML-DSA support, where cryptographic items like signatures have bigger sizes. However the current mailbox command loop reserves multiple buffers on the stack. For example, for the `INVOKE_DPE` command:

```Rust
    let mut resp = match CommandId::from(req_packet.cmd) {
        // ...
        CommandId::INVOKE_DPE => InvokeDpeCmd::execute(drivers, cmd_bytes),
        // ...
    }
```
while in the `execute` function:
```Rust
         let resp: Result<Response, DpeErrorCode> = match command {
                // ...
                Command::CertifyKey(cmd: &CertifyKeyCmd) => {
                    // ...
                    cmd.execute(dpe, &mut env, locality)
                }
         }
         let mut invoke_resp: InvokeDpeResp = InvokeDpeResp {
                hdr: MailboxRespHeader::default(),
                data_size: 0,
                data: [0u8; InvokeDpeResp::DATA_MAX_SIZE],
         };
         match resp {
                Ok(ref r: &Response) => {
                    let resp_bytes: &[u8] = r.as_bytes();
                    let data_size: usize = resp_bytes.len();
                    invoke_resp.data[..data_size].copy_from_slice(resp_bytes);
                    invoke_resp.data_size = data_size as u32;
                }
                // ...
         };
         Ok(MailboxResp::InvokeDpeCommand(invoke_resp))
```

There are several issues:
- For commands that invoke DPE, like `INVOKE_DPE` and `CERTIFY_KEY_EXTENDED`, the DPE response can be a few kilobytes, especially when ML-DSA is enabled.
- The mailbox response is owned and returned by-value. If the compiler doesn't optimize it, multiple instances will exist simultaneously on the stack.
- The DPE response is deserialized but immediately serialized into the mailbox response.

Some optimizations can be applied here to reduce the stack usage in order to make more room for PQC work:
- Take advantage of the [`execute_serialized`](https://github.com/chipsalliance/caliptra-dpe/blob/main/dpe/src/commands/mod.rs#L304) interface introduced in `caliptra-dpe/main`, where a buffer is provided by the caller for the command to dump its output.
[...]

PR #3860, opened 2026-06-10
runtime: support writing response to mailbox SRAM

pjhades:
This change introduces support for writing mailbox responses directly to the SRAM without stashing them on the stack. This pattern is then applied to commands that involve big buffers in the their responses, namely,

- `INVOKE_DPE`
- `CERTIFY_KEY`
- `DERIVE_CONTEXT`

Related to #3822

### DIFF
--- a/runtime/src/mailbox.rs
+++ b/runtime/src/mailbox.rs
@@ -173,4 +173,17 @@ impl Mailbox {
             )
         }
     }
+
+    /// Retrieve a mutable slice with the contents of the mailbox
+    pub fn raw_mailbox_contents_mut(&mut self) -> CaliptraResult<&mut [u8]> {
+        if !self.cmd_busy() {
+            return Err(CaliptraError::DRIVER_MAILBOX_INVALID_STATE);
+        }
+        unsafe {
+            Ok(slice::from_raw_parts_mut(
+                memory_layout::MBOX_ORG as *mut u8,
+                memory_layout::MBOX_SIZE as usize,
+            ))
+        }
+    }
 }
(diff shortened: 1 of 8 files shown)


runtime/src/mailbox.rs
```diff
@@ -173,4 +173,17 @@ impl Mailbox {
             )
         }
     }
+
+    /// Retrieve a mutable slice with the contents of the mailbox
+    pub fn raw_mailbox_contents_mut(&mut self) -> CaliptraResult<&mut [u8]> {
```
clundin25, line 178: I believe that the mailbox SRAM requires word-alignment. I believe this is unsound, since it returns a `&mut [u8]` which is not word aligned.

runtime/src/mailbox.rs
```diff
@@ -173,4 +173,17 @@ impl Mailbox {
...
+        if !self.cmd_busy() {
+            return Err(CaliptraError::DRIVER_MAILBOX_INVALID_STATE);
+        }
+        unsafe {
+            Ok(slice::from_raw_parts_mut(
+                memory_layout::MBOX_ORG as *mut u8,
+                memory_layout::MBOX_SIZE as usize,
+            ))
+        }
```
clundin25, line 187: Do you mind adding a safety comment?

runtime/src/mailbox.rs
pjhades: Yes, good callout! I chatted with @ternst-nv. He proposed a trait-based design for the DPE commands, where a trait controls how the response will be written. I think we can implement some dword buffering there.

PR #3861, opened 2026-06-10
hw-model: retry mailbox lock in subsystem mode to fix nightly race

swenson:
The MCU ROM performs post-runtime mailbox activity that briefly races host commands on the FPGA subsystem profile, causing intermittent UnableToLockMailbox failures across the FPGA Subsystem Suite (smoke_test, fake_collateral_boot_test, etc.).

Retry the mailbox lock in start_mailbox_execute when in subsystem mode, stepping the model between attempts so the MCU can release it. Extract the retry policy into a testable retry_acquire_mailbox_lock helper and add a unit test. Non-subsystem behavior is unchanged.


mhatrevi (approved):

zhalvorsen (approved):

swenson:
Thanks!

PR #3863, opened 2026-06-10
dpe context update prop

mhatrevi:


DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+When authorization succeeds and stashing is not skipped, Runtime creates the stashed DPE context, stores a fresh non-zero `measurement_handle` for the created context index, and returns the handle.
+
+When authorization fails or `SKIP_STASH` is set, Runtime returns an all-zero `measurement_handle`.
+
+## 4. Measurement Handle Mapping
+
+Runtime should store the handle mapping as DPE-side persistent metadata, parallel to the existing context tag side tables:
+
+```text
+measurement_handle_valid: [U8Bool; MAX_HANDLES]
```
zhalvorsen, line 121: nit: Lets use bitflags to support this so we can shrink this to a single u16

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+measurement_handle: u8[16]
+```
+
+When authorization succeeds and stashing is not skipped, Runtime creates the stashed DPE context, stores a fresh non-zero `measurement_handle` for the created context index, and returns the handle.
+
+When authorization fails or `SKIP_STASH` is set, Runtime returns an all-zero `measurement_handle`.
+
+## 4. Measurement Handle Mapping
+
+Runtime should store the handle mapping as DPE-side persistent metadata, parallel to the existing context tag side tables:
```
zhalvorsen, line 118: These handles will also need to be accessible in ROM for later releases so we should store this in the ROM partition of persistent data

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+Each valid entry maps one opaque `measurement_handle` to the DPE context at the same index in `fw.dpe.state.contexts`. The DPE context index is not returned to the caller.
+
+Handle generation rules:
+
+```text
+1. Generate 16 random bytes using the TRNG in Runtime.
+2. Reject the all-zero handle value.
+3. Check the candidate against all currently valid handles.
+4. If it collides, retry handle generation.
+5. If Runtime cannot produce a unique handle, fail the command.
```
zhalvorsen, line 134: In the future this will also apply to the ROM handler

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+A 16-byte random handle is a bearer capability for updating one stashed DPE measurement. It prevents a caller from selecting an update target by guessing or reusing a public `fw_id`, `metadata`, or `tci_type` value. If the handle is disclosed to another caller with access to `UPDATE_STASH`, that caller can request updates to the mapped measurement, so callers must treat the handle as sensitive.
+
+Runtime should keep the handle side table in sync with DPE context lifetime. A valid handle entry is only valid while the DPE context at the same index is active or retired. When Runtime clears or reuses a DPE context, it must also clear the handle entry at that index.
```
zhalvorsen, line 139: I don't think we need to stress this too much. `UPDATE_STASH` will check whether the context is still `Active` or `Retired`

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+
+When authorization fails or `SKIP_STASH` is set, Runtime returns an all-zero `measurement_handle`.
+
+## 4. Measurement Handle Mapping
+
+Runtime should store the handle mapping as DPE-side persistent metadata, parallel to the existing context tag side tables:
+
+```text
+measurement_handle_valid: [U8Bool; MAX_HANDLES]
+measurement_handles: [[u8; 16]; MAX_HANDLES]
```
zhalvorsen, line 122: ROM can only stash 8 measurements. I think it would be okay to use the same value here.

DPE_Context_Update_Proposal.md
```diff
@@ -0,0 +1,214 @@
...
+measurement: u8[48]
+svn: u32
+```
+
+The response should include a status result for the update. The exact response shape can mirror existing mailbox status conventions, but the command must fail if the handle is invalid or the mapped context cannot be updated.
+
+The update command should use the following lookup rule:
+
+```text
+index = the entry where:
```
zhalvorsen, line 158: We will need to store the index in persistent data as well when the context is created in DPE. This index won't directly correspond to the index in the DPE list of contexts. The DPE code isn't very explicit about choosing the index when creating a context. It uses a simple `iter.find()` call. As well, Caliptra creates a few contexts before creating the contexts from STASH_MEASUREMENT.

PR #3864, opened 2026-06-10
Expand PAUSER checks for a few mailbox operations

zhalvorsen:
- External Mailbox Commands
- External DPE AXI responses
- AES GCM Decrypt DMA requests


ternst-nv (approved):
LGTM

clundin25 (approved):
Maybe we should elevate this check out of the command implementations and into the command handler match statement?

zhalvorsen:
> Maybe we should elevate this check out of the command implementations and into the command handler match statement?

That could be a nice improvement. Then all of the PL0 checks can be coalesced into a single place for easy finding. Some of those checks are a bit more nuanced though which might make them harder to deal with. For example, the check that PL1 can't request a DPE X509 cert requires a bit of parsing that would be harder to pull out of context.

PR #3865, opened 2026-06-11
Libcaliptra: Adding length check to invoke DPE commands

### LINKED ISSUE #3704: libcaliptra: Validate data_size against struct size in Invoke DPE commands
Both caliptra_invoke_dpe_command and caliptra_invoke_dpe_mldsa87_command in libcaliptra should validate that req->data_size does not exceed the size of the request struct's data payload.

nquarton:


zhalvorsen (approved):

mhatrevi (approved):

nquarton:
I didn't look at this closely enough. The structs here use a variable length data field ([data[0]](https://github.com/chipsalliance/caliptra-sw/blob/d8c52d2eb1da47d191dc45e6b1c30dae569c0d57/libcaliptra/inc/caliptra_types.h#L628)). So this check doesn't work. I don't want to change anything and break current integrators and I can't think of a way to improve this without doing so.

So, I think I am going to drop this.

PR #3866, opened 2026-06-11
mldsa: Make indexing no-panic by construction

ternst-nv:
Change the various call sites where indexing was occuring against arrays to ensure the indexing was always provably valid to the compiler.  This allows the panics to be optimized out, a requirement for the caliptra runtime.

There are no functional changes.

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -193,10 +193,10 @@ fn scalar_ntt(s: &mut Scalar) {
         offset >>= 1;
         for (step_root, chunk) in NTT_ROOTS_MONTGOMERY[step..]
             .iter()
-            .zip(s.c.chunks_exact_mut(2 * offset))
+            .zip(s.c.chunks_exact_mut((2 * offset).min(K_DEGREE)))
         {
-            let (evens, odds) = chunk.split_at_mut(offset);
-            for j in 0..offset {
+            let (evens, odds) = chunk.split_at_mut(chunk.len() / 2);
+            for j in 0..evens.len() {
                 let even = evens[j];
                 let odd = reduce_montgomery((*step_root as u64).wrapping_mul(odds[j] as u64));
                 evens[j] = reduce_once(odd.wrapping_add(even));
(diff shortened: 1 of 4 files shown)


common/crypto/shake/src/shake_impl.rs
```diff
@@ -161,14 +165,27 @@ impl KeccakSt {
...
+                *state_word ^= u64::from_le_bytes([
+                    word_bytes[0],
+                    word_bytes[1],
+                    word_bytes[2],
+                    word_bytes[3],
+                    word_bytes[4],
+                    word_bytes[5],
+                    word_bytes[6],
+                    word_bytes[7],
+                ]);
```
zhalvorsen, line 187: nit: I think you should be able to simplify this line because you are using chunks_exact:


```suggestion
                *state_word ^= u64::from_le_bytes(word_bytes.try_into().unwrap());
```

zhalvorsen (dismissed):

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -692,7 +702,13 @@ fn hint_bit_pack(out: &mut [u8; OMEGA + 8], h: &Vector8) {
     for i in 0..8 {
         for j in 0..K_DEGREE {
             if h.v[i].c[j] != 0 {
-                out[index] = j as u8;
+                // By construction Signatures are sparsely populated and thus will fit within this
+                // construction.  If for some reason an invalid signature is constructed which does
+                // not obey this construction, this will merely result in a corrupted signature
+                // hint field, which would result in signature verification failing.
```
zhalvorsen: I assume we are going to do a `verify` every time we do a `sign`? Maybe say that here so its clear it should be an immediate failure rather than being unbounded

zhalvorsen (dismissed):

zhalvorsen (approved):

common/crypto/mldsa/src/mldsa87.rs
ternst-nv: Updated with a new comment, showing how we know offset wont be surpassed.

PR #3876, opened 2026-06-12
test: make CSR envelope stress test fail fast on a hung boot

swenson:
test_generate_csr_envelop_stress uses unbounded step_until calls, so when a boot fails to reach the expected state the test hangs until nextest's slow timeout (~90 min in nightly), which stalls and fails the whole nightly run.

Add HwModel::step_until_or_timeout (bounded like the existing step_until_boot_status, 30M cycles) and use it for the idevid_csr_ready, ready_for_mb_processing, and ready_for_runtime waits in the CSR envelope path. A hung boot now panics in seconds with the awaited condition named, instead of waiting for the harness timeout.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -819,6 +819,26 @@ fn retry_acquire_mailbox_lock(
     false
 }
 
+/// Step `ctx` until `predicate` is true, calling `step` between checks. Returns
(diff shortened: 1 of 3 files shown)


clundin25 (approved):

zhalvorsen (approved):

swenson:
Thanks!

PR #3885, opened 2026-06-15
[fpga]: Update UDS and FE offsets

clundin25:
Otherwise they collide with the DOT fuses.

This reflects the fuse changes from https://github.com/chipsalliance/caliptra-mcu-sw/pull/1628

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -300,6 +300,42 @@ fn finish_debug_unlock_mailbox_access(drivers: &mut Drivers) {
         .set_ss_dbg_unlock_tap_mailbox_available(false);
 }
 
+/// # Safety
+///
+/// This function is unsafe because it exposes internal cryptographic mailbox
+/// commands directly. It should ONLY be used by integration tests (e.g. the mock runtime).
+pub unsafe fn sha_init(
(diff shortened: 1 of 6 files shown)


jlmahowa-amd (dismissed):

runtime/src/lib.rs
```diff
@@ -300,6 +300,42 @@ fn finish_debug_unlock_mailbox_access(drivers: &mut Drivers) {
         .set_ss_dbg_unlock_tap_mailbox_available(false);
 }
 
+/// # Safety
+///
+/// This function is unsafe because it exposes internal cryptographic mailbox
+/// commands directly. It should ONLY be used by integration tests (e.g. the mock runtime).
+pub unsafe fn sha_init(
```
zhalvorsen, line 307: Is there are reason we don't want to expose this APIs? I think we can do so without the unsafe guards because they are accessible through the mailbox including to PL1

zhalvorsen (approved):

runtime/src/lib.rs
clundin25: I think it's okay to expose but I just wanted to carve them out for tests. 

I can send a follow up and remove the `unsafe`

PR #3887, opened 2026-06-15
[update] ROM: Gate fatal error reporting on recovery reset strap

mhatrevi:
Add SS_STRAP_GENERIC[3][1] as a ROM-controlled subsystem strap that delays fatal error reporting until the recovery interface requests a device reset via DEVICE_RESET.RESET_CTRL == 0x1.

When the strap is set, ROM stops WDT1 and waits for the recovery initiator to request Reset Device before writing CPTRA_FW_ERROR_FATAL. When the strap is clear, fatal error reporting remains unchanged.

### DIFF
--- a/common/src/error_handler.rs
+++ b/common/src/error_handler.rs
@@ -3,10 +3,23 @@ use caliptra_drivers::{
     cprintln, report_fw_error_fatal, report_fw_error_non_fatal, Aes, Ecc384, Hmac, KeyVault,
     Mailbox, MlKem1024, Mldsa87, Sha256, Sha2_512_384, Sha2_512_384Acc, Sha3, SocIfc,
 };
+#[cfg(feature = "rom")]
+use caliptra_drivers::{Dma, DmaRecovery};
+#[cfg(feature = "rom")]
+use caliptra_registers::soc_ifc::SocIfcReg;
 
 #[allow(clippy::empty_loop)]
 pub fn handle_fatal_error(code: u32) -> ! {
     cprintln!("Fatal Error: 0x{:08X}", code);
+
+    #[cfg(feature = "rom")]
+    {
+        let soc_ifc = SocIfc::new(unsafe { SocIfcReg::new() });
+        if soc_ifc.wait_for_device_reset_before_fatal_error() {
+            wait_for_device_reset_before_fatal_error(&soc_ifc);
+        }
+    }
+
     report_fw_error_fatal(code);
     // Populate the non-fatal error code too; if there was a
     // non-fatal error stored here before we don't want somebody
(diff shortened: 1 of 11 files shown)


common/src/error_handler.rs
```diff
@@ -3,10 +3,23 @@ use caliptra_drivers::{
...
+use caliptra_registers::soc_ifc::SocIfcReg;
 
 #[allow(clippy::empty_loop)]
 pub fn handle_fatal_error(code: u32) -> ! {
     cprintln!("Fatal Error: 0x{:08X}", code);
+
+    #[cfg(feature = "rom")]
+    {
+        let soc_ifc = SocIfc::new(unsafe { SocIfcReg::new() });
+        if soc_ifc.wait_for_device_reset_before_fatal_error() {
```
zhalvorsen: This check should probably have some CFI checks

drivers/tests/drivers_integration_tests/main.rs
```diff
@@ -828,6 +835,27 @@ fn test_status_reporter() {
     run_driver_test(&firmware::driver_tests::STATUS_REPORTER);
 }
 
+#[test]
+#[cfg_attr(
+    any(feature = "fpga_realtime", feature = "fpga_subsystem"),
+    ignore = "requires SS_STRAP_GENERIC_3 strap override, which is not supported on FPGA"
```
zhalvorsen, line 841: Is support for this going to be added to MCU ROM and then this can be enabled for subsystem?

common/src/error_handler.rs
```diff
@@ -3,10 +3,23 @@ use caliptra_drivers::{
...
+            wait_for_device_reset_before_fatal_error(&soc_ifc);
```
zhalvorsen: Does this have FIPS implications? I think we might be okay because we will eventually fatally error, but it seems a bit weird to do it with a delay.

What if we get a warm reset while waiting? Then this fatal error will never happen.

common/src/error_handler.rs
ternst-nv: Can this change be targeted to just the DOT operations?  I don't think we want to pause for general errors which are encountered.

Following up more on the DOT specific flow, I'm still not sure I understand how this operation is supposed to help during a cold boot scenario?  If we encounter an error during cold boot, how do we know that a BMC will be able to be started such that it can communicate to the Caliptra Core to tell it to continue with an error?

mhatrevi: Added.

drivers/tests/drivers_integration_tests/main.rs
mhatrevi: Yes, filed https://github.com/chipsalliance/caliptra-sw/issues/3889 for tracking.

common/src/error_handler.rs
mhatrevi: I don’t think this has a FIPS implication. The fatal condition is still terminal; this only delays when CPTRA_FW_ERROR_FATAL is reported so the recovery flow can coordinate a device reset first.

On the warm reset concern: I think the same kind of race can technically happen today, just with a much smaller window between entering the fatal path and writing the fatal error register.

common/src/error_handler.rs
mhatrevi: Discussed offline and we agreed to move the zeroization code in the fatal error handler before the new loop.

common/src/error_handler.rs
mhatrevi: Discussed offline.

zhalvorsen (dismissed):

zhalvorsen (approved):

PR #3904, opened 2026-06-18
[docs]: Define ExportCDI DPE profile behavior more clearly in RT Spec

### LINKED ISSUE #2274: [Caliptra 2.1] Define ExportCDI DPE profile behavior more clearly in RT Spec
The DPE iRoT profile defines the export CDI feature such that the returned value is that actual DICE CDI.

However Caliptra uses a different construction where the CDI is exported into a KeyVault slot and the returned value is jut a handle to that slot. Originally proposed [here](https://github.com/chipsalliance/caliptra-sw/issues/1807).

Caliptra should define a small Profile in the Runtime spec (inherited from the iRoT profile) and override the `export-cdi-format` profile attribute to clarify that the value returned from DeriveContext is a handle, not the actual CDI.

clundin25:
Define the Caliptra DPE Profile which inherits from the OCP DPE iRoT Profile (DPE_PROFILE_IROT_P384_SHA384) but overrides the export-cdi-format attribute to return a KeyVault slot handle instead of the actual CDI.

Update the mailbox API documentation to consistently use "exported CDI handle" instead of "exported CDI".

Fixes #2274

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -2983,7 +2983,7 @@ Command Code: `0x5357_4545` ("SWEE")
 | signature_r        | u8[48]   | The R BigNum of an ECDSA signature.                                        |
 | signature_s        | u8[48]   | The S BigNum of an ECDSA signature.                                        |
 
-The `exported_cdi` can be created by calling `DeriveContext` with the `export-cdi` and `create-certificate` flags.
+The `exported_cdi_handle` can be created by calling `DeriveContext` with the `export-cdi` and `create-certificate` flags.
 
 ### REVOKE\_EXPORTED\_CDI\_HANDLE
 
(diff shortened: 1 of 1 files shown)


clundin25:
@zhalvorsen FYI a follow up PR will document the ML-DSA profile.

runtime/README.md
```diff
@@ -3276,20 +3276,20 @@ At boot Caliptra Runtime FW consumes part of the PL0 active contexts (initially
 Further, it is not allowed for PL1 to call DeriveContext with the intent to change locality to PL0's locality; this would increase the number
 of active contexts in PL0's locality, and hence allow PL1 to DOS PL0.
 
-### DPE profile implementation
+### Caliptra DPE Profile
 
-The DPE iRoT profile leaves some choices up to implementers. This section
-describes specific requirements for the Caliptra DPE implementation.
+Caliptra defines the **Caliptra DPE Profile**, which inherits from the TCG DPE iRoT Profile (`DPE_PROFILE_IROT_P384_SHA384`) with some modifications. This section describes the profile attributes and specific requirements for the Caliptra DPE implementation.
```
zhalvorsen: This is an OCP profile instead of TCG. It would probably be good to link the spec here as well https://github.com/opencomputeproject/Security/tree/main/specifications/dpe-irot-profile

zhalvorsen (dismissed):

runtime/README.md
clundin25: Whoops! Done

zhalvorsen (approved):

Issue #3744, opened 2026-05-18
[PQC] Add software ML-DSA DpeCrypto backend


ternst-nv:
* Caliptra Version(s): 1.3
* Caliptra Mode: Core Mode

### Summary

Add a feature-gated software ML-DSA signer backend for DPE, deriving per-context ML-DSA seeds from the hardware-protected PQ root CDI in KeyVault slot 11.

### Scope

- [x] Add `Signer::MldsaSoftware` to `runtime/src/dpe_crypto.rs`.
- [x] Implement `DpeCrypto::new_mldsa87_sw()`.
- [x] Derive PQ root CDI into `KeyId::KeyId11` using HMAC-384 KDF and label `pq_devid_cdi`.
- [x] Derive per-context ML-DSA seeds from KeyVault slot 11 into software buffers.
- [x] Use `Mldsa87::pub_from_seed()` via the driver wrapper.
- [x] Use `Mldsa87::sign_deterministic()` via the driver wrapper.
- [x] Zeroize derived software seed buffers after use.
- [x] Implement required DPE crypto traits for the software ML-DSA variant.

### Acceptance Criteria

- [x] PQ root CDI is derived into KeyVault slot 11 using label `pq_devid_cdi` and remains hardware-protected.
- [x] Per-context ML-DSA seeds are derived from the PQ CDI and passed to the driver as software buffers.
- [x] ML-DSA public key derivation and signing go through the `Mldsa87` driver wrapper.
- [x] Derived software seed buffers are zeroized.
- [x] Feature-enabled DpeCrypto coverage exercises ML-DSA key derivation and signing.

Issue #3742, opened 2026-05-18
[PQC] Add SET_PQ_SEED Command


ternst-nv:
* Caliptra Version(s): 1.3
* Caliptra Mode: Core Mode

### Summary

Add feature-gated mailbox command definitions, request/response types and handlers for the SET_PQ_SEED Operation.  This should also include adding a section to the PersistentData which stores the PQC Seed.

### Scope

- [ ] Add `mldsa_attestation` feature to `runtime/Cargo.toml`.
- [ ] Add `SET_PQ_SEED`
- [ ] Confirm final command ID values before merging.
- [ ] Add `SetPqSeedReq`.
- [ ] Add `PqDevIdSeed` to `drivers/src/persistent.rs`.
- [ ] Carve storage from `reserved_memory` without shifting unrelated PersistentData fields.
- [ ] Add `pqc_mode` to runtime `Drivers`.
- [ ] Add PQC-specific `CaliptraError` values.

### Acceptance Criteria

- [ ] Final command ID values are documented in the API.
- [ ] Mailbox request/response struct sizes are locked by tests or compile-time assertions.
- [ ] PersistentData layout is unchanged without the feature.
- [ ] Feature-enabled PersistentData layout documents the reserved-memory carve-out.
- [ ] `pqc_mode` is runtime-only state and initializes to disabled.

PR #3913, opened 2026-06-22
persistent data: add PQ DevID seed placeholder

pjhades:
Add a placeholder for PQ DevID seed in persistent data to decouple development work (#3742) on the `SET_PQ_SEED` command and the ML-DSA backend (#3744).

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -46,6 +46,11 @@ pub const AUTH_MAN_IMAGE_METADATA_MAX_SIZE: u32 = 7 * 1024;
 pub const IDEVID_CSR_SIZE: u32 = 1024;
 pub const FMC_ALIAS_CSR_SIZE: u32 = 1024;
 pub const DPE_PL_CONTEXT_LIMITS_SIZE: u32 = 2;
+#[cfg(feature = "mldsa_attestation")]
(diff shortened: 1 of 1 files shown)


ternst-nv (dismissed):
LGTM

pjhades:
@zhalvorsen @clundin25 could you please take a look to make sure the change in persistent data makes sense? Do we need to keep the size of the persistent data consistent in both ML-DSA and no-ML-DSA cases?

clundin25:
@pjhades it could be a sharp edge if you hitlessly update from non-PQC to PQC in case a future item accidentally exists in two locations (because this slice is unreserved for non-PQC).

I think the happy path today it's not so bad but in a few years it might get hard to reason about hitless update compatibility.

IIRC the PQC key ladder starts in FMC, so maybe this isn't a real concern?

pjhades:
> hitless update compatibility.

That makes sense, thanks for the input.

> IIRC the PQC key ladder starts in FMC

It's all in runtime. The SoC manager sets the PQ seed via a mailbox command.

I'm going to make it consistent in both cases.

zhalvorsen (dismissed):

zhalvorsen (dismissed):

clundin25 (dismissed):

zhalvorsen (approved):

PR #3914, opened 2026-06-23
[ci]: Remove MCU commit check for the caliptra-mcu-sw smoke test

clundin25:
This makes the CI nice and green :)

### DIFF
--- a/.github/workflows/fpga-subsystem.yml
+++ b/.github/workflows/fpga-subsystem.yml
@@ -95,6 +95,7 @@ jobs:
 
   check_mcu_commit:
     runs-on: ubuntu-22.04
+    if: github.event_name != 'workflow_dispatch'
     steps:
       - name: Check MCU Commit validity
         run: |


.github/workflows/fpga-subsystem.yml
```diff
@@ -95,6 +95,7 @@ jobs:
 
   check_mcu_commit:
     runs-on: ubuntu-22.04
+    if: github.event_name != 'workflow_dispatch'
```
zhalvorsen, line 98: Does this still get called for this repo's CI?

zhalvorsen (approved):

.github/workflows/fpga-subsystem.yml
clundin25: Yea but this check is optional. It's intent is mainly a signal to reviewers that the MCU ref that is being committed is tracked in a branch that has been reviewed.

PR #3923, opened 2026-06-25
runtime: add ML-DSA DPE crypto backend

pjhades:
This commit introduces feature gated ML-DSA backend for DPE crypto. Major changes include:

- Derive PQ DevID CDI from the PQ seed and store it in key vault under key ID 11.
- Implement the crypto traits for ML-DSA.
- Extend existing crypto workflow to include the ML-DSA paths.

Issue: #3744


runtime/src/drivers.rs
```diff
@@ -487,6 +491,10 @@ impl Drivers {
         // Create a hash of critical initialization values
         let initialization_values_hash = Self::compute_initialization_values_hash(drivers)?;
 
+        // Derive the PQ DevID CDI when ML-DSA is enabled.
+        #[cfg(feature = "mldsa_attestation")]
```
zhalvorsen, line 496: This works on a warm reset when the seed is already present, but this portion of persistent data will be zeroized on a cold reset. What is the current plan for distributing the seed in time for this derivation?

runtime/src/dpe_crypto.rs
```diff
@@ -179,28 +253,56 @@ impl<'a> DpeCrypto<'a> {
...
+        signer: &mut Signer,
+        data: &SignData,
+        seed: Option<&Mldsa87Seed>,
+    ) -> Result<Signature, CryptoError> {
+        let Signer::Mldsa { rt_seed } = signer else {
+            return Err(CryptoError::MismatchedAlgorithm);
+        };
+        let mut sig = [0u8; MLDSA87_SIGNATURE_BYTES];
+        let seed = seed.unwrap_or(rt_seed);
+        Mldsa87::sign_deterministic(seed, data.as_slice(), &mut sig)
```
ternst-nv, line 270: This will need to be updated to handle the new `SignData::ResponseBuffer` variant, in either this or the response in mailbox PR.

runtime/src/dpe_crypto.rs
```diff
@@ -355,27 +489,54 @@ impl CdiManager for DpeCrypto<'_> {
...
+        rt_priv_key: KeyId,
+    },
+    #[cfg(feature = "mldsa_attestation")]
+    Mldsa { rt_seed: Zeroizing<Mldsa87Seed> },
+}
+
+enum DerivedKey {
+    Ec((KeyId, PubKey)),
+    #[cfg(feature = "mldsa_attestation")]
+    Mldsa((Zeroizing<Mldsa87Seed>, Mldsa87PubKey)),
```
ternst-nv: Do we want to store the `Mldsa87PubKey` since it is derivable from the `Seed`?  It takes some time to reconstruct, but also takes up ~4Kb of storage.

runtime/src/dpe_crypto.rs
```diff
@@ -64,13 +71,53 @@ impl<'a> DpeCrypto<'a> {
...
+    pub fn new_mldsa87(
+        sha384: &'a mut Sha384,
+        trng: &'a mut Trng,
+        hmac384: &'a mut Hmac384,
+        key_vault: &'a mut KeyVault,
+        key_id_rt_cdi: KeyId,
+        exported_cdi_slots: &'a mut ExportedCdiHandles,
+    ) -> CaliptraResult<Self> {
+        let mut output = Zeroizing::new(Array4x12::default());
+        hmac384_kdf(
```
ternst-nv: We should check if this kdf is required, since we are not deriving a RTAlias Certificate for the PQ ladder since we don't have space to execute within the FMC.

runtime/src/dpe_crypto.rs
```diff
@@ -342,8 +459,25 @@ impl CdiManager for DpeCrypto<'_> {
...
+                    &cdi,
+                    label,
+                    info,
+                    KEY_ID_DPE_PRIV_KEY,
+                )?));
+            }
+
+            #[cfg(feature = "mldsa_attestation")]
+            Signer::Mldsa { .. } => {
+                let mut seed = Zeroizing::new([0u8; MLDSA87_PRIVATE_SEED_BYTES]);
```
ternst-nv, line 453: [nit] It might be good to move the `Zeroize` to the Seed/Key types themselves to simplify this code a bit and ensure all utilizations are zeroized.

It does require some more refactoring though, so could be held off for a separate PR.

pjhades: Yeah, I was thinking about this too. I prefer handling this in a follow-up refactoring PR, and also including implementing other handy traits, like `Default`, on such types.

runtime/src/dpe_crypto.rs
pjhades: Good catch. The derivation is removed.

ternst-nv (approved):
LGTM

runtime/src/dpe_crypto.rs
```diff
@@ -64,13 +71,39 @@ impl<'a> DpeCrypto<'a> {
...
+
+    #[allow(clippy::too_many_arguments)]
+    #[allow(dead_code)]
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn new_mldsa87(
+        sha384: &'a mut Sha384,
+        trng: &'a mut Trng,
+        hmac384: &'a mut Hmac384,
+        key_vault: &'a mut KeyVault,
+        key_id_rt_cdi: KeyId,
```
zhalvorsen, line 95: If we are taking the key as a key vault slot. We shouldn't need to keep the seed in the persistent data right? Should we revert the change that allocated space in persistent data?

zhalvorsen (approved):

runtime/src/drivers.rs
pjhades: By the current design, the `SET_PQ_SEED` command will be the very first one handled by the runtime on a cold reset, before entering the real command loop, so the seed will be populated when DPE is initialized. Todd shared a nice write-up that includes the workflows: https://github.com/ternst-nv/caliptra-sw/blob/feature/pqc-plan/pqc_architecture_summary.md

runtime/src/dpe_crypto.rs
pjhades: So when ML-DSA is on, the SoC manager is expected to issue the `SET_PQ_SEED` command to set the seed as the very first command. The seed will be stored in persistent data to survive warm resets. The key ID here should always be 11, which points to the `PQ.DevID` (or equivalently `PQ.AliasRT`), which is derived from the seed. So no, we still need the seed in the persistent data.

zhalvorsen: The keyvault also persists across warm resets though, so once `PqDevId` is derived and stored in the keyvault, you shouldn't need the seed anymore

ternst-nv: Hmm, Jing correct me if I'm wrong but we are saving the CDI inside the key vault, not the Seed itself correct? If I understand correctly the KeyVault doesn't provide direct access to the keys themselves, which we need for the Seed, since we need it in software space to sign things.

pjhades: @ternst-nv Yes, we save the CDI in the key vault. The seed is in the persistent data.

@zhalvorsen Ah I see your point.

So if the only use of the seed is to derive the CDI, then it seems we don't need to keep it in the persistent data.

ternst-nv: I think we'll still need the PQ Seed to be able to sign the DPE leaf certificates.

pjhades: Maybe I misunderstood things. I was under the impression that downstream blobs are derived only from the CDI 🤔. Anyway this PR doesn't touch persistent data, but if it turns out necessary we can revert my previous persistent data PR.

ternst-nv: My apologies, I took another look at the RFC and think I was operating under a mistaken understanding of how the DPE Leaf Certificate would work.  I think the suggestion of just using the KeyVault makes sense (as stated we could handle in follow up PR).

pjhades: PR reverting the persistent data change: https://github.com/chipsalliance/caliptra-sw/pull/3941

PR #3924, opened 2026-06-25
Port DPE dependency bump to caliptra-2.0

parvathib:
Bump `caliptra-dpe` commit to latest and migrate the runtime integration to the renamed non-generic DPE APIs. 
Align CFI dependencies and feature wiring with the DPE-pinned CFI revision while preserving the caliptra-2.0 DPE state layout with 32 handles.


drivers/src/mldsa87.rs
```diff
@@ -157,6 +160,16 @@ impl Mldsa87 {
         Ok(pubkey)
     }
 
+    #[cfg(feature = "runtime")]
+    pub fn key_pair_no_pct(
```
zhalvorsen: I don't think we ever want to do this outside of this library right? We should probably remove this from public so it isn't used incorrectly

parvathib: Good catch! I removed the generic public  key_pair_no_pct  API and replaced it with a narrower runtime-only DPE helper that only re-derives the ML-DSA public key from a KV seed. It no longer accepts arbitrary seed variants or a private-key output buffer, so the no-PCT path isn’t exposed as a general driver API.

zhalvorsen (approved):

mhatrevi (approved):

PR #3925, opened 2026-06-25
runtime: response in mbox

ternst-nv:
Alter the runtime to allow responses to be written directly to the mailbox.  This allow greatly reduced stack space for commands with large responses.  Particularly this includes DPE commands which will be required to integrate with MLDSA functionality for the 1.x PQC retrofit.

This also includes the optimization of having responses be published from individual command arms, reducing stack optimizations for commands which previously allocated the union of responses on to the stack.

The overal statistics are as follows:

### Component Sizes (no-PQC baseline, image build)

| Component | Baseline text+data | Branch text+data | Δ bytes | Budget | Baseline Used | Branch Used |
|-----------|-------------------|-----------------|---------|--------|---------------|-------------|
| ROM       | 42,536            | 42,536          | 0       | 49,152 B (48 KiB) | 86.5% | 86.5% |
| FMC       | 20,300            | 20,300          | 0       | 24,576 B (24 KiB) | 82.6% | 82.6% |
| Runtime¹  | 83,228            | 84,716          | **+1,488** | 98,304 B (96 KiB) | 84.7% | 86.2% |

| Image Bundle | Baseline | Branch | Δ bytes | Budget | Baseline Used | Branch Used |
|---|---|---|---|---|---|---|
| image-bundle.bin | 109,412 B (106.85 KiB) | 110,900 B (108.30 KiB) | **+1,488** | 131,072 B (128 KiB) | 83.5% | 84.6% |

> ¹ Image build uses `APP_WITH_UART` feature set (`emu`, `fips_self_test`, `cfi`)
> — `mldsa_attestation` not included.

ROM and FMC are unaffected. The Runtime grows by **+1,488 bytes** (+1.5 pp of
the 96 KiB ICCM budget), primarily from `MboxResponseWriter`,
`finalize_mbox_buffer`, and the per-command direct-write logic.

### mldsa_attestation Feature Impact (Runtime, standalone builds)

| Runtime build             | Baseline text+data | Branch text+data | Budget Used |
|---------------------------|--------------------|-----------------|-------------|
| without mldsa_attestation | 83,228 B           | 82,764 B        | 84.2%       |
| with mldsa_attestation    | 83,228 B²          | 95,460 B        | 97.1%       |
| **Δ (feature cost)**      | —                  | **+12,696 B**   | **+12.9 pp** |

## Stack Usage

Runtime stack = **87,040 bytes** (both branches)

| Command | Baseline bytes | Branch bytes | Δ bytes | Δ % |
|---|---|---|---|---|
| INVOKE_DPE(CertifyKey/X509)  | 44,672 (51.3%) | 11,088 (12.7%) | **-33,584** | **-75%** |
| INVOKE_DPE(GetProfile)       | 36,976 (42.5%) |  6,640  (7.6%) | **-30,336** | **-82%** |
| CERTIFY_KEY_EXTENDED         | 34,256 (39.4%) |  6,592  (7.6%) | **-27,664** | **-81%** |
| SET_AUTH_MANIFEST            | 35,232 (40.5%) | 20,048 (23.0%) | **-15,184** | **-43%** |
| STASH_MEASUREMENT            | 33,680 (38.7%) | 13,184 (15.1%) | **-20,496** | **-61%** |
| AUTHORIZE_AND_STASH          | 33,216 (38.2%) | 12,704 (14.6%) | **-20,512** | **-62%** |
| LMS_VERIFY                   | 26,928 (30.9%) | 11,776 (13.5%) | **-15,152** | **-56%** |
| QUOTE_PCRS                   | 26,432 (30.4%) | 11,536 (13.3%) | **-14,896** | **-56%** |
| POPULATE_IDEV_CERT           | 25,056 (28.8%) |  9,888 (11.4%) | **-15,168** | **-61%** |
| SELF_TEST_GET_RESULTS(+KATs) | 23,936 (27.5%) |  8,768 (10.1%) | **-15,168** | **-63%** |
| SIGN_WITH_EXPORTED_ECDSA     | 22,880 (26.3%) |  7,888  (9.1%) | **-14,992** | **-66%** |
| GET_IDEV_CERT                | 22,128 (25.4%) |  6,976  (8.0%) | **-15,152** | **-68%** |
| GET_LDEV_CERT                | 21,520 (24.7%) |  6,352  (7.3%) | **-15,168** | **-71%** |
| GET_FMC_ALIAS_CERT           | 21,520 (24.7%) |  6,352  (7.3%) | **-15,168** | **-71%** |
| GET_RT_ALIAS_CERT            | 21,264 (24.4%) |  6,096  (7.0%) | **-15,168** | **-71%** |
| DISABLE_ATTESTATION          | 21,024 (24.2%) |  5,840  (6.7%) | **-15,184** | **-72%** |
| GET_PCR_LOG                  | 20,752 (23.8%) |  5,584  (6.4%) | **-15,168** | **-73%** |
| GET_FMC_ALIAS_CSR            | 20,656 (23.7%) |  5,504  (6.3%) | **-15,152** | **-73%** |
| ECDSA384_VERIFY              | 20,576 (23.6%) |  5,392  (6.2%) | **-15,184** | **-74%** |
| FW_INFO                      | 19,872 (22.8%) |  5,248  (6.0%) | **-14,624** | **-74%** |
| GET_IDEV_INFO                | 19,872 (22.8%) |  4,864  (5.6%) | **-15,008** | **-76%** |
| EXTEND_PCR                   | 20,256 (23.3%) |  5,088  (5.8%) | **-15,168** | **-75%** |
| ADD_SUBJECT_ALT_NAME         | 20,160 (23.2%) |  4,992  (5.7%) | **-15,168** | **-75%** |
| DPE_GET_TAGGED_TCI           | 19,904 (22.9%) |  4,928  (5.7%) | **-14,976** | **-75%** |
| DPE_TAG_TCI                  | 19,936 (22.9%) |  4,752  (5.5%) | **-15,184** | **-76%** |
| REVOKE_EXPORTED_CDI_HANDLE   | 19,936 (22.9%) |  4,768  (5.5%) | **-15,168** | **-76%** |
| REALLOCATE_DPE_CONTEXT_LIMITS| 19,904 (22.9%) |  4,736  (5.4%) | **-15,168** | **-76%** |
| INCREMENT_PCR_RESET_COUNTER  | 19,888 (22.8%) |  4,736  (5.4%) | **-15,152** | **-76%** |
| VERSION                      | 19,872 (22.8%) |  4,704  (5.4%) | **-15,168** | **-76%** |
| CAPABILITIES                 | 19,872 (22.8%) |  4,704  (5.4%) | **-15,168** | **-76%** |
| SELF_TEST_START              | 19,872 (22.8%) |  4,704  (5.4%) | **-15,168** | **-76%** |
| SHUTDOWN                     | 19,872 (22.8%) |  4,704  (5.4%) | **-15,168** | **-76%** |
| **Peak (worst case)**        | **44,672 (51.3%)** | **20,048 (23.0%)** | **-24,624** | **-55%** |

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -938,6 +938,7 @@ fn generate_priv_internal(
     );
 }
 
+#[inline(never)]
 fn sign_internal_with_mu(
     out_encoded_signature: &mut [u8; MLDSA87_SIGNATURE_BYTES],
     priv_key: &PrivateKey,
--- a/runtime/src/disable.rs
+++ b/runtime/src/disable.rs
@@ -14,7 +14,7 @@ Abstract:
 
 use crate::Drivers;
 use caliptra_cfi_derive::cfi_impl_fn;
-use caliptra_common::{keyids::KEY_ID_EXPORTED_DPE_CDI, mailbox_api::MailboxResp};
+use caliptra_common::keyids::KEY_ID_EXPORTED_DPE_CDI;
 use caliptra_drivers::{
     hmac384_kdf, Array4x12, CaliptraResult, Ecc384Seed, Hmac384Key, KeyId, KeyReadArgs, KeyUsage,
     KeyWriteArgs,
@@ -25,14 +25,14 @@ pub struct DisableAttestationCmd;
 impl DisableAttestationCmd {
     #[cfg_attr(feature = "cfi", cfi_impl_fn)]
     #[inline(never)]
-    pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<MailboxResp> {
+    pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<()> {
         Self::erase_keys(drivers)?;
         let key_id_rt_cdi = Drivers::get_key_id_rt_cdi(drivers)?;
         Self::zero_cdi(drivers, key_id_rt_cdi)?;
         Self::zero_cdi(drivers, KEY_ID_EXPORTED_DPE_CDI)?;
         Self::generate_dice_key(drivers)?;
         drivers.persistent_data.get_mut().attestation_disabled = U8Bool::new(true);
-        Ok(MailboxResp::default())
+        Ok(())
     }
 
     /// Erase the RT CDI and RT Private Key from the key vault
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -16,6 +16,7 @@ use caliptra_cfi_derive::cfi_impl_fn;
 use caliptra_common::keyids::{
     KEY_ID_DPE_CDI, KEY_ID_DPE_PRIV_KEY, KEY_ID_EXPORTED_DPE_CDI, KEY_ID_TMP,
 };
+use caliptra_dpe_response_buffer::ResponseBufError;
 use caliptra_drivers::{
     hmac384_kdf, sha384::DpeHasher, Array4x12, Ecc384, Ecc384PrivKeyIn, Ecc384PubKey, Ecc384Scalar,
     Ecc384Seed, ExportedCdiEntry, ExportedCdiHandles, Hmac384, KeyId, KeyReadArgs, KeyUsage,
@@ -224,6 +225,20 @@ impl<'a> DpeCrypto<'a> {
         };
         let digest = match data {
             SignData::Digest(Digest::Sha384(crypto::Sha384(digest))) => Ecc384Scalar::from(digest),
+            SignData::ResponseBuffer(buf, range) => {
+                let mut op = sha384
+                    .digest_init()
+                    .map_err(|_| CryptoError::HashError(0))?;
+                buf.read_range(range.clone(), &mut |d| {
+                    op.update(d).map_err(|_| ResponseBufError::Overflow)
+                })
+                .map_err(|_| CryptoError::HashError(0))?;
+
+                let mut digest = Array4x12::default();
+                op.finalize(&mut digest)
+                    .map_err(|_| CryptoError::HashError(0))?;
+                digest
+            }
             SignData::Raw(msg) => sha384.digest(msg).map_err(|_| CryptoError::HashError(0))?,
             _ => return Err(CryptoError::MismatchedAlgorithm),
         };
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -37,9 +37,8 @@ use dpe::{EcdsaAlgorithm, ExportedCdiHandle, U8Bool, MAX_EXPORTED_CDI_SIZE};
 use {
     caliptra_drivers::{
         Mldsa87, Mldsa87PubKey, Mldsa87Seed, MLDSA87_PRIVATE_SEED_BYTES, MLDSA87_PUBLIC_KEY_BYTES,
-        MLDSA87_SIGNATURE_BYTES,
     },
-    crypto::ml_dsa::{MldsaAlgorithm, MldsaPublicKey, MldsaSignature},
+    crypto::ml_dsa::{MldsaAlgorithm, MldsaPublicKey},
     zeroize::Zeroizing,
 };
 
@@ -280,11 +279,9 @@ impl<'a> DpeCrypto<'a> {
     }
 
     #[cfg(feature = "mldsa_attestation")]
-    fn sign_helper_mldsa(data: &SignData, seed: &Mldsa87Seed) -> Result<Signature, CryptoError> {
-        let mut sig = [0u8; MLDSA87_SIGNATURE_BYTES];
-        Mldsa87::sign_deterministic(seed, data.as_slice(), &mut sig)
-            .map_err(|e| CryptoError::CryptoLibError(u32::from(e)))?;
-        Ok(Signature::Mldsa(MldsaSignature(sig)))
+    fn sign_helper_mldsa(_data: &SignData, _seed: &Mldsa87Seed) -> Result<Signature, CryptoError> {
+        // TODO: Issue #3936 - Implement MLDSA DPE Hash.
+        Err(CryptoError::MismatchedAlgorithm)
     }
 }
 
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -28,6 +28,7 @@ pub mod handoff;
 mod hmac;
 pub mod info;
 mod invoke_dpe;
+pub mod mbox_response_writer;
 mod pcr;
 mod populate_idev;
 mod reallocate_dpe_context_limits;
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -81,6 +81,7 @@ pub use get_fmc_alias_csr::GetFmcAliasCsrCmd;
 pub use get_idev_csr::GetIdevCsrCmd;
 pub use info::{FwInfoCmd, IDevIdInfoCmd};
 pub use invoke_dpe::InvokeDpeCmd;
+pub use mbox_response_writer::MboxResponseWriter;
 pub use pcr::{GetPcrLogCmd, IncrementPcrResetCounterCmd};
 pub use reallocate_dpe_context_limits::ReallocateDpeContextLimitsCmd;
 pub use set_auth_manifest::SetAuthManifestCmd;
@@ -89,7 +90,7 @@ pub use stash_measurement::StashMeasurementCmd;
 pub use verify::Mldsa87VerifyCmd;
 pub use verify::{EcdsaVerifyCmd, LmsVerifyCmd};
 pub mod packet;
-use caliptra_common::mailbox_api::{CommandId, MailboxReqHeader, MailboxResp};
+use caliptra_common::mailbox_api::{CommandId, MailboxReqHeader, MailboxRespHeader};
 use packet::{copy_from_mbox, copy_to_mbox};
 use zerocopy::{FromZeros, IntoBytes};
 pub mod tagging;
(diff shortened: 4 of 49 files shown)


runtime/src/disable.rs
```diff
@@ -25,14 +25,14 @@ pub struct DisableAttestationCmd;
...
-    pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<MailboxResp> {
+    pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<()> {
         Self::erase_keys(drivers)?;
         let key_id_rt_cdi = Drivers::get_key_id_rt_cdi(drivers)?;
         Self::zero_cdi(drivers, key_id_rt_cdi)?;
         Self::zero_cdi(drivers, KEY_ID_EXPORTED_DPE_CDI)?;
         Self::generate_dice_key(drivers)?;
         drivers.persistent_data.get_mut().attestation_disabled = U8Bool::new(true);
-        Ok(MailboxResp::default())
+        Ok(())
```
zhalvorsen, line 35: This should probably write a blank `MailboxResp` to avoid misrepresenting information left over from the command

runtime/src/lib.rs
```diff
@@ -204,7 +206,8 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
         CommandId::STASH_MEASUREMENT => StashMeasurementCmd::execute(drivers),
         CommandId::DISABLE_ATTESTATION => {
             copy_from_mbox(drivers, MailboxReqHeader::new_zeroed().as_mut_bytes())?;
-            DisableAttestationCmd::execute(drivers)
+            DisableAttestationCmd::execute(drivers)?;
+            copy_to_mbox(drivers, MailboxRespHeader::default().as_mut_bytes())
```
zhalvorsen, line 213: Similar to the previous comment. I'd prefer to leave this logic in the command. This would keep it consistent with the other commands. If for some strange reason we decide to add response data to these commands, the developer will probably add it to the command and it could get overwritten because it is written in two places.

I think this should also be done in `CommandId::VERSION`, `CommandId::CAPABILITIES`, `CommandId::SELF_TEST_START`, and  `CommandId::SELF_TEST_GET_RESULTS` even though that might change what was done previously

runtime/src/sign_with_exported_ecdsa.rs
```diff
@@ -128,6 +126,10 @@ impl SignWithExportedEcdsaCmd {
         } else {
             return Err(CaliptraError::RUNTIME_SIGN_WITH_EXPORTED_ECDSA_INVALID_SIGNATURE);
         }
-        Ok(MailboxResp::SignWithExportedEcdsa(resp))
+
+        // Explicitely drop crypto and pdata so the multable borrow to drivers also ends.
```
zhalvorsen: nit: typo:
```suggestion
        // Explicitely drop crypto and pdata so the mutable borrow to drivers also ends.
```

runtime/src/mbox_response_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+
+impl MboxResponseWriter {
+    const BASE: usize = memory_layout::MBOX_ORG as usize;
+    const SIZE: usize = memory_layout::MBOX_SIZE as usize;
+
+    /// Return a pointer to `MBOX_ORG + byte_offset` cast to `*mut u32`.
+    ///
+    /// `byte_offset` must be word-aligned.
+    #[inline(always)]
+    fn word_ptr(&self, byte_offset: usize) -> *mut u32 {
```
zhalvorsen: All of the functions in `MboxResponseWriter` should probably be marked unsafe the caller is responsible for calling them in safe manners:

* `word_ptr` needs the offset to be within the mailbox
* `read_word` needs offset within mailbox and word aligned
* `write_word` needs offset within mailbox and word aligned

runtime/src/mbox_response_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+            .checked_add(bytes.len())
+            .ok_or(ResponseBufError::Overflow)?;
+        if end > Self::SIZE {
+            return Err(ResponseBufError::Overflow);
+        }
+
+        let mut src = bytes;
+        let mut pos = offset;
+
+        // Leading unaligned bytes.
```
zhalvorsen, line 76: It would be good to have some unit tests for these byte manipulations

runtime/src/mbox_response_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+            let leading = (4 - misalign).min(src.len());
+            for &b in &src[..leading] {
+                let abs_b = pos;
+                let word_off = abs_b & !3;
+                let lane = abs_b & 3;
+                let mut word = self.read_word(word_off);
+                word = (word & !(0xFF << (lane * 8))) | ((b as u32) << (lane * 8));
+                self.write_word(word_off, word);
+                pos += 1;
+            }
```
zhalvorsen: nit: this could be simplified a little and reduces the number of mailbox accesses


```suggestion
            let word_off = pos & !3;

            let mut word_bytes = self.read_word(word_off).to_le_bytes();            
            word_bytes[misalign..misalign + leading].copy_from_slice(&src[..leading]);
            self.write_word(word_off, u32::from_le_bytes(word_bytes));
            
            pos += leading;
            src = &src[leading..];
        }
```

runtime/src/mbox_response_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+        // Trailing partial word.
+        for &b in src {
+            let abs_b = pos;
+            let word_off = abs_b & !3;
+            let lane = abs_b & 3;
+            let mut word = self.read_word(word_off);
+            word = (word & !(0xFF << (lane * 8))) | ((b as u32) << (lane * 8));
+            self.write_word(word_off, word);
+            pos += 1;
+        }
```
zhalvorsen, line 115: nit: A similar change could be done here:


```suggestion
        if !src.is_empty() {
            let word_off = pos & !3;
            let mut word_bytes = self.read_word(word_off).to_le_bytes();
            word_bytes[..src.len()].copy_from_slice(src);
            self.write_word(word_off, u32::from_le_bytes(word_bytes));
        }
```

runtime/src/disable.rs
ternst-nv: Unfortunately this `execute` command is called by consumers outside of just the Command loop, and as such if we populate the response within the execute command we have various call stacks fail because the response is populated prematurely.

We could fix this be refactoring the `execute` body to just have the parse/publish and then a separate method handle actually implementing the logic and be generally callable.  I had thought that might be a bit much for this PR, but happy to make the change here/in a follow up if you think it is worth while.

zhalvorsen: Sounds good, thanks for keep it small. I think a follow up is a good option.

runtime/src/lib.rs
ternst-nv: I took a look at these, for `VERSION` it has the same multi-caller issue, I can add it to the refactor PR.  `CAPABILITIES` looks to put the Response on the buffer in the `execute` function, both of the `SELF_TEST` commands do not have execute functions.  I could add them as part of the refactor if that would be good?

zhalvorsen: Sounds good, thanks!

runtime/src/dpe_crypto.rs
```diff
@@ -266,8 +282,34 @@ impl<'a> DpeCrypto<'a> {
...
+                let mut shake256 = Shake256::new();
+                buf.read_range(range.clone(), &mut |d| {
+                    shake256.absorb(d);
+                    Ok(())
+                })
+                .map_err(|_| CryptoError::HashError(0))?;
+
+                let mut mu = [0; MLDSA87_MU_BYTES];
+                shake256.squeeze(&mut mu);
+                mu
```
zhalvorsen: Correct me if I'm wrong, I don't think this is the full mu calculation. Don't we need to absorb tr before the message? We would need a similar change in `Raw` as well.

zhalvorsen (dismissed):

runtime/src/dpe_crypto.rs
ternst-nv: Good catch, I've updated to expose the tr value and absorb it into the measurement.

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -1148,6 +1150,24 @@ fn verify_with_mu_context(
 
 /* Public API. */
 
+pub fn mldsa87_priv_key_hash(
```
zhalvorsen: Maybe I'm misunderstanding the internals. [Fips204](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf) calls it a hash of the public key:

> Before the message 𝑀 is signed, it is concatenated with the public-key hash 𝑡𝑟 and hashed down to a
64-byte message representative 𝜇 using H.

It might be good to have a function that calculates the mu given a public key and message (maybe one with a byte slice `&[u8]` and one with a `ResponseBuffer`).

```Rust
fn calculate_mu(pub_key: &[u8; MLDSA87_PUBLIC_KEY_BYTES], msg: &[u8]) -> [u8; MLDSA87_MU_BYTES] {
    let mut tr = [0u8; K_TR_BYTES];
    let mut shake256 = Shake256::new();
    shake256.absorb(pub_key);
    shake256.squeeze(&mut tr);

    let mut mu = [u8; MLDSA87_MU_BYTES];
    shake256.absorb(&tr);
    // Domain separator: "pure" (non-prehashed) ML-DSA
    shake256.absorb(&[0u8]);
    // Context length: 0 because it is empty
    shake256.absorb(&[0u8]);
    shake256.absorb(msg);
    shake256.squeeze(&mut mu);
    mu
}

zhalvorsen: @ternst-nv created https://github.com/chipsalliance/caliptra-sw/issues/3936 to address this later

zhalvorsen (dismissed):

zhalvorsen (approved):

PR #3927, opened 2026-06-26
Add gated SET_PQ_SEED runtime support

kaevans-nvidia:
Add the mldsa_attestation-gated SET_PQ_SEED mailbox API, persistent PQ IDevID seed storage carved from existing reserved firmware data, and a PL0-only runtime handler that enables PQC mode after storing the seed. Cover the new request layout, persistent-data carve-out, and PL0/PL1 runtime behavior with targeted tests.


pjhades:
Hi @kaevans-nvidia , can you please rebase the PR on the `caliptra-1.x` branch?

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
```diff
@@ -368,6 +372,72 @@ fn test_stash_measurement_cannot_be_called_from_pl1() {
     );
 }
 
+#[cfg(feature = "mldsa_attestation")]
+#[test]
+fn test_set_pq_seed_from_pl0_enables_pqc_mode_once() {
```
zhalvorsen: In general, this file only contains checks for PAUSER specific tests like the second test. Usually we create a separate test file for each command. This test would probably fit better in the separate file.

runtime/src/drivers.rs
```diff
@@ -145,6 +155,8 @@ impl Drivers {
             lms: Lms::default(),
             trng,
             persistent_data: PersistentDataAccessor::new(),
+            #[cfg(feature = "mldsa_attestation")]
+            pqc_mode: PqcMode::Disabled,
```
zhalvorsen: These values do not persist across warm resets or hitless updates. I think we need this to persist. Can you move this to `persistent_data`? It would also probably be good to turn this into a bit flag to save memory

zhalvorsen (dismissed):

api/src/mailbox.rs
```diff
@@ -10,6 +10,10 @@ use crate::CaliptraApiError;
 use caliptra_registers::mbox;
 use ureg::MmioMut;
 
+/// PQ IDevID seed size in bytes for SET_PQ_SEED.
+#[cfg(feature = "mldsa_attestation")]
+pub const SET_PQ_SEED_SEED_SIZE: usize = 32;
```
pjhades: Should we reuse `PQ_DEVID_SEED_SIZE` from drivers/src/persistent.rs here?

api/src/mailbox.rs
```diff
@@ -26,6 +30,9 @@ impl CommandId {
     pub const STASH_MEASUREMENT: Self = Self(0x4D454153); // "MEAS"
     pub const INVOKE_DPE: Self = Self(0x44504543); // "DPEC"
     pub const DISABLE_ATTESTATION: Self = Self(0x4453424C); // "DSBL"
+    /// Set the PQ IDevID seed. Command ID is "PQSD" (0x50515344).
```
pjhades: Nit: maybe removing this comment since the pattern is already self-explanatory?

runtime/src/set_pq_seed.rs
```diff
@@ -0,0 +1,45 @@
...
+            Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL)?;
+        }
+
+        if drivers.pqc_mode != PqcMode::Disabled {
+            Err(CaliptraError::RUNTIME_SET_PQ_SEED_ALREADY_SET)?;
+        }
+
+        let mut cmd = SetPqSeedReq::new_zeroed();
+        crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())
+            .map_err(|_| CaliptraError::RUNTIME_SET_PQ_SEED_INVALID_REQUEST)?;
```
pjhades: `copy_from_mbox` already returns `CaliptraResult<()>` so we can just do `?` on it to keep consistent with other commands.

runtime/src/drivers.rs
```diff
@@ -73,6 +73,13 @@ pub enum PauserPrivileges {
     PL1,
 }
 
+#[cfg(feature = "mldsa_attestation")]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum PqcMode {
```
pjhades: Is there any specific reason of using an enum here?

error/src/lib.rs
```diff
@@ -1263,6 +1263,16 @@ impl CaliptraError {
             0x000E0062,
             "Runtime Error: ML-DSA-87 signature verification failed"
         ),
+        (
+            RUNTIME_SET_PQ_SEED_INVALID_REQUEST,
+            0x000E009F,
```
pjhades: I probably missed some context and others can comment here. Should we use consecutive error values?

pjhades (commented):
LGTM, left some comments.

runtime/src/lib.rs
```diff
@@ -255,6 +259,8 @@ fn handle_command(drivers: &mut Drivers) -> CaliptraResult<MboxStatusE> {
             FipsShutdownCmd::execute(drivers)
         }
         CommandId::SET_AUTH_MANIFEST => SetAuthManifestCmd::execute(drivers),
+        #[cfg(feature = "mldsa_attestation")]
+        CommandId::SET_PQ_SEED => SetPqSeedCmd::execute(drivers),
```
ternst-nv: We do not want `SET_PQ_SEED` to run from the main command loop, since it should execute prior to the DPE initialization.  Could we have this return an invalid or unsupported command here?

In the follow issue we'll patch this into the initialization flow, so that it is called prior to dpe init.

runtime/tests/runtime_integration_tests/test_pauser_privilege_levels.rs
kaevans-nvidia: Duly noted, thanks. I Moved this out of the PAUSER-specific test file. Since SET_PQ_SEED should not execute from the runtime command loop yet, I replaced the PL0/PL1 runtime behavior tests with a dedicated SET_PQ_SEED test that verifies the command is rejected as unimplemented until the pre-DPE init flow wires it in.

runtime/src/drivers.rs
kaevans-nvidia: Done. I removed the runtime-only `Drivers::pqc_mode state` and moved this into `persistent_data` as a single status flag byte carved from reserved memory. Thanks for the catch.

api/src/mailbox.rs
kaevans-nvidia: Good point. The API crate cannot directly reuse `caliptra-drivers::PQ_DEVID_SEED_SIZE` without reversing the dependency direction, so I changed the API constant to use `MLDSA87_PRIVATE_SEED_BYTES` instead of duplicating the literal 32.

api/src/mailbox.rs
kaevans-nvidia: Np and agreed. Done. Thanks

runtime/src/set_pq_seed.rs
kaevans-nvidia: Yeah thats a good point. Changed so `copy_from_mbox` now uses ? directly and preserves the underlying mailbox error.

runtime/src/drivers.rs
kaevans-nvidia: I suppose not specific reason. I changed so it no longer uses an enum. This is now a persistent bit flag.

error/src/lib.rs
kaevans-nvidia: I suppose we should. 

I removed the now-unused invalid-request error and assigned the remaining `SET_PQ_SEED` error the next consecutive runtime value.

runtime/src/lib.rs
kaevans-nvidia: I see, yeah good point. Changed so `SET_PQ_SEED` returns `RUNTIME_UNIMPLEMENTED_COMMAND` from the runtime command loop. I left the handler staged for the follow-up that will call it before DPE initialization.

ternst-nv (dismissed):
LGTM

kaevans-nvidia:
Had to rebase over new `dpe/ml-dsa` feature

ternst-nv (approved):
LGTM

PR #3929, opened 2026-06-29
drivers: add trait impls and APIs to ML-DSA types

pjhades:
Add useful trait impls like `Default` and `Zeroize`, together with helper APIs to ML-DSA types to enable code simplification on user side.

### DIFF
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -48,6 +48,10 @@ pub struct Mldsa87PrivKey([u8; MLDSA87_PRIVATE_KEY_BYTES]);
 #[derive(KnownLayout, Immutable, FromBytes)]
 pub struct Mldsa87Signature([u8; MLDSA87_SIGNATURE_BYTES]);
 
+#[repr(transparent)]
+#[derive(KnownLayout, Immutable, FromBytes)]
+pub struct Mldsa87Mu([u8; MLDSA87_MU_BYTES]);
+
 macro_rules! impl_helper_traits {
     ($name:ty) => {
         impl $name {
@@ -85,8 +89,7 @@ impl_helper_traits!(Mldsa87Seed);
 impl_helper_traits!(Mldsa87PubKey);
 impl_helper_traits!(Mldsa87PrivKey);
 impl_helper_traits!(Mldsa87Signature);
-
-pub type Mldsa87Mu = [u8; MLDSA87_MU_BYTES];
+impl_helper_traits!(Mldsa87Mu);
 
 /// Software ML-DSA-87 engine.
 ///
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -19,6 +19,8 @@ Abstract:
 
 use crate::CaliptraResult;
 use caliptra_mldsa::Mldsa87 as Mldsa87Sw;
+use core::ops::{Deref, DerefMut};
+use zeroize::Zeroize;
 
 pub use caliptra_mldsa::{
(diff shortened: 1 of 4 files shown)


drivers/src/mldsa87.rs
```diff
@@ -19,23 +19,74 @@ Abstract:
...
+pub struct Mldsa87Signature([u8; MLDSA87_SIGNATURE_BYTES]);
+
+macro_rules! impl_helper_traits {
+    ($name:ty) => {
+        impl $name {
+            pub const fn new(data: [u8; core::mem::size_of::<$name>()]) -> Self {
+                Self(data)
+            }
+
+            pub fn from_ref(data: &[u8; core::mem::size_of::<$name>()]) -> &Self {
```
ternst-nv: Could we use one of the `ZeroCopy` traits (I think FromBytes is the one we'd want), instead?  It does things like ensure the type is transitively transparent/creatable from bytes, so we don't have to worry about future changes accidentally violating this constraint.

ternst-nv (commented):
LGTM

drivers/src/mldsa87.rs
pjhades: Thanks! I didn't realize `ZeroCopy` at all. Will make the change.

zhalvorsen (dismissed):

ternst-nv (dismissed):
LGTM

drivers/src/mldsa87.rs
```diff
@@ -19,23 +19,72 @@ Abstract:
...
-pub type Mldsa87PubKey = [u8; MLDSA87_PUBLIC_KEY_BYTES];
+#[repr(transparent)]
+#[derive(KnownLayout, Immutable, FromBytes)]
+pub struct Mldsa87PubKey([u8; MLDSA87_PUBLIC_KEY_BYTES]);
 
 /// ML-DSA-87 FIPS 204 encoded private key (4,896 bytes).
-pub type Mldsa87PrivKey = [u8; MLDSA87_PRIVATE_KEY_BYTES];
+#[repr(transparent)]
+#[derive(Zeroize)]
+pub struct Mldsa87PrivKey([u8; MLDSA87_PRIVATE_KEY_BYTES]);
```
clundin25: Does it make sense to make this zeroize on drop?

pjhades: Since the private key is currently only generated in KATs, the `Zeroize` was defensive rather than strictly necessary. Removed in 4f04190.

clundin25: Hmm that seems like a footgun, should that invariant change

pjhades: Sorry could you please be a bit more specific?

clundin25: I mean that should new code get introduced that uses `Mldsa87PrivKey` Zeroizing it on drop may be a miss

zhalvorsen: Maybe there is confusion on what `Zeroize` does. That only adds functionality to call `zeroize` or `from_zeros` on the object. Carl is saying it would be good to also add `ZeroizeOnDrop` which will automatically zeroize the memory backing the object once it goes out of scope.

```Rust
#[derive(Zeroize, ZeroizeOnDrop)]
```

pjhades: OK, it seems I misunderstood the trait itself. I thought deriving `Zeroize` would be sufficient but it turned out that I should derive `ZeroizeOnDrop` instead.

pjhades: @zhalvorsen oops I didn't see your comment before posting my own..yeah exactly I had the confusion before.

zhalvorsen (approved):

clundin25 (approved):

pjhades:
Thanks for the review! Can we merge this?

PR #3934, opened 2026-07-01
[docs] Add stable key derivation diagrams

mhatrevi:
Add SVG diagrams for ROM stable root derivation and the CM_DERIVE_STABLE_KEY mailbox flow, then link them from the ROM and runtime READMEs.


rom/dev/doc/svg/cm-derive-stable-key.svg
bluegate010, line 1: Will just say that it would be nice if new diagrams were made using [Draw.io](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio). Keeps the diagram source embedded within the .svg file, so others can edit without needing a separate Vizio source file.

Won't block this PR though.

bluegate010 (approved):

clundin25 (approved):

rom/dev/doc/svg/cm-derive-stable-key.svg
mhatrevi: Ack; I'll looking into creating a separate PR later to convert all the current diagrams via draw.io.

PR #3935, opened 2026-07-01
Feature gates to support multiple HW revisions on main

embediver:
Introduce a `cfg` flag that switch between multiple HW revisions, by having multiple generated register crates under `hw/`.
The cfg flag can be either set via the `CALIPTRA_HW_REV` environment variable or by specifying via the --rev flag in _xtask_ and the _caliptra-builder_ . The accepted revisions are `latest` and `2.1` for now. They point to the same commit at this point, until we move `latest` forward.

For now I added the v2.1.2 tagged HW as revision, so we could then move `latest` towards v2.2.
The registers crate also checks the `2.0` revision, but that is currently only a placeholder (yielding a `todo!` compile error) until we decide to integrate v2.0 specific changes into `main`.

The workflow for supporting the `revision` cfg in other crates that might need it, is is as follows:
- Add a `[lints.rust.unexpected_cfgs]` section to the _Cargo.toml_
- Add a _build.rs_ that parses the `CALIPTRA_HW_REV` env-var and emits corresponding `cargo::rustc-cfg` statements

See the `registers` crate for reference.

Also various `build.sh` scripts have been cleaned up and been replaced with a `xtask build` command.


api/test-fw/Cargo.toml
```diff
@@ -17,6 +17,8 @@ zerocopy.workspace = true
 [features]
 emu = ["caliptra-test-harness/emu"]
 riscv = ["caliptra-test-harness/riscv"]
+rev-latest = ["caliptra-registers/rev-latest", "caliptra-api/rev-latest"]
+rev-2_1 = ["caliptra-registers/rev-2_1", "caliptra-api/rev-2_1"]
```
zhalvorsen: nit: The 1.x branch uses "hw-1.0" which I think I slightly prefer. What do you think?

```suggestion
hw-2.1 = ["caliptra-registers/hw-2.1", "caliptra-api/hw-2.1"]
```

api/test-fw/Cargo.toml
```diff
@@ -17,6 +17,8 @@ zerocopy.workspace = true
 [features]
 emu = ["caliptra-test-harness/emu"]
 riscv = ["caliptra-test-harness/riscv"]
+rev-latest = ["caliptra-registers/rev-latest", "caliptra-api/rev-latest"]
```
zhalvorsen: Should we assume "latest" unless we are given a different feature flag otherwise? That could reduce the number of feature flags by one at least.

sw-emulator/app/Cargo.toml
```diff
@@ -20,4 +20,8 @@ clap.workspace = true
 gdbstub_arch.workspace = true
 gdbstub.workspace = true
 hex.workspace = true
-tock-registers.workspace = true
\ No newline at end of file
+tock-registers.workspace = true
+
+[features]
+rev-latest = ["caliptra-registers/rev-latest", "caliptra-hw-model/rev-latest"]
```
zhalvorsen: If possible, it would be nice to have these be selected at runtime for the `sw-emulator`. This would probably be best as a separate change though

zhalvorsen (dismissed):

clundin25:
Using feature flags like this leaks the hardware revision to the rest of the creates and adds a lot of overhead (everytime we add a revision will need to repeat all of this work).

What about using the build script + env variable to select the HW revision? We will need this to download the correct bitstream, and it reduces how the feature flags proliferate.

Here is an example: https://github.com/clundin25/caliptra-sw/commit/aefe2f36c55f562e3756e6a5f1f5b327c477dac6

CC @zhalvorsen

zhalvorsen:
> Using feature flags like this leaks the hardware revision to the rest of the creates and adds a lot of overhead (everytime we add a revision will need to repeat all of this work).
> 
> What about using the build script + env variable to select the HW revision? We will need this to download the correct bitstream, and it reduces how the feature flags proliferate.
> 
> Here is an example: [clundin25@aefe2f3](https://github.com/clundin25/caliptra-sw/commit/aefe2f36c55f562e3756e6a5f1f5b327c477dac6)
> 
> CC @zhalvorsen

That does simplify it a lot. We could also set the environment for the whole workspace pretty easily using the .[cargo/config](https://github.com/chipsalliance/caliptra-sw/blob/caliptra-1.x/.cargo/config) similar to the way we were setting the MAX_HANDLES variable in 1.x

embediver:
> Using feature flags like this leaks the hardware revision to the rest of the creates and adds a lot of overhead (everytime we add a revision will need to repeat all of this work).
> 
> What about using the build script + env variable to select the HW revision? We will need this to download the correct bitstream, and it reduces how the feature flags proliferate.
> 
> Here is an example: [clundin25@aefe2f3](https://github.com/clundin25/caliptra-sw/commit/aefe2f36c55f562e3756e6a5f1f5b327c477dac6)
> 
> CC @zhalvorsen

Valid point, I was already thinking about how we could simplify this.
But can we get away without feature flags in most of the crates?
If we need different logic depending on hw revisions in some of the crates, we would then also have to set this up in the build.rs there right?

clundin25:
> > Using feature flags like this leaks the hardware revision to the rest of the creates and adds a lot of overhead (everytime we add a revision will need to repeat all of this work).
> > What about using the build script + env variable to select the HW revision? We will need this to download the correct bitstream, and it reduces how the feature flags proliferate.
> > Here is an example: [clundin25@aefe2f3](https://github.com/clundin25/caliptra-sw/commit/aefe2f36c55f562e3756e6a5f1f5b327c477dac6)
> > CC @zhalvorsen
> 
> Valid point, I was already thinking about how we could simplify this. But can we get away without feature flags in most of the crates? If we need different logic depending on hw revisions in some of the crates, we would then also have to set this up in the build.rs there right?

Yes that's a good point. I think it makes sense to use the `build.rs` approach since Rust doesn't support "workspace" wide feature flags. 

It may also be useful to have some kind of facade that handles disparities between 2.0 and 2.1 (e.g. register renames)

ternst-nv:
I'm curious if we want to place this within `build.rs` or within the `cargo xtask` framework, and then have it appropriately set the environment variables so they are exported to each crate/binary being compiled.  That allows us to avoid doing both the `feature_flag` work in each toml and the build.rs for each crate, as well as allowing targeted compilation in both functional crates and the facade crate between registers.

The downside to this is it places more logic within the xtask, meaning those who choose to do manual builds have a more complex CLI they need to keep in mind, but xtask is how subsystem handles builds and would be a simple standard to unify on.

clundin25:
> I'm curious if we want to place this within `build.rs` or within the `cargo xtask` framework, and then have it appropriately set the environment variables so they are exported to each crate/binary being compiled. That allows us to avoid doing both the `feature_flag` work in each toml and the build.rs for each crate, as well as allowing targeted compilation in both functional crates and the facade crate between registers.
> 
> The downside to this is it places more logic within the xtask, meaning those who choose to do manual builds have a more complex CLI they need to keep in mind, but xtask is how subsystem handles builds and would be a simple standard to unify on.

+1, xtask should be the preferred method of building. We should clean up the CI & remove scripts / makefiles, so we can centralize on xtask.

embediver:
I updated the PR description and changed the PR to ready.
Currently only the build CI workflow runs both `latest` and `2.1` HW revisions.
I would like to get this PR done now and then do the remaining CI work for this in a follow up PR.

builder/bin/image_gen.rs
```diff
@@ -48,12 +48,20 @@ fn main() {
...
     let args = cmd.get_matches_mut();
 
     // Print help if the provided args did not create anything.
     let mut valid_cmd = false;
 
+    let rev = args.get_one::<String>("hw-revision");
+    if let Some(rev) = rev {
+        unsafe {
+            std::env::set_var("CALIPTRA_HW_REV", rev);
+        }
```
clundin25, line 64: Can you add a safety comment that clarifies this is safe because at this stage of the binary it is single threaded

xtask/src/main.rs
```diff
@@ -125,6 +146,14 @@ fn main() {
         LevelFilter::Info
     };
     let _ = SimpleLogger::new().with_level(level).init();
+
+    if let Some(rev) = cli.rev {
+        log::debug!("Setting CALIPTRA_HW_REV to `{}`", rev);
+        unsafe {
+            std::env::set_var("CALIPTRA_HW_REV", rev);
+        }
```
clundin25, line 156: Similar safety comment here

clundin25 (dismissed):

builder/bin/image_gen.rs
embediver: Sure.

The alternative would be to set the variable for every command that gets run, but this seemed like the easier more maintainable way.
Tell me if you would prefer the other method, or if we should avoid the arguments completely.

clundin25: I think it's reasonable given the program is not multi-threaded, so the code is safe.

That said, moving it to the invocation point may make it easier to determine _why_ the environment variable is set and what it mutates. 

As is, it may not be clear that this modifies the register build script.

I don't feel very strongly about this and am happy to merge as is with or without a followup

ternst-nv:
The approach for build looks good to me, thanks!  What are your thoughts on extending this to running tests?

registers/src/lib.rs
```diff
@@ -1,5 +1,18 @@
 // Licensed under the Apache-2.0 license
 //
 #![no_std]
+#![cfg_attr(revision = "latest", doc = "Revision: _latest_")]
+#![cfg_attr(revision = "2.1", doc = "Revision: _2.1_")]
```
zhalvorsen: A similar comment to one I had earlier. I'm worried "rev" or "revision" could be too generic. Without proper context it could sound like FMC version, RT version or ROM version. Can we change these to reference hardware more directly? Something like:

```Rust
#![cfg_attr(hw = "2.1", doc = "Hardware Revision: _2.1_")]
```

zhalvorsen (dismissed):

embediver:
> The approach for build looks good to me, thanks! What are your thoughts on extending this to running tests?

I implemented the build job with a simple matrix and that seems to be a good option for other jobs too.
I thought a bit how this could be improved, but running tests for every hw revision is probably unavoidable.
Maybe some of the tests can be moved to a nightly schedule so only `latest` tests run on every push.

registers/src/lib.rs
embediver: changed it to `hw_rev`, that way it aligns with the env variable `CALIPTRA_HW_REV`.

zhalvorsen (approved):
Great job, thanks man!

clundin25 (approved):

clundin25:
> > The approach for build looks good to me, thanks! What are your thoughts on extending this to running tests?
> 
> I implemented the build job with a simple matrix and that seems to be a good option for other jobs too. I thought a bit how this could be improved, but running tests for every hw revision is probably unavoidable. Maybe some of the tests can be moved to a nightly schedule so only `latest` tests run on every push.

When we add this for FPGAs, let's only run latest on PR and do the rest as part of the nightly.

PR #3942, opened 2026-07-03
dpe: enable hybrid feature

pjhades:
Enable the "hybrid" feature of caliptra-dpe, and bump the size of receive buffers for INVOKE_DPE and CERTIFY_KEY commands to match caliptra-sw 2.0. This will be the last step of upgrading caliptra-dpe to the version with ML-DSA support.

Issue: #3741

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -730,7 +730,7 @@ pub struct CertifyKeyExtendedResp {
     pub certify_key_resp: [u8; CertifyKeyExtendedResp::CERTIFY_KEY_RESP_SIZE],
 }
 impl CertifyKeyExtendedResp {
-    pub const CERTIFY_KEY_RESP_SIZE: usize = 12672;
+    pub const CERTIFY_KEY_RESP_SIZE: usize = 25152;
 }
 impl Response for CertifyKeyExtendedResp {}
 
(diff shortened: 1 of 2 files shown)


ternst-nv (approved):
LGTM!

clundin25 (approved):

api/src/mailbox.rs
```diff
@@ -730,7 +730,7 @@ pub struct CertifyKeyExtendedResp {
     pub certify_key_resp: [u8; CertifyKeyExtendedResp::CERTIFY_KEY_RESP_SIZE],
 }
 impl CertifyKeyExtendedResp {
-    pub const CERTIFY_KEY_RESP_SIZE: usize = 12672;
+    pub const CERTIFY_KEY_RESP_SIZE: usize = 25152;
```
zhalvorsen, line 733: With the newest DPE revision, you should be able to decrease this value because we are only using 32 contexts. https://github.com/chipsalliance/caliptra-dpe/pull/597 changes MAX_CERT_SIZE based on the number of contexts set

zhalvorsen (approved):

api/src/mailbox.rs
pjhades: Thanks for the heads-up! Will do that in a follow-up PR.

PR #3944, opened 2026-07-06
runtime: store PQ.DevID.CDI in persistent data

pjhades:
Only stores `PQ.DevID.CDI` in persistent data, so that on cold resets, during the handling of `SET_PQ_SEED` command, the seed is consumed once to derive the CDI, and the CDI is stored. On warm resets, fetch the CDI from persistent data.

This works around the problem that the CDI cannot be read out of the key vault by software after written, as the ML-DSA logic doesn't have hardware support.


ternst-nv (approved):
LGTM

clundin25 (approved):

zhalvorsen (approved):

PR #3945, opened 2026-07-06
mldsa-dpe: Add skeleton for mldsa87 DPE Commands

ternst-nv:
Add the skeletons for the MLDSA related DPE commands.  These currently return unimplemented and have tests verifying the plumbing works as expected.

### DIFF
--- a/api/src/mailbox.rs
+++ b/api/src/mailbox.rs
@@ -42,6 +42,12 @@ impl CommandId {
     pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
     pub const ADD_SUBJECT_ALT_NAME: Self = Self(0x414C544E); // "ALTN"
     pub const CERTIFY_KEY_EXTENDED: Self = Self(0x434B4558); // "CKEX"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const INVOKE_DPE_MLDSA87: Self = Self(0x4D4C4450); // "MLDP"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const GET_PQ_CSR: Self = Self(0x50514353); // "PQCS"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const CERTIFY_KEY_EXTENDED_MLDSA87: Self = Self(0x434B454D); // "CKEM"
 
     /// FIPS module commands.
     /// The status command.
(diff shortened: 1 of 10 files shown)


pjhades (approved):

runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -73,6 +73,25 @@ pub fn run_rt_test_return_fw(args: RuntimeTestArgs) -> (DefaultHwModel, ImageBun
     run_rt_test_base(args, false)
 }
 
+// Boot the ML-DSA attestation runtime image and wait until it is ready for
+// commands. Shared by the per-command PQC DPE integration tests.
+#[cfg(feature = "mldsa_attestation")]
+pub fn run_pqc_rt_test() -> DefaultHwModel {
```
clundin25, line 79: This could be a parameter (or wrap) `run_rt_test_base`. Having a common entry for all the runtime tests makes it easier to parameterize the test matrix with this.

clundin25 (approved):

runtime/tests/runtime_integration_tests/common.rs
ternst-nv: Good point, I'll take a look at doing so in a follow up PR.  This function should also be removed in the near future when the `mldsa_attestation` implementation is complete and we no longer need to conditionally keep it removed.

api/src/mailbox.rs
```diff
@@ -42,6 +42,12 @@ impl CommandId {
     pub const EXTEND_PCR: Self = Self(0x50435245); // "PCRE"
     pub const ADD_SUBJECT_ALT_NAME: Self = Self(0x414C544E); // "ALTN"
     pub const CERTIFY_KEY_EXTENDED: Self = Self(0x434B4558); // "CKEX"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const INVOKE_DPE_MLDSA87: Self = Self(0x4D4C4450); // "MLDP"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const GET_PQ_CSR: Self = Self(0x50514353); // "PQCS"
+    #[cfg(feature = "mldsa_attestation")]
+    pub const CERTIFY_KEY_EXTENDED_MLDSA87: Self = Self(0x434B454D); // "CKEM"
```
zhalvorsen, line 50: fyi: You'll also need an ML-DSA variant of `SIGN_WITH_EXPORTED_ECDSA`, but feel free to add it in a follow-on

zhalvorsen (approved):

PR #3949, opened 2026-07-07
mldsa: csr cmd

ternst-nv:
Add the command to retrieve the mldsa csr.

### DIFF
--- a/runtime/src/get_pq_csr.rs
+++ b/runtime/src/get_pq_csr.rs
@@ -31,7 +31,7 @@ use caliptra_drivers::{
 use caliptra_x509::{MlDsa87CsrBuilder, PqDevIdCsrTbsMlDsa87, PqDevIdCsrTbsMlDsa87Params};
 use crypto::Digest;
 use zerocopy::{FromZeros, IntoBytes};
-use zeroize::{Zeroize, Zeroizing};
+use zeroize::Zeroizing;
 
 pub struct GetPqCsrCmd;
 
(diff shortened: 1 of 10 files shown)


runtime/src/get_pq_csr.rs
```diff
@@ -11,21 +11,142 @@ Abstract:
...
+        };
+        let tbs = PqDevIdCsrTbsMlDsa87::new(&params);
+
+        // Sign the TBS deterministically with the PQ.DevID private key.
+        let mut sig = Mldsa87Signature::default();
+        Mldsa87::sign_deterministic(&seed, tbs.tbs(), &mut sig)?;
+
+        // Seed and derived key material are no longer needed.
+        drop(seed);
+        public_key.zeroize();
```
zhalvorsen: We can skip this step since it is the public key

runtime/tests/runtime_integration_tests/test_get_pq_csr.rs
```diff
@@ -1,26 +1,113 @@
...
+    // Deterministic: regenerating from the same CDI yields the identical CSR.
+    let response2 = model
+        .mailbox_execute(u32::from(CommandId::GET_PQ_CSR), payload.as_bytes())
+        .unwrap()
+        .unwrap();
+    let csr_resp2 = GetPqCsrResp::ref_from_bytes(response2.as_bytes()).unwrap();
+    assert_eq!(
+        csr_bytes,
+        &csr_resp2.data[..csr_resp2.data_size as usize],
+        "GET_PQ_CSR should be deterministic across calls"
```
zhalvorsen, line 164: Feel free to do in a follow-on, but it would be good to add a test to derive the expected public key and make sure it matches the public key in the CSR so we know the firmware is deriving it correctly.

zhalvorsen (dismissed):

pjhades (dismissed):

zhalvorsen (approved):

Issue #3743, opened 2026-05-18
[PQC] Defer DPE init and implement boot-phase SET_PQ_SEED


ternst-nv:
* Caliptra Version(s): 1.3
* Caliptra Mode: Core Mode

### Summary

Restructure RT cold-reset flow to create a first-command seed window. If the first command is `SET_PQ_SEED`, enable PQC mode. If any other command arrives first, leave the command in mailbox hardware, initialize DPE in non-PQC mode, and process the held command normally.

### Scope

- [x] Create feature-gated `runtime/src/set_pq_seed.rs`.
- [ ] Validate seed length and reject zero seed.
- [ ] Store the PQ seed in PersistentData.
- [ ] Set `drivers.pqc_mode = true` on successful first-command seed receipt.
- [ ] Split cold-reset flow into pre-command ECDSA setup and post-command DPE init.
- [ ] Peek first mailbox command without consuming non-PQ commands.
- [ ] Initialize DPE after the seed window.
- [ ] Report `RtDpeInitComplete`.
- [ ] Reject `SET_PQ_SEED` from the normal command loop.
- [ ] Detect persisted seed during warm reset and auto-enable PQC mode.

### Acceptance Criteria

- [ ] First cold-reset command `SET_PQ_SEED` succeeds and enables PQC mode.
- [ ] First cold-reset command other than `SET_PQ_SEED` leaves PQC mode disabled.
- [ ] A held non-PQ first command is processed normally after DPE init.
- [ ] `SET_PQ_SEED` is rejected in the normal command loop.
- [ ] Malformed, short, and zero seeds are rejected.
- [ ] Existing ECDSA DPE commands work after the deferred init path.
- [ ] Warm reset with a persisted seed auto-enables PQC mode.
- [ ] `RtDpeInitComplete` is reported after deferred DPE initialization.

PR #3950, opened 2026-07-07
runtime: hook SET_PQ_SEED to the command loop

pjhades:
Handle `SET_PQ_SEED` in the normal command loop, but reject it once a seed is successfully set and a `PQ.DevID.CDI` is successfully derived. Also reject an all-zero seed, and update corresponding unit tests.

Issue: #3743

Although I'm referencing the ticket, the ticket itself doesn't really reflect the new design that `SET_PQ_SEED` is regarded as a normal mailbox command. For DPE initialization, we follow the way that 2.x uses - using an ECDSA-based environment, because the initialization is algorithm-agnostic and only creates the context tree. In the follow-up PRs that implement the DPE commands like `INVOKE_DPE`, when such an command is executed, we create an algorithm-dependent DPE environment.

### DIFF
--- a/runtime/src/set_pq_seed.rs
+++ b/runtime/src/set_pq_seed.rs
@@ -35,10 +35,6 @@ impl SetPqSeedCmd {
         let mut cmd = SetPqSeedReq::new_zeroed();
         crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())?;
 
-        if cmd.seed == [0; SET_PQ_SEED_SEED_SIZE] {
-            return Err(CaliptraError::RUNTIME_INVALID_PQ_SEED);
-        }
-
         let mut out = Array4x12::default();
         Self::derive_pq_devid_cdi(drivers, &cmd.seed, &mut out)?;
 
--- a/runtime/src/set_pq_seed.rs
+++ b/runtime/src/set_pq_seed.rs
@@ -14,7 +14,7 @@ Abstract:
 
 use crate::{Drivers, PauserPrivileges};
 use caliptra_cfi_derive::cfi_impl_fn;
-use caliptra_common::mailbox_api::{MailboxResp, SetPqSeedReq, SET_PQ_SEED_SEED_SIZE};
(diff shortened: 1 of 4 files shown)


runtime/src/set_pq_seed.rs
```diff
@@ -23,25 +23,29 @@ pub struct SetPqSeedCmd;
...
 
         if drivers.persistent_data.get().pqc_mode_enabled() {
-            Err(CaliptraError::RUNTIME_SET_PQ_SEED_ALREADY_SET)?;
+            return Err(CaliptraError::RUNTIME_SET_PQ_SEED_ALREADY_SET);
         }
 
         let mut cmd = SetPqSeedReq::new_zeroed();
         crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())?;
 
+        if cmd.seed == [0; SET_PQ_SEED_SEED_SIZE] {
```
ternst-nv: While I agree a production PQ seed should not be all 0s, I don't think it is the responsibility of Caliptra Core to reject poor seed values.  I'm curious others thoughts?

ternst-nv (dismissed):
LGTM, with a question about rejecting an all-zero seed.

runtime/src/set_pq_seed.rs
pjhades: I added the check because the linked issue listed it. I think the check feels a bit incomplete: either we check all possible bad values (or patterns), or we don't do it at all if it's outside the responsibility scope.

ternst-nv: Yeah, that was my concern as well.  I think we should remove it since it would be hard to describe a list of bad values.

pjhades: Alrighty, removed in [c325141](https://github.com/chipsalliance/caliptra-sw/pull/3950/commits/c3251417e4dace61fe97c7458dd388f381ff7666).

ternst-nv (approved):
LGTM

clundin25 (approved):

PR #3960, opened 2026-07-09
runtime: mark certain functions never inline

pjhades:
Mark a few functions never inline, reducing the code size by roughly 400 bytes.

Before (caliptra-1.x):
```
   text    data     bss     dec     hex filename
  96636     128   89088  185852   2d5fc target/riscv32imc-unknown-none-elf/firmware/caliptra-runtime
```

After (this branch):
```
   text    data     bss     dec     hex filename
  96284     128   89088  185500   2d49c target/riscv32imc-unknown-none-elf/firmware/caliptra-runtime
```

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -276,6 +276,7 @@ impl Drivers {
     }
 
     /// Validate DPE and disable attestation if validation fails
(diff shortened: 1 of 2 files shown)


ternst-nv (approved):
LGTM

PR #3962, opened 2026-07-10
mldsa: Provide storage for exported cdi

ternst-nv:
Allow for the mldsa exported cdi to be stored in persistent data, since it cannot be stored within the key vault to access contraints.

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -50,7 +50,8 @@ pub const PQ_DEVID_CDI_SIZE: u32 = 48;
 pub const PQC_STATUS_FLAGS_SIZE: u32 = 1;
 #[cfg(feature = "mldsa_attestation")]
 pub const PQC_MODE_ENABLED_FLAG: u8 = 1;
-pub const RESERVED_MEMORY_SIZE: u32 = (3 * 1024) - 2 - PQ_DEVID_CDI_SIZE - PQC_STATUS_FLAGS_SIZE;
+pub const RESERVED_MEMORY_SIZE: u32 =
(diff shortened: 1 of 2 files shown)


drivers/src/persistent.rs
```diff
@@ -330,6 +353,9 @@ pub struct PersistentData {
     #[cfg(not(feature = "mldsa_attestation"))]
     pqc_status_flags: u8,
 
+    #[cfg(all(feature = "mldsa_attestation", feature = "runtime"))]
+    pub mldsa_exported_cdi_slots: MldsaExportedCdiEntry,
```
clundin25, line 364: IMO it would be good to make `DPE_DCCM_STORAGE` the same shape with and without `mldsa_attestion` to simplify hitlessly updating to ML-DSA enabled firmware.

clundin25 (dismissed):

drivers/src/persistent.rs
zhalvorsen: I agree, we might as well allocate at it even though we aren't using it at the time

ternst-nv: Yeah, we've been doing that with the other mldsa fields, but that didn't seem to be the pattern for the `runtime` fields.  I'll go ahead and use the mldsa formulation for this field as well.

clundin25: It would also be useful to have some test case that hitlessly updates from non-mldsa -> mldsa to make sure this is working

ternst-nv: Yeah, agreed, I think we should put together a series of tests for that in a follow on PR.

zhalvorsen (approved):

PR #3963, opened 2026-07-10
[dpe] Make SVN in `DeriveContext` optional

### LINKED ISSUE #3957: Make SVN in DPE's `DeriveContext` optional.
* Caliptra Version(s) - 1.3

DPE made a breaking change to the `DeriveContext` command by adding an SVN field at the end. To avoid making this a breaking change to Caliptra integrators, we should add some logic to make the SVN field optional within the `INVOKE_DPE_*` commands. If the command received is a `DeriveContext` command without an SVN field, we will extend the command with an SVN of `0`.

zhalvorsen:
DPE added an SVN field to `DeriveContext`. To avoid making a breaking change to Caliptra integrators, this will append a default SVN value (`0`) to the command if it is not present.

This also adds a test to make sure it is handled correctly.

### DIFF
--- a/runtime/src/invoke_dpe.rs
+++ b/runtime/src/invoke_dpe.rs
@@ -20,7 +20,10 @@ use caliptra_cfi_derive::{cfi_impl_fn, cfi_mod_fn};
 use caliptra_common::mailbox_api::{InvokeDpeReq, MailboxRespHeader, MailboxRespHeaderVarSize};
 use caliptra_drivers::{CaliptraError, CaliptraResult};
 use dpe::{
-    commands::{CertifyKeyCommand, Command, CommandExecution, InitCtxCmd},
+    commands::{
+        CertifyKeyCommand, Command, CommandExecution, CommandHdr, DeriveContextCmd,
+        DeriveContextCmdV1, InitCtxCmd,
+    },
     context::ContextState,
     error::DpeErrorCode,
     response::ResponseHdr,
@@ -28,7 +31,7 @@ use dpe::{
 };
 use platform::MAX_OTHER_NAME_SIZE;
 use ufmt::derive::uDebug;
-use zerocopy::{FromZeros, IntoBytes};
+use zerocopy::{FromBytes, FromZeros, IntoBytes};
 #[cfg(feature = "mldsa_attestation")]
 use {crate::mldsa_dpe_env, caliptra_common::mailbox_api::InvokeDpeMldsa87Req};
 
@@ -59,7 +62,7 @@ impl InvokeDpeCmd {
         execute(
             drivers,
             CaliptraDpeProfile::Ecc384,
-            &cmd.data,
+            &mut cmd.data,
             cmd.data_size as usize,
         )
     }
@@ -111,7 +114,7 @@ impl InvokeDpeMldsa87Cmd {
         execute(
             drivers,
             CaliptraDpeProfile::Mldsa,
-            &cmd.data,
+            &mut cmd.data,
             cmd.data_size as usize,
         )
     }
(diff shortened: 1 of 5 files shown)


runtime/src/invoke_dpe.rs
```diff
@@ -57,6 +64,25 @@ impl InvokeDpeCmd {
...
+                size_of::<CommandHdr>() + size_of::<DeriveContextCmd>() - size_of::<u32>();
+            let cmd_size = cmd.data_size as usize;
+
+            let is_derive_context_cmd = hdr.cmd_id == Command::DERIVE_CONTEXT;
+            let is_missing_svn = cmd_size == expected_no_svn_len;
+            let has_space_for_svn = cmd_size + size_of::<u32>() <= cmd.data.len();
+
+            if is_derive_context_cmd && is_missing_svn && has_space_for_svn {
+                cmd.data[cmd_size..cmd_size + size_of::<u32>()].fill(0);
+                cmd.data_size += size_of::<u32>() as u32;
```
clundin25: What do you think about parsing what we know about the command and serializing a new one; I think it makes the code a bit easier to read / maintainable, but a much more verbose.

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
```diff
@@ -532,3 +532,62 @@ fn test_certify_key_with_max_contexts() {
...
+
+    let mut cmd_data = [0u8; InvokeDpeReq::DATA_MAX_SIZE];
+    let cmd_hdr = CommandHdr::new(DpeProfile::P384Sha384, Command::DERIVE_CONTEXT);
+    let cmd_hdr_buf = cmd_hdr.as_bytes();
+    cmd_data[..cmd_hdr_buf.len()].copy_from_slice(cmd_hdr_buf);
+    let dpe_cmd_buf = derive_ctx_cmd.as_bytes();
+    // Strip the last 4 bytes (svn u32 field) to simulate caller without SVN
+    let dpe_cmd_buf_no_svn = &dpe_cmd_buf[..dpe_cmd_buf.len() - 4];
+    cmd_data[cmd_hdr_buf.len()..cmd_hdr_buf.len() + dpe_cmd_buf_no_svn.len()]
+        .copy_from_slice(dpe_cmd_buf_no_svn);
```
clundin25, line 581: Alternatively could re-define the older version of the command struct, rather than manipulating the serialized form

clundin25 (dismissed):

runtime/src/invoke_dpe.rs
```diff
@@ -57,6 +64,25 @@ impl InvokeDpeCmd {
...
+        // missing. This is to support older versions of the DPE that do not include the SVN in the
+        // DeriveContext command.
+        if let Ok((hdr, _)) = CommandHdr::read_from_prefix(&cmd.data[..cmd.data_size as usize]) {
+            let expected_no_svn_len =
+                size_of::<CommandHdr>() + size_of::<DeriveContextCmd>() - size_of::<u32>();
+            let cmd_size = cmd.data_size as usize;
+
+            let is_derive_context_cmd = hdr.cmd_id == Command::DERIVE_CONTEXT;
+            let is_missing_svn = cmd_size == expected_no_svn_len;
+            let has_space_for_svn = cmd_size + size_of::<u32>() <= cmd.data.len();
```
ternst-nv: This check here would mean the message still has to be provided with additional space, which would require a change from integrators right?

One thing we could do is detect if a `DeriveContext` command is being executed and allocate a full buffer then copy over the slice and parse from the new buffer.  It would give us a default value of SVN, and not require an integrator change, though that seems a bit brittle to me.

Another option, would be to provide the ability for the DPE library to handle both provided and non-provided svns, likely by providing a way to deserialize either?

This does raise a good point of how we wish to handle DPE command versioning though.

clundin25: Suppose we add another set of invoke dpe commands? That helps make the breaking change explicit and gives us a way to translate the previous API in an obvious way.

There is some boiler plate for a new command like this but hopefully it doesn't result in too much code space, since we can just translate the command and call the updated API

zhalvorsen: I decided to mock up handling it in the DPE library. I think I like it more, but it does break the mold a little bit. What do you guys think https://github.com/chipsalliance/caliptra-dpe/pull/613?

clundin25: I'd prefer handling the translation in caliptra since it's strictly an issue with the caliptra API, and not necessarily the DPE API

runtime/src/invoke_dpe.rs
zhalvorsen: That's a good idea. It makes it more explicit about what the conversion is and how to handle it if further changes happen.

Done

runtime/tests/runtime_integration_tests/test_invoke_dpe.rs
zhalvorsen: Done

runtime/src/invoke_dpe.rs
zhalvorsen: It doesn't actually require a change from integrators because we allocate a default `InvokeDpeReq` onto the stack and then fill in as much as the requestor sends. This means they only give us a sub-portion of the command, but we can still use the rest of the buffer.

clundin25 (dismissed):

runtime/src/invoke_dpe.rs
```diff
@@ -245,3 +276,31 @@ fn clear_tags_for_inactive_contexts(
         }
     });
 }
+
+#[repr(C, align(4))]
+#[derive(Debug, PartialEq, Eq, FromBytes, IntoBytes, Immutable, KnownLayout)]
+struct DeriveContextCmdV1 {
```
ternst-nv: [nit] Can this move to the DPE crate?

I do think we should probably put together a design doc on how we want to handle versioning for DPE commands, as its likely something we'll need to deal with.

ternst-nv (dismissed):

runtime/src/invoke_dpe.rs
zhalvorsen: That makes sense, especially so we can use it in the test. Created https://github.com/chipsalliance/caliptra-dpe/pull/616

clundin25 (dismissed):

ternst-nv (dismissed):
LGTM

ternst-nv (approved):

Issue #3965, opened 2026-07-10
Runtime image includes segments of ELF metadata


pjhades:
Current runtime ELF contains the following segments:
```
16:22:16:jing-dev:caliptra-sw (caliptra-1.x) $ riscv32-unknown-elf-readelf -Wl ./target/riscv32imc-unknown-none-elf/firmware/caliptra-runtime

Elf file type is EXEC (Executable file)
Entry point 0x40005940
There are 10 program headers, starting at offset 52

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  PHDR           0x000034 0x40005034 0x40005034 0x00140 0x00140 R   0x4
  LOAD           0x000000 0x40005000 0x40005000 0x00174 0x00174 R   0x1000
  LOAD           0x000800 0x40005800 0x40005800 0x12fa4 0x12fa4 R E 0x1000
  LOAD           0x0137a4 0x400187a4 0x400187a4 0x00fbc 0x00fbc R   0x1000
  LOAD           0x014c00 0x50009c00 0x40019760 0x00080 0x00080 RW  0x1000
  LOAD           0x015400 0x5000a400 0x5000a400 0x00000 0x15400 RW  0x1000
  LOAD           0x015800 0x5001f800 0x5001f800 0x00000 0x00400 RW  0x1000
  LOAD           0x015c00 0x5001fc00 0x5001fc00 0x00000 0x00400 RW  0x1000
  GNU_STACK      0x000000 0x00000000 0x00000000 0x00000 0x00000 RW  0
  RISCV_ATTRIBUT 0x015d5b 0x00000000 0x00000000 0x0003d 0x0003d R   0x1

 Section to Segment mapping:
  Segment Sections...
   00
   01
   02     .text
   03     .rodata
   04     .data
   05     .stack
   06     .estack
   07     .nstack
   08
   09     .riscv.attributes
```
The first `LOAD` segment, given its offset and file size, contains ELF header and the program header table, which is not interesting to the firmware image. However the image builder still pulls it into the firmware image.

I ran into this issue while working on a [dev branch](https://github.com/pjhades/caliptra-sw/tree/3745-mldsa-cert-chain). With the metadata segment, the image appeared to be larger, causing the verifier to fail [this length check](https://github.com/pjhades/caliptra-sw/blob/caliptra-1.x/runtime/src/fips.rs#L104-L106) while running the FIPS self test. 

[...]

PR #3966, opened 2026-07-11
image: skip segments containing ELF metadata

pjhades:
When creating the firmware image, do not consider ELF segments that include ELF metadata. This prevents irrelevant segments from being written to the firmware image, and fixes the load address calculation.

Issue: #3965


ternst-nv (approved):
LGTM

pjhades:
@clundin25 @zhalvorsen if you guys could take a look and double check, it would be nice.

clundin25 (approved):
Nice find!

PR #3967, opened 2026-07-13
runtime: fix source of PQ.DevID.CDI

pjhades:
When passing the CDI to the HMAC KDF, fetch it from the persistent data instead of the key vault.


ternst-nv (approved):
LGTM

clundin25 (approved):

PR #3969, opened 2026-07-13
mldsa: revoke mldsa cdi

ternst-nv:
Add the command to revoke an exported MLDSA cdi, in a similar fashion to the ecdsa cdi.  Also add unit tests validating the same functionality.

Note: Additional tests will be added once the export mldsa cdi infrastructure is implemented.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_revoke_exported_cdi_handle_mldsa.rs
+++ b/runtime/tests/runtime_integration_tests/test_revoke_exported_cdi_handle_mldsa.rs
@@ -1,0 +1,86 @@
+// Licensed under the Apache-2.0 license
+
+use caliptra_api::SocManager;
+use caliptra_builder::firmware::APP_MLDSA_ATTESTATION;
+use caliptra_common::mailbox_api::{
+    CommandId, MailboxReq, MailboxReqHeader, RevokeExportedCdiHandleReq,
+};
+use caliptra_error::CaliptraError;
+use caliptra_hw_model::HwModel;
+use caliptra_runtime::RtBootStatus;
+
+use crate::common::{assert_error, run_pqc_rt_test, run_rt_test, RuntimeTestArgs};
+
+#[test]
+fn test_revoke_exported_cdi_handle_mldsa_not_found() {
+    // Exercises the unified REVOKE_EXPORTED_CDI_HANDLE command on the ML-DSA
+    // attestation build, where it also checks the ML-DSA exported-CDI slot.
+    let mut model = run_pqc_rt_test();
+
+    // Revoking a non-existent handle should return NOT_FOUND.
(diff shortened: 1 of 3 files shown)


api/src/mailbox.rs
```diff
@@ -1475,6 +1495,46 @@ pub struct RevokeExportedCdiHandleResp {
     pub hdr: MailboxRespHeader,
 }
 
+// REVOKE_EXPORTED_CDI_HANDLE_MLDSA
+#[cfg(feature = "mldsa_attestation")]
+#[repr(C)]
+#[derive(Debug, IntoBytes, FromBytes, KnownLayout, Immutable, PartialEq, Eq)]
+pub struct RevokeExportedCdiHandleMldsaReq {
+    pub hdr: MailboxReqHeader,
+    pub exported_cdi_handle: [u8; Self::EXPORTED_CDI_MAX_SIZE],
```
clundin25: Since this handle is unique per exported context could we just implement this in the existing command?

runtime/tests/runtime_integration_tests/test_revoke_exported_cdi_handle_mldsa.rs
```diff
@@ -0,0 +1,95 @@
...
+fn boot_mldsa_model() -> caliptra_hw_model::DefaultHwModel {
+    let mut model = run_rt_test(RuntimeTestArgs {
+        test_fwid: Some(&APP_MLDSA_ATTESTATION),
+        ..Default::default()
+    });
+    model.step_until(|m| {
+        m.soc_ifc().cptra_boot_status().read() == u32::from(RtBootStatus::RtReadyForCommands)
+    });
+    model
+}
```
clundin25: nit: Seems like a useful helper function, maybe it should live in `common.rs`?

api/src/mailbox.rs
ternst-nv: Yes, we could.  I went back and forth, but thought a symmetrical command best fit what the rest of the structure was so went with that approach.  Do you feel strongly we should unify them?

clundin25: Hmm, I think it doesn't have to be coupled to the underlying algorithm, and hopefully that will save a little bit of code space

pjhades: I think I agree with the point that this command isn't conceptually algorithm-dependent. So personally I'm inclined to unify them.

ternst-nv: Makes sense, I've unified the command.

clundin25 (approved):
I think we should also add a test that verifies if we revoke a handle of one type (ML-DSA) it doesn't impact the other.

ternst-nv:
> I think we should also add a test that verifies if we revoke a handle of one type (ML-DSA) it doesn't impact the other.

Agreed, to fully test the mldsa cdi we'll need the `invoke_dpe` mldsa variant to be populated, so I'll make sure that the test is added when that is available.

PR #3978, opened 2026-07-15
runtime: implement sign_with_alias for ML-DSA

pjhades:
Add the missing implementation of `DpeCrypto::sign_with_alias` for the ML-DSA call path.

Tests exercising this interface will be added along with the upcoming PR for `INVOKE_DPE` implementation.

### DIFF
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -35,6 +35,7 @@ use crypto::{
 use dpe::{EcdsaAlgorithm, ExportedCdiHandle, U8Bool, MAX_EXPORTED_CDI_SIZE};
 #[cfg(feature = "mldsa_attestation")]
 use {
+    crate::dice,
     caliptra_drivers::{
         Hmac384Key, Mldsa87, Mldsa87Mu, Mldsa87PubKey, Mldsa87Seed, Mldsa87Signature,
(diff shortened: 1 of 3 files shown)


runtime/src/dpe_crypto.rs
```diff
@@ -485,7 +485,27 @@ impl Crypto for DpeCrypto<'_> {
             }
 
             #[cfg(feature = "mldsa_attestation")]
-            Signer::Mldsa => Err(CryptoError::MismatchedAlgorithm),
+            Signer::Mldsa => {
```
ternst-nv, line 489: Looks like this might duplicate some logic in `fn derive_devid_seed`, in particular I'm worried about making sure `label` parameter to the kdf matches.  Would it be possible to have a utility function used by both which just accepts an initial CDI?

ternst-nv (dismissed):
Logic looks good to me, question about using a utility function.

runtime/src/dpe_crypto.rs
pjhades: Yep, that's a good point.

ternst-nv (approved):
LGTM

clundin25 (approved):

zhalvorsen (approved):

PR #3979, opened 2026-07-15
[fpga-image]: Create stable mac address by hashing boot SD serial number

clundin25:
* This assumes the SD serial numbers are unique
* This resolves a longstanding problem, where the FPGAs all have the same MAC address

This should provide a stable and unique MAC solving some longstanding problems and inconveniences with the FPGAs.


zhalvorsen (approved):
Great idea!

PR #3981, opened 2026-07-16
mldsa: pqc certify cmd

ternst-nv:
Add support for the certify extended key command, including updates to the DpeCrypto to support the required implementation.  This includes a battery of unit tests validating the certify key functionality.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -938,7 +938,7 @@ pub trait HwModel: SocManager {
     /// Wait for the response to a previous call to `start_mailbox_execute()`.
     fn finish_mailbox_execute(&mut self) -> std::result::Result<Option<Vec<u8>>, ModelError> {
         // Wait for the microcontroller to finish executing
-        let mut timeout_cycles = 40000000; // 100ms @400MHz
+        let mut timeout_cycles = 400000000; // 1s @400MHz
         while self.soc_mbox().status().read().status().cmd_busy() {
             self.step();
             timeout_cycles -= 1;
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -935,10 +935,23 @@ pub trait HwModel: SocManager {
         Ok(())
     }
 
+    /// Number of `step()` iterations to wait for a mailbox command to complete
+    /// before declaring a `MailboxTimeout`.
+    ///
+    /// On the emulator each `step()` advances a single emulated core cycle, so a
+    /// large count is needed to accommodate long-running commands (e.g. ML-DSA).
+    /// On real hardware (FPGA) the core runs in real time and each `step()` is a
+    /// real-time MMIO poll, so this count acts as a busy-wait bound and must stay
+    /// small. Backends override this as appropriate; the default matches the
+    /// original Caliptra 1.x value (100ms @400MHz).
+    fn mailbox_timeout_cycles(&self) -> u32 {
+        40000000 // 100ms @400MHz
+    }
+
     /// Wait for the response to a previous call to `start_mailbox_execute()`.
     fn finish_mailbox_execute(&mut self) -> std::result::Result<Option<Vec<u8>>, ModelError> {
         // Wait for the microcontroller to finish executing
-        let mut timeout_cycles = 400000000; // 1s @400MHz
+        let mut timeout_cycles = self.mailbox_timeout_cycles();
         while self.soc_mbox().status().read().status().cmd_busy() {
             self.step();
             timeout_cycles -= 1;
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -945,7 +945,7 @@ pub trait HwModel: SocManager {
     /// small. Backends override this as appropriate; the default matches the
     /// original Caliptra 1.x value (100ms @400MHz).
     fn mailbox_timeout_cycles(&self) -> u32 {
-        40000000 // 100ms @400MHz
+        40_000_000 // 100ms @400MHz
     }
 
     /// Wait for the response to a previous call to `start_mailbox_execute()`.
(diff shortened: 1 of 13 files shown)


hw-model/src/lib.rs
```diff
@@ -938,7 +938,7 @@ pub trait HwModel: SocManager {
     /// Wait for the response to a previous call to `start_mailbox_execute()`.
     fn finish_mailbox_execute(&mut self) -> std::result::Result<Option<Vec<u8>>, ModelError> {
         // Wait for the microcontroller to finish executing
-        let mut timeout_cycles = 40000000; // 100ms @400MHz
+        let mut timeout_cycles = 400000000; // 1s @400MHz
```
zhalvorsen: Yikes :)

runtime/src/certify_key_extended.rs
```diff
@@ -28,7 +32,21 @@ impl CertifyKeyExtendedCmd {
     pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<()> {
         let mut cmd = CertifyKeyExtendedReq::new_zeroed();
         crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())?;
+        Self::certify_key(
+            drivers,
+            cmd.flags,
+            &cmd.certify_key_req[..],
```
pjhades: nit: `[..]` could be removed.

runtime/src/certify_key_extended_mldsa.rs
```diff
@@ -13,19 +13,42 @@ Abstract:
...
+            &mut drivers.soc_ifc,
+            caliptra_common::WdtTimeout::new_const(200_000_000),
+        );
+
+        // The certify-key logic is shared with the ECDSA variant; only the DPE
+        // profile (and thus identity / signature algorithm) differs.
+        CertifyKeyExtendedCmd::certify_key(
+            drivers,
+            cmd.flags,
+            &cmd.certify_key_req[..],
```
pjhades: nit: ditto

hw-model/src/lib.rs
```diff
@@ -935,10 +935,23 @@ pub trait HwModel: SocManager {
...
+    /// On the emulator each `step()` advances a single emulated core cycle, so a
+    /// large count is needed to accommodate long-running commands (e.g. ML-DSA).
+    /// On real hardware (FPGA) the core runs in real time and each `step()` is a
+    /// real-time MMIO poll, so this count acts as a busy-wait bound and must stay
+    /// small. Backends override this as appropriate; the default matches the
+    /// original Caliptra 1.x value (100ms @400MHz).
+    fn mailbox_timeout_cycles(&self) -> u32 {
+        40000000 // 100ms @400MHz
+    }
+
```
pjhades, line 950: Would it make sense to move this change to a separate PR?

Also, nit: maybe `40_000_000` is clearer? It took me a moment to realize how many 0s are there, compared to the value used in the trait impls.

ternst-nv: Good point on the format, I'll switch that.  I'd prefer to land this as part of this PR, as otherwise the FPGA tests do not complete.

hw-model/src/model_emulated.rs
```diff
@@ -262,6 +262,12 @@ impl HwModel for ModelEmulated {
         }
     }
 
+    fn mailbox_timeout_cycles(&self) -> u32 {
+        // Each step() advances a single emulated core cycle, so allow extra
+        // cycles to accommodate long-running commands (e.g. ML-DSA). 1s @400MHz.
+        400000000
```
pjhades: nit: format this number as well? (And the one in hw-model/src/model_verilated.rs)

pjhades (dismissed):
LGTM!

pjhades (dismissed):

runtime/src/dpe_crypto.rs
```diff
@@ -101,63 +105,97 @@ impl<'a> DpeCrypto<'a> {
...
+    fn derive_cdi_inner_mldsa(
+        &mut self,
+        measurement: &Digest,
+        info: &[u8],
+    ) -> Result<Zeroizing<Array4x12>, CryptoError> {
+        // Copy root CDI to the stack before borrowing self.hmac384/trng.
+        let root_copy: Array4x12 = match &self.rt_cdi {
+            Cdi::Mldsa(root) => **root,
+            _ => return Err(CryptoError::MismatchedAlgorithm),
+        };
```
zhalvorsen: I'm a little worried there will still be a lingering copy on the stack here. Can we wrap the copy in a `Zeroizing`?

(warning: I haven't tried building this)


```suggestion
        let root_copy: Zeroizing<Array4x12> = match &self.rt_cdi {
            Cdi::Mldsa(root) => Zeroizing<Array4x12>::new(*root),
            _ => return Err(CryptoError::MismatchedAlgorithm),
        };
```

runtime/src/dpe_crypto.rs
```diff
@@ -486,13 +563,13 @@ impl Crypto for DpeCrypto<'_> {
             }
 
             #[cfg(feature = "mldsa_attestation")]
-            Signer::Mldsa => {
-                let cdi = match self.rt_cdi {
-                    RtCdi::Mldsa(ref pq_devid_cdi) => pq_devid_cdi,
+            Signer::Mldsa { .. } => {
+                let cdi: Array4x12 = match self.rt_cdi {
```
zhalvorsen: Similar potential lingering copy here

zhalvorsen (dismissed):

hw-model/src/lib.rs
ternst-nv: Yeah, this isn't the worst case either.  I think we're going to have to spend some time tuning the performance of the MLDSA library.

zhalvorsen (approved):

PR #3988, opened 2026-07-17
runtime: Harden DPE index cache

parvathib:
Follow-up to #3729.

- Update reset initializes the cache only if the marker is missing, supporting 2.1.1 -> 2.1.2 once, without rescanning on later 2.1.2+ updates.

- Switched cached-index users from raw u32 tci_type to CaliptraManagedDpeContext enum.

- Moved measurement unchanged check into update_caliptra_managed_measurement() so CCIV/MCFW no-op updates do not extend cumulative state.

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -98,12 +98,48 @@ pub type AuthManifestImageMetadataList =
     [AuthManifestImageMetadata; AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT];
 
 #[cfg(feature = "runtime")]
-#[derive(Clone, Copy, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize, Default)]
+#[derive(Clone, Copy, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize)]
 #[repr(C)]
 pub struct CaliptraManagedDpeContextIndices {
+    pub initialized: U8Bool,
     pub cciv: u8,
     pub mcu_rt: u8,
-    pub reserved: [u8; 2],
+    pub reserved: [u8; 1],
+}
+
+#[cfg(feature = "runtime")]
+impl CaliptraManagedDpeContextIndices {
+    pub const INVALID_INDEX: u8 = 0xff;
+
+    pub fn invalidate(&mut self) {
+        *self = Self::default();
+    }
+
+    pub fn is_initialized(&self) -> bool {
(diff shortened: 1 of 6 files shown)


drivers/src/persistent.rs
```diff
@@ -98,12 +98,49 @@ pub type AuthManifestImageMetadataList =
     [AuthManifestImageMetadata; AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT];
 
 #[cfg(feature = "runtime")]
-#[derive(Clone, Copy, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize, Default)]
+#[derive(Clone, Copy, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize)]
 #[repr(C)]
 pub struct CaliptraManagedDpeContextIndices {
+    pub initialized: u8,
```
zhalvorsen: fyi: We have a `U8Bool` that we use pretty frequently in persistent data that could work well here

drivers/src/persistent.rs
```diff
@@ -98,12 +98,49 @@ pub type AuthManifestImageMetadataList =
...
     pub cciv: u8,
     pub mcu_rt: u8,
-    pub reserved: [u8; 2],
+    pub reserved: [u8; 1],
```
zhalvorsen, line 107: Should we allocate some more for future contexts like the potential `SOMO` and `SOMV` contexts?

zhalvorsen (dismissed):

drivers/src/persistent.rs
parvathib: That's upcoming PR. I'll extend it there :)

zhalvorsen: Sounds good to me!

drivers/src/persistent.rs
parvathib: done

zhalvorsen (dismissed):

mhatrevi (dismissed):

mhatrevi (approved):
As a follow up change, please add an integration test that does the following:
fw-2.0.2 cold boot
-> establish DPE state and MCFW
-> update to new Runtime
-> verify CCIV and MCFW cache-backed updates
-> perform another update
-> verify persisted cache remains valid

Also, the PR description should say fw-2.0.2 -> fw-2.0.3, not 2.1.1 -> 2.1.2.

PR #3989, opened 2026-07-17
Dpe soc manifest contexts

### LINKED ISSUE #3758: Add SoC Manifest claims to attestation evidence and anchor manifest preamble in DPE chain
Caliptra RT does not currently track or stash the SoC Authorization Manifest [preamble fields](https://github.com/chipsalliance/caliptra-sw/blob/main/auth-manifest/README.md#preamble) such as policy, signing keys, and SVN. Because of this, manifest authorization state is not anchored in the chain of trust or reflected in attestation evidence.

This creates a gap where a verifier can determine what MCU RT is running, but not who authorized it.

# Proposal

- Track manifest preamble authorization data in Caliptra RT
- Create separate DPE contexts for vendor and owner policy/key material
- Include this authorization state in the DPE certificate chain and attestation evidence

## Caliptra-managed DPE chain

| State | DPE Lineage |
| --- | --- |
| Current | `RTMR -> CCIV -> ROM_STASH_1..N -> MCFW` |
| Proposed | `RTMR -> CCIV -> ROM_STASH_1..N -> SOMV -> SOMO -> MCFW` |

## Proposed SoC Manifest Preamble DPE Measurements
|Preamble Field | DPE Context | Measurement | Fields included |
|--- | --- | --- | --- |
| Vendor fields | `SOMV` | `SHA384(version \|\| svn \|\| flags \|\| vendor_pub_keys)` | `version`, <br> `svn`, <br>`flags`, <br>`vendor_pub_keys` |
| Owner fields | `SOMO` | `SHA384(owner_pub_keys)` | `owner_pub_keys` |

**Affected versions**
* Caliptra Version(s) (2.x)
* Caliptra Mode (Subsystem Mode)

parvathib:
**Changes:**

- Add SOMV/SOMO cached DPE context indices and TCI types.
- Create SOMV/SOMO during recovery SET_AUTH_MANIFEST and update existing contexts on mailbox SET_AUTH_MANIFEST.
- Measure SOMV from vendor preamble policy/key fields and SOMO from owner key fields.
- Report manifest SVN on SOMV TCB info and keep SOMO SVN at 0.

**Hitless / update policy:**
- Recovery/cold boot creates `SOMV` and `SOMO` before `MCFW`.
- Mailbox `SET_AUTH_MANIFEST` only updates existing cached `SOMV`/`SOMO` contexts; it does not create missing contexts mid-chain.
- On `SET_AUTH_MANIFEST`, runtime recomputes SOMV/SOMO measurements and compares them against the current DPE TCI.
- If the recomputed digest is unchanged, the update is a no-op:
  - no DPE current TCI update
  - no DPE cumulative/journey update
  - no PCR31 extension
- If the recomputed digest changes:
  - DPE current TCI is updated
  - DPE cumulative/journey TCI is extended
  - PCR31 is extended
- Vendor preamble field changes (`version || svn || flags || vendor_pub_keys`) update only `SOMV`.
- Owner key changes update only `SOMO`.

**Tests:**

- Verify SOMV/SOMO/MCFW TCBInfo presence, SVN, and digest contents in DPE certs.
- Verify same-manifest update is a no-op for SOMV/SOMO and PCR31.
- Verify SVN-only update changes SOMV/PCR31 but not SOMO.
- Verify owner-key update changes SOMO/PCR31 but not SOMV.

Fixes #3758

### DIFF
--- a/runtime/src/drivers.rs
+++ b/runtime/src/drivers.rs
@@ -19,7 +19,7 @@ use crate::debug_unlock::ProductionDebugUnlock;
 use crate::dpe_crypto::DpeCrypto;
 #[cfg(feature = "fips_self_test")]
 pub use crate::fips::fips_self_test_cmd::SelfTestStatus;
-use crate::invoke_dpe::dpe_error_detail;
+use crate::invoke_dpe::{dpe_error_detail, invoke_dpe_cmd};
 use crate::recovery_flow::RecoveryFlow;
 use crate::{
     dice, CaliptraDpeEnv, CaliptraDpeProfile, DisableAttestationCmd, DpePlatform, Mailbox,
@@ -38,7 +38,7 @@ use caliptra_common::cfi_check;
 use caliptra_common::crypto::Crypto;
 use caliptra_common::dice::{copy_ldevid_ecc384_cert, copy_ldevid_mldsa87_cert};
 use caliptra_common::mailbox_api::AddSubjectAltNameReq;
-use caliptra_dpe::commands::DeriveContextCmd;
+use caliptra_dpe::commands::{Command, DeriveContextCmd};
 use caliptra_dpe::context::{Context, ContextState, ContextType};
 use caliptra_dpe::response::DeriveContextResp;
 use caliptra_dpe::tci::TciMeasurement;
@@ -73,7 +73,10 @@ use caliptra_registers::{
 use caliptra_ureg::MmioMut;
 use caliptra_x509::{NotAfter, NotBefore};
 
-use core::cmp::Ordering::{Equal, Greater};
+use core::{
+    cmp::Ordering::{Equal, Greater},
+    mem::size_of,
+};
 use zerocopy::IntoBytes;
 
 pub const MCI_TOP_REG_RESET_REASON_OFFSET: u32 = 0x38;
@@ -101,6 +104,8 @@ impl From<McuResetReason> for u32 {
 #[derive(Clone, Copy)]
 pub(crate) enum CaliptraManagedDpeContext {
     Cciv,
+    Somv,
+    Somo,
     McuRt,
 }
 
@@ -108,7 +113,27 @@ impl CaliptraManagedDpeContext {
     fn tci_type(self) -> u32 {
         match self {
             Self::Cciv => Drivers::CCIV_TCI_TYPE,
+            Self::Somv => Drivers::SOMV_TCI_TYPE,
+            Self::Somo => Drivers::SOMO_TCI_TYPE,
             Self::McuRt => Drivers::MCU_RT_TCI_TYPE,
+        }
+    }
+
+    fn cached_index(self, indices: &CaliptraManagedDpeContextIndices) -> u8 {
+        match self {
+            Self::Cciv => indices.cciv,
+            Self::Somv => indices.somv,
+            Self::Somo => indices.somo,
+            Self::McuRt => indices.mcu_rt,
+        }
+    }
+
+    fn set_cached_index(self, indices: &mut CaliptraManagedDpeContextIndices, idx: u8) {
+        match self {
+            Self::Cciv => indices.set_cciv(idx),
+            Self::Somv => indices.set_somv(idx),
+            Self::Somo => indices.set_somo(idx),
+            Self::McuRt => indices.set_mcu_rt(idx),
         }
     }
 }
@@ -183,6 +208,8 @@ pub struct Drivers {
 
 impl Drivers {
     pub const CCIV_TCI_TYPE: u32 = u32::from_be_bytes(*b"CCIV");
+    pub const SOMV_TCI_TYPE: u32 = u32::from_be_bytes(*b"SOMV");
+    pub const SOMO_TCI_TYPE: u32 = u32::from_be_bytes(*b"SOMO");
     pub const MCU_RT_TCI_TYPE: u32 = u32::from_be_bytes(*b"MCFW");
 
     /// # Safety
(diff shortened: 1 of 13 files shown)


runtime/src/drivers.rs
```diff
@@ -410,6 +435,90 @@ impl Drivers {
...
+            );
+        }
+
+        if !create_if_missing {
+            return Ok(());
+        }
+
+        self.is_dpe_context_threshold_exceeded(PauserPrivileges::PL0)?;
+
+        let flags = DeriveContextFlags::MAKE_DEFAULT
```
zhalvorsen, line 480: This does imply that the configuration still has only a default context in the PL0 locality. Can we make that assumption in the case of hitless update? Will it go through this flow on a 2.1.1 -> 2.1.2 hitless update flow?

runtime/src/drivers.rs
```diff
@@ -410,6 +435,90 @@ impl Drivers {
...
+        let indices = &mut self
+            .persistent_data
+            .get_mut()
+            .caliptra_managed_dpe_context_indices;
+        match caliptra_managed_context {
+            CaliptraManagedDpeContext::Cciv => indices.set_cciv(idx),
+            CaliptraManagedDpeContext::Somv => indices.set_somv(idx),
+            CaliptraManagedDpeContext::Somo => indices.set_somo(idx),
+            CaliptraManagedDpeContext::McuRt => indices.set_mcu_rt(idx),
+        }
```
zhalvorsen: nit: should there be a helper function for this? I think we might already be doing something similar elsewhere?

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -219,6 +222,10 @@ fn test_authorize_and_stash_cmd_deny_authorization() {
     hasher.update(rt_current_pcr);
     hasher.update(cptra_config_init_vals_hash);
     if model.subsystem_mode() {
+        let (somv_measurement, somo_measurement) =
+            default_soc_manifest_measurements(FwVerificationPqcKeyType::LMS, 0);
```
zhalvorsen: Should we default to ML-DSA?

runtime/src/set_auth_manifest.rs
```diff
@@ -69,6 +75,60 @@ impl SetAuthManifestCmd {
             .ok_or(err)
     }
 
+    fn digest_to_measurement(digest: &ImageDigest384) -> [u8; SHA384_DIGEST_BYTE_SIZE] {
```
zhalvorsen, line 78: nit: optional: this could be simplified by utilizing `Array4x12`:

```Rust
Array4x12::from(digest).into()
```

zhalvorsen (dismissed):

runtime/src/drivers.rs
parvathib: No, we should not assume that for hitless update. This flow is recovery/cold-boot only.
A 2.1.1 -> 2.1.2 hitless update goes through UpdateReset, so it does not create missing SOMV/SOMO contexts. It only updates existing cached contexts if they are already present.
If an existing chain does not have SOMV/SOMO, runtime does not insert them mid-chain.

parvathib: No, we should not assume that for hitless update. This flow is recovery/cold-boot only.
A 2.1.1 -> 2.1.2 hitless update goes through UpdateReset, so it does not create missing SOMV/SOMO contexts. It only updates existing cached contexts if they are already present.
If an existing chain does not have SOMV/SOMO, runtime does not insert them mid-chain.

zhalvorsen: Okay, good. That sounds good to me

runtime/src/drivers.rs
parvathib: Good idea! added helpers for this now:  `cached_index()` and `set_cached_index()` on `CaliptraManagedDpeContext` .

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
parvathib: Agree. I kept the repo-wide default unchanged to avoid unrelated test churn but updated this SoC manifest DPE test path to use ML-DSA.

runtime/src/set_auth_manifest.rs
parvathib: done

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
zhalvorsen: Sounds good. If you want to later, you could cherry pick https://github.com/chipsalliance/caliptra-sw/pull/3955 to the 2.0 branch to take care of the fw-signing keys.

runtime/src/drivers.rs
```diff
@@ -410,6 +449,113 @@ impl Drivers {
...
+            ueid,
+            Some(locality),
+            &mut resp_buf,
+        )
+        .or_else(|e| {
+            if e == DpeErrorCode::InvalidArgument {
+                let cmd = DeriveContextCmd {
+                    handle: ContextHandle::default(),
+                    data: TciMeasurement(*measurement),
+                    flags: DeriveContextFlags::RECURSIVE,
```
zhalvorsen: Which flag here is causing the `InvalidArgument` error? I think we want to keep all of those flags if we can

zhalvorsen (approved):

runtime/src/drivers.rs
parvathib: Good point. The fallback was too broad and dropped more flags than intended. I removed it so this path keeps the original creation flags. If the DPE derive is invalid, we now surface the error instead of retrying with only RECURSIVE.

mhatrevi (approved):

PR #3992, opened 2026-07-20
mldsa: safe pqc seed access

ternst-nv:
Make accessing the PQC seed safe, by enforcing a check on the seed enable prior to providing a reference to the data.  This prevents future and current code from inadvertently accessing the data prior to it being provided.


pjhades (dismissed):

clundin25 (dismissed):

drivers/src/persistent.rs
```diff
@@ -386,8 +382,22 @@ impl PersistentData {
...
+            .then_some(&self.pq_devid_cdi)
+            .ok_or(CaliptraError::RUNTIME_PQC_NOT_INITIALIZED)
+    }
+
+    /// Stores the PQ.DevID CDI and atomically marks PQC mode enabled, upholding
+    /// the invariant that a readable CDI always has its status flag set.
     #[cfg(feature = "mldsa_attestation")]
-    pub fn set_pqc_mode_enabled(&mut self) {
+    pub fn set_pq_devid_cdi(&mut self, cdi: PqDevIdCdi) {
+        self.pq_devid_cdi = cdi;
```
zhalvorsen, line 410: For a similar reason, should this error if it is already set?

drivers/src/persistent.rs
```diff
@@ -386,8 +382,22 @@ impl PersistentData {
...
+    /// `RUNTIME_PQC_NOT_INITIALIZED` otherwise, ensuring uninitialized CDI bytes
+    /// can never be read.
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn pq_devid_cdi(&self) -> CaliptraResult<&PqDevIdCdi> {
+        self.pqc_mode_enabled()
+            .then_some(&self.pq_devid_cdi)
+            .ok_or(CaliptraError::RUNTIME_PQC_NOT_INITIALIZED)
+    }
+
+    /// Stores the PQ.DevID CDI and atomically marks PQC mode enabled, upholding
```
zhalvorsen, line 401: nit typo: 
```suggestion
    /// Stores the PQ.DevID CDI and automically marks PQC mode enabled, upholding
```

drivers/src/persistent.rs
```diff
@@ -386,8 +382,22 @@ impl PersistentData {
         self.pqc_status_flags & PQC_MODE_ENABLED_FLAG != 0
     }
 
+    /// Returns the provisioned PQ.DevID CDI, but only once a seed has been set
+    /// via `set_pq_devid_cdi` (i.e. the PQC-mode-enabled flag is set). Returns
+    /// `RUNTIME_PQC_NOT_INITIALIZED` otherwise, ensuring uninitialized CDI bytes
+    /// can never be read.
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn pq_devid_cdi(&self) -> CaliptraResult<&PqDevIdCdi> {
```
zhalvorsen: What do you think of making `PqDevIdCdi` `ZeroizeOnDrop`? Or to not mess with persistent data, either a new type could be made or use `Mldsa87Seed` which already does this

zhalvorsen (dismissed):

drivers/src/persistent.rs
ternst-nv: Hmm, I think "atomically" works as well here, and works with the mutex style paradigm being established.

drivers/src/persistent.rs
ternst-nv: Fair enough, I went with wrapping the return in `Zeroizing`.

drivers/src/persistent.rs
zhalvorsen: Ha ha to be honest I read that as automatically. Whoops, ignore me

zhalvorsen (approved):

PR #3997, opened 2026-07-20
mldsa87: remove 1 KiB copied data from stack

pjhades:
`scalar_mul_add(out, a, b, c)` computes `out = a + b * c` but expects to receive `out` and `a` as different references. When `out` and `a` points to the same object, the caller has to copy `out` on the stack, using 1 KiB stack space.

Rewriting `scalar_mul_add` as `scalar_mul_add_assign` such that it computes `out += b * c`, allowing the caller to remove the copy from stack.

The stack frame of `matrix87_expand_mul` before this change:
```
4000f408 <_ZN14caliptra_mldsa7mldsa8719matrix87_expand_mul17he9ae2717b4b6ea47E>:
4000f408:       7111                    addi    sp,sp,-256
...
4000f41e:       8b010113                addi    sp,sp,-1872
```
After this change:
```
4000f408 <_ZN14caliptra_mldsa7mldsa8719matrix87_expand_mul17he9ae2717b4b6ea47E>:
4000f408:       bb010113                addi    sp,sp,-1104
```

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -193,9 +193,9 @@ fn scalar_mul(out: &mut Scalar, lhs: &Scalar, rhs: &Scalar) {
     }
 }
 
-fn scalar_mul_add(out: &mut Scalar, a: &Scalar, b: &Scalar, c: &Scalar) {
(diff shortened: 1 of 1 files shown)


zhalvorsen (approved):

ternst-nv (approved):
LGTM

PR #3998, opened 2026-07-20
Clear `PqDevIdCdi` when attestation is disabled

zhalvorsen:
This overwrites the `PqDevIdCdi` with a default "zeroized" value. This models the other attestation forms when disabled. This allows the system to stay alive and respond to commands, but in a degraded state. It can respond, but attestations will fail. This can help with debugging.

This now returns an error when trying to set the `PqDevIdCdi` when attestation is disabled to avoid re-enabling attestation.

The ECDSA variants also re-derive public keys that are stored in persistent data so the verify checks during signing will still work. This doesn't apply to ML-DSA because we lazily generate the keys in all use cases.

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -404,12 +404,26 @@ impl PersistentData {
     /// `RUNTIME_SET_PQ_SEED_ALREADY_SET` and leaves the existing value intact.
     #[cfg(feature = "mldsa_attestation")]
     pub fn set_pq_devid_cdi(&mut self, cdi: PqDevIdCdi) -> CaliptraResult<()> {
+        #[cfg(feature = "runtime")]
+        if self.attestation_disabled.get() {
+            return Err(CaliptraError::RUNTIME_SET_PQ_SEED_ATTESTATION_DISABLED);
+        }
         if self.pqc_mode_enabled() {
             return Err(CaliptraError::RUNTIME_SET_PQ_SEED_ALREADY_SET);
         }
         self.pq_devid_cdi = cdi;
         self.pqc_status_flags |= PQC_MODE_ENABLED_FLAG;
         Ok(())
+    }
+
+    /// When PQC mode is enabled, overwrites the PQ.DevID CDI with dummy bytes
+    /// and zeroizes the exported ML-DSA CDI slots; otherwise, does nothing.
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn erase_pq_devid_cdi(&mut self, dummy_cdi: PqDevIdCdi) {
+        if self.pqc_mode_enabled() {
+            self.pq_devid_cdi = dummy_cdi;
+            self.mldsa_exported_cdi_slots.zeroize();
+        }
     }
 
     pub fn assert_matches_layout() {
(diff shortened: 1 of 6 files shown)


clundin25 (dismissed):

drivers/src/persistent.rs
```diff
@@ -412,6 +416,14 @@ impl PersistentData {
         Ok(())
     }
 
+    /// Overwrites the PQ.DevID CDI with dummy bytes (e.g. when attestation is disabled)
+    /// and zeroizes the exported ML-DSA CDI slots.
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn erase_pq_devid_cdi(&mut self, dummy_cdi: PqDevIdCdi) {
```
ternst-nv, line 422: Its a bit confusing to me we set the cdi even in the case the `pqc_enabled` is false, and thus this cdi could never be reached.

I think its functionally correct, so maybe we include a comment about it, or make an explicit if check prior to setting the CDI?

runtime/src/disable.rs
```diff
@@ -99,4 +101,29 @@ impl DisableAttestationCmd {
...
     }
+
+    /// Zero out the PQ DevID CDI and exported CDI slots in persistent data.
+    ///
+    /// # Arguments
+    ///
+    /// * `drivers` - Drivers
+    #[cfg(feature = "mldsa_attestation")]
+    #[cfg_attr(feature = "cfi", cfi_impl_fn)]
+    fn zero_pq_devid_cdi(drivers: &mut Drivers) -> CaliptraResult<()> {
```
ternst-nv, line 123: [nit] Is there a way we could deduplicate this with `runtime/src/disable.rs:zero_cdi`, perhaps by making the `key` parameter be a `KeyWriteArgs` parameter?  

That would allow us to only define the label once, and cut down on additional imem usage which is at a premium.

runtime/src/set_pq_seed.rs
```diff
@@ -24,6 +24,10 @@ impl SetPqSeedCmd {
     #[cfg_attr(feature = "cfi", cfi_impl_fn)]
     #[inline(never)]
     pub(crate) fn execute(drivers: &mut Drivers) -> CaliptraResult<()> {
+        if drivers.persistent_data.get().attestation_disabled.get() {
```
ternst-nv: Isn't this redundant with the check in `persistent.rs`?

ternst-nv (dismissed):
LGTM - A couple small questions

drivers/src/persistent.rs
zhalvorsen: That's a fair point. I changed it to only overwrite it when PQC mode is enabled and updated the comment

runtime/src/disable.rs
zhalvorsen: Deduplicated

runtime/src/set_pq_seed.rs
zhalvorsen: Yeah, removed

ternst-nv (approved):
LGTM

clundin25 (approved):

PR #3999, opened 2026-07-20
[builder]: Support downloading prebuilt Caliptra artifacts

clundin25:
Establish a pattern to reference prebuilt firmware as a FWID. This makes it easier to write integration tests against older released firmware.

### DIFF
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -3,7 +3,7 @@
 // Centralized list of all firmware targets. This allows us to compile them all
 // ahead of time for executing tests on hosts that can't compile rust code.
 
-use crate::FwId;
+use crate::{FirmwareType, FwId};
 
 pub fn rom_from_env() -> &'static FwId<'static> {
     match std::env::var("CPTRA_ROM_TYPE").as_ref().map(|s| s.as_str()) {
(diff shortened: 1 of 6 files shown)


builder/src/lib.rs
```diff
@@ -484,9 +493,45 @@ pub fn rom_for_fw_integration_tests_fpga(fpga: bool) -> io::Result<Cow<'static,
...
+    let bytes = build_firmware_rom(id)?;
+    let leaked: &'static [u8] = Box::leak(bytes.into_boxed_slice());
+    cache.insert(*id, leaked);
+    Ok(leaked)
+}
+
 pub fn build_firmware_rom(id: &FwId<'static>) -> io::Result<Vec<u8>> {
-    let elf_bytes = build_firmware_elf(id)?;
-    elf2rom(&elf_bytes)
+    if id.features.contains(&"prebuilt") {
```
zhalvorsen: Can we add a separate field instead of overriding the `features` field?

zhalvorsen (dismissed):
This is going to be super helpful. I would like to start adding hitless update tests to make sure functionality can work from one release to the next and this would be crucial.

builder/src/lib.rs
clundin25: Done!

zhalvorsen (dismissed):

clundin25:
@ternst-nv PTAL in case we want to use this pattern in 1.x

builder/src/lib.rs
```diff
@@ -85,6 +86,13 @@ pub fn run_cmd_stdout(cmd: &mut Command, input: Option<&[u8]>) -> io::Result<Str
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq, Ord, PartialOrd)]
+pub enum FirmwareType {
```
ternst-nv: Source seems to be the only enum type where `features` is used, as such would it be possible to move it into the enum?

builder/src/release_artifacts.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    let rom = std::fs::read(&rom_path)?;
+    let fw = std::fs::read(&fw_path)?;
+
+    Ok(ReleaseArtifacts { rom, fw })
+}
+
+/// Convenience helper to fetch Caliptra 2.0.1 release artifacts from GitHub Releases.
+pub fn get_2_0_1_release_artifacts() -> anyhow::Result<ReleaseArtifacts> {
```
ternst-nv: Might it help to have the release specific path extension included in the Enum defintion?  Then we can be agnostic over the specific release for our retrieval logic.

builder/src/lib.rs
```diff
@@ -484,9 +503,53 @@ pub fn rom_for_fw_integration_tests_fpga(fpga: bool) -> io::Result<Cow<'static,
...
+pub fn get_prebuilt_firmware_static(id: &FwId<'static>) -> io::Result<&'static [u8]> {
+    static CACHE: PrebuiltCacheMap = Lazy::new(|| Mutex::new(HashMap::new()));
+
+    let mut cache = CACHE.lock().unwrap();
+    if let Some(&bytes) = cache.get(id) {
+        return Ok(bytes);
+    }
+
+    let bytes = build_firmware_rom(id)?;
+    let leaked: &'static [u8] = Box::leak(bytes.into_boxed_slice());
```
ternst-nv, line 545: [nit] Couldn't we have the bytes stored within a Box instead of leaking them? I suppose since we don't clear the cache and the process is short lived it shouldn't make too much of a difference, but would prevent surprises down the line if those assumptions don't hold.

ternst-nv (dismissed):
Approach LGTM - A couple of small comments

builder/src/lib.rs
clundin25: I want to fit the FWID into the existing model API which uses a `&'static [u8]`.  

Perhaps we should consider migrating to using the FWID type instead of a raw byte slice in the model, but this makes the test code a bit easier to write

builder/src/release_artifacts.rs
clundin25: Done!

builder/src/lib.rs
clundin25: Done!

builder/src/firmware.rs
```diff
@@ -42,129 +42,187 @@ pub fn fake_rom(fpga: bool) -> &'static FwId<'static> {
...
+
+/// Pre-built 2.0.1 ROM release binary target identifier.
+/// Represents the downloaded 2.0.1 release ROM artifact (caliptra-rom-with-log.bin)
+/// from the GitHub caliptra-sw release package.
+pub const ROM_2_0_1_RELEASE: FwId = FwId {
+    crate_name: "caliptra-rom-2_0_1-release",
+    bin_name: "caliptra-rom-with-log.bin",
+    fw_type: FirmwareType::TaggedReleaseFromNetwork {
+        version_tag: "fw-2.0.1",
+        url: "https://github.com/chipsalliance/caliptra-sw/releases/download/fw-2.0.1/caliptra_release_v20260210_0-2.0.zip",
```
zhalvorsen: Should these refer to the ROM tag:

```suggestion
        version_tag: "rom-2.0.1",
        url: "https://github.com/chipsalliance/caliptra-sw/releases/download/rom-2.0.1/caliptra_release_v20260209_0-2.0.zip",
```

clundin25: Done!

zhalvorsen (approved):

PR #4001, opened 2026-07-21
mldsa87: merge mu generation into signing

pjhades:
Based on the current usage, the only purpose of `mldsa87_generate_mu` is to generate a mu that is then immediately signed with `mldsa87_sign_mu`. However both APIs call `generate_key_internal` under the hood such that two key generation passes on the same seed are involved.

This commit introduces `mldsa87_generate_sign_mu_deterministic` as one combined step in order to reuse the private key generated in the first pass, reducing the number of cycles burnt on the call path.

Here's the comparison of cycles spent on command `CERTIFY_KEY_EXTENDED_MLDSA87` before and after this change (23882879 cycles fewer):

```
command                                cycles   command                                cycles
----------------------------------------------  ----------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87         83821879   CERTIFY_KEY_EXTENDED_MLDSA87         59939000
```

Impact on code size:

Before:
```
│ Runtime with ML-DSA    │ 101,248 │ 128  │ 103,684 │ 97.8% │ 2,308  │
│ Runtime without ML-DSA │ 79,984  │ 128  │ 103,684 │ 77.3% │ 23,572 │
```
After:
```
│ Runtime with ML-DSA    │ 101,224 │ 128  │ 103,684 │ 97.8% │ 2,332  │
│ Runtime without ML-DSA │ 79,984  │ 128  │ 103,684 │ 77.3% │ 23,572 │
```

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -1415,25 +1415,28 @@ pub fn mldsa87_sign_mu_from_sk(
 }
 
 /// Compute `mu` (the 64-byte message representative) from a private key seed
-/// and a message supplied via a [`ResponseBuffer`].
+/// and a message supplied via a [`ResponseBuffer`] and then sign it, reusing
+/// the private key from the generation.
 ///
 /// Implements FIPS 204 HashML-DSA with an empty (zero-length) context:
 /// `mu = SHAKE256(tr || 0x00 || 0x00 || M)` where `tr = SHAKE256(pk)` and
 /// `pk` is derived from `seed`.
-pub fn mldsa87_generate_mu(
+pub fn mldsa87_generate_sign_mu_deterministic(
+    out_encoded_signature: &mut [u8; MLDSA87_SIGNATURE_BYTES],
     private_key_seed: &[u8; MLDSA87_PRIVATE_SEED_BYTES],
     msg: &dyn ResponseBuffer,
     msg_range: core::ops::Range<usize>,
-    out_mu: &mut [u8; MLDSA87_MU_BYTES],
 ) -> Result<(), ResponseBufError> {
-    let mut pub_key = [0u8; MLDSA87_PUBLIC_KEY_BYTES];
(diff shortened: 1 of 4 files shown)


common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -1417,33 +1417,39 @@ pub fn mldsa87_sign_mu_from_sk(
...
-/// and a message supplied via a [`ResponseBuffer`].
+/// and a message supplied via a [`ResponseBuffer`] and then sign it, reusing
+/// the private key from the generation.
 ///
 /// Implements FIPS 204 HashML-DSA with an empty (zero-length) context:
 /// `mu = SHAKE256(tr || 0x00 || 0x00 || M)` where `tr = SHAKE256(pk)` and
 /// `pk` is derived from `seed`.
-pub fn mldsa87_generate_mu(
+pub fn mldsa87_generate_sign_mu_deterministic(
+    out_encoded_signature: &mut [u8; MLDSA87_SIGNATURE_BYTES],
```
ternst-nv, line 1425: [nit] Could we move the out parameter to be the last one?

ternst-nv (approved):
LGTM

common/crypto/mldsa/src/mldsa87.rs
pjhades: Yeah I noticed that, but I was trying to keep it consistent with other APIs, like https://github.com/chipsalliance/caliptra-sw/blob/caliptra-1.x/common/crypto/mldsa/src/mldsa87.rs#L820 and https://github.com/chipsalliance/caliptra-sw/blob/caliptra-1.x/common/crypto/mldsa/src/mldsa87.rs#L945. I think it's better sorting them out in a separate style-fixing PR.

ternst-nv: Good point, agree on a follow up PR

zhalvorsen (approved):

PR #4004, opened 2026-07-21
mldsa: refactor size optimizations

ternst-nv:
A collection of minor code optimizations and refactoring changes which together reduce the instruction memory usage of the mldsa_attestation feature by ~600 Bytes.

### DIFF
--- a/runtime/src/dpe_crypto.rs
+++ b/runtime/src/dpe_crypto.rs
@@ -24,6 +24,7 @@ use caliptra_drivers::{
 };
 use caliptra_error::CaliptraResult;
 use constant_time_eq::constant_time_eq;
+use core::marker::PhantomData;
 use crypto::{
     ecdsa::{
         curve_384::{EcdsaPub384, EcdsaSignature384},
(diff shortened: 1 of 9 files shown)


runtime/src/dpe_crypto.rs
```diff
@@ -56,65 +57,61 @@ enum Cdi {
...
+    pub ecc384: &'a mut Ecc384,
+    pub hmac384: &'a mut Hmac384,
+    pub key_vault: &'a mut KeyVault,
+}
+
 pub struct DpeCrypto<'a> {
-    trng: &'a mut Trng,
-    hmac384: &'a mut Hmac384,
-    key_vault: &'a mut KeyVault,
+    engines: CryptoEngines<'a>,
```
clundin25, line 75: uber nit: I don't think these need to be plural (even though under the hood there are several crypto peripherals).

runtime/src/invoke_dpe.rs
```diff
@@ -110,27 +106,32 @@ impl InvokeDpeMldsa87Cmd {
...
 fn execute(
     drivers: &mut Drivers,
     profile: CaliptraDpeProfile,
-    command: Command<'_>,
+    data: &[u8],
+    data_size: usize,
 ) -> CaliptraResult<()> {
+    if data_size > data.len() {
+        return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS);
+    }
```
clundin25, line 132: q: Should the caller just slice `data` to `data_size`?

clundin25 (approved):

runtime/src/invoke_dpe.rs
zhalvorsen: https://github.com/chipsalliance/caliptra-sw/pull/3963 does a very similar abstraction, but relies on the caller sending the full buffer instead of the truncated slice. So I would prefer to leave it as is

runtime/src/dpe_crypto.rs
ternst-nv: Resolved offline

pjhades (approved):

PR #4012, opened 2026-07-22
[emu]: Support running emulator against 2.0 and 2.1 firmware

clundin25:
Emulator selects major Caliptra version with a runtime flag, letting it operate against both 2.0 and 2.1 firmware binaries.

### DIFF
--- a/hw-model/src/lib.rs
+++ b/hw-model/src/lib.rs
@@ -70,6 +70,7 @@ pub use api::mailbox::{mbox_write_fifo, mbox_write_fifo_with_limit};
 pub use api_types::{DbgManufServiceRegReq, DeviceLifecycle, Fuses, SecurityState, U4};
 pub use caliptra_emu_bus::BusMmio;
 pub use caliptra_emu_cpu::{CodeRange, ImageInfo, StackInfo, StackRange};
+pub use caliptra_hw_model_types::CaliptraHwVersion;
 pub use output::ExitStatus;
 pub use output::Output;
 
@@ -247,6 +248,8 @@ impl Default for SubsystemInitParams<'_> {
 }
 
 pub struct InitParams<'a> {
+    pub hw_version: CaliptraHwVersion,
+
     // Fuse settings
     pub fuses: Fuses,
 
@@ -355,6 +358,7 @@ impl Default for InitParams<'_> {
                 Box::new(RandomEtrngResponses::new_from_stdrng())
             };
         Self {
+            hw_version: Default::default(),
             fuses: Default::default(),
             rom: Default::default(),
             dccm: Default::default(),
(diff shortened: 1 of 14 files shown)


hw-model/src/lib.rs
```diff
@@ -247,6 +248,8 @@ impl Default for SubsystemInitParams<'_> {
 }
 
 pub struct InitParams<'a> {
+    pub version: CaliptraVersion,
```
zhalvorsen: We should keep the terminology consistent with the build infrastructure


```suggestion
    pub hw_rev: CaliptraHwRevision,
```

sw-emulator/app/src/main.rs
```diff
@@ -259,6 +260,11 @@ fn main() -> io::Result<()> {
                 .required(true)
                 .value_parser(value_parser!(u32)),
         )
+        .arg(
+            arg!(--"caliptra-version" <VERSION> "Caliptra hardware version (2.0, 2.1, 2.2)")
```
zhalvorsen: Same here:


```suggestion
            arg!(--"hw-rev" <VERSION> "Caliptra hardware revision (2.0, 2.1, 2.2)")
```

sw-emulator/lib/types/src/lib.rs
```diff
@@ -39,3 +39,17 @@ emu_enum!(
     };
     Invalid
 );
+
+/// Caliptra Hardware Version
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum CaliptraVersion {
```
zhalvorsen: Can we put this in a place where it can have a single source of truth?

sw-emulator/lib/periph/src/root_bus.rs
```diff
@@ -386,8 +493,16 @@ impl CaliptraRootBus {
         );
 
         let sha512 = HashSha512::new(clock, key_vault.clone());
-        let ml_dsa87 = Abr::new(clock, key_vault.clone(), sha512.clone());
-        let sha3 = HashSha3::new();
+        let crypto_engine = match version {
+            CaliptraVersion::V2_0 => {
```
zhalvorsen: I'm not sure if this is already on your radar, also feel free to do this in a follow-on, but another HW change that happened between 2.0 and 2.1 is the mailbox size is reduced in subsystem mode in 2.1. We should handle that appropriately, as well.

zhalvorsen (commented):
This is super awesome!

sw-emulator/lib/periph/src/root_bus.rs
clundin25: No we will need to fix that. I think there will be a few more changes needed, but 2.0 boots atleast

zhalvorsen (approved):

PR #4014, opened 2026-07-22
runtime: store PQ.DevID public key hash in persistent data

pjhades:
Calculate the SHA-256 hash of the PQ.DevID public key after setting PQ seed and store it in persistent data. This saves a full key generation pass at the entry of `mldsa_dpe_env`, which is called at the beginning of each DPE command invocation.

Performance before and after this change:
```
command                                cycles   command                                cycles
----------------------------------------------  ----------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87        114696878   SELF_TEST_GET_RESULTS(+KATs)         48587836
SELF_TEST_GET_RESULTS(+KATs)         48577427   CERTIFY_KEY_EXTENDED_MLDSA87         35417842
```

The change takes 300+ bytes of additional instruction memory:
Before:
```
│ Runtime with ML-DSA    │ 101,308 │ 128  │ 103,684 │ 97.8% │ 2,248  │
```
after:
```
┌────────────────────────┬─────────┬──────┬─────────┬────────┬────────┐
│         Binary         │  Text   │ Data │ Budget  │ Used%  │  Free  │
├────────────────────────┼─────────┼──────┼─────────┼────────┼────────┤
│ Runtime with ML-DSA    │ 101,684 │ 128  │ 103,684 │ 98.19% │ 1,872  │
└────────────────────────┴─────────┴──────┴─────────┴────────┴────────┘
```

### DIFF
--- a/drivers/src/persistent.rs
+++ b/drivers/src/persistent.rs
@@ -49,11 +49,16 @@ pub const IDEVID_CSR_SIZE: u32 = 1024;
 pub const FMC_ALIAS_CSR_SIZE: u32 = 1024;
 pub const DPE_PL_CONTEXT_LIMITS_SIZE: u32 = 2;
 pub const PQ_DEVID_CDI_SIZE: u32 = 48;
+pub const PQ_DEVID_PUB_KEY_DIGEST_SIZE: u32 = 32;
 pub const PQC_STATUS_FLAGS_SIZE: u32 = 1;
 #[cfg(feature = "mldsa_attestation")]
 pub const PQC_MODE_ENABLED_FLAG: u8 = 1;
-pub const RESERVED_MEMORY_SIZE: u32 =
-    (3 * 1024) - 2 - PQ_DEVID_CDI_SIZE - PQC_STATUS_FLAGS_SIZE - MLDSA_EXPORTED_CDI_HANDLES_SIZE;
+pub const RESERVED_MEMORY_SIZE: u32 = (3 * 1024)
+    - 2
+    - PQ_DEVID_CDI_SIZE
+    - PQ_DEVID_PUB_KEY_DIGEST_SIZE
+    - PQC_STATUS_FLAGS_SIZE
+    - MLDSA_EXPORTED_CDI_HANDLES_SIZE;
 
 pub const PCR_LOG_MAX_COUNT: usize = 17;
 pub const FUSE_LOG_MAX_COUNT: usize = 62;
@@ -358,6 +363,7 @@ pub struct PersistentData {
     // `pq_devid_cdi()`, which hands out a reference exclusively once the seed
     // has been provisioned (see `pqc_mode_enabled`).
     pq_devid_cdi: PqDevIdCdi,
+    pq_devid_pub_key_digest: [u8; PQ_DEVID_PUB_KEY_DIGEST_SIZE as usize],
     pqc_status_flags: u8,
 
     #[cfg(all(feature = "mldsa_attestation", feature = "runtime"))]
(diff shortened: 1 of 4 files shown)


pjhades:
Hmm..using pattern matching gains some code size but is slower:
```diff
 use caliptra_cfi_derive::cfi_impl_fn;
 use caliptra_common::mailbox_api::{MailboxRespHeader, SetPqSeedReq, SET_PQ_SEED_SEED_SIZE};
 use caliptra_drivers::{hmac384_kdf, Array4x12, CaliptraError, CaliptraResult};
+use crypto::{Digest, Sha256};
 use zerocopy::{FromZeros, IntoBytes};

 pub struct SetPqSeedCmd;
@@ -42,16 +43,13 @@ impl SetPqSeedCmd {
         // Calculate the digest of the PQ.DevID public key and cache it
         // in the persistent data to avoid repeated key generation passes
         // involved in DPE invocation.
-        let (_, _, digest) = drivers.compute_mldsa_key_material()?;
+        let (_, _, Digest::Sha256(Sha256(digest))) = drivers.compute_mldsa_key_material()? else {
+            return Err(CaliptraError::RUNTIME_PQ_INVALID_PUBKEY_DIGEST);
+        };
         drivers
             .persistent_data
             .get_mut()
-            .set_pq_devid_pub_key_digest(
-                digest
-                    .as_slice()
-                    .try_into()
-                    .map_err(|_| CaliptraError::RUNTIME_PQ_INVALID_PUBKEY_DIGEST)?,
-            )?;
+            .set_pq_devid_pub_key_digest(digest)?;

         crate::packet::copy_to_mbox(drivers, MailboxRespHeader::default().as_mut_bytes())
     }
```

Performance:
```
command                                cycles
----------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87         76909419
```

Code size:
```
┌────────────────────────┬─────────┬──────┬─────────────────┬───────┬────────┐
│         Binary         │  Text   │ Data │     Budget      │ Used% │  Free  │
├────────────────────────┼─────────┼──────┼─────────────────┼───────┼────────┤
│ Runtime with ML-DSA    │ 101,540 │ 128  │ 103,684         │ 98.1% │ 2,016  │
└────────────────────────┴─────────┴──────┴─────────────────┴───────┴────────┘
```

ternst-nv:
The MLDSA algorithm relies on eventually converging to a small enough delta, and thus changes to the measurement (in this case runtime), can change the sampling loops.  My guess is the code itself isn't much slower, its instead the that the measurement had to be sampled more times due to the runtime `mu`.

To get a good sample we could run ~100 runs with slightly different labels and see if the distribution of performance changed.

zhalvorsen (dismissed):

clundin25 (dismissed):

pjhades:
I'm including the diff I pasted above since offline test results (average over 100 runs) show that it isn't significantly slow and it gives us back some imem room.

drivers/src/persistent.rs
```diff
@@ -426,6 +443,18 @@ impl PersistentData {
         }
     }
 
+    #[cfg(feature = "mldsa_attestation")]
+    pub fn set_pq_devid_pub_key_digest(
```
ternst-nv, line 447: [nit] Might be good to make a single `set_...` function to avoid having the cdi and digest set separately.

ternst-nv (approved):

drivers/src/persistent.rs
pjhades: Yeah I thought about it. But currently `Drivers::derive_devid_seed` pulls the CDI from persistent data, so it relies on the fact of CDI already being set. Having a single function to set both the CDI and digest means we need some plumbing. I tried this little refactoring:

```Rust
    pub(crate) fn derive_devid_seed(
        &mut self,
        seed: &mut Mldsa87Seed,
        pq_devid_cdi: Option<&PqDevIdCdi>,
    ) -> CaliptraResult<()>;

    pub fn compute_mldsa_key_material(
        &mut self,
        pq_devid_cdi: Option<&PqDevIdCdi>,
    ) -> CaliptraResult<(Mldsa87Seed, Mldsa87PubKey, Digest)>;
```

but the resulting code size is slightly bigger

```
┌────────────────────────┬─────────┬──────┬─────────┬────────┬────────┐
│         Binary         │  Text   │ Data │ Budget  │ Used%  │  Free  │
├────────────────────────┼─────────┼──────┼─────────┼────────┼────────┤
│ Runtime/ML-DSA  after  │ 100,780 │ 128  │ 103,684 │ 97.32% │ 2,776  │
├────────────────────────┼─────────┼──────┼─────────┼────────┼────────┤
│ Runtime/ML-DSA  before │ 100,632 │ 128  │ 103,684 │ 97.2%  │ 2,924  │
└────────────────────────┴─────────┴──────┴─────────┴────────┴────────┘
```

PR #4015, opened 2026-07-22
dpe: Update dpe library

ternst-nv:
Update the dpe library to use the recent imem optimizations.


zhalvorsen (dismissed):
It shouldn't affect anything but it is probably good practice to always update the golang DPE library for the verifier at the same time

zhalvorsen (approved):

Issue #4010, opened 2026-07-22
Need to check stack size with SIGN_WITH_EXPORTED_MLDSA CMD.


daeyoungkim-87:
* Caliptra Version(s) : **FW-2.1.1**
* Caliptra Mode Subsystem Mode

Hello,

https://github.com/chipsalliance/caliptra-sw/pull/3679
I performed the MLDSA test using the code that had been merged at that time.

After booting Caliptra Core RT, I generated a DPE Exported MLDSA Certificate using a DPE command.

When I then generate a signature using the **SIGN_WITH_EXPORTED_MLDSA** command, it appears that an error occurs while executing the sign_var() function in mldsa87.rs, possibly due to insufficient stack size.

Could you please investigate this issue?

For reference, the function call path I used is as follows.
-> Caliptra SIGN_WITH_EXPORTED_MLDSA
 -> SignWithExportedMldsaCmd::execute
  -> SignWithExportedMldsaCmd::mldsa_sign
   -> DpeCrypto::derive_key_pair_exported
    -> DpeCrypto::derive_key_pair_inner
     -> Mldsa87::key_pair
      -> Mldsa87::pct
       -> Mldsa87::sign_var
        -> trng.generate16()
         -> NMI error

zhalvorsen:
I tried to reproduce this stack overflow, but I wasn't able to. I created https://github.com/chipsalliance/caliptra-sw/pull/4019 which would relieve the pressure a little bit. Can you share some more details about what happened? Which commit are you running from, and log output? Maybe specific commands sequences and structures.

Maybe the best way to share your results would be through a unit test that can reproduce the error. Would you be able to put something like that together?

daeyoungkim-87:
[0001-test_debug_export_cdi_mldsa.zip](https://github.com/user-attachments/files/30338415/0001-test_debug_export_cdi_mldsa.zip)

I uploaded a patch file that includes debug code to calculate the available stack size in the affected test case and at each relevant code location, and then write the calculated values to the FW_EXT registers.

I ran the Caliptra-SW emulator using the following command:
`
~/cptra21_p4v_rt/caliptra-sw$ cargo test --manifest-path ~/cptra21_p4v_rt/caliptra-sw/Cargo.toml   -p caliptra-runtime   --features "itrng"   --test runtime_integration_tests   test_sign_with_exported_cdi_mldsa_raw_repro_flow   -- --nocapture`

[...]

PR #4019, opened 2026-07-23
Add stack optimizations to `SIGN_WITH_EXPORTED_MLDSA`

zhalvorsen:
#4010 saw a potential stack overflow while executing this command. This should relieve the pressure on the stack by ~7KB bytes.


clundin25 (approved):
@zhalvorsen do you know why tests didn't catch this?

ternst-nv (approved):
LGTM

pjhades (approved):

zhalvorsen:
> @zhalvorsen do you know why tests didn't catch this?

I'm not sure, it sounds like it is FPGA only which surprises me, but that also should have been caught with the tests. I'll keep looking into it further

PR #4022, opened 2026-07-23
mldsa: Implement libcaliptra mldsa bindings

ternst-nv:
Implement the bindings for the mldsa variant commands for lib caliptra. Include a battery of tests to validate functionality.

### DIFF
--- a/libcaliptra/inc/caliptra_api.h
+++ b/libcaliptra/inc/caliptra_api.h
@@ -233,6 +233,32 @@ int caliptra_sign_with_exported_ecdsa(struct caliptra_sign_with_exported_ecdsa_r
 // Revoke Exported CDI Handle
 int caliptra_revoke_exported_cdi_handle(struct caliptra_revoke_exported_cdi_handle_req *req, bool async);
 
+// PQC / ML-DSA-87 commands
+
+// Set PQ Seed (provision the PQ.DevID ML-DSA seed)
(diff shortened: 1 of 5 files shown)


libcaliptra/inc/caliptra_api.h
```diff
@@ -233,6 +233,32 @@ int caliptra_sign_with_exported_ecdsa(struct caliptra_sign_with_exported_ecdsa_r
 // Revoke Exported CDI Handle
 int caliptra_revoke_exported_cdi_handle(struct caliptra_revoke_exported_cdi_handle_req *req, bool async);
 
+// PQC / ML-DSA-87 commands
+
+// Set PQ Seed (provision the PQ.DevID ML-DSA seed)
+int caliptra_set_pq_seed(struct caliptra_set_pq_seed_req *req, bool async);
```
clundin25, line 245: This reminds me. @ternst-nv is there an API to query if Caliptra supports ML-DSA?

Should we add a bit in the Capabilities command?

ternst-nv: Hmm, thats a good question.  I had initially been thinking it would be based on version, i.e. if 1.3 PQC supported, but adding a capability bit seems like it would be a good idea as well.

clundin25 (approved):

pjhades (approved):

PR #4026, opened 2026-07-24
runtime: merge functions to reduce code size

pjhades:
- merge `*_dice_sign_*`
- merge `expand_*_ntt_mul_scalar`

Reduce code size by 230 bytes:

```
│ Runtime with ML-DSA    │ 100,212 │ 128  │ 103,684 │ 96.7% │ 3,344  │
│ Runtime with ML-DSA    │ 99,980  │ 128  │ 103,684 │ 96.5% │ 3,576  │
```

### DIFF
--- a/runtime/src/dice.rs
+++ b/runtime/src/dice.rs
@@ -153,20 +153,13 @@ impl GetRtAliasCertCmd {
 ///
 /// # Returns
 ///
-/// * `Ecc384Scalar` - The s portion of the LDevId cert signature
-fn get_dice_sign_portion(
-    handle: HandOffDataHandle,
-    dv: &DataVault,
-) -> CaliptraResult<Ecc384Scalar> {
-    let ds: DataStore = handle
-        .try_into()
-        .map_err(|_| CaliptraError::RUNTIME_FMC_CERT_HANDOFF_FAILED)?;
-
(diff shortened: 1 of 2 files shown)


runtime/src/dice.rs
```diff
@@ -144,59 +144,29 @@ impl GetRtAliasCertCmd {
...
 ///
 /// # Arguments
 ///
-/// * `persistent_data` - PersistentData
-/// * `dv` - DataVault
-///
-/// # Returns
-///
-/// * `Ecc384Scalar` - The r portion of the LDevId cert signature
-fn ldevid_dice_sign_r(
```
ternst-nv, line 157: The reduction here makes sense, but this will change the error being returned for Ldevid.  Could we setup the calling functions so they return the previously expected errors so we don't change behavior.

pjhades: Yeah, good catch, I missed the error type.

pjhades: The fix of error code even produces the windfall of 20 more bytes.
```
│ Runtime with ML-DSA    │ 99,960  │ 128  │ 103,684 │ 96.5% │ 3,596  │
```

ternst-nv (dismissed):
LGTM

clundin25 (approved):

PR #4036, opened 2026-07-29
hw-model: force cfi-test feature on caliptra-cfi-lib for host builds

chris-hinson:
caliptra-image-types / caliptra-lms-types pull in caliptra-cfi-lib with only the `cfi`/`cfi-counter` features. That transitive copy emits the firmware-only externs CFI_STATE_ORG and cfi_panic_handler without `cfi-test`, which the host link cannot resolve: debug builds fail with unresolved-symbol errors (LNK2019/LNK1120 on Windows), while release only links because dead-code elimination happens to drop the references.

Add caliptra-cfi-lib as a direct hw-model dependency with `cfi-test` enabled. Cargo unifies features additively across a single resolved instance of a crate, so this turns CFI_STATE into a real static and cfi_panic into a `panic!` for every consumer of hw-model, fixing a Windows debug link failure without touching the firmware-only call sites.

### DIFF
--- a/hw-model/Cargo.toml
+++ b/hw-model/Cargo.toml
@@ -18,6 +18,11 @@ coverage = ["dep:caliptra-coverage"]
 ocp-lock = []
 
 [dependencies]
+# Enable cfi-test for host builds. This dep exists solely to add the feature
+# to the transitive copy pulled in by caliptra-image-types/caliptra-lms-types.
+# Without this, debug builds fail with unresolved CFI_STATE_ORG/cfi_panic_handler.
+caliptra-cfi-lib = { workspace = true, features = ["cfi-test"] }
+
 anyhow.workspace = true
 bitfield.workspace = true
 bitflags.workspace = true


hw-model/Cargo.toml
```diff
@@ -18,6 +18,11 @@ coverage = ["dep:caliptra-coverage"]
 ocp-lock = []
 
 [dependencies]
+# Enable cfi-test for host builds. This dep exists solely to add the feature
+# to the transitive copy pulled in by caliptra-image-types/caliptra-lms-types.
+# Without this, debug builds fail with unresolved CFI_STATE_ORG/cfi_panic_handler.
+caliptra-cfi-lib = { workspace = true, features = ["cfi-test"] }
```
clundin25, line 24: Hmm, should we instead update the crate that is pulling in the CFI feature, to not pull it in?

chris-hinson: Since the "real" users of cfi-lib are caliptra-image-types and caliptra-lms-types, those consumers need the cfi-test feature excluded and as such we cannot change the feature list at the calitpra-sw Cargo.toml level.   These two crates are also the gateway by which hw-model pulls in cfi lib (according to cargo tree)

hw-model
  ├── caliptra-api
  │   └── caliptra-image-types
  │       └── caliptra-cfi-lib  ←
  │
  ├── caliptra-image-fake-keys
  │   └── caliptra-image-gen
  │       └── caliptra-image-types
  │           └── caliptra-cfi-lib  ←
  │
  └── (via caliptra-image-types)
      └── caliptra-lms-types
          └── caliptra-cfi-lib  ←

Messing with the deps that pull it in seems like it would be more headache than its worth since they both explicity want cfi-test to NOT be present (so the proper handlers can be linked into fw) 

This extra dep "hack" was the cleanest way i could find to get around cargo's feature unification rules and fix the linker errors for hw-model.   If there's a cleaner solution I'm happy to rewrite

clundin25: Hmm let me take a look tomorrow, I think `cfi` should not be getting included by default anymore, so I must have missed some crate previously

clundin25: Opened https://github.com/chipsalliance/caliptra-sw/pull/4039. Checked ROM and Runtime crate trees and CFI is still properly pulled in.

ROM hashes should not change, that should catch any mistakes.

chris-hinson: very nice - I like the solution of 4039.  Looks like its having some CI issues currently, but good to close this PR without merging once 4039 goes in

clundin25:
Closing in favor of #4039. Thanks for raising this!

PR #4038, opened 2026-07-29
drivers: dynamic power handling of entropy_src

### LINKED ISSUE #2868: [RFC] Dynamic power handling of entropy source

GwangbaeCHOI:
Keep entropy_src disabled outside of CSRNG Reseed windows to reduce power consumption, per RFC #2868. Caliptra RT leaves entropy_src MODULE_ENABLE=False by default and only enables it around a Reseed, with a headroom so the health-test startup is ready when the Reseed fires. RESEED_INTERVAL stays at the hardware default — the saving comes from gating entropy_src, not from shortening the interval.

reseed() waits for the CSRNG main SM to return to MainSmIdle before returning, since send_command returns when the Reseed is accepted into the staging FIFO, not when the CSRNG finishes executing it.

generate12/generate16 surface a runtime health-check failure via check_runtime_health, which calls check_for_alert_state only while entropy_src is enabled — calling it with entropy_src disabled would hang waiting for a main-sm state that never comes.

Fixes #2868

### DIFF
--- a/drivers/src/csrng.rs
+++ b/drivers/src/csrng.rs
@@ -28,8 +28,6 @@ use caliptra_registers::csrng::CsrngReg;
 use caliptra_registers::entropy_src::{self, regs::AlertFailCountsReadVal, EntropySrcReg};
 use caliptra_registers::soc_ifc::{self, SocIfcReg};
 
-use core::mem::MaybeUninit;
-
 // https://opentitan.org/book/hw/ip/csrng/doc/theory_of_operation.html#command-description
 pub const MAX_SEED_WORDS: usize = 12;
 const WORDS_PER_BLOCK: usize = 4;
@@ -102,31 +100,28 @@ impl Csrng {
         seed: Seed,
         persistent_data: PersistentDataAccessor,
     ) -> CaliptraResult<Self> {
-        const FALSE: u32 = MultiBitBool::False as u32;
-        const TRUE: u32 = MultiBitBool::True as u32;
-
         let mut result = Self { csrng, entropy_src };
         let e = result.entropy_src.regs_mut();
 
         // Configure and enable entropy_src if not already enabled.
         // If already enabled, assume it was configured correctly by a previous call.
-        if e.module_enable().read().module_enable() == FALSE {
+        if e.module_enable().read().module_enable() == MULTIBIT_FALSE {
             // Configure entropy_src
             let entropy_cfg = read_entropy_configuration(&soc_ifc.regs(), persistent_data);
             set_health_check_thresholds(e, entropy_cfg.entropy_cfg.clone());
 
             let rng_bit_enable = if entropy_cfg.entropy_cfg_extension.rng_bit_enable() {
-                TRUE
+                MULTIBIT_TRUE
             } else {
-                FALSE
+                MULTIBIT_FALSE
             };
 
             e.conf().write(|w| {
-                w.fips_enable(TRUE)
-                    .entropy_data_reg_enable(FALSE)
+                w.fips_enable(MULTIBIT_TRUE)
+                    .entropy_data_reg_enable(MULTIBIT_FALSE)
                     // THRESHOLD_SCOPE=FALSE so adaptive-proportion and Markov health
                     // tests score each of the 4 RNG bus lanes individually.
-                    .threshold_scope(FALSE)
+                    .threshold_scope(MULTIBIT_FALSE)
                     .rng_bit_enable(rng_bit_enable)
                     .rng_bit_sel(entropy_cfg.entropy_cfg_extension.rng_bit_sel())
             });
@@ -138,9 +133,9 @@ impl Csrng {
                 & 1
                 == 1
             {
-                e.entropy_control().modify(|w| w.es_type(TRUE));
+                e.entropy_control().modify(|w| w.es_type(MULTIBIT_TRUE));
             }
-            e.module_enable().write(|w| w.module_enable(TRUE));
+            e.module_enable().write(|w| w.module_enable(MULTIBIT_TRUE));
             check_for_alert_state(result.entropy_src.regs())?;
 
             // Lock entropy_src configuration if not in debug mode.
@@ -153,18 +148,47 @@ impl Csrng {
                 e.sw_regupd().modify(|w| w.sw_regupd(false));
             }
         }
-
         let c = result.csrng.regs_mut();
 
-        if c.ctrl().read().enable() == FALSE {
-            c.ctrl()
-                .write(|w| w.enable(TRUE).sw_app_enable(TRUE).read_int_state(TRUE));
+        if c.ctrl().read().enable() == MULTIBIT_FALSE {
+            c.ctrl().write(|w| {
+                w.enable(MULTIBIT_TRUE)
+                    .sw_app_enable(MULTIBIT_TRUE)
+                    .read_int_state(MULTIBIT_TRUE)
+            });
         }
 
         send_command(&mut result.csrng, Command::Uninstantiate)?;
         send_command(&mut result.csrng, Command::Instantiate(seed))?;
 
         Ok(result)
+    }
+
+    /// Return 4 randomly generated [`u32`]s as a tuple.
+    ///
+    /// The return type is a tuple of 4 `u32`s which fits entirely in registers
+    /// on RV32, guaranteeing no hidden out-pointer and no `memcpy` on the
+    /// return path regardless of optimization level. Safe to call before CFI
+    /// counters are initialized.
+    pub fn generate4(&mut self) -> CaliptraResult<(u32, u32, u32, u32)> {
+        let token = self.manage_entropy_src_and_reseed()?;
+        self.do_generate4(token)
+    }
+
+    fn do_generate4(&mut self, _token: GenerateToken) -> CaliptraResult<(u32, u32, u32, u32)> {
+        send_command(
+            &mut self.csrng,
+            Command::Generate {
+                num_128_bit_blocks: 1,
+            },
+        )?;
+        wait::until(|| self.csrng.regs().genbits_vld().read().genbits_vld());
+        Ok((
+            self.csrng.regs().genbits().read(),
+            self.csrng.regs().genbits().read(),
+            self.csrng.regs().genbits().read(),
+            self.csrng.regs().genbits().read(),
+        ))
     }
 
     /// Return 12 randomly generated [`u32`]s.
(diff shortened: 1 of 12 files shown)


drivers/src/csrng.rs
```diff
@@ -185,7 +184,7 @@ impl Csrng {
     /// }
     /// ```
     pub fn generate12(&mut self) -> CaliptraResult<[u32; 12]> {
-        check_for_alert_state(self.entropy_src.regs())?;
```
GwangbaeCHOI, line 188: `generate12` only uses the CSRNG and the CSRNG doesn't interact with the entropy source to generate random bits. Checking the entropy source state is not necessary.

drivers/src/csrng.rs
```diff
@@ -235,7 +287,7 @@ impl Csrng {
     ///
     /// Returns an error if the internal generate command fails.
     pub fn generate16(&mut self) -> CaliptraResult<[u32; 16]> {
-        check_for_alert_state(self.entropy_src.regs())?;
```
GwangbaeCHOI, line 238: Same with the `generate12` case.
https://github.com/chipsalliance/caliptra-sw/pull/4038/changes#r3670202720

runtime/src/main.rs
```diff
@@ -103,6 +103,10 @@ pub extern "C" fn entry_point() -> ! {
     if !drivers.persistent_data.get().rom.fht.is_valid() {
         handle_fatal_error(caliptra_drivers::CaliptraError::RUNTIME_HANDOFF_FHT_NOT_LOADED.into());
     }
+
+    cprintln!("[rt] Disable entropy source");
+    drivers.trng.disable_entropy_source();
```
GwangbaeCHOI, line 108: Disable the entropy source once the RT is loaded. Therefore, disabling the entropy source won't delay the boot time.

drivers/src/csrng.rs
```diff
@@ -222,6 +221,59 @@ impl Csrng {
         }
     }
 
+    /// Number of Generate calls before the reseed interval at which entropy_src
+    /// is pre-enabled, so it is ready by the time the reseed is needed.
+    /// Covers the ~2 ms entropy_src startup at typical generate cadence.
```
GwangbaeCHOI, line 226: Tests were done on FPGA

sw-emulator/lib/periph/src/csrng.rs
GwangbaeCHOI, line 1: Addding reseeding related logics into the sw emulator

clundin25:
@GwangbaeCHOI can you commit the new ROM hash using `cargo xtask update-frozen-images`?

drivers/src/csrng.rs
```diff
@@ -222,6 +221,59 @@ impl Csrng {
...
+    pub const ENTROPY_SRC_PRE_ENABLE_HEADROOM: u32 = 5;
+
+    /// Disable entropy_src.
+    ///
+    /// Must be called only after any in-flight Reseed has consumed its seed
+    /// from the esfinal FIFO; disabling mid-Reseed clears that FIFO and hangs
+    /// the CSRNG in `MainSmReseedPrep`. Pair with `wait_for_csrng_idle`.
+    pub fn disable_entropy_source(&mut self) {
+        const FALSE: u32 = MultiBitBool::False as u32;
```
clundin25: nit: I think this and `TRUE` are already defined in `with_seed`.

Maybe we should just elevate that scope so we can reuse the constant definition?

drivers/src/csrng.rs
```diff
@@ -222,6 +221,59 @@ impl Csrng {
...
+    /// is pre-enabled, so it is ready by the time the reseed is needed.
+    /// Covers the ~2 ms entropy_src startup at typical generate cadence.
+    pub const ENTROPY_SRC_PRE_ENABLE_HEADROOM: u32 = 5;
+
+    /// Disable entropy_src.
+    ///
+    /// Must be called only after any in-flight Reseed has consumed its seed
+    /// from the esfinal FIFO; disabling mid-Reseed clears that FIFO and hangs
+    /// the CSRNG in `MainSmReseedPrep`. Pair with `wait_for_csrng_idle`.
+    pub fn disable_entropy_source(&mut self) {
```
clundin25: Can we track this state so we don't inadvernately do this? Preferably at compile time since I think this would be a logic bug (Type state pattern may be a good match?)

drivers/src/csrng.rs
```diff
@@ -222,6 +221,59 @@ impl Csrng {
...
+    /// Called before every Generate command. Keeps entropy_src disabled except
+    /// during a short window around each reseed:
+    ///
+    /// Phase 1 — pre-enable (`counter >= interval - HEADROOM`): enable
+    /// entropy_src in the background so its startup health checks overlap with
+    /// the next few Generates.
+    ///
+    /// Phase 2 — reseed (`counter >= interval`): reseed from entropy_src,
+    /// then disable it again. The Generate runs with entropy_src off.
+    fn manage_entropy_src_and_reseed(&mut self) -> CaliptraResult<()> {
```
clundin25: Similarly, I think we could split the Generate commands to a different type that this function returns, guaranteeing we never forget to reseed

clundin25:
@zhalvorsen @vmhatre PTAL

GwangbaeCHOI:
> @GwangbaeCHOI can you commit the new ROM hash using `cargo xtask update-frozen-images`?

I just added it.

drivers/src/csrng.rs
GwangbaeCHOI: Fixed

drivers/src/csrng.rs
GwangbaeCHOI: Were you thinking of something like Csrng&lt;S> or a different approach? I guess the former would propagate the type parameter into Trng, Drivers, and every ROM/RT function taking &mut Drivers. So, I wonder if this was the approach that you were thinking.

GwangbaeCHOI: For now, changed `pub` to `pub(crate)` on disable_entropy_source and enable_entropy_source so they are not accessible outside the crate.

clundin25:
@GwangbaeCHOI looks like there is a merge conflict, do you mind fixing it?

drivers/src/csrng.rs
clundin25: Good point, that sounds good to me. We can consider a future refactor since it will touch so much code.

I think you are right, we would need to change the public API to make it worthwhile

drivers/src/csrng.rs
GwangbaeCHOI: Update it.

GwangbaeCHOI:
> @GwangbaeCHOI looks like there is a merge conflict, do you mind fixing it?

I updated it. I guess the conflict is resolved.

clundin25:
PR LGTM but looks like there are some test failures. @GwangbaeCHOI can you take a look?

drivers/test-fw/src/bin/csrng_runtime_health_fail_tests.rs
```diff
@@ -36,10 +36,10 @@ fn test_runtime_health_check_failure() {
...
         .expect("CSRNG should pass startup health testing");
 
-    // First generate should start seeing bad entropy
-    // The emulator will be configured to provide good entropy for startup,
-    // then bad entropy for runtime operations
-    let result = csrng.generate12();
+    // Reseed from entropy_src — this enables entropy_src, consumes from it,
+    // and is where the health check fires. Bad entropy injected by the test
+    // harness will be detected here.
+    let result = csrng.reseed(CsrngSeed::EntropySrc);
```
GwangbaeCHOI, line 42: Changed to `reseed` as `generate12` never cause a entropy source failure.

drivers/src/csrng.rs
```diff
@@ -167,6 +164,33 @@ impl Csrng {
         Ok(result)
     }
 
+    /// Return 4 randomly generated [`u32`]s as a tuple.
+    ///
+    /// The return type is a tuple of 4 `u32`s which fits entirely in registers
+    /// on RV32, guaranteeing no hidden out-pointer and no `memcpy` on the
+    /// return path regardless of optimization level. Safe to call before CFI
+    /// counters are initialized.
+    pub fn generate4(&mut self) -> CaliptraResult<(u32, u32, u32, u32)> {
```
GwangbaeCHOI, line 173: Added to avoid memcpy before the CFI initialization. Dedicated function is added to avoid using a unsafe block as it was done in old `generate12`.

GwangbaeCHOI:
> PR LGTM but looks like there are some test failures. @GwangbaeCHOI can you take a look?

Failed testcases are addressed.

clundin25 (approved):

PR #4039, opened 2026-07-29
[hw-model]: Remove CFI from dependency tree

clundin25:
The CFI crates are annoying to compile on non-firmware targets. The CFI should not get pulled into non-firwmare and this cleans up the crate defaults that are pulling it in. 

Now CFI no longer gets pulled into the caliptra-hw-model

```
caliptra-hw-model v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/hw-model) 
├── anyhow v1.0.98
├── bit-vec v0.6.3
├── bitfield v0.14.0
├── bitflags v2.11.0
├── caliptra-api v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/api)
│   ├── bitflags v2.11.0
│   ├── caliptra-api-types v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/api/types)
│   │   └── caliptra-image-types v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/image/types)
│   │       ├── caliptra-error v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/error)
│   │       ├── caliptra-lms-types v0.1.0 (/usr/local/google/home/clundin/code/caliptra-sw/lms-types)
│   │       │   ├── serde v1.0.228
│   │       │   │   ├── serde_core v1.0.228
│   │       │   │   └── serde_derive v1.0.228 (proc-macro)
│   │       │   │       ├── proc-macro2 v1.0.95
│   │       │   │       │   └── unicode-ident v1.0.18
│   │       │   │       ├── quote v1.0.40
│   │       │   │       │   └── proc-macro2 v1.0.95 (*)
│   │       │   │       └── syn v2.0.117
│   │       │   │           ├── proc-macro2 v1.0.95 (*)
│   │       │   │           ├── quote v1.0.40 (*)
│   │       │   │           └── unicode-ident v1.0.18
│   │       │   ├── serde_derive v1.0.228 (proc-macro) (*)
│   │       │   ├── zerocopy v0.8.26
│   │       │   │   └── zerocopy-derive v0.8.26 (proc-macro)
│   │       │   │       ├── proc-macro2 v1.0.95 (*)
│   │       │   │       ├── quote v1.0.40 (*)
│   │       │   │       └── syn v2.0.117 (*)
│   │       │   └── zeroize v1.8.2
│   │       │       └── zeroize_derive v1.4.2 (proc-macro)
[...]

PR #4041, opened 2026-07-30
[cherry-pick][hw-model]: Remove CFI from dependency tree (#4039)

clundin25:
The CFI crates are annoying to compile on non-firmware targets. The CFI should not get pulled into non-firwmare and this cleans up the crate defaults that are pulling it in. 

Now CFI no longer gets pulled into the caliptra-hw-model

(cherry picked from commit 13e460a5deac9ca6094f60b4c1fdc2a2871d5058)

### DIFF
--- a/coverage/Cargo.toml
+++ b/coverage/Cargo.toml
@@ -18,4 +18,4 @@ caliptra-builder.workspace = true
 elf.workspace = true
 regex.workspace = true
 caliptra-image-types.workspace = true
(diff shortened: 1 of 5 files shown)


clundin25:
@chris-hinson FYI - shoot me a ping if this is remains broken

chris-hinson:
@clundin25 looks like CI is failing on the cargo.lock check

zhalvorsen (approved):

PR #4049, opened 2026-07-31
Build 1.0 and etrng FPGA variants for 1.x nightly

nquarton:
Also requires infra repo change: https://github.com/chipsalliance/caliptra-infra/pull/74


.github/workflows/bitstream-build.yml
```diff
@@ -73,45 +97,57 @@ jobs:
...
 
       - name: Create Pull Request
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          BASE_URL="https://storage.googleapis.com/caliptra-github-ci-bitstreams/v1/${{ github.sha }}"
+          BODY="Automated update for run ${{ github.run_id }}."$'\n'
+          for VARIANT in $(echo '${{ needs.setup.outputs.variants }}' | jq -r '.[]'); do
+            BODY+="- [${VARIANT}](${BASE_URL}/manifest-${VARIANT}.toml)"$'\n'
+          done
```
clundin25, line 148: @nquarton FYI this check is expected to be broken (Changes look good though!)

.github/workflows/bitstream-build.yml
```diff
@@ -73,45 +97,57 @@ jobs:
...
-          
-          # Retry for up to 5 minutes (30 * 10s)
-          for i in {1..30}; do
-            if curl -s -f -o "$TARGET_PATH" "$URL"; then
-              echo "Successfully downloaded manifest!"
-              break
-            fi
-            echo "Waiting for manifest to be uploaded... (attempt $i)"
-            sleep 10
+          BASE_URL="https://storage.googleapis.com/caliptra-github-ci-bitstreams/v1/${{ github.sha }}"
```
clundin25, line 102: Hmm the upload logic might step over itself. I wonder if we should tweak https://github.com/chipsalliance/caliptra-infra/pull/74 to instead be the namespace? That way we have one folder per variant, where each variant has artifacts with the same name.

WDYT?

clundin25 (commented):
Looks good! I think that the upload step may overwrite items currently

.github/workflows/bitstream-build.yml
nquarton: Ah yeah it definitely will. I like that approach better. Just pushed changes for it.

nquarton:
Changing to have each variant use its own directory/namespace. Caliptra-infra change: https://github.com/chipsalliance/caliptra-infra/pull/75

clundin25 (approved):

PR #4057, opened 2026-08-04
mldsa: Add workflow integration test

ternst-nv:
Add two addtional tests which flush out the full functionality of the mldsa attestation flow.  This includes setting the pq seed, obtaining the CSR, populating the certificate, and then performing SPDM like operations against the DPE context.

### DIFF
--- a/runtime/tests/runtime_integration_tests/test_mldsa_attestation_workflow.rs
+++ b/runtime/tests/runtime_integration_tests/test_mldsa_attestation_workflow.rs
@@ -96,24 +96,45 @@ fn new_cert_builder(subject: &X509NameRef, serial: u32, pubkey: &PKey<Public>) -
     b
 }
 
-/// Build a self-signed ML-DSA test Root CA certificate.
-fn build_root_ca(ca_key: &PKey<Private>, cn: &str) -> X509 {
-    let name = x509_name(cn);
-    let ca_pub = PKey::public_key_from_der(&ca_key.public_key_to_der().unwrap()).unwrap();
-    let mut b = new_cert_builder(&name, 1, &ca_pub);
-    b.set_issuer_name(&name).unwrap();
-    b.sign(ca_key, MessageDigest::null()).unwrap();
-    b.build()
+/// A test post-quantum Root CA: an ML-DSA-87 key pair and its self-signed CA
+/// certificate. Guides the "stand up a CA, then issue certificates from it"
+/// flow so a DevID cert can only be produced from a CA.
+struct TestPqCa {
+    key: PKey<Private>,
+    cert: X509,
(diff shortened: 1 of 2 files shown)


pjhades (dismissed):

runtime/tests/runtime_integration_tests/test_mldsa_attestation_workflow.rs
```diff
@@ -0,0 +1,447 @@
...
+    execute_dpe_cmd, get_pq_csr, mldsa_csr_public_key, provision_pq_seed, run_pqc_rt_test,
+    DpeResult, TEST_DIGEST_MLDSA, TEST_LABEL,
+};
+
+const PROFILE: CaliptraDpeProfile = CaliptraDpeProfile::Mldsa;
+
+/// Subject CN the ML-DSA DPE stamps as the leaf certificate's Issuer CN (see
+/// `runtime/src/dpe_platform.rs`). The populated DevID cert uses the same Subject
+/// CN so the leaf's issuer name resolves to it.
+const DEVID_SUBJECT_CN: &str = "Caliptra 1.0 Rt Alias";
```
clundin25, line 56: I wonder if we should bump the version number in these certificates?

runtime/tests/runtime_integration_tests/test_mldsa_attestation_workflow.rs
```diff
@@ -0,0 +1,447 @@
...
+}
+
+/// Issue a DevID certificate certifying `devid_pubkey` (the PQ.DevID key),
+/// signed by the Root CA. The Subject CN matches the DPE leaf issuer CN.
+fn issue_devid_cert(ca_key: &PKey<Private>, ca_cert: &X509, devid_pubkey: &PKey<Public>) -> X509 {
+    let subject = x509_name(DEVID_SUBJECT_CN);
+    let mut b = new_cert_builder(&subject, 2, devid_pubkey);
+    b.set_issuer_name(ca_cert.subject_name()).unwrap();
+    b.sign(ca_key, MessageDigest::null()).unwrap();
+    b.build()
```
clundin25: nit: I think it would be useful to wrap this with in a type. I think this is a valuable test case, and using a `struct` to "guide" the call flow may make it easier to re-use

clundin25 (dismissed):

clundin25 (approved):

runtime/tests/runtime_integration_tests/test_mldsa_attestation_workflow.rs
ternst-nv: Good point, lets discuss offline and put up another PR if we decide to bump this.

PR #4060, opened 2026-08-04
Add PR CI sanity check with legacy 1.x ROMs

### LINKED ISSUE #1793: PR actions should run at least the smoke test against older ROMS
We run the full test suite against older, frozen ROMS on the nightly/release testing. We should add some small set of tests to run against the older ROMs for each PR to catch some basic issues earlier. Starting with the smoke test may make the most sense.

nquarton:
Previously, legacy ROM testing was not performed until the nightly release. This allows basic compatibility issues to be caught prior to merging a PR without adding too much testing overhead.

### DIFF
--- a/.github/workflows/fw-test-emu.yml
+++ b/.github/workflows/fw-test-emu.yml
@@ -22,6 +22,10 @@ on:
       runner:
         default: '["ubuntu-22.04"]'
         type: string
+      test-filter:
+        default: ""
+        type: string
+        description: "nextest -E filter expression; empty means run all tests"
 
 jobs:
   build_and_test:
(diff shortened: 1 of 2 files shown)


.github/workflows/fw-test-emu.yml
```diff
@@ -89,7 +93,8 @@ jobs:
           cargo-nextest nextest run \
             --features="${{ inputs.extra-features }}" \
             --no-fail-fast \
-            --profile=nightly
+            --profile=nightly \
+            ${{ inputs.test-filter != '' && format('-E ''{0}''', inputs.test-filter) || '' }}
```
clundin25, line 97: I think this may override the filter-set from the profile?

nquarton: That's a good point. I looked into this and I think they both apply [according to the docs](https://nexte.st/docs/filtersets/#filtersets-with-the-default-set) unless you explicitly pass "ignore-default-filter"

"_filtersets on the command line are intersected with the default filter"_

clundin25: TIL! Thanks!

clundin25 (approved):

PR #4062, opened 2026-08-05
mldsa: Increase pqc watchdog timer

ternst-nv:
The PQC operations can take extensive amounts of time due to their performing software calculations.  For MLDSA commands which can trigger this behavior, increase the Watchdog timer to appropriately handle it.

Also adjust the workflow tests so similar issues will be detected automatically.

### DIFF
--- a/runtime/src/certify_key_extended_mldsa.rs
+++ b/runtime/src/certify_key_extended_mldsa.rs
@@ -34,12 +34,10 @@ impl CertifyKeyExtendedMldsa87Cmd {
             return Err(CaliptraError::RUNTIME_PQC_NOT_INITIALIZED);
         }
 
-        // The mldsa variant of the certify key extended can be quite long, the current longest
-        // measurement over 400_000_000 cycles. As such support 800_000_000 cycles prior to timeout.
-        caliptra_common::wdt::start_wdt(
-            &mut drivers.soc_ifc,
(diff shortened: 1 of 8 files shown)


pjhades (approved):

clundin25:
I think we are setting the WD via registers in the MCU that wraps Caliptra core. Does it make sense to recommend a longer WDT to integrators?

Now it seems we have to track two different WDT timeouts.

CC @zhalvorsen

ternst-nv:
Hmm, my understanding was the watchdog timer was set here: https://github.com/chipsalliance/caliptra-sw/blob/8a97f42afec4c7b535b9bad2902c63e34ba8ed83/runtime/src/lib.rs#L416.  Are there additional locations an MCU wdt is being set?

zhalvorsen (approved):

zhalvorsen:
The initial WDT settings are given by SoC manager during boot, but that just sets the WDT for the length of time from release from reset to RT starting. Then RT does its own watchdog management by starting it and stopping it while executing commands

clundin25:
> The initial WDT settings are given by SoC manager during boot, but that just sets the WDT for the length of time from release from reset to RT starting. Then RT does its own watchdog management by starting it and stopping it while executing commands

Ah, my mistake, that was exactly what I was thinking about!

clundin25 (approved):

PR #4063, opened 2026-08-05
[build] update Rust toolchain to 1.95

mhatrevi:
Update the pinned Rust toolchain, ROM MSRV, documentation, and FPGA workflows to Rust 1.95.


runtime/src/invoke_dpe.rs
```diff
@@ -196,11 +196,9 @@ impl InvokeDpeCmd {
 
         // Check if command can be executed
         match command {
-            Command::InitCtx(cmd) => {
+            Command::InitCtx(cmd) if InitCtxCmd::flag_is_simulation(cmd) => {
```
ternst-nv, line 200: This behavior looks to be the same, but I'm curious what will happen if the command is not executed and we slide through this.  Should we be returning an error?

clundin25 (dismissed):

zhalvorsen (dismissed):

ternst-nv (dismissed):
LGTM - one question

zhalvorsen (dismissed):

zhalvorsen (dismissed):

runtime/src/invoke_dpe.rs
mhatrevi: Good question. This match only applies Caliptra-specific preflight checks; it does not dispatch the command. A non-simulation InitCtx falls through to the wildcard arm and is still passed unconditionally to invoke_dpe_cmd below. DPE then returns ArgumentNotSupported, which is placed in the response header. The follow-up commit 9eaf8ed8 added the clarifying comment, and test_invoke_dpe::test_dpe_header_error_code covers this exact path; I reran it on the current head and it passes.

zhalvorsen (approved):

PR #3958, opened 2026-07-08
ML-DSA firmware implementation optimizations

zhalvorsen:
1. (Persistent State Compression):  s1_ntt  and  s2_ntt  removed;  sigma  stored instead. 
  2. (Hint Data Structure Reduction): Dense bitmasks replaced with sparse struct.
  3. (Key Generation Memory Aliasing):  s1_ntt ,  s2 , and  t1  unionized; accumulator t computed in-place over output buffer  priv_key.t0_ntt .
  4. (Streaming Matrix Multiplication): Evaluate column-by-column, eliminating the 7 KB full mask vector allocation during signing. 
  5. (Streaming Digest Absorption): Eliminated 8 KB  w1  vector and 1,024-byte  w1_encoded  buffer; scalars encoded directly into SHAKE-256 per iteration. 
  6. (Component-by-Component Signature Generation): Eliminated 47+ KB of dynamic intermediate vectors ( s1_ntt ,  cs1 ,  y / z ,  s2_ntt ,  cs2 ,  r0 ,  ct0 ) in favor of 1 KB loop-local scalars.
  7. Section 7 (Buffer Sharing): Reused  tr_mu  across API wrappers and shared  tmp  and  VerifyState  stack storage.

PR #4066, opened 2026-08-06
mldsa87: check z-max and continue early

pjhades:
Check if z-max is valid early in the rejection sampling loop and continue early if it's invalid. Skip the remaining computation in the same iteration to speed up worst cases for about 4 million cycles (averaged across 100 test runs, each with a slightly different label):

```
command                                cycles
----------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87         74352703

command                                cycles
----------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87         70815760
```

It's not a substantial performance boost but a nice-to-have since it's almost free.

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -1095,6 +1095,11 @@ fn sign_internal_with_mu(
         // Copy of rho_prime; dead from here, and this runs per rejection iteration.
         mask_seed.zeroize();
 
+        if ct_ge(z_max, GAMMA1.wrapping_sub(BETA)) != 0 {
+            kappa += 7;
+            continue;
+        }
+
         let mut r0_max = 0u32;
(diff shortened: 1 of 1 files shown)


zhalvorsen (approved):

ternst-nv (commented):
The optimization looks good to me, only concern if short circuiting the additional multiplication could somehow expose a timing attack by exposing at what point in the rejection loop things fail?

pjhades:
Would need some input from a crypto expert. @zhalvorsen could you comment? It seems the code before your [previous stack optimization](https://github.com/chipsalliance/caliptra-sw/pull/3958/changes) included multiple continue points. Was there a concern about the previous implementation?

zhalvorsen:
I'll ping someone on our side to take a look. We had a crypto expert look at the previous continue points

PR #4070, opened 2026-08-07
Cherry-pick DPE context updates to main

parvathib:
Cherry-picks the DPE context update series from `caliptra-2.0`  to  `main`.

Commits:

• #3729:  runtime: Anchor MCU RT in Caliptra-managed DPE chain 
• #3988:  runtime: Harden DPE index cache 
• #3996:  runtime: Test DPE index cache across hitless update 
• #3989:  runtime: Add SoC manifest preamble DPE contexts 

Main-branch conflict adaptations:

• Adapted persistent DPE data changes to main ’s  `fw.dpe`  layout.
• Preserved owner authorization manifest handling already present on main .
• Avoided stack overflow in recovery `SET_AUTH_MANIFEST` by borrowing  `AuthManifestPreamble`  from the manifest buffer instead of copying it.

### DIFF
--- a/runtime/src/stash_measurement.rs
+++ b/runtime/src/stash_measurement.rs
@@ -86,6 +86,25 @@ impl StashMeasurementCmd {
         };
 
         if let DpeErrorCode::NoError = dpe_result {
+            if metadata == &[2, 0, 0, 0] {
+                let mcu_rt_dpe_context_idx = drivers
+                    .persistent_data
+                    .get()
+                    .fw
+                    .dpe
+                    .state
+                    .get_active_context_pos(&ContextHandle::default(), locality)
+                    .map_err(|_| CaliptraError::RUNTIME_DPE_CONTEXT_NOT_FOUND)?;
+                drivers
+                    .persistent_data
+                    .get_mut()
+                    .fw
+                    .dpe
+                    .caliptra_managed_dpe_context_indices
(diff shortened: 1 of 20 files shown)


runtime/src/stash_measurement.rs
```diff
@@ -86,6 +86,26 @@ impl StashMeasurementCmd {
         };
 
         if let DpeErrorCode::NoError = dpe_result {
+            if metadata == &[2, 0, 0, 0] {
```
zhalvorsen, line 89: I just realized, doesn't this allow MCU to update its own measurement by calling stash measurement with that `metadata`? We should enforce a way that this portion is only done when Caliptra calls it internally versus coming from the mailbox

zhalvorsen (dismissed):

runtime/src/stash_measurement.rs
parvathib: Good catch. Yes, we should not allow the mailbox path to use the reserved MCU FW metadata value to update the Caliptra-managed MCFW context

I’ll address this with a follow-up PR by separating mailbox vs internal stash behavior and rejecting the reserved MCU FW metadata value on the mailbox path

zhalvorsen: Sounds good, thanks!

zhalvorsen (dismissed):

zhalvorsen (approved):

mhatrevi (approved):

PR #4071, opened 2026-08-07
Add stack sampler to emulated CPU and generate flamegraph for PQC tests

pjhades:
Introduce an optional stack sampler to the CPU, which traces JAL and
JALR instructions to maintain a shadow call stack, and record a stack
sample periodically under the given sample rate.

Then, wire the stack sampler to potentially slow PQC commands, namely,
`CERTIFY_KEY_EXTENDED`, `GET_PQ_CSR` and `MLDSA87_SIGNATURE_VERIFY`. Render a
flamegraph after each command completes.

Example flamegraph for `CERTIFY_KEY_EXTENDED`: 
CERTIFY_KEY_EXTENDED_MLDSA87
[CERTIFY_KEY_EXTENDED_MLDSA87.svg.zip](https://github.com/user-attachments/files/30835313/CERTIFY_KEY_EXTENDED_MLDSA87.svg.zip)

This is not currently wired to the emulator's CLI interface due to the following reasons:
- The emulator is separated from the builder and expects pre-built binaries from CLI. So it needs additional CLI arguments to read the ELF symbols.
- The emulator already supports an instruction tracer. Thus we could maybe refactor the tracer so that the emulator takes a `Vec<InstrTracer>` where both instruction logging and stack sampling could happen.


runtime/tests/runtime_integration_tests/common.rs
```diff
@@ -609,3 +614,46 @@ pub fn calculate_cptra_config_init_vals_hash<T: HwModel>(
 
     hasher.finalize().into()
 }
+
+#[allow(dead_code)]
```
ternst-nv: [nit] There look to be some unnecessary `allow` statements here and in later code, can they be removed?

runtime/tests/runtime_integration_tests/test_command_timing.rs
```diff
@@ -47,15 +51,39 @@ impl CommandSampler for CycleSampler {
...
+}
+
+fn measure_timing(args: RuntimeTestArgs) {
+    // The binary is cached so we'll be using the same ELF as the one used in test.
+    let fwid = args
+        .test_fwid
+        .unwrap_or(if cfg!(feature = "fpga_realtime") {
+            &APP_WITH_UART_FPGA
+        } else {
+            &APP_WITH_UART
```
ternst-nv: This isn't right for the default, since we're currently using the MLDSA attestation fwid.  Could we fail if this isn't set?

ternst-nv (dismissed):
LGTM - A couple small nits.

I also wonder if we could tie this functionality to the model specifically.  Then this could be easily available to all tests which wanted to profile themselves.  All they would have to do is enable the stack trace and then clear and gen flame graph.

This is really cool functionality, thanks for adding Jing!

runtime/tests/runtime_integration_tests/common.rs
pjhades: This is because the function is currently only called by PQC tests. I'll move it somewhere else while refactoring it.

runtime/tests/runtime_integration_tests/test_command_timing.rs
pjhades: Good catch! Will fix it.

pjhades:
I've done some refactoring here to wire the functionality to the emulated model. A widespread change is to make `caliptra_builder::build_and_sign_image` return an additional `struct Elfs` that contains the ELF byte streams of the runtime and FMC for later symbol resolution. Now in tests, once we have a model, we could do 
```Rust
        model.reset_stack_sampler();
        measure_hdr(sampler, model, CommandId::ABC);
        model.gen_flamegraph("ABC").unwrap();
```
to render a flamegraph named `ABC.svg` under `target/flamegraphs`.

clundin25:
@pjhades could you add an API that includes this new option, that the existing API wraps? That will reduce the amount of code this touches.

E.g.

```Rust
fn caliptra_builder::build_and_sign_image_ext(opts: BuilderOptions) -> (ImageBundle, Elfs) {
 . . .
} 

fn caliptra_builder::build_and_sign_image() - > ImageBundle {
 let out, _ =  caliptra_builder::build_and_sign_image_ext
 out
}
```

clundin25 (approved):

PR #4075, opened 2026-08-07
runtime: Reject reserved MCU RT firmware ID on STASH_MEASUREMENT

parvathib:
Firmware ID 2 is reserved for the Caliptra-managed MCU RT DPE context. STASH_MEASUREMENT does no authorization, so MCU (the PL0 caller in subsystem mode) could stash under the reserved ID and update its own measurement.

Add StashAuthorization to distinguish measurements Caliptra has authorized from raw mailbox input, and reject the reserved ID for unauthorized ones.

AUTHORIZE_AND_STASH keeps working with firmware ID 2, since it only stashes after verifying the image against the signed SoC manifest and is required by the hitless MCU update flow.

Also replace the magic [2, 0, 0, 0] literals with MCU_RT_RESERVED_FW_ID.

Addresses review feedback on #4070.

### DIFF
--- a/runtime/src/stash_measurement.rs
+++ b/runtime/src/stash_measurement.rs
@@ -17,7 +17,9 @@ use crate::{
 };
 #[cfg(feature = "cfi")]
 use caliptra_cfi_derive::cfi_impl_fn;
-use caliptra_common::mailbox_api::{MailboxRespHeader, StashMeasurementReq, StashMeasurementResp};
+use caliptra_common::mailbox_api::{
+    ActivateFirmwareReq, MailboxRespHeader, StashMeasurementReq, StashMeasurementResp,
+};
 use caliptra_dpe::{
     commands::{Command, DeriveContextCmd, DeriveContextFlags},
     context::ContextHandle,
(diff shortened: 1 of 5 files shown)


runtime/src/stash_measurement.rs
```diff
@@ -30,23 +32,63 @@ use zerocopy::{FromBytes, IntoBytes};
...
+/// reachable with a measurement Caliptra has authorized against the signed SoC
+/// manifest (or produced itself).
+///
+/// This deliberately has no `Default` impl and no conversion from raw mailbox
+/// bytes so that a mailbox caller cannot forge [`StashAuthorization::Authorized`].
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(crate) enum StashAuthorization {
+    /// The measurement was authorized against the signed SoC manifest, or was
+    /// produced by a Caliptra-internal flow such as recovery.
+    Authorized,
```
zhalvorsen: nit: authorized might be a little ambiguous here, could we use something like `CaliptraManaged` or something more specific?

zhalvorsen (dismissed):

zhalvorsen (approved):

runtime/src/stash_measurement.rs
parvathib: Good call! renamed to  `CaliptraManagedContextAccess`  with variants  `Allowed / Denied` (param is now  caliptra_managed_access ).

zhalvorsen: That's perfect, thanks!

mhatrevi (approved):

PR #4077, opened 2026-08-10
runtime: allow PL1 callers to use AUTHORIZE_AND_STASH with SKIP_STASH

GwangbaeCHOI:
Previously, PL1 callers were rejected unconditionally. Now PL1 is permitted when the SKIP_STASH flag is set, since stashing into DPE is the only PL0-exclusive operation. Adds a test covering the PL1 rejection path when SKIP_STASH is not set.

Also fixes README: LoadAddress and ImageStagingAddress source values were documented as 3/4 but are actually 2/3.

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -1529,7 +1529,7 @@ Command Code: `0x4154_5348` ("ATSH")
 | context     | u8[48]   | Context field for `svn`; e.g., a hash of the public key that authenticated the SVN. |
 | svn         | u32      | The version of the image |
 | flags       | u32      | See AUTHORIZE_AND_STASH_FLAGS below |
-| source      | u32      | This field identifies the source of the digest to be used to compare with the SoC's<br />SHA digest in the SoC Manifest<br /><br />Values<br />1 - InRequest - Use the hash in the 'measurement' field of this command<br /><br />3 - LoadAddress - The image located in the `ImageLoadAddress` will be streamed to the SHA Accelerator to <br />               retrieve the digest that will be used for authorization.<br />4 - ImageStagingAddress - The image located in the `StagingAddress` will be streamed to the SHA Accelerator to<br />               retrieve the digest that will be used for authorization |
+| source      | u32      | This field identifies the source of the digest to be used to compare with the SoC's<br />SHA digest in the SoC Manifest<br /><br />Values<br />1 - InRequest - Use the hash in the 'measurement' field of this command<br /><br />2 - LoadAddress - The image located in the `ImageLoadAddress` will be streamed to the SHA Accelerator to <br />               retrieve the digest that will be used for authorization.<br />3 - ImageStagingAddress - The image located in the `StagingAddress` will be streamed to the SHA Accelerator to<br />               retrieve the digest that will be used for authorization |
 | image_size   | u32      | The size of the image to hash. Only valid if source is `ImageLoadAddress` or `StagingAddress` |
 
 *Table: `AUTHORIZE_AND_STASH_FLAGS` input flags*
--- a/runtime/src/authorize_and_stash.rs
+++ b/runtime/src/authorize_and_stash.rs
@@ -56,16 +56,17 @@ impl AuthorizeAndStashCmd {
         resp: &mut [u8],
     ) -> CaliptraResult<usize> {
         let caller_privilege_level = drivers.caller_privilege_level();
-        match caller_privilege_level {
-            // Only PL0 can call STASH_MEASUREMENT
-            PauserPrivileges::PL0 => (),
-            PauserPrivileges::PL1 => {
-                return Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL);
-            }
-        }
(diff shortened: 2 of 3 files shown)


runtime/src/authorize_and_stash.rs
```diff
@@ -56,16 +56,17 @@ impl AuthorizeAndStashCmd {
...
-                return Err(CaliptraError::RUNTIME_INCORRECT_PAUSER_PRIVILEGE_LEVEL);
-            }
-        }
         let locality = drivers.mbox.id();
 
         if let Ok(cmd) = AuthorizeAndStashReq::ref_from_bytes(cmd_args) {
+            let auth_and_stash_flags: AuthAndStashFlags = cmd.flags.into();
+            // PL1 callers may only use AUTHORIZE_AND_STASH with SKIP_STASH set;
+            // stashing into DPE requires PL0.
+            if caller_privilege_level == PauserPrivileges::PL1
```
GwangbaeCHOI, line 65: PL1 user is only allowed to use the authorize portion of the command

runtime/README.md
```diff
@@ -1520,7 +1520,7 @@ Command Code: `0x4154_5348` ("ATSH")
...
+| source      | u32      | This field identifies the source of the digest to be used to compare with the SoC's<br />SHA digest in the SoC Manifest<br /><br />Values<br />1 - InRequest - Use the hash in the 'measurement' field of this command<br /><br />2 - LoadAddress - The image located in the `ImageLoadAddress` will be streamed to the SHA Accelerator to <br />               retrieve the digest that will be used for authorization.<br />3 - ImageStagingAddress - The image located in the `StagingAddress` will be streamed to the SHA Accelerator to<br />               retrieve the digest that will be used for authorization |
```
GwangbaeCHOI, line 1532: Updated together as these numbers do not match with the implementation:

https://github.com/chipsalliance/caliptra-sw/blob/main/api/src/mailbox.rs#L2629-L2658

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
```diff
@@ -1684,3 +1684,33 @@ fn test_verify_invalid_manifest() {
     let authorize_and_stash_resp = AuthorizeAndStashResp::read_from_bytes(resp.as_slice()).unwrap();
     assert_eq!(authorize_and_stash_resp.auth_req_result, IMAGE_AUTHORIZED);
 }
+
+#[test]
+fn test_authorize_and_stash_pl1_without_skip_stash_fails() {
```
clundin25, line 1695: Do we have a test that confirms authorize will work for PL1?

clundin25 (dismissed):

runtime/src/authorize_and_stash.rs
zhalvorsen, line 65: This should also return an error if the source is anything besides `InRequest`

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
zhalvorsen: It would be good to have such a test

zhalvorsen (dismissed):

runtime/src/authorize_and_stash.rs
GwangbaeCHOI: I'm not sure if allowing other options could be harmful. Can you share your insight?

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
GwangbaeCHOI: RIght, it will be better to have it. I will update it.

runtime/src/authorize_and_stash.rs
zhalvorsen: The problem I'm thinking of is allowing PL1 to execute a DMA operation external to the chip that could make it halt if that address is either invalid or not accessible at the time. This would create a denial of service attack for PL1. We have been limiting that trust to only PL0 callers.

When `source` is `ImageHashSource::LoadAddress` or `ImageHashSource::StagingAddress` Caliptra will use the SHA DMA engine to calculate the hash of the image.

GwangbaeCHOI: I see. I think we don't need to worry about such problem as the address is already fixed in the SoC manifest, and a PL1 user can cause a denial of service more easily by acquiring the mailbox lock and waiting. So, I guess the risk is quite low.

runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs
GwangbaeCHOI: It's added.

runtime/src/authorize_and_stash.rs
zhalvorsen: That's fair, the address and size has to be signed by the vendor and owner

clundin25 (approved):

zhalvorsen (approved):

PR #4081, opened 2026-08-10
drivers: Reduce PersistentData size

ternst-nv:
Reduce the reserved area of PersistentData by 4Kb.  This can be utilized for MLDSA optimization operations.

This also removes sparse space reservations for a number of data types, which are defined either through the ROM, Hardware or FMC.

Note:  This does change the location of some runtime specific Persistent Data fields.  This is acceptable since Hitless Update cannot be supported for this release due to the changes FMC.

### DIFF
--- a/drivers/src/memory_layout.rs
+++ b/drivers/src/memory_layout.rs
@@ -24,9 +24,9 @@ pub const CFI_STATE_ORG: u32 = 0x500003E4; // size = 6 words
 pub const BOOT_STATUS_ORG: u32 = 0x500003FC;
 pub const PERSISTENT_DATA_ORG: u32 = 0x50000400;
 
-pub const DATA_ORG: u32 = 0x50009C00;
+pub const DATA_ORG: u32 = 0x50009400;
 
-pub const STACK_ORG: u32 = 0x5000A400;
+pub const STACK_ORG: u32 = 0x50009C00;
 pub const ROM_STACK_ORG: u32 = 0x5001C000;
 
 pub const ESTACK_ORG: u32 = 0x5001F800;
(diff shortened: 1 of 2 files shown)


clundin25 (dismissed):

drivers/src/persistent.rs
```diff
@@ -43,17 +43,21 @@ pub const PCR_LOG_SIZE: u32 = 1024;
 pub const MEASUREMENT_LOG_SIZE: u32 = 1024;
 pub const FUSE_LOG_SIZE: u32 = 1024;
 pub const DPE_SIZE: u32 = 5 * 1024;
-pub const PCR_RESET_COUNTER_SIZE: u32 = 1024;
+pub const PCR_RESET_COUNTER_SIZE: u32 = 128;
 pub const AUTH_MAN_IMAGE_METADATA_MAX_SIZE: u32 = 7 * 1024;
-pub const IDEVID_CSR_SIZE: u32 = 1024;
-pub const FMC_ALIAS_CSR_SIZE: u32 = 1024;
+pub const IDEVID_CSR_SIZE: u32 = 516;
+pub const FMC_ALIAS_CSR_SIZE: u32 = 772;
```
zhalvorsen: Do these affect the ROM portion of persistent data? That may not be obviously until the nightly runs, but it could make the firmware incompatible with previous ROMs.

drivers/src/memory_layout.rs
```diff
@@ -54,7 +54,7 @@ pub const ICCM_SIZE: u32 = 128 * 1024;
 pub const DCCM_SIZE: u32 = 128 * 1024;
 pub const ROM_DATA_SIZE: u32 = 996;
 pub const DATA_SIZE: u32 = 2 * 1024;
-pub const STACK_SIZE: u32 = 85 * 1024;
+pub const STACK_SIZE: u32 = 89 * 1024;
```
zhalvorsen: I thought we were okay on stack space now. Did something change? Another option we could do is split the difference and expand the number of DPE contexts to 64. That has been a common ask from a lot of people over the years

drivers/src/persistent.rs
```diff
@@ -43,17 +43,21 @@ pub const PCR_LOG_SIZE: u32 = 1024;
 pub const MEASUREMENT_LOG_SIZE: u32 = 1024;
 pub const FUSE_LOG_SIZE: u32 = 1024;
 pub const DPE_SIZE: u32 = 5 * 1024;
-pub const PCR_RESET_COUNTER_SIZE: u32 = 1024;
+pub const PCR_RESET_COUNTER_SIZE: u32 = 128;
```
zhalvorsen: If we are going to break persistent data compatibility. We should also add add a version field while we are at it and maybe a marker magic value. This would make it is easier to ensure compatibility in later firmwares.

zhalvorsen (approved):

PR #4091, opened 2026-08-12
fpga: Adds a boot sequence for external I3C host

lstrz:
Depending on the external I3C boot parameter, a configuration sequence will be executed that sets things up for Caliptra to boot from the external I3C device.

### DIFF
--- a/hw-model/src/model_fpga_subsystem.rs
+++ b/hw-model/src/model_fpga_subsystem.rs
@@ -7,7 +7,7 @@ use crate::api_types::{DeviceLifecycle, Fuses};
 use crate::bmc::Bmc;
 use crate::fpga_regs::{
     Control, FifoData, FifoRegs, FifoStatus, FlashControl, FlashCtrlRegs, FlashOpStatus,
-    ItrngFifoStatus, WrapperRegs,
+    ItrngFifoStatus, SpareI3cControlSts, WrapperRegs,
 };
 use crate::keys::{DEFAULT_LIFECYCLE_RAW_TOKENS, DEFAULT_MANUF_DEBUG_UNLOCK_RAW_TOKEN};
 use crate::mcu_boot_status::McuBootMilestones;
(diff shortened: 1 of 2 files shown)


hw-model/src/model_fpga_subsystem.rs
```diff
@@ -587,6 +589,7 @@ impl ModelFpgaSubsystem {
             .regs()
             .mcu_reset_vector
             .set(FPGA_MEMORY_MAP.rom_offset);
+       self.wrapper.regs().spare_i3c_control_sts.modify(SpareI3cControlSts::UseExtI3cHost::SET);
```
clundin25: Should this be gated by the boolean?

hw-model/src/model_fpga_subsystem.rs
```diff
@@ -2296,6 +2308,29 @@ impl HwModel for ModelFpgaSubsystem {
...
+            cb(self);
+        }
+
+        // Notify MCU ROM it can start loading firmware.
+        let gpio = &self.wrapper.regs().mci_generic_input_wires[1];
+        let current = gpio.extract().get();
+        // MCU ROM will wait after reaching the mailbox for this bit before booting RT.
+        gpio.set(current | 1 << 31);
+
+        // ironically, we don't need to support this
```
clundin25: I think we can delete this comment

lstrz: Done.

hw-model/src/model_fpga_subsystem.rs
lstrz: ~I meant for this init sequence to be used specifically in scenarios where we know we want to call that, hence the name of the function indicating that.~
~The condition you're talking about will thus be checked before this call happens in the first place, by the caller. I have a subsequent commit that shows its use better. I'm not strictly opposed to modifying this though, if you feel strongly.~

lstrz: My bad, I think this line being here is a mistake, as we already have it in enable_external_i3c_upload_firmware_rri()

clundin25 (approved):

PR #4096, opened 2026-08-13
mldsa: cache polys

ternst-nv:
Implement an algorithm for caching the `A` matrix during MLDSA signing operations.  This can greatly reduce the cost of Rejection loop processing, as it avoids duplicated calculation of the `A` matrix polynomials which relies on the Keccak algorithm.

As part of this include a series of tests validating the full functionality, and which can produce timing results.

Stack and Performance metrics:
image

image

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -54,6 +54,17 @@ const K_GAMMA_2: u32 = (K_PRIME - 1) / 32;
 const BETA: u32 = 120;
 const OMEGA: usize = 75;
 
+/// Number of ML-DSA-87 matrix A polynomials cached across the signing rejection
+/// loop. A is 8 rows x 7 columns = 56 polynomials and depends only on `rho`, so
+/// caching avoids re-expanding these via SHAKE128 on every iteration — the
+/// dominant per-iteration cost. Each cached polynomial is 24-bit packed (768
+/// bytes vs 1024 unpacked) and the cache lives on the sign stack.
+const A_CACHE_POLYS: usize = 50;
+/// Bytes per 24-bit-packed cached polynomial (256 coeffs * 3 bytes).
+const A_CACHE_POLY_BYTES: usize = K_DEGREE * 3;
+/// Total A-cache size in bytes.
+const A_CACHE_BYTES: usize = A_CACHE_POLYS * A_CACHE_POLY_BYTES;
+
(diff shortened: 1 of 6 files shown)


zhalvorsen:
It looks like this adds about 40KB to stack usage. Is that right?

ternst-nv:
Yes, that would be about right.  The stack usage is directly tied to the number of polynomials we cache, so we can directly trade between performance and DCCM usage if we wish to keep greater buffer.

zhalvorsen (dismissed):

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -680,9 +725,17 @@ fn matrix87_expand_mul_mask(
...
-            scalar_from_keccak_vartime(&mut m_ij, &derived_seed);
+            let idx = j * 8 + i;
+            if idx < A_CACHE_POLYS {
+                scalar_unpack_24(
+                    &mut m_ij,
+                    &a_cache[idx * A_CACHE_POLY_BYTES..(idx + 1) * A_CACHE_POLY_BYTES],
+                );
+            } else {
+                derived_seed[K_RHO_BYTES + 1] = i as u8;
+                derived_seed[K_RHO_BYTES] = j as u8;
```
pjhades, line 736: nit: it looks like this assignment can be moved to the outer loop?

runtime/tests/runtime_integration_tests/test_command_timing.rs
```diff
@@ -111,3 +131,241 @@ fn measure_runtime_command_timing() {
...
+    use dpe::{
+        commands::{CertifyKeyCommand, CertifyKeyFlags, CertifyKeyMldsa87Cmd},
+        context::ContextHandle,
+    };
+    use zerocopy::IntoBytes;
+
+    /// Number of distinct-label CERTIFY_KEY calls to sample.
+    const N: usize = 50;
+
+    let mut model = run_rt_test(RuntimeTestArgs {
```
pjhades: nit: is it possible for this test to reuse `sample_certify_worker`?

pjhades (dismissed):

common/crypto/mldsa/src/mldsa87.rs
ternst-nv: Hmm, derived seed is only used in this block, so I'd prefer to keep it here.

zhalvorsen (approved):

PR #4106, opened 2026-08-13
Add execution of CSRNG KAT before usage in CFI init

### LINKED ISSUE #4037: [FIPS-140-3] - CTR_DRBG (CSRNG) executed before its SP 800-90A CAST in ROM boot — violates FIPS 140-3 / ISO 19790 AS10.27
* Caliptra Version(s) (2.1)
* Caliptra Mode (Subsystem Mode)

## Problem

In Caliptra ROM boot, the SP 800-90A **CTR_DRBG (CSRNG)** is executed (`Generate()`) to seed the CFI random-delay PRNG and the ABR SCA countermeasure at `rom/dev/src/main.rs:82-88` and `:101-104` — **before** the DRBG's known-answer self-test (`CsrngKat`, `kat/src/lib.rs:80`) runs.

```rust
CfiCounter::reset()                          cfi_counter.rs:77-80
  └─ prng().mix_entropy(entropy_gen)
       entropy_gen = || trng.generate4()     rom/dev/src/main.rs:82-85
            └─ Trng::generate4()             drivers/src/trng.rs
                 └─ csrng.generate12()       drivers/src/csrng.rs:187   ← CSRNG / CTR_DRBG Generate
``` 
| Self-test | Standard | IG 10.3.A item | Runs before first DRBG `Generate()`? |
|---|---|---|---|
| Entropy-source health tests (repcnt/adaptp/markov) | SP 800-90B | Resolution #7 (are CASTs, AS10.34) | **YES — hardware-gated** |
| CTR_DRBG known-answer test (`CsrngKat`) | SP 800-90A | Resolution #6 (§11.3) | **NO — runs later** |

## AS violated

- **ISO/IEC 19790:2025 AS10.27** — *"The conditional cryptographic algorithm self-test shall be executed prior to the first operational use of the cryptographic algorithm after power-on…"* (mirrored in **FIPS 140-3 IG 10.3.A [10.27]**).
- Reinforced by **IG 10.3.A GeneralNote8** — *"a CAST shall be performed prior to first time algorithm execution in the approved mode,"* and **Resolution 6** (SP 800-90A DRBG CAST required per §11.3 for the `Generate` function that is the one being invoked).

The CTR_DRBG (an approved algorithm; a random bit generator is a "security function" per **ISO 3.126**) is executed before its CAST completes — contrary to AS10.27's plain, unqualified text.

## Caveat

SP 800-90A §11.3 full text not locally verified; confirm it doesn't already permit pre-KAT DRBG execution when no approved output is consumed.

nquarton:
Run the CSRNG KAT prior to any use.

We believe the existing use before the KAT should be permitted given it is not a FIPS approved use case. But this change removes any debate or doubt during CMVP review.

### DIFF
--- a/rom/dev/src/main.rs
+++ b/rom/dev/src/main.rs
@@ -76,6 +76,12 @@ pub extern "C" fn rom_entry() -> ! {
         Err(e) => handle_fatal_error(e.into()),
     };
 
+    // Run CSRNG KAT before any CSRNG output is consumed
+    #[cfg(not(feature = "fake-rom"))]
+    if let Err(e) = CsrngKat::default().execute(&mut trng) {
+        handle_fatal_error(e.into());
+    }
+
     // Initialize CFI before creating the rest of the environment
     // (AesGcm::new runs KATs which have CFI annotations)
     if cfg!(feature = "cfi") {
--- a/rom/dev/tests/rom_integration_tests/test_cfi.rs
+++ b/rom/dev/tests/rom_integration_tests/test_cfi.rs
@@ -9,6 +9,7 @@ use caliptra_common::RomBootStatus;
 use caliptra_emu_cpu::CoverageBitmaps;
 use caliptra_hw_model::{BootParams, HwModel, InitParams};
 
+#[allow(dead_code)]
 fn find_symbol<'a>(symbols: &'a [Symbol<'a>], name: &str) -> &'a Symbol<'a> {
     symbols
         .iter()
@@ -57,8 +58,7 @@ fn test_memcpy_not_called_before_cfi_init() {
 
         hw.step_until_boot_status(RomBootStatus::CfiInitialized.into(), true);
 
-        assert_symbol_not_called(&hw, find_symbol(&symbols, "memcpy"));
-        assert_symbol_not_called(&hw, find_symbol(&symbols, "memset"));
+        // Ideally we would also check for memcpy and memset, but the CSRNG KAT has to happen before the CFI init which uses it
         assert_symbol_not_called(&hw, find_symbol_containing(&symbols, "read_volatile_slice"));
         assert_symbol_not_called(
             &hw,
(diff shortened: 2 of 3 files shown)


zhalvorsen (dismissed):

RaunakGu:
@nquarton @zhalvorsen, Could you please help to add this under ROM-2.1.3 milestone

rom/dev/tests/rom_integration_tests/test_cfi.rs
```diff
@@ -57,8 +58,7 @@ fn test_memcpy_not_called_before_cfi_init() {
 
         hw.step_until_boot_status(RomBootStatus::CfiInitialized.into(), true);
 
-        assert_symbol_not_called(&hw, find_symbol(&symbols, "memcpy"));
-        assert_symbol_not_called(&hw, find_symbol(&symbols, "memset"));
+        // Ideally we would also check for memcpy and memset, but the CSRNG KAT has to happen before the CFI init which uses it
```
nquarton, line 61: I unfortunately had to remove this portion of the check since the CSRNG KAT introduces both of these prior to CFI init. My understanding is there is no degradation in security since that KAT didn't have CFI annotations to begin with. But this means this test is less likely to catch other code being added prior to CFI init.

clundin25 (dismissed):

rom/dev/src/main.rs
```diff
@@ -76,6 +76,12 @@ pub extern "C" fn rom_entry() -> ! {
         Err(e) => handle_fatal_error(e.into()),
     };
 
+    // Run CSRNG KAT before any CSRNG output is consumed
+    #[cfg(not(feature = "fake-rom"))]
+    if let Err(e) = CsrngKat::default().execute(&mut trng) {
+        handle_fatal_error(e.into());
+    }
```
clundin25, line 83: Where did this run before? Shouldn't there be some code to delete?

It would be nice to make the KAT happen in the CSRNG constructor, but that is a broader change

nquarton: It ran with the rest of the KATs and it is still there. I can't fully remove it from there since it needs to run when invoking the self tests (which shares that code). We could add some args that specify whether or not to run it, but given this is a very quick KAT (~10k cycles or 0.025ms) I figured the safer approach for this patch was to just add an extra execution of it. The same patterns exists for the SHA256 KAT that has to run prior to the ROM integrity test. That also runs twice.

Vishal also was looking into refactoring KAT execution so they can run on-demand instead of like this.

Overall, I am not really opposed to optimizing this. But I went for the fewest ROM changes here since a refactor of this is expected.

clundin25: That sounds good, I don't have any feedback that is actionable, just was interested in raising it

mhatrevi (dismissed):

mhatrevi (approved):

clundin25 (approved):

PR #3930, opened 2026-06-30
Record subsystem mode in ROM DeviceStatus measurement

v-focutter:
Summery:
* Added the subsystem to the device status
* Added Tests for both states of the new flag

Task [3916](https://github.com/chipsalliance/caliptra-sw/issues/3916)

PR #4112, opened 2026-08-15
[nightly] Support older ROMs in DICE and PCR0 derivation for nightly CI

zhalvorsen:
#3930 ("Record subsystem mode in ROM DeviceStatus measurement") added subsystem_mode to the ROM DeviceStatus PCR measurement (extending it from 17 to 18 bytes) and to gen_fmc_alias_vendor_device_info_hash. Follow-up commit #4065 ("Fix smoke test failures caused by missing subsystem mode in TCB info and PCR0 derivation") updated test derivation logic unconditionally.

Nightly release workflows run test suites against frozen older ROMs (2.1.0 and 2.1.1 via CPTRA_CI_ROM_VERSION). Because older ROMs did not include subsystem_mode in DeviceStatus or vendor device info TCB digests, smoke_test failed on both:

1. TCB Info comparison between the ROM-generated certificate and the FMC-generated CSR (where FMC from HEAD hashed subsystem_mode while the older ROM did not).
2. Expected PCR0 / FMC Alias Key derivation in test helpers.

This change:

1. Adds `SocIfc::rom_fw_rev_id()` to inspect CPTRA_FW_REV_ID[0] at runtime.
2. Updates `common::dice::gen_fmc_alias_vendor_device_info_hash` to only include subsystem_mode when running on ROM > 2.1.2.
3. Conditionalizes `subsystem_mode` inclusion in `test::derive::Pcr0::derive` and `smoke_test::smoke_test` based on `caliptra_builder::get_ci_rom_version()`.
4. Updates `test_derive_pcr0` unit test to assert golden PCR0 values for both older and latest ROM configurations.

Tested:

- `smoke_test::smoke_test` with CPTRA_CI_ROM_VERSION=2.1.0
- `smoke_test::smoke_test` with CPTRA_CI_ROM_VERSION=2.1.1
- `smoke_test::smoke_test` with default (latest) ROM
- `cargo check --all-targets --workspace`


builder/src/lib.rs
```diff
@@ -508,6 +527,35 @@ pub fn rom_for_fw_integration_tests_fpga(fpga: bool) -> io::Result<Cow<'static,
                 Err(other_err(format!("Unexpected ROM fwid {rom_from_env:?}")))
             }
         }
+        CiRomVersion::Rom2_1_2 => {
```
clundin25, line 530: Out of scope for this PR, we should replace this logic with: https://github.com/chipsalliance/caliptra-sw/blob/main/builder/src/firmware.rs#L739

clundin25 (approved):

builder/src/lib.rs
zhalvorsen: Oh yeah I forgot you added that. That would be easy better

PR #4120, opened 2026-08-17
runtime: fix aliasing UB

pjhades:
Rely on zerocopy traits to cast the raw command buffer to a `&mut [AuthManifestImageMetadata]` in a safe way.

### DIFF
--- a/runtime/src/set_auth_manifest.rs
+++ b/runtime/src/set_auth_manifest.rs
@@ -363,16 +363,9 @@ impl SetAuthManifestCmd {
             .get_mut(..metadata_size)
             .ok_or(CaliptraError::RUNTIME_AUTH_MANIFEST_IMAGE_METADATA_LIST_INVALID_SIZE)?;
 
-        let entry_count = u32::from_le_bytes(
-            buf.get(..size_of::<u32>())
-                .ok_or(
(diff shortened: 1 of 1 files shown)


clundin25 (dismissed):

runtime/src/set_auth_manifest.rs
```diff
@@ -360,23 +360,28 @@ impl SetAuthManifestCmd {
...
         // Resize the buffer to the metadata size.
         let buf = cmd_buf
-            .get(..metadata_size)
+            .get_mut(..metadata_size)
             .ok_or(CaliptraError::RUNTIME_AUTH_MANIFEST_IMAGE_METADATA_LIST_INVALID_SIZE)?;
 
-        // Typecast the mailbox buffer to the image metadata collection.
-        let metadata_mailbox =
-            unsafe { &mut *(buf.as_ptr() as *mut AuthManifestImageMetadataCollection) };
+        let entry_count = u32::from_le_bytes(
```
ternst-nv: [nit] Could we rely on the `ImageMetadataCollection` to provide the address in the buf of the entry count?  We could also use `zerocopy` for u32 to shrink this code a bit.

ternst-nv (dismissed):
LGTM

ternst-nv (approved):
LGTM

PR #4136, opened 2026-08-20
docs:  Add PL0/PL1 documentation

ternst-nv:
Document whether a command is restricted, and if so any PL1 permissions in both a easily viewable table and as documentation on each command.

This is intended to serve both as a convenient notice for integrators/ developers on what commands are available.


clundin25 (approved):

nquarton (approved):
Looks good to me.

I worry a little that having the into in two places (both the table and the command reference) may eventually end up out-of-sync. But I don't feel that strongly that it needs to be changed.

PR #4137, opened 2026-08-21
mldsa: Add some additional zeroizing

ternst-nv:
As a defence in depth precaution add some additional zeroizing of sensitive information calculated as part of the MLDSA library internals and utilizing callers.

This does not effect stack usage at all, increases imem by ~400 bytes, with more than 5Kb of imem remaining, and does slightly increase the runtime of MDLSA operations.  The marginal performance regression is determined to be worth the defense in depth provided.

### DIFF
--- a/runtime/src/set_pq_seed.rs
+++ b/runtime/src/set_pq_seed.rs
@@ -30,14 +30,16 @@ impl SetPqSeedCmd {
         drivers.ensure_pl0()?;
 
         let mut cmd = SetPqSeedReq::new_zeroed();
-        crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())?;
+        crate::packet::copy_from_mbox(drivers, cmd.as_mut_bytes())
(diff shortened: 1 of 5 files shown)


runtime/src/set_pq_seed.rs
```diff
@@ -35,13 +36,16 @@ impl SetPqSeedCmd {
...
         Self::derive_pq_devid_cdi(drivers, &cmd.seed, &mut out)?;
 
         drivers
             .persistent_data
             .get_mut()
-            .set_pq_devid_cdi(out.into())?;
+            .set_pq_devid_cdi((*out).into())?;
+
+        // The request buffer still holds the caller-supplied seed.
+        cmd.seed.zeroize();
```
clundin25, line 50: I think this will be skipped on an early return on line 40 or 42.

ternst-nv: Good catch, added zeroizing for those, thanks!

clundin25 (approved):

PR #4139, opened 2026-08-21
fpga: Adds a test for external I3C host boot

lstrz:
The test is ignored by default and needs to be explicitly invoked with something such as
`cargo xtask-subsystem fpga test --target-host root@192.168.0.12 --test-filter='package(caliptra-runtime) and test(test_boot_external)" --run-ignored all -E "none()'`

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -79,6 +79,10 @@ slow-timeout = { period = "30s", terminate-after = 100 }
 filter = 'test(test_pl0_pl1_reallocation_range)'
 slow-timeout = { period = "30s", terminate-after = 100 }
 
+[[profile.fpga-subsystem.overrides]]
+filter = 'test(test_boot_external_i3c_host)'
+slow-timeout = { period = "1m", terminate-after = 60 }
+
 [profile.fpga-subsystem.junit]
 path = "/tmp/junit.xml"
 store-success-output = true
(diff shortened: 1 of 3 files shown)


.config/nextest.toml
```diff
@@ -79,6 +79,10 @@ slow-timeout = { period = "30s", terminate-after = 100 }
 filter = 'test(test_pl0_pl1_reallocation_range)'
 slow-timeout = { period = "30s", terminate-after = 100 }
 
+[[profile.fpga-subsystem.overrides]]
+filter = 'test(test_boot_external_i3c_host)'
+slow-timeout = { period = "30m", terminate-after = 2 }
```
clundin25: 60 minutes seems like a fairly long timeout, is that expected?

It could also be nice to chunk the period to say 30s -1 min and increate the terminate-after count so we can see that the test hasn't hung

lstrz: The way my workflow worked usually is running this test and then going to the BMC and copying images, enumerating the device over I3C, running OCP Recovery, retrying if something fails, using a logic analyzer to look at I3C traffic, etc.

I made it a big timeout specifically so that I have plenty of time to iterate, knowing that there's an active device on the other side listening to I3C traffic.

I lowered the period to 1m and increased terminate-after proportionally.

clundin25 (approved):

PR #4141, opened 2026-08-22
runtime: AES-XTS requirement compliance

### LINKED ISSUE #3526: [OCP-LOCK]: Check AES-XTS half equality
The AES-XTS halves should not be equal. We must introduce a check in firmware. 

More details are in https://github.com/chipsalliance/Caliptra/issues/618.

GwangbaeCHOI:
AES-XTS requires that the two halves of a key be distinct. Enforce this at generation by redrawing until the key is valid, and again just before decryption, as equality between the halves is preserved across AES-ECB decryption.

The halves are checked at both 128-bit and 256-bit granularity, so the AES-XTS-{128,256} requirements are both satisfied.

Resolves #3526

### DIFF
--- a/runtime/src/ocp_lock/mod.rs
+++ b/runtime/src/ocp_lock/mod.rs
@@ -557,11 +557,16 @@ pub struct Mek {
 impl Mek {
     /// Generate a new `Mek`
     pub fn generate(trng: &mut Trng) -> CaliptraResult<Self> {
-        let key = {
-            let seed = trng.generate16()?;
-            transmute!(seed)
-        };
-        Ok(Self { key })
+        loop {
+            let key = {
+                let seed = trng.generate16()?;
+                transmute!(seed)
+            };
+
+            if is_valid_xts_key(&key) {
+                return Ok(Self { key });
+            }
+        }
     }
 
     /// Consumes `Mek` and encrypts `Mek` with the `Mdk` and returns a `SingleEncryptedMek`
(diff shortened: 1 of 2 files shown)


runtime/src/ocp_lock/mod.rs
```diff
@@ -1460,6 +1480,11 @@ fn timeout_to_mtime(command_timeout: Millisecond, clock_period: Picosecond) -> u
     walltime_to_mtime(bounded_command_timeout, period)
 }
 
+/// Check if the given key is a valid key for AES-XTS-{128, 256}
+fn is_valid_xts_key(key: &[u8; 64]) -> bool {
```
GwangbaeCHOI, line 1507: Compare both 128-bit and 256-bit granularity. So, a key is a valid for both AES-XTS-128 and AES-XTS-256

runtime/src/ocp_lock/mod.rs
```diff
@@ -1079,6 +1084,13 @@ impl OcpLockContext {
             cfi_assert!(expect_mek_checksum == checksum);
         }
 
+        // Check validity just before AES-ECB decryption.
+        // Equality between both halves is preserved across decryption,
+        // this is the case because the half-size is a multiple of the 128-bit AES block size.
+        if !is_valid_xts_key(&mek_seed.as_ref().into()) {
```
clundin25, line 1113: I think we should use the CFI code to harden this check against a glitch attack

clundin25 (approved):
LGTM - I think we could consider hardening the if block with a CFI check but it seems we would need to find more codespace to make it work

clundin25:
@GwangbaeCHOI do you mind rebasing? I don't seem to have the necessary permission

runtime/src/ocp_lock/mod.rs
GwangbaeCHOI: Which one should I call? `cfi_launder`?

GwangbaeCHOI:
> @GwangbaeCHOI do you mind rebasing? I don't seem to have the necessary permission

I did it

clundin25 (approved):

PR #4142, opened 2026-08-22
drivers: shrink ML-KEM-1024 KAT by replacing decaps vectors with a digest

GwangbaeCHOI:
Drop the two full decapsulation known-answer vectors (REJECT_DECAP and VALID_DECAP, ~4.7 KB each) and the DecapParams helper. The implicit-rejection path now reuses the keygen-derived dk and corrupts the freshly generated ciphertext in place (first 4 bytes zeroed), then compares the decapsulation output against a single 64-byte SHA-512 digest (KAT_IMPLICIT_REJECTION_SK_DIGEST). Also removes the EK digest check.

Frees ~9.5 KB of .rodata to reclaim runtime image space.


drivers/src/kats/mlkem1024_kat.rs
```diff
@@ -430,11 +59,11 @@ const KAT_SK_DIGEST: Array4x16 = Array4x16::new([
...
 ]);
 
-struct DecapParams {
-    dk: MlKem1024DecapsKey,
-    c: MlKem1024Ciphertext,
-    e: MlKem1024SharedKey,
-}
+// Computed after corrupting first 4-byte of expected ciphertext
+const KAT_IMPLICIT_REJECTION_SK_DIGEST: Array4x16 = Array4x16::new([
+    0x0A93BDDE, 0x596B7198, 0xDC291AAB, 0x8A1CE69E, 0x468F7255, 0xC0B9AD38, 0x30B19523, 0x5E4CD0DB,
```
GwangbaeCHOI, line 64: This digest is computed and tested with external libraries.

drivers/src/kats/mlkem1024_kat.rs
```diff
@@ -455,20 +84,17 @@ fn kat_keygen_encaps_decaps(
...
 
     // Verify keygen outputs via SHA-512 digests
-    let ek_digest = sha2
-        .sha512_digest(ek.as_bytes())
-        .map_err(|_| CaliptraError::KAT_SHA384_DIGEST_FAILURE)?;
     let dk_digest = sha2
         .sha512_digest(dk.as_bytes())
         .map_err(|_| CaliptraError::KAT_SHA384_DIGEST_FAILURE)?;
 
-    if ek_digest != KAT_EK_DIGEST || dk_digest != KAT_DK_DIGEST {
```
GwangbaeCHOI, line 465: encapsulation key comparison can be skipped, and it's a part of decapusulation key. 

From FIPS 40-3 IG Note22
>  if the module implements ML-KEM key generation, the module shall have an ML-KEM key generation 
CAST. The ML-KEM key generation does not take input and outputs an encapsulation key (ek) and a 
decapsulation key (dk).  The CAST shall use the ML-KEM key generation algorithm (i.e., Algorithm 
19 in FIPS 203) or ML-KEM key generation internal algorithm (i.e., Algorithm 16 in FIPS 203), and 
for a KAT, using fixed/predetermined random values (i.e., z and d) to compare the resulting output to 
the pre-computed dk value. It is not necessary to compare the values of ek, as dk contains a copy of 
ek.

drivers/src/kats/mlkem1024_kat.rs
```diff
@@ -492,28 +118,36 @@ fn kat_keygen_encaps_decaps(
...
-            &decap_params.dk,
-            &decap_params.c,
-            MlKem1024SharedKeyOut::Array(&mut shared_key_dec),
-        )
-        .map_err(|_| CaliptraError::KAT_MLKEM1024_DECAPSULATE_FAILURE)?;
-
-    if shared_key_dec != decap_params.e {
-        return Err(CaliptraError::KAT_MLKEM1024_SHARED_KEY_MISMATCH);
+    // Implicit rejection path
+    ciphertext.0[0] = 0; // Corrupt the first 4-byte of ciphertext
```
GwangbaeCHOI, line 138: Testing the implicit rejection path by corrupting the ciphertext

drivers/src/kats/mlkem1024_kat.rs
zhalvorsen: Can we add a unit test as further proof and to make sure this doesn't change

zhalvorsen (approved):

clundin25 (approved):

drivers/src/kats/mlkem1024_kat.rs
clundin25: Chatted with @zhalvorsen and adding a test to the ML-KEM driver tests will capture this

PR #4144, opened 2026-08-24
[mjolnir]: Add Mjolnir GitHub Workflow and Config

clundin25:


clundin25:
FYI @ShadowBearVR

.github/workflows/mjolnir_audit.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: e2-standard-2-reporter
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install Nix
```
zhalvorsen, line 24: Could we use a third party library here?

```
  uses: cachix/install-nix-action@v27
  with:
    extra_nix_config: "experimental-features = nix-command flakes"
```

zhalvorsen (approved):

.github/workflows/mjolnir_audit.yml
clundin25: I took a look and I didn't see any compelling reasons to use this. Happy to follow up and add this, I just generally prefer avoiding new dependencies if possible

Issue #4102, opened 2026-08-13
PQC intermediate secret values not actively zeroized


mojtaba-bisheh:
ML-KEM seed_d, seed_z, message, and shared_key values are never zeroized (the type derives Zeroize, not ZeroizeOnDrop); the 64-byte ML-KEM seed in drivers/src/mlkem.rs is unwiped on every production decap() call, unlike the equivalent pattern used elsewhere (e.g. drivers/src/mldsa87.rs:238-240). A related confidentiality gap: an unredacted Debug derive exists on secret-holding aliases in the decaps-key path, though no active logging call site was found exercising it.
PS: MLKEM Decaps Boolean check should also be actively zeroized

•	Source: caliptra-sw:drivers/src/mlkem.rs, lines 100–104; runtime/src/cryptographic_mailbox.rs (decrypt_mlkem_seeds, mlkem_key_gen).

RaunakGu:
@clundin25 Can you help to add this under ROM-2.1.3 milestone, it has PR targeted for 2.1.3 ROM.

Issue #4099, opened 2026-08-13
ML-DSA Signature buffer size mismatch


mojtaba-bisheh:
The mailbox-visible ML-DSA-87 signature buffer is 4628 bytes, one byte larger than the spec's 4627-byte value. Only one call site (common/src/dice.rs, line 127) correctly truncates to the spec size before further use.

a check is required to drop the command in case of non-zero value for that extra byte.

PR #4148, opened 2026-08-26
Adding check for Buffer Size mismatch and Zeoring intermediate secret data after use. 

v-focutter:
* To address task https://github.com/chipsalliance/caliptra-sw/issues/4099 the code now checks and failures if the extra byte is none-zero.
* To address task https://github.com/chipsalliance/caliptra-sw/issues/4102 seed_d, seed_z, message, and shared_key are zeroed after use, and the debug derives have been removed from the related structures.


drivers/src/ml_kem.rs
```diff
@@ -58,7 +58,7 @@ const DECAPS: u32 = 3;
 const KEYGEN_DECAPS: u32 = 4;
 
 /// ML-KEM-1024 Seeds
-#[derive(Debug, Copy, Clone)]
+#[derive(Copy, Clone)]
 pub enum MlKem1024Seeds<'a> {
```
clundin25, line 61: Keeping this as a `Copy` makes a `move` leave this memory behind. Should we remove that trait so we can ensure we clean up anything left over?

RaunakGu:
@v-focutter, @vmhatre, @zhalvorsen 
This seems to also change Caliptra ROM, Could you please help to add Caliptra ROM target release(Possibly v2.1.3) for this?

common/src/crypto.rs
```diff
@@ -58,7 +58,7 @@ pub const CMK_MAX_KEY_SIZE_BITS: usize = 512;
 pub const UNENCRYPTED_CMK_SIZE_BYTES: usize = 80;
 
 #[repr(C)]
-#[derive(Clone, FromBytes, Immutable, IntoBytes, KnownLayout)]
+#[derive(Clone, FromBytes, Immutable, IntoBytes, KnownLayout, Zeroize)]
```
zhalvorsen: It would be worth adding `ZeroizeOnDrop` here so it can be cleared automatically. It should reduce some of the manual zeroization logic you have later

mhatrevi:
`ZeroizeOnDrop` clears the returned `UnencryptedCmk`, but the local `plaintext` buffer still contains a separate copy of the decrypted key material and is not wiped before this function returns. Could we explicitly zeroize it on both the success and parse-error paths?

```suggestion
        let result = UnencryptedCmk::read_from_bytes(&plaintext)
            .map_err(|_| CaliptraError::CMB_HMAC_INVALID_DEC_CMK);
        plaintext.zeroize();
        result
```

This ensures the original AES output buffer does not leave ML-KEM seeds or other CMK key material on the stack.

mhatrevi (dismissed):

mhatrevi (dismissed):

mhatrevi (approved):

PR #4150, opened 2026-08-26
ROM: halt on post-load Runtime verification failure

mhatrevi:
## Summary

- Treat a Runtime digest mismatch detected after loading ICCM as fatal, because the previously authenticated Runtime has already been overwritten.
- Preserve the existing non-fatal fallback for image-verification failures that occur before ICCM is modified.
- Add a subsystem regression test that mutates the staged Runtime after verification and confirms ROM reports a fatal digest mismatch without launching FMC.

## Testing

- `cargo fmt --all`
- `cargo test -p caliptra-rom --test rom_integration_tests test_update_reset_post_load_digest_mismatch_is_fatal -- --nocapture`
- `cargo test -p caliptra-rom --test rom_integration_tests test_update_reset_verify_image_failure -- --nocapture`
- `cargo test -p caliptra-rom --test rom_integration_tests test_update_reset_success -- --nocapture`
- `cargo clippy -p caliptra-rom --tests --no-deps -- -D warnings`

Full dependency Clippy remains blocked by a pre-existing `collapsible_if` warning in `runtime/src/ocp_lock/mod.rs`.

### DIFF
--- a/rom/dev/src/flow/update_reset.rs
+++ b/rom/dev/src/flow/update_reset.rs
@@ -179,7 +179,13 @@ impl UpdateResetFlow {
                 &mut env.dma,
                 staging_addr,
             )?;
-            crate::flow::loaded_image::verify_runtime(&manifest, &mut env.sha2_512_384)?;
+            if let Err(err) =
+                crate::flow::loaded_image::verify_runtime(&manifest, &mut env.sha2_512_384)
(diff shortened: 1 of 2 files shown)


mhatrevi:
Closing for now; this will be re-opened once the change has completed review. No action needed here.

PR #4160, opened 2026-08-27
[mjolnir]: Add Mjolnir GitHub Workflow and Config

ShadowBearVR:
This PR closely mirrors https://github.com/chipsalliance/caliptra-sw/pull/4144 and adds Mjolnir support to the `caliptra-2.0` branch.

### DIFF
--- a/tools/mjolnir/project.nix
+++ b/tools/mjolnir/project.nix
@@ -1,0 +1,15 @@
+# Licensed under the Apache-2.0 license
+# SPDX-License-Identifier: Apache-2.0
+{
+  name = "Caliptra SW";
+  repoName = "caliptra-sw";
+  repoUrl = "https://github.com/chipsalliance/caliptra-sw.git";
+  threatModel = ./threat_model.md;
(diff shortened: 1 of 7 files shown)


tools/mjolnir/project.nix
```diff
@@ -0,0 +1,15 @@
+# Licensed under the Apache-2.0 license
+# SPDX-License-Identifier: Apache-2.0
+{
+  name = "Caliptra SW";
+  repoName = "caliptra-sw";
+  repoUrl = "https://github.com/chipsalliance/caliptra-sw.git";
```
clundin25, line 6: We may want to come up with a strategy to surface firmware versions (e.g. 2.0.x, 2.1.x) so we can differentiate it easily when reading a report.

clundin25 (approved):

tools/mjolnir/project.nix
clundin25: Tracked in https://github.com/chipsalliance/mjolnir/issues/109

PR #4161, opened 2026-08-27
[mjolnir]: enable full scan workflow

ShadowBearVR:
This PR adds the ability to initiate manual codebase-wide scans via GitHub Actions, as well as scheduled weekly scans.

### DIFF
--- a/.github/workflows/mjolnir_full_scan.yml
+++ b/.github/workflows/mjolnir_full_scan.yml
@@ -1,0 +1,51 @@
+# Licensed under the Apache-2.0 license
+# SPDX-License-Identifier: Apache-2.0
+
+name: Mjolnir Security Audit (Full Scan)
+
+on:
+  workflow_call:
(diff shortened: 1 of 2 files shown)


.github/workflows/mjolnir_full_scan.yml
```diff
@@ -0,0 +1,43 @@
...
+  mjolnir-full-scan:
+    name: Mjolnir Security Audit (Full Scan)
+    runs-on: e2-standard-2-reporter
+    timeout-minutes: 300
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
clundin25, line 33: As we talked about offline, we'll need to add a dispatch variable that sets the checkout ref so we can configure the cron trigger from the main branch

clundin25 (approved):

PR #4167, opened 2026-08-30
rom: set FIPS status on ML-DSA verify response

iyernaveenr:
`ECDSA384_SIGNATURE_VERIFY` sets `fips_status` to `FIPS_STATUS_NOT_APPROVED_USER_SUPPLIED_DIGEST` on success, because the caller supplies the digest rather than the message being signed (`rom/dev/src/flow/cold_reset/fw_processor/ecdsa_verify.rs`).

`MLDSA87_SIGNATURE_VERIFY` is the same class of operation, but its handler took the response buffer as `_resp` and left it untouched, with the comment "Zero value of response buffer is good". Since `FIPS_STATUS_APPROVED = 0` and `FIPS_STATUS_NOT_APPROVED_USER_SUPPLIED_DIGEST = 0x5553_5244` ("USRD") (`api/src/mailbox.rs:1158-1159`), that response reported **approved** status for an operation its ECDSA counterpart reports as **not approved**.

This populates the field the same way `ecdsa_verify.rs` does. No behavioural change beyond the reported status; the error path is untouched.

## Verification

`cargo check -p caliptra-rom` passes on the host target, which typechecks the modified source. The change also builds for `riscv32imc-unknown-none-elf` under `--profile firmware --no-default-features --features cfi`, invoked from `rom/dev`.

## Note: `main` does not build with `--no-default-features`

Worth flagging separately, as it is unrelated to this change:

```
error: cannot find macro `cfi_assert` in this scope
  --> rom/dev/src/flow/cold_reset/ocp_lock.rs:33:9
error[E0425]: cannot find function `cfi_launder` in this scope
  --> rom/dev/src/flow/cold_reset/ocp_lock.rs:30:8
```

plus the same pair at `:142` and `:145`. `ocp_lock.rs:6-9` gates its `cfi_assert` / `cfi_launder` imports behind `#[cfg(feature = "cfi")]` but uses them unconditionally at those four sites. Since `rom/dev/Cargo.toml:69` sets `default = ["std", "cfi"]`, dropping default features removes the imports but not the uses. Other ROM files import the same helpers ungated, for example `fw_processor/mod.rs:27-28`.

Reproduces identically on unmodified `main`, verified with a clean worktree as a control. Happy to file it separately.

### DIFF
--- a/rom/dev/src/flow/cold_reset/fw_processor/mldsa_verify.rs
+++ b/rom/dev/src/flow/cold_reset/fw_processor/mldsa_verify.rs
@@ -14,6 +14,7 @@ Abstract:
 
 use caliptra_common::mailbox_api::MailboxRespHeader;
 use caliptra_drivers::{report_fw_error_non_fatal, CaliptraResult, Mldsa87};
+use zerocopy::FromBytes;
 
 pub struct MldsaVerifyCmd;
 impl MldsaVerifyCmd {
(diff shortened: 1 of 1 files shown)


iyernaveenr:
Closing - the premise is wrong.

`ECDSA384_SIGNATURE_VERIFY` takes a caller-supplied digest (`EcdsaVerifyReq.hash`), which is why it reports `FIPS_STATUS_NOT_APPROVED_USER_SUPPLIED_DIGEST`.

`MLDSA87_SIGNATURE_VERIFY` takes a message (`MldsaVerifyReq.message_size` / `.message`) and verifies it via `verify_var` with `stream_msg`, so the module performs the hashing itself. It is not a user-supplied-digest operation, the existing `fips_status` of zero is correct, and the "Zero value of response buffer is good" comment is accurate. `rom/dev/tests/rom_integration_tests/test_mldsa_verify.rs` already asserts `FIPS_STATUS_APPROVED` here, so this change would have broken a passing test.

I checked the constants, the ECDSA precedent and the build, but not the one thing the claim rested on - what the two commands actually take as input - and did not look for an existing test. Apologies for the noise.

PR #4170, opened 2026-08-31
pl0 checks

ternst-nv:
Change ensure pl0 to be a helper function to save ~1.3Kb of instruction memory.  Include some commands.

### DIFF
--- a/runtime/src/activate_firmware.rs
+++ b/runtime/src/activate_firmware.rs
@@ -40,6 +40,9 @@ impl ActivateFirmwareCmd {
         cmd_args: &[u8],
         resp: &mut [u8],
     ) -> CaliptraResult<usize> {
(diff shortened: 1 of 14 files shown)


clundin25 (dismissed):

mhatrevi (changes requested):
Could we explicitly set image_opts.pqc_key_type to FwVerificationPqcKeyType::LMS here? ImageOptions::default() selects MLDSA, but run_rt_test() defaults to LMS and programs the fuses accordingly without overriding the supplied image options.

That mismatch prevents this test from reaching the runtime privilege check, so neither the PL1 rejection nor the unchanged attestation state gets tested. Alternatively, use run_rt_test_pqc() with a matching key type.

mhatrevi (approved):

PR #4177, opened 2026-09-01
mldsa87: reject contexts longer than 255 bytes

pjhades:
Explicitly return an error if a given context is longer than the maximum length defined by FIPS 204, following the the pseudo code provided by the same spec.

### DIFF
--- a/common/crypto/mldsa/src/mldsa87.rs
+++ b/common/crypto/mldsa/src/mldsa87.rs
@@ -55,9 +55,6 @@ const K_GAMMA_2: u32 = (K_PRIME - 1) / 32;
 const BETA: u32 = 120;
 const OMEGA: usize = 75;
 
-/// Maximum length of the context string defined in Section 5.2 and 5.3 of FIPS 204.
-const MAX_CONTEXT_BYTES: usize = 255;
-
 /// Number of ML-DSA-87 matrix A polynomials cached across the signing rejection
 /// loop. A is 8 rows x 7 columns = 56 polynomials and depends only on `rho`, so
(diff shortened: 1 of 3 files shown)


clundin25 (dismissed):

common/crypto/mldsa/src/mldsa87.rs
```diff
@@ -1429,7 +1436,11 @@ pub fn mldsa87_sign_with_context_from_sk(
     randomizer: &[u8; MLDSA87_RANDOMIZER_BYTES],
     msg: &[u8],
     context: &[u8],
-) {
+) -> Mldsa87Result {
+    if context.len() > MAX_CONTEXT_BYTES {
```
ternst-nv: [nit] Could we utilize `let context_len = u8::try_from(context.len()).map_err(...)` instead here?  That would provide us a validated variable we can use, instead of relying on this check executing prior to the cast.

pjhades: Yeah, that's a good point. I'm leaning toward keeping the `if` check. My thoughts:
1) The return type `Mldsa87Result` is actually an enum, so I can't simply do `try_from(..).map_err(..)?`.
2) With the check we'd have some minor win in terms of code size.
3) Checking it at the beginning matches the pseudo code from the spec.

pjhades: Alright, updated based on offline discussion. Thanks.

ternst-nv (approved):
LGTM

clundin25 (approved):

PR #4179, opened 2026-09-01
mldsa: Add additional zeroization

ternst-nv:
Add an additional zeroization logic to remove secrets from stack.  This includes both shake and MLDSA87


clundin25 (approved):

common/crypto/mldsa/src/mldsa87.rs
pjhades, line 1109: Should we zeroize `cs_i` before continuing from here?

ternst-nv: Hmm, I think this should be fine, because we are going to be ovewriting the value on the next iteration.

pjhades (approved):

PR #4180, opened 2026-09-01
Adding pairwise consistency test to MLDSA

### LINKED ISSUE #4168: Add Pairwise Consistency Test to Caliptra 1.3 MLDSA
FIPS requires a pairwise consistency test for key pairs prior to use. Per [FIPS CMVP Implementation Guidance](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf) 10.3.a "Additional Comments" number 1, performing this once should be sufficient even though re re-derive the key pair on each use.

_"a PCT is only required to be performed on the key pair prior to the first exportation of the seed or key pair or prior to the first operational use of the key pair after the seed has been generated within the module. Subsequent generations of the key pair from the stored seed are considered a reconstitution of the key pair and a PCT is not required."_

So, performing a PCT only during SET_PQ_SEED should meet the requirements and minimize the performance hit.

nquarton:
Add a pairwise consistency test to MLDSA.

FIPS requires this to be performed prior to use of or export of the keys. However, it only requires it be performed once even if the keys are re-derived from the same seed (which we do to avoid storing them).

This adds execution of the PCT when setting the PQ seed as well as in DPE commands that derive MLDSA keys.

### DIFF
--- a/drivers/src/mldsa87.rs
+++ b/drivers/src/mldsa87.rs
@@ -17,7 +17,7 @@ Abstract:
 
 --*/
 
-use crate::CaliptraResult;
+use crate::{CaliptraError, CaliptraResult};
 use caliptra_mldsa::Mldsa87 as Mldsa87Sw;
 use core::ops::{Deref, DerefMut};
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
@@ -95,9 +95,9 @@ impl_helper_traits!(Mldsa87Mu);
 /// Software ML-DSA-87 engine.
 ///
 /// This is a stateless wrapper; there is no hardware to own. The methods return
-/// [`CaliptraResult`] for consistency with the other crypto drivers and to keep
-/// the API stable if validation (e.g. a pairwise-consistency self-test) is added
-/// later. The underlying software operations are currently infallible.
+/// [`CaliptraResult`] for consistency with the other crypto drivers though is
+/// only currently leveraged by the pairwise-consistency self-test. All other
+/// underlying software operations are currently infallible.
 ///
 /// Each operation is marked `#[inline(never)]` so that the large (tens of KB)
 /// stack frames used by keygen and signing do not get inlined into — and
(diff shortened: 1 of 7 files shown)


drivers/src/mldsa87.rs
```diff
@@ -203,6 +203,41 @@ impl Mldsa87 {
...
+            crate::FipsTestHook::corrupt_data_if_hook_set(
+                crate::FipsTestHook::MLDSA87_PCT_FAILURE,
+                sig_scratch,
+            )
+        };
+
+        // Verify in a nested #[inline(never)] fn so pub_key lives in its own stack frame
+        // and is never simultaneously alive with the ML-DSA sign algorithm frames above
+        #[inline(never)]
+        fn verify(seed: &Mldsa87Seed, sig: &Mldsa87Signature, msg: &[u8]) -> CaliptraResult<()> {
```
ternst-nv: [nit] I'd prefer to see this defined as a separate function, just to make it clear what is going on here.

runtime/src/dpe_crypto.rs
```diff
@@ -657,28 +671,44 @@ impl crypto::Signer for DpeCrypto<'_> {
             }
 
             Signer::Mldsa { .. } => {
-                let Some(DerivedKey::Mldsa(seed)) = &self.derived_key else {
+                let Some(DerivedKey::Mldsa { seed, pct_done }) = &mut self.derived_key else {
                     return Err(CryptoError::CryptoLibError(3));
                 };
+                if !*pct_done {
+                    let Signature::Mldsa(MldsaSignature(buf)) = &mut *out else {
```
ternst-nv: [nit] Do we need the de-reference and then re-borrow here?

ternst-nv (dismissed):
LGTM - a couple nits

drivers/src/mldsa87.rs
nquarton: I was on the fence about this. I'll switch it back to a separate function.

runtime/src/dpe_crypto.rs
nquarton: Good catch. That should not be needed here

drivers/src/mldsa87.rs
```diff
@@ -203,6 +203,41 @@ impl Mldsa87 {
...
+            )
+        };
+
+        Self::pct_verify(seed, sig_scratch, &PCT_MESSAGE)
+    }
+
+    // Separate #[inline(never)] fn so pub_key lives in its own stack frame
+    // and is never simultaneously alive with the ML-DSA sign frames in pct().
+    #[inline(never)]
+    fn pct_verify(seed: &Mldsa87Seed, sig: &Mldsa87Signature, msg: &[u8]) -> CaliptraResult<()> {
```
zhalvorsen, line 232: optional: A potential optimization here would be to give the pub key as the input instead of the seed so key gen only happens once, but I think the bigger bottleneck is signing so it may not be too worth while

ternst-nv (approved):
LGTM

runtime/src/dpe_crypto.rs
```diff
@@ -633,7 +640,14 @@ impl CdiManager for DpeCrypto<'_> {
...
                 self.derive_key_pair_mldsa((&*cdi_bytes).into(), label, info, &mut seed)?;
-                self.derived_key = Some(DerivedKey::Mldsa(seed));
+                // Skip overwriting if seed matches existing derived_key. Preserves pct_done=true from the first call.
+                if !matches!(&self.derived_key, Some(DerivedKey::Mldsa { seed: prev, .. }) if constant_time_eq(&**prev, &*seed))
+                {
+                    self.derived_key = Some(DerivedKey::Mldsa {
+                        seed,
+                        pct_done: false,
+                    });
+                }
```
zhalvorsen, line 650: nit: optional: what do you think of making this a helper function so it can be used here and above?

zhalvorsen (approved):

drivers/src/mldsa87.rs
nquarton: There may be an optimization that can support this, but it actually can't fit on the stack during signing right now. If every caller already had it on the stack, it could be an option to pass it in at the pct() level. But I think there was at least one case where that wasn't true and I was having to add it.

runtime/src/dpe_crypto.rs
nquarton: I agree it would be better if this was not duplicated. Opened a small PR for this: https://github.com/chipsalliance/caliptra-sw/pull/4183

drivers/src/mldsa87.rs
zhalvorsen: That makes sense, thanks!

PR #4205, opened 2026-09-14
runtime: Allow ML-KEM keys in HKDF

embediver:
Allow 32 bytes IKM as CM_HKDF_EXTRACT input, to be able to use 32 byte ML-KEM keys in HKDF.
Additionally different IKM length are now allowed regardless of selected hash algorithm.


zhalvorsen (dismissed):

runtime/src/cryptographic_mailbox.rs
```diff
@@ -2017,10 +2017,8 @@ impl Commands {
         let ikm = Self::decrypt_hmac_key(drivers, &cmd.ikm)?;
         let salt = Self::decrypt_hmac_key(drivers, &cmd.salt)?;
 
-        match (cm_hash_algorithm, ikm.length) {
-            (CmHashAlgorithm::Sha384, 48) => {}
-            (CmHashAlgorithm::Sha512, 64) => {}
-            _ => return Err(CaliptraError::RUNTIME_MAILBOX_INVALID_PARAMS)?,
+        if !matches!(ikm.length, 32 | 48 | 64) {
```
zhalvorsen, line 2020: Can you add a test for 32 byte IKMs as well?

zhalvorsen (dismissed):

zhalvorsen:
It looks like `test_cryptographic_mailbox::test_hkdf_random` is failing

embediver:
> It looks like `test_cryptographic_mailbox::test_hkdf_random` is failing

CM_IMPORT is used in the test and also had the restriction.
I pushed a commit that allows 32 byte keys for HMAC usage on CM_IMPORT.
Let me know if that is ok.

zhalvorsen (dismissed):

runtime/src/cryptographic_mailbox.rs
parvathib: is this check redundant? `decrypt_hmac_key` is already validating the lengths?

runtime/src/cryptographic_mailbox.rs
```diff
@@ -517,7 +517,7 @@ impl Commands {
         match (key_usage, cmd.input_size) {
             (CmKeyUsage::Aes | CmKeyUsage::Mldsa, 32) => (),
             (CmKeyUsage::Ecdsa, 48) => (),
-            (CmKeyUsage::Hmac, 48 | 64) => (),
+            (CmKeyUsage::Hmac, 32 | 48 | 64) => (),
```
parvathib, line 520: Caliptra already returns the ML-KEM derived secret as a CMK. Do we need 32-byte HMAC imports beyond the HKDF tests, or could we leave CM_IMPORT unchanged?

runtime/tests/runtime_integration_tests/test_cryptographic_mailbox.rs
```diff
@@ -2330,103 +2330,105 @@ fn test_hkdf_random() {
...
-        const KEYS: usize = 16;
-        let mut keys = vec![];
-        let mut cmks = vec![];
-        for _ in 0..KEYS {
-            let mut key = vec![0u8; size];
-            seeded_rng.fill_bytes(&mut key);
-            cmks.push(import_key(&mut model, &key, CmKeyUsage::Hmac));
-            keys.push(key);
-        }
+    for size in [32, 48, 64] {
```
parvathib, line 2333: If there’s no production need for importing 32-byte HMAC keys, could we cover the 32-byte IKM tests using an ML-KEM-derived CMK instead?

runtime/src/cryptographic_mailbox.rs
parvathib, line 2020: Is this check redundant? `decrypt_hmac_key` is already validating the ikm lengths.

runtime/README.md
```diff
@@ -1929,7 +1929,7 @@ Implements HKDF-Extract as specified in [RFC 5869](https://www.rfc-editor.org/rf
 The CMKs for IKM and salt must have been created for HMAC usage. The output will be tagged for HMAC usage.
 
 Use CM_IMPORT to import non-secret (plaintext) salt or IKMs to use
-with HKDF-Extract after right-padding to 48 or 64 bytes with zeros.
+with HKDF-Extract after right-padding to 32, 48 or 64 bytes with zeros.
```
parvathib, line 1932: If 32-byte HMAC import is only needed by tests, could we keep both the production  CM_IMPORT  behavior and this documentation unchanged?

runtime/src/cryptographic_mailbox.rs
embediver: Yes, it is. Is it acceptable to rely on that? I would be fine with it.

zhalvorsen (approved):

runtime/src/cryptographic_mailbox.rs
embediver: I discarded that change now and wrote a dedicated test for 32 byte IKM that uses ML-KEM.

parvathib (approved):

PR #4223, opened 2026-09-18
[ATTESTED CSR] update CSR EAT CDDL compliance and use build templates

parvathib:
Update CSR EAT implementation to conform to the latest OCP Device Identity Provisioning specification (PEN 42623, claim -70002 map, claim -70003 discovery) and optimize runtime memory by using build-time byte templates.

- caliptra-ocp-eat: Update OCP Security PEN to 42623; encode key-attributes-map claim (-70002) as map { TaggedOid => bitfield }; add keypair-inventory claim (-70003) with encode/decode support and keypair-id validation.
- runtime: Handle key_id == 0 discovery in GET_ATTESTED_CSR returning signed keypair-inventory token; generate CSR EAT byte templates at build time in build.rs; sign pure ML-DSA-87 and stream ECDSA directly to SHA-384 with in-place buffers, eliminating 16KB stack buffers and caliptra-ocp-eat runtime dependency.
- tests: Add unit and integration tests for inventory discovery, map format, and cryptographic verification of ML-DSA-87 COSE Sign1 envelopes.


common/eat/src/csr_eat.rs
```diff
@@ -205,9 +211,122 @@ impl<'a> CsrEatClaims<'a> {
...
+
+    pub fn with_nonce(entries: &'a [KeyPairInventoryEntry<'a>], nonce: &'a [u8]) -> Self {
+        Self {
+            nonce: Some(nonce),
+            entries,
+        }
+    }
+
+    pub fn validate(&self) -> Result<(), EatError> {
+        if let Some(nonce) = self.nonce {
```
clundin25, line 320: Why not validate the input in the constructor?

common/eat/src/csr_eat.rs
```diff
@@ -205,9 +211,122 @@ impl<'a> CsrEatClaims<'a> {
...
+        }
+
+        size += 5; // Inventory key + array overhead
+        for entry in self.entries {
+            size += 3 + 2; // entry array header + keypair_id uint
+            size += 5; // attributes map header
+            for attr in entry.attributes {
+                size += 3 + attr.oid.oid.len();
+                size += 9;
+            }
```
clundin25: Is it possible to push this logic into the entry type?

common/eat/src/csr_eat.rs
```diff
@@ -205,9 +211,122 @@ impl<'a> CsrEatClaims<'a> {
...
+        }
+
+        Ok(())
+    }
+
+    pub fn estimate_buffer_size(&self) -> usize {
+        let mut size = 10; // Map overhead
+
+        if let Some(nonce) = self.nonce {
+            size += 5 + nonce.len();
```
clundin25: What does the `5` mean?

runtime/src/attested_csr/mod.rs
```diff
@@ -68,45 +45,54 @@ impl TryFrom<u32> for DevIdKeyType {
...
+        if buf.len() < 3 {
+            return Err(CaliptraError::RUNTIME_INSUFFICIENT_MEMORY);
+        }
+        buf[0] = 0x59;
+        buf[1] = (len >> 8) as u8;
+        buf[2] = (len & 0xff) as u8;
+        Ok(3)
+    } else {
+        Err(CaliptraError::RUNTIME_ATTESTED_CSR_EAT_ENCODING_ERROR)
     }
```
clundin25: Do you mind using constants for these?

e.g. `u8::MAX`, `u16::MAX`

Also it would be good to define `0x40, 0x58, 0x59`

runtime/src/attested_csr/mod.rs
```diff
@@ -116,136 +102,236 @@ impl DevIdKeyType {
...
-    pub fn generate_attested_ecc_csr(
-        &self,
-        drivers: &mut Drivers,
-        payload: &[u8],
-        rt_key_id: &[u8; 20],
-        signed_eat_buffer: &mut [u8],
-    ) -> CaliptraResult<usize> {
-        // Get RT public key
-        let rt_pub_key = drivers.persistent_data.get().rom.fht.rt_dice_ecc_pub_key;
+        let prefix_len = CSR_EAT_PREFIX.len() + 32 + CSR_EAT_CLAIM_KEY_CSR.len();
```
clundin25:
```suggestion
        let prefix_len = CSR_EAT_PREFIX.len() + nonce.len() + CSR_EAT_CLAIM_KEY_CSR.len();
```

parvathib: Applied suggestion, thanks!

runtime/src/attested_csr/mod.rs
parvathib: Done! Defined standard `RFC 8949 byte-string header` constants in  `caliptra-ocp-eat::cbor::bstr_header`  and exported them through the build-time template into runtime so there are no magic numbers. 
Also switched the bounds checks to  `u8::MAX`  and  `u16::MAX`

common/eat/src/csr_eat.rs
parvathib: That was an estimate for encoding the negative integer claim key (` -70003` , which takes 5 bytes in CBOR). 
But I totally agree that hardcoding it can be obscure.
Replaced the magic numbers across `csr_eat` with  `CborEncoder::estimate_int_size(...)`  and  `estimate_uint_size(...)` , and delegated entry sizing so it's calculated dynamically. Thanks for pointing that out.

common/eat/src/csr_eat.rs
parvathib: Done. Pushed the CBOR encoding and size estimation into  `KeyPairInventoryEntry::encode`  and  `estimate_buffer_size`.

common/eat/src/csr_eat.rs
parvathib: Good point. Updated the constructors ( `new`  and  `with_nonce`) to validate inputs immediately and return  `Result<Self, EatError>` , ensuring invalid claims or nonce lengths cannot be instantiated.

clundin25 (approved):
This is a great change :)

PR #4229, opened 2026-09-21
[FPGA-CI]: Improve 2.0 smoke test reliability

ArthurHeymans:
The 2.0 subsystem smoke test can start after its one-day build artifacts have expired, and it uses the nightly nextest profile without the FPGA retry policy.

Retain intermediate artifacts for seven days and retry failed FPGA tests three times while preserving the existing 2.0 test filters.


.github/workflows/fpga-subsystem.yml
```diff
@@ -371,7 +371,8 @@ jobs:
               "${COMMON_ARGS[@]}" \
               --test-threads=1 \
               --no-fail-fast \
-              --profile=nightly
+              --profile=nightly \
+              --retries 3
```
clundin25, line 375: I don't love this. This is a bandaid over figuring out why the test is actually flaky.

clundin25 (approved):

